PR #5, opened 2022-01-23
Allow configuration of indicator column

ddeclerck:


nberth:
Why bother? `nl` is so easy to use.

ddeclerck:
Because :
- this avoids a transformation of the source programs (client wishes)
- the GNUCobol developers intended to have such feature (according to the comment at cobc.c:124)
- @lefessan thinks it's a good idea to have such feature

lefessan:
Maybe you should only add the option in the `gcos` files instead of all of them ?

ddeclerck:
> Maybe you should only add the option in the `gcos` files instead of all of them ?

Alas, you can't do that. Configuration files must give a value to all existing config options.

lefessan:
> Why bother? `nl` is so easy to use. In our case, `nl -s "" -n rz` does the trick.

That's assuming:
* the user is under a Unix environment
* the user knows this command (that I just discovered with your post :-) )
In general, the idea is to have direct support in GNUCobol of most GCOS features, and not rely on any external command in the translation.

nberth:
Yes sure it assumes some specific environment and some work on user side. I'm fine this it as it's done. My main concerns were mostly "principled", in that :
- adding this option to GNUCobol makes yet another range of input formats valid in the already large family of "COBOL programing languages", and those formats may not be portable to any other compiler (yet);
- adding this feature adds more lines of code, that naturally increases number of potential bugs, although some dedicated tool already does the job of reformatting with line numbers nicely and has been tested extensvely.

lefessan:
Portability to other non-open-source compilers is not a problem for us :-)
Anyway, the trick you mentionned with `nl` would work for other compilers to easily port such files...

nberth:
No worries I'm just being fussy to publicize the indeed little-known little `nl` ;-)

PR #8, opened 2022-01-24
[GCOS] Support EBCDIC symbolic characters in alphanumeric literals

nberth:


cobc/scanner.l
```diff
@@ -1271,11 +1271,71 @@ error_literal (const char *type, const char *literal)
 	cb_error ("%s", err_msg);
 }
 
+#define WITH_EBCDIC_LITERALS 1
```
lefessan, line 1274: Why use both a macro and a dynamic flag ?

nberth: I initially did not know to what extent this change would break scanning of literals from other standards. In particular, as of now any alphanumeric literal that is immediately followed by a digit without a separator triggers a warning or error if the `gcos-ebcdic-literals` feature is not enabled. In such a situation I though it's better to be able to preproces it out, so I left the macros.
However, using another variant of `CB_CONFIG_*` I'm still not aware of may be more suitable to silently disable ebcdic-scanning; I'll look it up.

nberth: Ok now `gcos-ebcdic-literals` is boolean so the behavior is as before except for `gcos` dialects.

cobc/scanner.l
```diff
@@ -1271,11 +1271,67 @@ error_literal (const char *type, const char *literal)
 	cb_error ("%s", err_msg);
 }
 
+/* EBCDIC to ASCII table, used in read_literal */
+static const cob_u8_t cob_ebcdic_ascii[256] = {
```
lefessan, line 1275: It would make sense to add a comment saying that this table is the same as generated at codegen.c:3195 by `output_ebcdic_to_ascii_table`. Also, why not use the one generated by `output_full_ebcdic_table` ? What's the difference ?

nberth: Actually there are many more ways this should be done, but first I went for the obvious one: The standard says "At object time, each symbolic-character in the non-numeric literal is replaced by the character from the native character set
it represents." I went straight to ASCII as native character set for simplicity.
Now that you mention it I just found the macro to support both ASCII and EBCDIC native character sets…

cobc/scanner.l
nberth: Yes that's already done (that change is outdated).

PR #9, opened 2022-01-25
[GCOS] Add GCOS-specific mnemonics TERMINAL, ALTERNATE-CONSOLE, ALTERNATE CONSOLE

ddeclerck:


config/gcos-strict.conf
```diff
@@ -39,6 +41,9 @@ pretty-display:			yes
 # Allow complex OCCURS DEPENDING ON
 complex-odo:			no
 
+# Adjust position of items following OCCURS DEPENDING
```
nberth, line 44: I think these should belong to another pull request ;-)

config/gcos-strict.conf
```diff
@@ -231,6 +236,7 @@ records-mismatch-record-clause:		error
 record-delimiter:			ok
 sequential-delimiters:			ok
 record-delim-with-fixed-recs:		unconformable
+record-sequential-advancing:		ok
```
nberth, line 239: As above.

config/gcos-strict.conf
lefessan: To be honest, it's very likely that will be squashing all these PRs into a single commit ("add GCOS dialect") sent as a single patch to the official GNUCobol team. So, no need to be extract careful about splitting PRs and commits for now. We may split it at some point to extract things that are not specifically tied to GCOS.

PR #11, opened 2022-01-25
[GCOS] Parse WITH CONVERSION in DISPLAY

ddeclerck:
Parse the WITH CONVERSION clause in DISPLAY. Actual implementation left for later (this is a minor visual effect on the output).

### DIFF
--- a/cobc/parser.y
+++ b/cobc/parser.y
@@ -12140,6 +12140,7 @@ display_body:
 	cb_emit_command_line ($1);
   }
 | screen_or_device_display _common_exception_phrases
+| with_conversion screen_or_device_display _common_exception_phrases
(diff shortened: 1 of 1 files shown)


cobc/parser.y
```diff
@@ -3308,6 +3308,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token WIDTH_IN_CELLS		"WIDTH-IN-CELLS"
 %token WINDOW
 %token WITH
+//%token WITH_CONVERSION          "WITH CONVERSION"
```
lefessan, line 3311: Shouldn't this comment be removed ?

ddeclerck: Indeed.

PR #21, opened 2022-02-02
[GCOS] Improve support for sections and paragraphs that do not end with a period

nberth:


cobc/pplex.l
```diff
@@ -437,7 +437,7 @@ DEFNUM_LITERAL	[+-]?[0-9]*[\.]*[0-9]+
 	return COPY;
 }
 
-"REPLACE"		{
+<*>"REPLACE"		{
```
lefessan, line 440: is it specific to GCOS or standard ?

nberth: This is to support replace statements in control divisions, so it is specific to GCOS.
This edit allows us to remove one rule below in `pplex.l`.

lefessan: My point was more about the fact that it is modified for all dialects, not only GCOS. Isn't it a problem ?

nberth: Hard to assess as it depends on where the various dialects allow `REPLACE` to appear, and all may not be clear about that. Also most other start conditions relate to compiler directives where the `REPLACE` word is forbidden AFAIK.

But the case may still by open for `COPY_STATE` and `PSEUDO_STATE`. In `COPY_STATE`, a `COPY` statement is scanned, and there I think "`REPLACE`" is forbidden too. For `PSEUDO_STATE`, indeed it's not much clearer…
So while in doubt I've pushed a change to explicitly restrict to previous behaviors for the other dialects.

PR #25, opened 2022-03-21
GCOS Support

lefessan:


libcob/common.h
```diff
@@ -1272,7 +1273,7 @@ typedef struct __cob_file_key {
 
 
 /* File version (likely can be removed from cob_file in the future) */
-#define	COB_FILE_VERSION	1
+#define	COB_FILE_VERSION	2
```
GitMensch, line 1276: no need for an updated file version here - only later bits are added and the default is unchanged

GitMensch:
It is a bit hard to follow that diff, please adjust the Changelog files (and recheck, you likely can keep build_aux untouched).

GitMensch:
Could you please try to split this into smaller separate patches along with the changelog entries and tests? This would help _a lot_ in reviewing  it.

cobc/config.def
```diff
@@ -51,7 +51,10 @@ CB_CONFIG_STRING (cb_reserved_words, "reserved-words", _("use of complete/fixed
 CB_CONFIG_INT (cb_tab_width, "tab-width", 1, 12, CB_XRANGE(1,12),
 	_("set number of spaces that are assumed for tabs"))
 
-CB_CONFIG_INT (cb_config_text_column, "text-column", 72, 255, CB_XRANGE(72,255),
+CB_CONFIG_INT (cb_config_indicator_column, "indicator-column", 1, 16, CB_XRANGE(1,16),
```
GitMensch, line 54: This one isn't actually used in this patchset, is there a reason to add it?

GitMensch: Nevermind, I've just stumbled over "very big patch" together with "GH not displaying the bigger diffs".

cobc/warning.def
```diff
@@ -114,5 +114,8 @@ CB_ONWARNDEF (cb_warn_dialect, "dialect",
 CB_NOWARNDEF (cb_warn_source_after_code, "dangling-text",
 	_("  -Wdangling-text       warn about source text after program-area"))
 
+CB_ONWARNDEF (cb_missing_newline, "missing-newline",
```
GitMensch, line 117: You may want to move this out as a separate minimal patch at https://sourceforge.net/p/gnucobol/patches/ - it cannot be "significant" enough for copyright but when you care enough about that to do the change+changelog+testsuite entry I'd include that soon.
Note: as [GCC dropped that warning altogether](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=14331) and the COBOL standard never required it (use of some preprocessors and editors did) let's only include it with `-Wall` (CB_WARNDEF) or `-Wextra` (CB_NOWARNDEF) and with the llvm name [`-Wnewline-eof`](https://clang.llvm.org/docs/DiagnosticsReference.html#wnewline-eof).

cobc/config.def
```diff
@@ -385,3 +404,14 @@ CB_CONFIG_SUPPORT (cb_vsam_status, "vsam-status",
 CB_CONFIG_SUPPORT (cb_self_call_recursive, "self-call-recursive",
 	_("CALL to own PROGRAM-ID implies RECURSIVE attribute"))
 
+CB_CONFIG_SUPPORT (cb_record_contains_depending_clause, "record-contains-depending-clause",
+	_("DEPENDING clause in RECORD CONTAINS"))
+
+CB_CONFIG_BOOLEAN (cb_relax_paragraph_name_uniqueness, "relax-paragraph-name-uniqueness",
```
GitMensch, line 410: This seems to implement https://sourceforge.net/p/gnucobol/feature-requests/260/ - doesn't it? In this case you'd likely reference that in the changelog entry. In any case that issue may have some relevant data to consider in.

cobc/error.c
```diff
@@ -842,6 +842,31 @@ cb_verify (const enum cb_support tag, const char *feature)
...
+	case CB_OBSOLETE:
+		return 1;
+	case CB_SKIP:
+	case CB_IGNORE:
+	case CB_ERROR:
+	case CB_UNCONFORMABLE:
+	default:
+		return 0;
+	}
+#ifndef _MSC_VER
```
GitMensch, line 865: Which compiler raises a diagnostic if this isn't added?

cobc/typeck.c
```diff
@@ -2017,6 +2020,31 @@ cb_build_assignment_name (struct cb_file *cfile, cb_tree name)
 	}
 }
 
+cb_tree
```
GitMensch, line 2023: Please add a comment what this actually should build the name to giving one "special" example (and test case).

cobc/field.c
```diff
@@ -581,15 +581,15 @@ cb_build_full_field_reference (struct cb_field* field)
...
 	while (field) {
-		if (!field->flag_filler) {
-			cb_tree rchain = cb_build_reference (field->name);
-			if (ref) {
-				CB_REFERENCE (ref)->chain = rchain;
-			} else {
-				ret = rchain;
-			}
-			ref = rchain;
+
```
GitMensch, line 584: This is a change of behaviour, the old code goes on with inspecting the parent field when the current one is a filler, the new code here stops the check.

Was that planned?

nberth:
> Could you please try to split this into smaller separate patches along with the changelog entries and tests? This would help _a lot_ in reviewing it.

To help with this, we have started creating a new `gcos4gnucobol-3.x` branch here (that has only three commits above `gnucobol-3.x` already), and separate pull requests for each feature. We fill in ChangeLog entries (and test suites) along the way.

With this I think we can close or leave this particular PR as is.

GitMensch:
Totally agreed.

PR #26, opened 2022-03-25
[GCOS] Update dialect configuration files

ddeclerck:
Co-authored-by: Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>


config/gcos.words
```diff
@@ -103,7 +103,7 @@ reserved:	COMP-15=FLOAT-LONG		# GCOS
 reserved:	COMPUTATIONAL=COMPUTATIONAL-6
 reserved:	COMPUTATIONAL-1=BINARY-SHORT	# GCOS
 reserved:	COMPUTATIONAL-2=BINARY-LONG	# GCOS
-reserved:	COMPUTATIONAL-3=COMPUTATIONAL-6	# GCOS
+reserved:	COMPUTATIONAL-3			# GCOS
```
GitMensch, line 106: That won't work; you'd have to alias the long form to the shorter one, see ibm.words:

```

reserved:	COMP
reserved:	COMP-1
reserved:	COMP-2
reserved:	COMP-3
reserved:	COMP-4
reserved:	COMP-5
reserved:	COMPUTATIONAL=COMP
reserved:	COMPUTATIONAL-1=COMP-1
reserved:	COMPUTATIONAL-2=COMP-2
reserved:	COMPUTATIONAL-3=COMP-3
reserved:	COMPUTATIONAL-4=COMP-4
reserved:	COMPUTATIONAL-5=COMP-5
```

config/gcos.words
```diff
@@ -93,17 +93,17 @@ reserved:	COMMUNICATION
...
-#reserved:	COMP-3=COMP-6			# GCOS
+reserved:	COMP-3				# GCOS
 reserved:       COMP-6				# GCOS
 reserved:	COMP-5=PACKED-DECIMAL		# GCOS
 reserved:	COMP-8=PACKED-DECIMAL		# GCOS
 reserved:	COMP-9=FLOAT-SHORT		# GCOS
 reserved:	COMP-10=FLOAT-LONG		# GCOS
 reserved:	COMP-15=FLOAT-LONG		# GCOS
-reserved:	COMPUTATIONAL=COMPUTATIONAL-6
+reserved:	COMPUTATIONAL=COMP-6
```
GitMensch, line 103: From the manual there seems to be no COMP-6, but others:

| USAGE | Machine Description |PICTURE|
|-------------------|---------------------------|-------|
| DISPLAY | EBCDIC byte | R |
|-------------------|---------------------------|-------|
|* COMPUTATIONAL | Packed decimal (possibly | R |
| or COMP | without sign position | |
| | depending on PICTURE) | |
|-------------------|---------------------------|-------|
| PACKED DECIMAL | Packed decimal (always | R |
| | with sign position) | |
|-------------------|---------------------------|-------|
| BINARY | 16, 32, 48 or 64 bit | R |
| | fixed binary | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-1 | 16-bit fixed binary | NR |
| or COMP-1 | (possibly 32-bit fixed | |
| | binary depending on | |
| | PICTURE) | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-2 | 32-bit fixed binary | NR |
| or COMP-2 | | |
|-------------------|---------------------------|-------|
|* COMPUTATIONAL-3 | Packed decimal (possibly | R |
| or COMP-3 | without sign position | |
| | depending on PICTURE) | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-5 | Packed decimal (always | R |
| or COMP-5 | with sign position, sign | |
| | has ASCII representation) | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-8 | Packed decimal (always | R |
| or COMP-8 | with sign position) | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-9 | Floating binary | NA |
| or COMP-9 | single precision | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-10 | Floating binary | NA |
| or COMP-10 | double precision | |
|-------------------|---------------------------|-------|
| COMPUTATIONAL-15 | Floating binary | NA |
| or COMP-15 | quadruple precision | |
|-------------------|---------------------------|-------|
| BIT | 1 bit per character | R |
| | position | |
|-------------------|---------------------------|-------|
| POINTER | 32 bit direct ITS | NA |
|-------------------|---------------------------|-------|
| INDEX | 6 bytes | NA |

ddeclerck: Yup, that's what we used to alias the usages to whichever was the closest. COMP is an alias of COMP-3, which we remapped to COMP-6, because we thought GnuCobol's COMP-6 was closest to GCOS' COMP-3. Turns out we realized this was not so good, so we just removed the alias, but forgot to remove COMP-6 and set COMP back to COMP-3.

Anyway, I'm not very satisfied with using aliases to deal with usages. I plan to add a configuration feature to make the COMP* usages configurable (i.e. allow to map each COMP* usage with an "internal" usage), in the spirit of the existing `binary-comp-1` option.

That's an other subject, but the `binary-size` is also different in GCOS (neither 2-4-8, 1-2-4-8, nor 1--8). It's close to 2-4-8, but with 48-bits only for integers between 10 and 14 digits (that would be 2-4-6-8 ?). We just use 2-4-8 for now.

In any case, that's the closest we can get to the GCOS specificities using just a config file.

nberth:

config/gcos.words
GitMensch: Let's just get the GCOS dialect included as noted in the other issue - I'd prefer a single commit with the newest config and a single ChangeLog entry for this - and please include your note about "not 100% matching" within the file, this ensures we and others are aware of this when checking.
And I agree - the internal mapping for the `USAGEs` should come later - and should also get a discussion and a rough draft on the GnuCOBOL dev list (that's a private one, will add you three sometime in the next weeks) - before you take the time of the actual implementation.

nberth: David, the original author of this PR, is not currently available; as I'm taking over and cannot update this PR, I'll open a one with the aforementioned changes.

PR #27, opened 2022-03-25
[GCOS] Add support for EBCDIC symbolic characters in alphanumeric literals

nberth:


GitMensch:
That's an interesting one and the implementation seems complete. Can you point to some documentation on this GCOS extension, please?
I do wonder if there's only one actual EBCDIC->ASCII table as there are also multiple encodings in ASCII and things like that are commonly used for extended characters, not :alnum:.

cobc/scanner.l
```diff
@@ -1257,11 +1257,76 @@ error_literal (const char *type, const char *literal)
 	cb_error ("%s", err_msg);
 }
 
+/* EBCDIC to ASCII table, used in read_literal; this table was copied from
+   `codegen.c:output_ebcdic_to_ascii_table`.  */
+static const cob_u8_t cob_ebcdic_ascii[256] = {
```
GitMensch, line 1262: To drop an unused warning you likely want to surround that by `#ifndef COB_EBCDIC_MACHINE`.

GitMensch:
I see. The manual also provides the full list of expected mappings at "B. Collating Sequences", you likely want to check if the current translation table matches that.

From looking at the documentation, please rename `gcos-ebcdic-literals` to `ebcdic-symbolic-character-string` or similar. Whe possible we prefer to not include the implementation in the name but to match the actual extension.

Also from the docs it seems that ´""123, 256""´ and ´""123  256""´should raise a syntax error, while ´""123,256""´ and ´""123 256""´ shouldn't. Can you verify this with a GCOS environment?

nberth:
OK I'll change the name accordingly.
For the non-unique separators in symbolic character strings, it turns out I cannot really test with a GCOS environment. Is it an issue for you if we support a slightly less strict syntax like `"  "135,151,151"bar"195 194"Z"`, as long as the strict version is supported?

GitMensch:
> For the non-unique separators in symbolic character strings, it turns out I cannot really test with a GCOS environment. Is it an issue for you if we support a slightly less strict syntax like `" "135,151, 151"bar"195 ,194"Z"`, as long as the strict version is supported?

Not an issue, more a preference. We made bad experience with extending extensions when people rely on that and either face problems when compiling for the original environment later or when we later on make the rules stricter (= matching the original environment).

I guess you have sample sources to inspect, if there's only either single comma or space separated ebcdic-symbolic-characters, then let's go with this.

nberth:
For now I went for a warning when such extraneous separators are seen.

nberth:
> That's an interesting one and the implementation seems complete. Can you point to some documentation on this GCOS extension, please? I do wonder if there's only one actual EBCDIC->ASCII table as there are also multiple encodings in ASCII and things like that are commonly used for extended characters, not :alnum:.

After a thorough investigation I have observed that the "EBCDIC" character set as per GCOS7 does not match the "EBCDIC" already encoded in GNUCobol.  For this reason I have added an option `ebcdic-literals-charset` to enable the customization of the character set used in EBCDIC symbolic characters.

I tried, but failed, in identifying what set is encoded in the translation tables already in GNUCobol — I initially thought those matched one from IBM, but none of IBM docs I found mention a one-to-one mapping to ASCII as implemented. For this reason I (temporarily) named it "default".

nberth:
@GitMensch this PR might also be ready now.

cobc/scanner.l
```diff
@@ -1257,11 +1257,189 @@ error_literal (const char *type, const char *literal)
 	cb_error ("%s", err_msg);
 }
 
+#ifndef COB_EBCDIC_MACHINE
```
GitMensch, line 1260: Don't we have a conceptional issue here?
EBCDIC on zLinux ("in the cloud", that's the most likely EBCDIC you'd move an old program to) won't match GCOS EBCDIC expectations, would it?

cobc/cobc.h
```diff
@@ -237,6 +237,14 @@ enum cb_dpc_in_data_options {
 	CB_DPC_IN_ALL
 };
 
+/* Character set used to decode EBCDIC symbolic characters in alphanumeric
+   literals */
+enum cb_ebcdic_literal_charset {
```
GitMensch, line 242: I somehow think that if we add this, then there should be more than only those two sets... Adding IBM default for US and Europe seems to be reasonable.
Note: I _guess_ the translation table was coded similar to MF initially.

cobc/scanner.l
```diff
@@ -1283,11 +1461,46 @@ read_literal (const char mark, const char *type)
...
+		}
+
 		if (c == mark && (c = input ()) != (int)mark) {
 			if (c == '-') {
 				/* Free format continuation ("a"- 'b'- ) */
 				/* Hack it as concatenation */
 				unput ('&');
+			} else if (type[0] == '\0' && cb_ebcdic_literals &&
+				   '0' <= c && c <= '9') {
+				/* GCOS 7: symbolic EBCDIC character in
```
GitMensch, line 1499: As far as known: Is this GCOS only - or is there a similar extension with IBM COBOL or BS2000?

GitMensch (commented):
This PR seems to be "more far away" from being ready - but we may solve some issues or bring them up to consideration; and possibly change to a draft.

cobc/scanner.l
nberth: I have not seen any mention of such symbolic characters in documentations of other dialects.

cobc/scanner.l
nberth: `COB_EBCDIC_MACHINE` is probably not precise enough. Could it then be worth using a predefined set of values for a variable like `COB_MACHINE_CHARS`, with, for now, `ZLINUX` (kind of IBM EBCDIC, if there's only one), and (ASCII8)?

GitMensch:
We could replace that with a configure built variable. So far no one has given this a try.

cobc/cobc.h
nberth: Am I right in assuming that to convert from, *eg*, CP037 (EBCDIC-US) would require a special handling for non-representable characters? At the beginning I just dealt with the few "total" conversions I could find: the one in GCOS7 documentation, and the one already in GnuCOBOL's source code.

nberth:
> We could replace that with a configure built variable. So far no one has given this a try.

I would assume someone with an access to a zlinux system might be in a better position to try this, and I don't have access to any. In any case I'm not sure how to check that with host infos.

nberth:
> In any case I'm not sure how to check that with host infos.
`libcob/common.h`:
```c
/* EBCDIC determination */

#if ' ' == 0x40
#define	COB_EBCDIC_MACHINE
#else
#undef	COB_EBCDIC_MACHINE
#endif
```
I see…

cobc/scanner.l
```diff
@@ -1257,11 +1257,189 @@ error_literal (const char *type, const char *literal)
...
 
+#ifndef COB_EBCDIC_MACHINE
+/* EBCDIC GCOS7 8-bit to ASCII conversion table, used in read_literal below in
+   case symbolic character literals are encoded with the "GCOS7" version of
+   EBCDIC:
+
+   https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN27-oct2009/47A205UL04.pdf,
+   p627.  Note one page is missing from this documentation, but the full table
+   can be found in the French version. */
+static const cob_u8_t cob_gcos7ebcdic_ascii[256] = {
```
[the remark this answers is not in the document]
nberth: I've added a module names `ebcdics` in`cobc`. Alternative names could avoid referencing EBCDIC, like `cconvs` maybe. You probably have a better name in mind ;-)

cobc/cobc.h
nberth: For timing reasons I would like to postpone the introduction of additional charsets, for which tedious research of proper documentations might be required — also a proper policy for handling partial conversions might be required (error? warn and use `space`…).
I also have the feeling that, going down that line further may lead to rather substantial reworks for parameterising the collating sequences (used in `SORT`, for instance).

cobc/scanner.l
```diff
@@ -1257,11 +1257,189 @@ error_literal (const char *type, const char *literal)
 	cb_error ("%s", err_msg);
 }
 
+#ifndef COB_EBCDIC_MACHINE
+/* EBCDIC GCOS7 8-bit to ASCII conversion table, used in read_literal below in
+   case symbolic character literals are encoded with the "GCOS7" version of
```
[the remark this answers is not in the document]
nberth: Here the idea was mainly to deal with symbolic characters given in literals in GCOS-compatible source code when executing on a non-GCOS platform. In this case, this means converting the characters to the native encoding (the "default" EBCDIC on zlinux, or the 8-bit ASCII already used in GnuCOBOL).

nberth:
@GitMensch Despite the work on Area A enforcement being more important for the community, if possible I would like to prioritize this particular PR first, without introduction of new character sets that can be done later.

cobc/scanner.l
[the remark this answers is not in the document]
GitMensch: `cconv` for character-set conversion would be good

cobc/cconv.h
```diff
@@ -0,0 +1,53 @@
...
+   along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
+*/
+/* CHECKME: I think adding this particular header file makes sense, but this
+   could also move to cobc.h for consistency with all other declarations. */
+#ifndef CB_CCONV_H
+#define CB_CCONV_H
+
+#include "config.h"
+
+#include <stdio.h>	  /* CHECKME: needed for definitions in cobc.h; could be
```
GitMensch, line 28: this include and the ones above ad below should be replaced by `include "libcob/common.h"`.

GitMensch:
I'm confused by the rules for symbolic-character-1:
* there may be multiple ones, separated by separator comma
* a single one may not contain of more than 30 digits ?!?

Definitely: those can be "embedded" with in existing strings - for this alone I'd like to wait for the "main feature" until we have embedded characters in there in general, this should be the case with PIC U and utf8-literals (`à la IBM, will be "soon" in discussion for COBOL 202x [the one after COBOL 2022]).

--------

But we (ok, you) _could_ take one part out of this PR _now_, which make future PRs possibly smaller and the current cobc code better in any case: `cconv.[ch]` along with the change to codegen.c; ideally cconv.h also includes the official IBM tables https://www.ibm.com/docs/en/iis/11.3?topic=tables-ascii-ebcdic https://www.ibm.com/docs/en/iis/11.3?topic=tables-ebcdic-ascii, all that are currently used, and the gcos one.

I trust you enough that you can directly commit this part, which is mostly "copy what  is in here already, add with a new single entry to cobc/Changelog).

Not necessary but then more useful: a new flag (flag.def) `-febcdic-table` - an internal enum that takes `resticted-gc` (effectively current `cb_flag_alt_ebcdic` which would be removed and the option be placed in cobc.c `long_options` instead), `default` (the current common one), `ibm` (new table), `gcos`.
If you see any way to cut-and-paste + adjust this until Saturday we could include that part in 3.2 (and the optional gcos setting is likely useful for _something_ in your project, too).

nberth:
> * a single one may not contain of more than 30 digits ?!?

I missed that one rule. Maybe the designers of GCOS allowed 30 digits just because they could; or else they planned for extensions :-D

I'll see what I can do about `cconv` in the short term: for sure it will help us with dedicated support for more collating sequences (or should I say codesets, is this what that's called?).

Still, our goal is also to parse those symbolic characters *à là* GCOS: even though the current implementation hard-codes the translation into internal character representation as part of the scanner, do you think the current lexing rules could conflict with what's envisioned for `PIC U`/embedded characters (which I know nothing about yet)?

GitMensch:
> I'll see what I can do about `cconv` in the short term

Thank you very much, I agree that this will be helpful, especially if you can add `-febcdic-table` on the way (allowing to also convert by GCOS and IBM rules with a new flag - and for 3.2 all this without changing the existing ruleset [which will happen with 4.x when this flag is moved to a dialect option after we made the "additional research"]).

> for sure it will help us with dedicated support for more collating sequences (or should I say codesets, is this what that's called?).

That overlaps but is not necessary the same. The codeset define what hex value to expect for a specific single-byte character. This also allows a conversion from one to another (which not necessarily works with all characters, see the IBM approach of "not matching = 0x1B [if I remember correctly] and therefore also is not possible without loosing information [converting in one direction and then back does not necessarily result in the original value). The codeset has always a full range 0x00 - 0xFF.

The collating sequence is about the ordering, relevant for comparisons (mostly less/greater, but also equal). Its lowest position in COBOL is 1 and its highest value depends on the actual collation - GnuCOBOL's default is equal to the "native codeset" (0x00-0xFF, in this case 1 - 256).
ACUCOBOL-GT's for example is not "complete" (it contains less than 256 entries) which was a big gotcha for me when doing a `SORT` of a table with pointers - and finding that the result is not what I've wanted it to be (the fix was to define a full alphabet and use that as program collation).
And for national and unicode data items / literals the collation normally contains more than 256 entries.
[just a note here: at least MF explicit document that for UTF-8 comparisions only the hex values are used = there is no utf8 character comparision via collating sequence]

> Still, our goal is also to parse those symbolic characters _à là_ GCO

That's totally clear and fine. It just won't be in the official 3.2. From what I remember when giving this PR a deeper review the last time I think your approach presented here is fine for a local-adjusted installation of GnuCOBOL.

> do you think the current lexing rules could conflict with what's envisioned for `PIC U`/embedded characters

I'm not sure on this, but I know that for unicode embedded characters there will be a scanning with embedded escape sequence; and before that is in don't want to add others. See https://www.ibm.com/docs/en/cobol-zos/6.3?topic=cobol-using-utf-8-literals and https://www.microfocus.com/documentation/visual-cobol/vc80/DevHub/GUID-ED574236-40D7-4D2E-8BD7-E36DC883E54C.html for its rules.
If you want to give that a go you could create a PR for adding u'' (with both escaped sequences) and ux'' literals with some syntax tests and a single runtime test that uses the `HEX-OF` intrinsic function - we'd then have some "basis" for tackling the GCOS EBCDIC symbolic characters.

lefessan:
I updated this PR as it is used a lot in the GCOS project we are migrating to GnuCOBOL.

When the option `ebcdic-symbolic-character` is `no`, the PR does not modify the interpretation of literals, so it does not prevent the addition of UTF-8 or other encodings in the future for non-GCOS configurations, as discussed in the previous comments.

cobc/ChangeLog
```diff
@@ -538,6 +546,14 @@
 	* typeck.c (validate_occurs): change level 01/77 check back to plain
 	  dialect verification fixing #854
 
+2022-10-05  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```
GitMensch, line 549: duplicate entry

cobc/scanner.l
```diff
@@ -141,6 +141,19 @@ struct cb_top_level_78 {
 	struct cb_level_78	*lev_78_ptr;
 };
 
+enum cb_literal_type {
```
GitMensch, line 144: I'm fine with the internal refactoring to use the literal type enum, feel free to commit that along with a Changelog entry to svn, which will reduce the changes in this PR a bit.

cobc/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2022-12-16  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```
GitMensch, line 2: enum literal type refactoring is missing here

tests/testsuite.src/run_extensions.at
```diff
@@ -124,6 +124,77 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+AT_CHECK([$COMPILE -febcdic-symbolic-characters -febcdic-table=ebcdic500_latin1 prog.cob], [0], [], [])
+AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [])
+
+AT_CLEANUP
+
+
+AT_SETUP([GCOS literals with EBCDIC symbols])
+AT_KEYWORDS([extensions gcos ebcdic])
+
+AT_DATA([prog.cob], [
```
GitMensch, line 172: that's a syntax check, please move to the syn_...at file, that contains the other literal checks, then including also the compile without the new flag

tests/testsuite.src/run_extensions.at
```diff
@@ -124,6 +124,77 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+           MOVE 2 TO T
+           MOVE "  "135,151,151"bar"195 194"Z" TO X
+           IF X NOT EQUAL "  foobarBAZ" PERFORM SHOW.
+
+           MOVE 3 TO T
+           MOVE ' '135 151,151'bar'195,194'Z' TO X
+           IF X NOT EQUAL " foobarBAZ" PERFORM SHOW.
+
+           *> GCOS7-specific: should give a ! in between square brackets:
+           *> note square brackets need to be doubled due to m4 preprocessing.
```
GitMensch, line 152: Can you double the [[ ]] of the prog.cob content instead? Possibly much easier: replace [] by () / {} ?

cobc/scanner.l
```diff
@@ -1377,19 +1465,20 @@ read_literal (const char mark, const char *type)
 	if (!i) {
 		cb_verify (cb_zero_length_lit, _("zero-length literal"));
 		cb_warning (COBC_WARN_FILLER,
-			    type[0] == 'N' ?
-			    _("national literal has zero length; a SPACE will be assumed") :
-			    _("alphanumeric literal has zero length; a SPACE will be assumed"));
+			    _("%s literal has zero length; a SPACE will be assumed"),
```
GitMensch, line 1468: We don't replace the two existing msgids by 1 shortly before the final release; this would need to be postponed after the release but also then only be done when changed for all of the similar ones to something like `"%s literal has zero length; %s will be assumed"`.
Please revert this part for now and feel free to keep it around in a separate PR.

cobc/scanner.l
```diff
@@ -1328,11 +1380,47 @@ read_literal (const char mark, const char *type)
...
+				ebcdic_flag = CB_SYM_EBCDIC_INACTIVE;
+			} else if ('0' <= c && c <= '9') {
+				plex_buff[i++] = scan_ebcdic_char (c);
+				ebcdic_flag = CB_SYM_EBCDIC_ACTIVE;
+			} else if (c != ' ' && c != ',') {
+				cb_error (_("invalid character '%c' in sequence "
+					    "of symbolic EBCDIC characters"), c);
+				ebcdic_flag = CB_SYM_EBCDIC_ACTIVE;
+			} else if (ebcdic_flag == CB_SYM_EBCDIC_AFTER_SEP) {
+				cb_warning (COBC_WARN_FILLER,
```
GitMensch, line 1401: Isn't that an error, both here and above?

GitMensch (changes requested):
I'm quite sure that we can't get that into the final 3.2 release, as we're on post rc2 state already; but I do understand your need to have that integrated in your local version and/or an official dev one, so I took the time to review the current version.
Please adjust as noted, possibly commit the refactoring of the internal literal-type enum (without msgid change) and re-request the review, then I'll have another look.

cobc/ChangeLog
lefessan: fixed

cobc/ChangeLog
lefessan: done

cobc/scanner.l
lefessan: done

tests/testsuite.src/run_extensions.at
lefessan: done using [[...]]

tests/testsuite.src/run_extensions.at
lefessan: I created a file `syn_literals.at` and moved as many tests as possible there, otherwise they would have to be moved into `syn_misc.at` again...

cobc/scanner.l
lefessan: Gcos documentation is not very explicit, it says that separators MUST be used between symbolic characters, but it does not say if using several of them or extra at the end is an error. Maybe we should have a warning only in non-relaxed syntax ?

GitMensch: That's a 100% vendor extension with big use for that and little use for any non-GCOS code - so it should match the original as best as possible; testing on GCOS with presumably invalid data is necessary.

tests/testsuite.src/run_extensions.at
```diff
@@ -123,95 +123,6 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
 
 AT_CLEANUP
 
-
-AT_SETUP([HP COBOL octal literals])
```
GitMensch, line 127: Those are runtime tests which actually need to be executed to verify that the translation is done correctly. Therefore those need to stay in here (and the one necessary runtime check with [[]] should be added below.

tests/testsuite.src/run_extensions.at
GitMensch: You likely have thought about that already:
* the new file must be included in tests/Makefile.am
* the "new" file with the old content should have an entry in tests/ChangeLog
* as noted before: the two run tests must stay in run_extensions, but we _may_ have or miss syntax tests for those, in this case these should be moved/added to the new file
* to decrease syn_misc and have more structure the split test source is very good - to decrease the range of this PR - can you please separate that out to a new one which we can fast-inspect and then distribute with the 3.2 final already?

lefessan: Ok, I must admit I didn't really understand at first the reason why some of those tests were in `run_extensions`, but now, it makes much more sense. I will try to work on such a PR tomorrow.

cobc/scanner.l
lefessan: Ok, I will have to ask David to perform such tests...

lefessan: So, on GCOS, misplaced commas generate an error (duplicates, before first char, after last char), but the behavior is different for spaces: multiple spaces are allowed, and space after last char too. I will try to mimic this behavior in the next version.

lefessan: done

lefessan: I think the new version mimic this behavior:

* spaces are ok everywhere
* extra commas generate errors

cobc/scanner.l
```diff
@@ -1257,11 +1257,189 @@ error_literal (const char *type, const char *literal)
...
+{
+	char buff[10];	       /* Arbitrary limit, mostly for error-reporting */
+	unsigned int j = 0;
+	do {
+		buff[j++] = c;
+	} while ((c = input ()) != EOF && '0' <= c && c <= '9' && j < 10);
+	unput (c);
+	buff[j++] = 0;
+	c = atoi (buff);
+	if (c < 1 || c >= 257) {
```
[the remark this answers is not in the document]
GitMensch: whoa, July 2022, we've come a long way already :-)
No need fo a direct push to fix this but whenever the next one comes please have this "syntactic sugar only" change included

cobc/ChangeLog
```diff
@@ -1,3 +1,10 @@
+2022-02-20  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```
GitMensch, line 1: 2022-02? If that change is _really_ that old please move the entry, but I guess it is just a typo.

cobc/scanner.l
```diff
@@ -1310,12 +1316,51 @@ error_literal (const char *type, const char *literal, unsigned int literal_error
...
+	}
+        c--; // in ordinal, i.e. 1..256 -> 0..255
+#ifdef COB_EBCDIC_MACHINE
+	return (cob_u8_t) c;
+#else
+	if( !ebcdic_to_ascii_initialized ){
+		if (cob_load_collation (cb_ebcdic_table,
+                                        ebcdic_to_ascii,
+                                        NULL) < 0) {
+                        cobc_err_exit (_("invalid parameter: %s"), "-febcdic-table");
```
GitMensch, line 1348: Can we get that error here? If not then please use `(void)cob_load_collation` here, in any case move `ebcdic_to_ascii_initialized = 1;` above because we won't have a different result when getting here the next time and really want to do the loading only once.

tests/testsuite.src/syn_literals.at
```diff
@@ -1511,3 +1511,27 @@ prog.cob:5: error: missing terminating ' character
...
+
+           STOP RUN.
+])
+AT_CHECK([$COMPILE -febcdic-symbolic-characters prog.cob], [1], [], [prog.cob:1: note: free format detected
+prog.cob:4: error: symbolic character 0 out of allowed EBCDIC range (1...256)
+prog.cob:5: error: symbolic character 257 out of allowed EBCDIC range (1...256)
+prog.cob:6: error: invalid character 'x' in sequence of symbolic EBCDIC characters
+prog.cob:7: error: extraneous character ',' in sequence of symbolic EBCDIC characters
+prog.cob:7: error: extraneous separator ',' at end of symbolic EBCDIC characters
+])
```
GitMensch, line 1535: please add a program with valid values (similar to the one in run_extensions), then compile it with and without `-febcdic-symbolic-characters` and one time with `-febcdic-symbolic-character -febcdic-table=dummyNotThere`.

cobc/scanner.l
```diff
@@ -1310,12 +1316,51 @@ error_literal (const char *type, const char *literal, unsigned int literal_error
...
+
+	unsigned int j = 0;
+	do {
+		buff[j++] = c;
+	} while ((c = input ()) != EOF && '0' <= c && c <= '9' && j < 10);
+	unput (c);
+	buff[j++] = 0;
+	c = atoi (buff);
+	if (c < 1 || c >= 257) {
+		cb_error (_("symbolic character %s out of allowed EBCDIC range (1...256)"),
```
GitMensch, line 1337: please reword that msgid to make it usable also for unicode and others later (drop EBCDIC, keep range, possibly adjust "symbolic character")

GitMensch (changes requested):
The good message: I'm fine with including this after the changes outlined in GC 3.2 final.

We add some msgids which is "bad" at this state, but we'll need them here and it isn't too much to be a big offense to the translators that did their work to have a 100% translation ready.

As the changes are now they don't effect either compile performance much and outside of two conditions that will never be true don't do anything.

Well done :-)

cobc/scanner.l
lefessan: From the tests that I added following your advises, it looks like we can indeed get the error and exit there. Anyway, I swapped the `ebcdic_to_ascii_initialized = 1;` to always be done.

cobc/scanner.l
```diff
@@ -1310,12 +1316,51 @@ error_literal (const char *type, const char *literal, unsigned int literal_error
...
+			  buff);
+	}
+        c--; // in ordinal, i.e. 1..256 -> 0..255
+#ifdef COB_EBCDIC_MACHINE
+	return (cob_u8_t) c;
+#else
+	if( !ebcdic_to_ascii_initialized ){
```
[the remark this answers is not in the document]
lefessan: I put the definition within an `#ifndef` clause to get the same effect.

cobc/scanner.l
```diff
@@ -1310,12 +1316,51 @@ error_literal (const char *type, const char *literal, unsigned int literal_error
...
+static  int ebcdic_to_ascii_initialized = 0 ;
+
+	unsigned int j = 0;
+	do {
+		buff[j++] = c;
+	} while ((c = input ()) != EOF && '0' <= c && c <= '9' && j < 10);
+	unput (c);
+	buff[j++] = 0;
+	c = atoi (buff);
+	if (c < 1 || c >= 257) {
```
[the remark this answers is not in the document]
lefessan: done

cobc/scanner.l
```diff
@@ -1310,12 +1316,51 @@ error_literal (const char *type, const char *literal, unsigned int literal_error
...
+		cb_error (_("symbolic character %s out of allowed EBCDIC range (1...256)"),
+			  buff);
+	}
```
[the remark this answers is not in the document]
lefessan: done

cobc/scanner.l
GitMensch: Let's just cb_error + memset the table as '?' to let the program parse until end (because of the swapped setting we won't get in there anymore and just interpret everything as '?''.

tests/testsuite.src/syn_literals.at
```diff
@@ -1511,3 +1511,40 @@ prog.cob:5: error: missing terminating ' character
...
+AT_DATA([prog2.cob],[
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       PROCEDURE        DIVISION.
+           DISPLAY ""135,151,151"bar"195,194,234" = foobarBAZ"
+           STOP RUN.
+])
+AT_CHECK([$COMPILE -febcdic-symbolic-characters prog2.cob], [0], [], [])
+AT_CHECK([$COMPILE prog2.cob], [0], [], [prog2.cob:5: warning: alphanumeric literal has zero length; a SPACE will be assumed
```
GitMensch, line 1544: minor formatting issue: the first output line is wrapped (of course still starting with [ as seen with the other testcases.

GitMensch (approved):
3 minor things open, but you can solve them when integrating it in svn

tests/testsuite.src/syn_literals.at
lefessan: fixed

cobc/scanner.l
lefessan: `cb_error` done, I used a negative value for `ebcdic_to_ascii_initialized` instead of `memset` to return `'?'`

lefessan:
Merged in SVN, thanks Simon !

PR #28, opened 2022-03-28
Allow configuration of indicator column

ddeclerck:


GitMensch:
I guess the changed used_binaries.at will be added - but so far I don't see the purpose of this change, is it the basis for a later patch? I this directly GCOS related?

ddeclerck:
> I guess the changed used_binaries.at will be added - but so far I don't see the purpose of this change, is it the basis for a later patch? I this directly GCOS related?

Not sure if it is GCOS-specific, but our customer happens to have Cobol files where the indicator is in column 1.

GitMensch:
Rechecking both the language reference and the user guide I've found no option to adjust the indicator area.

I therefore _guess_ that when the library files containing the COBOL sources were unloaded, there was possibly a wrong type used.
Is that a GCOS (Bull) -> GnuCOBOL (GNU/Linux) conversion or was there "something else" in between?

Note that GnuCOBOL already has for example fixed format with `-fmf-comment` (which should be moved to a dialect flag) and open feature requests concerning the format.
Implementing those may solve the customer issue already.

https://sourceforge.net/p/gnucobol/feature-requests/29/
https://sourceforge.net/p/gnucobol/feature-requests/230/

lefessan:
Also, we initially thought we could solve the issue by using the free format to parse these files, but they also contain continuation lines, that are not allowed in the free format...

nberth:

cobc/config.def
```diff
@@ -51,7 +51,10 @@ CB_CONFIG_STRING (cb_reserved_words, "reserved-words", _("use of complete/fixed
 CB_CONFIG_INT (cb_tab_width, "tab-width", 1, 12, CB_XRANGE(1,12),
 	_("set number of spaces that are assumed for tabs"))
 
-CB_CONFIG_INT (cb_config_text_column, "text-column", 72, 255, CB_XRANGE(72,255),
+CB_CONFIG_INT (cb_config_indicator_column, "indicator-column", 1, 16, CB_XRANGE(1,16),
```
GitMensch, line 54: please adjust the text here and for the next to the style "indicator column number for fixed-form reference-format", and for the previous drop "set "

cobc/pplex.l
```diff
@@ -1817,14 +1817,14 @@ start:
 			line_overflow = 0;
 		}
 
-		memset (buff, ' ', (size_t)6);
-		/* Note we allow directive lines to start at column 7 */
-		bp = &buff[6];
+		memset (buff, ' ', (size_t)(cb_indicator_column - 1));
+		/* Note we allow directive lines to start in the indicator column */
+		bp = &buff[(cb_indicator_column - 1)];
```
GitMensch, line 1822: style: please always use the same as in line 1907 (in brackets without the parenthesis)

cobc/pplex.l
```diff
@@ -1945,7 +1946,8 @@ start:
 	/* Skip comments that follow after AUTHORS, etc. */
 	if (unlikely (within_comment)) {
 		/* Check all of "Area A" */
-		for (ipchar = 7; ipchar < (n - 1) && ipchar < 11; ++ipchar) {
+		for (ipchar = cb_indicator_column;
+		     ipchar < (n - 1) && ipchar < cb_indicator_column + 4; ++ipchar) {
```
GitMensch, line 1950: please adjust to use something like the following before the `for` (it doesn't matter performance-wise here but is easier to follow)

```c
const int margin_b = cb_indicator_column + 4;
const int ipchar_max = (n - 1 < margin_b)  ? n - 1 : margin_b;
```

cobc/config.def
GitMensch: ... and as all dialects use the same version it seems more reasonable to have that in flags.def; wouldn't it be reasonable to move `cb_indicator_column` from cobc.c only to flags.def and use this everywhere (which would mean one variable less and no changes under "config" at all)?

cobc/cobc.c
```diff
@@ -228,7 +228,7 @@ FILE			*cb_listing_file = NULL;
 
 /* TODO: add new compiler configuration flags for this*/
```
GitMensch, line 229: This comment (which also was misleading - SORRY) should be dropped.

GitMensch (commented):
The changes within the parser and cobc look fine.
The general approach (as this came from my misleading comment which I'm really sorry for) of the dialect option seems bad. The only part which is dialect specific is the question "Is there a margin A + B as in COBOL85?", not the actual margin numbers.

Adding a flag may be useful, but even there I'm not sure.
The likely best option would be to add the acuterminal and cobolx formats - see https://sourceforge.net/p/gnucobol/feature-requests/29/#c2d0/8d2f (both formats are identical with one exception: acuterminal debugging lines start with `\D`, cobolx debugging lines with `D ` and to create a new flag "reference format" which may get one of the options fixed/free/variable/acuterm/cobolx (and then in cobc.c adjust the externalized `cb_indicator_column` and a new internal setting that takes a new enum with these values (to decide for the debugging indicator).

GitMensch:
@ddeclerck The redesign to the flag is well done, looks good.
So far, this leaves us with an "accepted and reviewed patch", but I'm not sure if we should upstream it this way.
 
Obviously this leaves the question: Are you OK with pushing it the "missing piece" further and instead of `-findicator-column` - which I understand was the point of this PR - to have `-fformat` with the options fixed, free, variable, acuterm, cobolx?

nberth:
> Obviously this leaves the question: Are you OK with pushing it the "missing piece" further and instead of `-findicator-column` - which I understand was the point of this PR - to have `-fformat` with the options fixed, free, variable, acuterm, cobolx?

Is the idea just to extend `cb_format` and remove the `-findicator-column` flag? (and effectively have `--free` an alias of `-fformat=free` and similarly for `--fixed`). If that's all it boils down to, then that may be achieved indeed.

GitMensch:
> Is the idea just to extend `cb_format` and remove the `-findicator-column` flag? (and effectively have `--free` an alias of `-fformat=free` and similarly for `--fixed`). If that's all it boils down to, then that may be achieved indeed.

Yes, that's the idea, seems you like it, too.

tests/testsuite.src/run_extensions.at
```diff
@@ -5070,6 +5070,54 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+
+AT_DATA([prog.cob], [
+       >>SOURCE FORMAT COBOLX
+ IDENTIFICATION DIVISION.
+ PROGRAM-ID. prog.
+* blah blah blah
+ PROCEDURE DIVISION.
+                                                        DISPLAY "Hello!"
+     .
+ END PROGRAM prog.
```
GitMensch, line 5084: Is "END PROGRAM" GCOS style? If not please remove that part and test it (as below) using `-fformat`. Please also test the debugging line for cobolx.

cobc/pplex.l
```diff
@@ -1323,6 +1330,67 @@ ppparse_error (const char *err_msg)
...
+static int
+cobc_parse_source_format (enum cb_format *const out, const char *const sfname) {
+	enum cb_format format;
+	if (!cb_strcasecmp (sfname, "FIXED")) {
+		format = CB_FORMAT_FIXED;
+	} else if (!cb_strcasecmp (sfname, "FREE")) {
+		format = CB_FORMAT_FREE;
+	} else if (!cb_strcasecmp (sfname, "VARIABLE")) {
+		format = CB_FORMAT_VARIABLE;
+	} else if (!cb_strcasecmp (sfname, "ACUTERM")) {
```
GitMensch, line 1374: Rechecked MF docs- after their full integration of ACUCOBOL-GT they actually added a
`SOURCEFORMAT"TERMINAL"` compiler directive which means exactly this, so lets use `TERMINAL` instead of `ACUTERM` for the text and stay with `CB_FORMAT_ACUTERM` for the enum value, despite my initial "this likely raises conflicts" thoughts - and handle the conflicts when/if they occur for a user.

GitMensch: ... ok the conflicts are already mentioned in the MF docs:

> ACUCOBOL-GT Terminal format is compatible with the VAX COBOL terminal source format, except for the introduction of the identification area, which VAX COBOL does not support. ACUCOBOL-GT Terminal mode is also compatible with ICOBOL terminal source format, except that area A must start in column 1 (in ICOBOL area A may start in column 2).

Any thoughts about possibly handling this? We should at least "somewhere" copy the facts of this text (not the text) into a comment in cobc, maybe `set_source_format()` is the best place.

cobc/flag.def
```diff
@@ -74,6 +74,10 @@ CB_FLAG_RQ (cb_max_errors, 1, "max-errors", 128, 7,
 
 /* Flags with required parameter and no associated variable */
 
+CB_FLAG_NQ (1, "format", 14,
```
GitMensch, line 77: That's the correct place, guess that looks nice in the generated gnucobol pdf document (I _could_ have a look if the CI update would be live ;-). Please have a look.

The biggest issue: No one but us knows what those "magic values" mean; but luckily there's a way to fix this: https://github.com/OCamlPro/gnucobol/blob/3b31b75dc01dadfd5d4a90d3d045381e56a08fba/doc/gnucobol.texi#L406-L427 (currently missing in this PR); this _may_ could include the _facts_ about TERMINAL from non-acu, too.

cobc/ChangeLog
```diff
@@ -1,4 +1,26 @@
...
+	  format configuration into preprocessor lexer
+	* pplex.l (ppinput): add support for ACU terminal indicators
+	* ppparse.y: extend SOURCEFORMAT directive for ACUTERM and COBOLX
+
+2022-06-24  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* cobc.c, config.def, flag.def, ppparse.y: requalify indicator-column as
+	  a flag instead of a dialect option
+
+2022-02-04  David Declerck <david.declerck@ocamlpro.com>
```
GitMensch, line 19: The February entry is fine as-is, but please move it down to the appropriate date.

GitMensch (changes requested):
I really like the direction of this PR and am looking forward to see it integrated in GnuCOBOL. While it isn't merge-ready yet it is on the finish line, see review comments.

Also: `-fformat` is so big, it should be mentioned in the NEWS entry (for details we can point to the manual).

cobc/cobc.c
```diff
@@ -4985,7 +4988,7 @@ set_listing_header_code (void)
 			"PG/LN  A...B..............................."
 			".............................");
 		if (cb_listing_wide) {
-			if (cb_listing_file_struct->source_format == CB_FORMAT_FIXED
+			if (cb_listing_file_struct->source_format != CB_FORMAT_VARIABLE
```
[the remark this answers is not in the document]
nberth: Actually no change is needed here. I was a bit confused with the test for FIXED format and some failing tests. Also I'm not sure to clearly understand why the behavior should be different based on such a restrictive condition on indicator and text columns: as far as I can tell only checking for `cb_listing_file_struct->source_format == CB_FORMAT_FIXED` may be enough.

cobc/pplex.l
nberth: Does "area A may start in column 2" mean that margin B is "sometimes" shifted one character right in ICOBOL, like if the line starts with a space? The "may" is confusing. If it's a "must", it's COBOLX with `\D`, right?

GitMensch: For ICOBOL I had a look at [their docs](https://www.icobol.com/ftproot/546/referic.pdf). Finding out there are three formats:

* "ANSI Card Format" -> -fformat-fixed
* "xCard (extended card)" format - which is identical to MF's VARIABLE format (but we should increase the right margin to 255, if possible)
* Free-Form Format (CRT) - _identical_ to acuterm but without ` \` before the `D`, so as with ACU there's a "floating" area B (always 4 characters from the indicator area and indicator area does only exist if there's an appropriate value in there)

GitMensch: ICOBOL would be 255 while MF would be 250, and as we don't have `-fformat=variable` until "very soon" people actually used `-ftext-column=250` instead.

But we may consider dropping that option in 4.x (dropping `-fixed`/`-free` is likely too intrusive so those will stay even in 4.x).

cobc/ppparse.y
```diff
@@ -1145,32 +1138,18 @@ refmod_directive:
...
-  {
-	cb_source_format = CB_FORMAT_FREE;
-  }
-| VARIABLE
-  {
-	cb_source_format = CB_FORMAT_FIXED;
-	cb_text_column = 500;
-  }
-| GARBAGE
+| _format _is GARBAGE
```
GitMensch, line 1150: Do we still get in the GARBAGE path with the VARIABLE_NAME token being based on WORD?

cobc/ppparse.y
```diff
@@ -612,6 +612,8 @@ ppparse_clear_vars (const struct cb_define_struct *p)
 %token FIXED
 %token FREE
 %token VARIABLE
+%token ACUTERM
```
GitMensch, line 615: looks like two unused tokens

cobc/pplex.l
```diff
@@ -2337,6 +2466,7 @@ check_listing (const char *text, const unsigned int comment)
 		fprintf (cb_listing_file, "%6d%c", ++listing_line, c);
 	}
 
+	/* FIXME: what's the condition here? */
 	if (requires_listing_line && cb_source_format != CB_FORMAT_FREE &&
 	    strlen (text) > 6) {
```
GitMensch, line 2471: That's `cb_indicator_column - 1` and a check for "is there something to print afterwards (thinking about it: if the length is too small it should point to `""` instead of sequence area, that was buggy before)

cobc/cobc.h
```diff
@@ -353,11 +358,14 @@ struct list_files {
...
 extern struct list_files	*cb_current_file;
 
-extern enum cb_format		cb_source_format;
 #if 0 /* ancient OSVS registers that need special runtime handling - low priority */
 extern enum cb_current_date	current_date;
 #endif
+extern int			cb_indicator_column;
 extern int			cb_text_column;	/* end of area B (in single-byte characters) */
+extern int			cb_floating_area_b; /* whether indicator is optional */
+extern int			cb_fill_literals; /* whether trucated alphanumeric literals should
```
GitMensch, line 367: What are "truncated alphanumeric"?

cobc/cobc.h
```diff
@@ -353,11 +358,14 @@ struct list_files {
 
 extern struct list_files	*cb_current_file;
 
-extern enum cb_format		cb_source_format;
 #if 0 /* ancient OSVS registers that need special runtime handling - low priority */
 extern enum cb_current_date	current_date;
 #endif
+extern int			cb_indicator_column;
 extern int			cb_text_column;	/* end of area B (in single-byte characters) */
+extern int			cb_floating_area_b; /* whether indicator is optional */
```
GitMensch, line 366: Do we need those two variables in cobc.c for listing purposes? If not move to pplex.l and drop the `cb_` prefix.

cobc/ChangeLog
```diff
@@ -1,4 +1,24 @@
 
+2022-07-04  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```
GitMensch, line 2: I think somewhere here should be a reference to the FRs 29 and 130 (also "extended as `[feature-requests:#29], [feature-requests:#130]` in the svn commit message).

doc/gnucobol.texi
```diff
@@ -414,26 +414,55 @@ Place the output into @var{file}.
...
+@item -fformat=variable
+Microfocus' Variable format.  Identical to the fixed format above except
+for the program-text area which extends up to column 250 instead of 72.
+
+@item -fformat=xcard
+ICOBOL xCard format.  Variable format with right margin set at column
+255 instead of 250.
+
+@item -fformat=crt
+ICOBOL Free-form format (CRT).  The program-text area may start in
```
GitMensch, line 444: Isn't this the  X/Open free-form format (FR 130)? If it is we may move this up one entry, reference X/Open first, then ICOBOL.

Note: you can get the definition without cost (just registration needed) from https://publications.opengroup.org/c192

GitMensch (changes requested):
There are minor changes requested - but in general this PR is awesome!

Feel free to either directly commit once solved (no need for a re-review from my side) or explicit request a re-review as you like.

cobc/ppparse.y
nberth: Actually we can get a LITERAL; I'll fix this.

doc/gnucobol.texi
nberth: X/Open required a slightly different debug indicator, so I've added a test for this format as well.
I changed the order of the formats in the documentation, but I'm not sure what's the main driver behind a specific choice; apart from fixed and free that come first, I was using a very loose and subjective measure of readability/usability of tghe sources ;-)

nberth:
I've added yet another format with a new debug indicator, so I think a quick review is in order ;-)
After that it's probably be ready for commit.

doc/gnucobol.texi
nberth: BTW thanks for the link.

cobc/pplex.l
```diff
@@ -1945,7 +2090,12 @@ start:
 	/* Skip comments that follow after AUTHORS, etc. */
 	if (unlikely (within_comment)) {
 		/* Check all of "Area A" */
-		for (ipchar = 7; ipchar < (n - 1) && ipchar < 11; ++ipchar) {
+		/* CHECKME: MF's doc on TERMINAL format reads "Area B starts in
+		   column 5 and extends to the end of the line."  Does this hold
```
GitMensch, line 2094: Will check that with some old ACU on a VM tomorrow...

GitMensch (changes requested):
Some minor things again...

cobc/config.def
```diff
@@ -49,10 +49,10 @@ CB_CONFIG_STRING (cb_reserved_words, "reserved-words", _("use of complete/fixed
 /* Integer flags */
 
 CB_CONFIG_INT (cb_tab_width, "tab-width", 1, 12, CB_XRANGE(1,12),
-	_("set number of spaces that are assumed for tabs"))
+	_("number of spaces that are assumed for tabs"))
 
-CB_CONFIG_INT (cb_config_text_column, "text-column", 72, 255, CB_XRANGE(72,255),
-	_("set right margin for source (fixed format only)"))
+CB_CONFIG_INT (cb_config_text_column, "text-column", 66, 255, CB_XRANGE(66,255),
```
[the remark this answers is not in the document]
nberth: Indeed it's not needed anymore. That was due to the previous `indicator-column` configuration.

cobc/pplex.l
```diff
@@ -128,6 +128,10 @@ struct plex_stack {
 
 /* Global variables */
 
+/* Format-specific columns; should only be assigned in cobc_set_source_format
```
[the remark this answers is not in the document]
nberth: These format-specific attributes were involved in several macros that I had to redefine within `cobc.c`.
To help in balancing readability and performance, these are encapsulated behind pure functions in `pplex.l`.

cobc/pplex.l
```diff
@@ -1945,7 +2093,12 @@ start:
 	/* Skip comments that follow after AUTHORS, etc. */
 	if (unlikely (within_comment)) {
 		/* Check all of "Area A" */
-		for (ipchar = 7; ipchar < (n - 1) && ipchar < 11; ++ipchar) {
+		/* CHECKME: MF's doc on TERMINAL format reads "Area B starts in
```
GitMensch, line 2096: Did the check with ACU 3.2.1, result: It depends on the indicator width:

* if none specified (width = 0): starts at col 5
* if specified, for example a `-` (width = 1), starts at col 6
* for debugging lines with an indicator of `\D` (width = 2): starts at col 7

GitMensch: The last commit - indicator width did actually fix that and only the testcase is missing, correct?
Do we have a chance to actually have that before "one of the next bigger PRs 'area a/b handling' as prerequisite for #35"?

nberth: That last commit did handle indicator width the wrong way (width above 1 was counted as part of area A). I've thus added a test for the `TERMINAL` format to check this point, along with a fix.

GitMensch (approved):
Good test and a good FIXME note. That's good to commit.

nberth:
Ok, commit time ;-)
That PR indeed took quite a turn, but that was pretty fun to work and navigate through all the now newly supported formats.

nberth:
Now upstream :-)

PR #29, opened 2022-03-28
[GCOS] Support DEPENDING clause in RECORD CONTAINS

ddeclerck:


config/gcos-strict.conf
```diff
@@ -239,6 +239,7 @@ assign-ext-dyn:				unconformable
 assign-disk-from:			unconformable
 vsam-status:				unconformable
 self-call-recursive:			skip
+record-contains-depending-clause:	ok
```
GitMensch, line 242: The GCOS docs say "The DEPENDING ON clause in format 3 is kept for compatibility with previous releases" - so should this be marked as "obsolete"? This would only raise a warning if explicit requested.

ddeclerck: Yes indeed. Thanks for pointing this out.

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2022-02-07  David Declerck <david.declerck@ocamlpro.com>
```
GitMensch, line 1: There are ChangeLog files in most directories, please move the entries that are not related to the build system (root) to the files in the matching directories.

Exception: commonly we don't add extra entries in tests/Changelog when adjusting/adding a test (only for bigger changes to the testsuite or helper files).

ddeclerck: Indeed, hadn't noticed. I'll make the change to all my PRs.

nberth:

GitMensch:
Please drop a note once conflicts are solved to get a review.

nberth:
I think this one can be reviewed.

cobc/parser.y
```diff
@@ -6150,12 +6150,18 @@ record_clause:
...
   }
 | RECORD _contains integer TO integer _characters
+  _record_depending /* GCOS extension */
   {
 	check_repeated ("RECORD", SYN_CLAUSE_4, &check_duplicate);
 	if (current_file->organization == COB_ORG_LINE_SEQUENTIAL) {
 		cb_warning (cb_warn_additional, _("RECORD clause ignored for LINE SEQUENTIAL"));
 	} else {
 		set_record_size ($3, $5);
+		if (current_file->record_depending) {
```
GitMensch, line 6160: if there's a duplicate record clause this may be triggered without the clause actually given, to fix this change `_record_depending` to either return its `$3` or (in case of not specified) `NULL`, then adjust this if with using the new token `$7`.
Or: duplicate the complete entry, including _record_depending here, which would allow you to set `current_file->record_depending` based on `cb_verify`.

GitMensch (approved):
This can be included upstream. Please consider adjusting parser.y as mentioned - but that's not mandatory.

GitMensch (commented):
https://github.com/OCamlPro/gnucobol/pull/29/files/2d0a77bfdfa798f7913331e1749d63ef5cb8e5b1..c5d09368e442be140108c6b070676a0b0cda8e74 is nice and clean!

GitMensch:
Good to commit.

nberth:
No upstream

PR #30, opened 2022-03-28
[WIP] [GCOS] Add mnemonics TERMINAL, ALTERNATE-CONSOLE, ALTERNATE CONSOLE

ddeclerck:


cobc/parser.y
```diff
@@ -4184,6 +4184,11 @@ mnemonic_name_clause:
   mnemonic_choices
 ;
 
+word_or_terminal:
+  WORD     { $$ = $1; }
+  /* under GCOS, this reserved name can also be a system name */
+| TERMINAL { $$ = cb_build_reference("TERMINAL"); }
```
GitMensch, line 4190: I'm confused, is this reserved in GCOS now or not? If not then it would get into the "WORD" match with `-std=gcos-strict` (and then gcos.conf should explicit unreserve it after including all words).

ddeclerck: Actually, my first try was to add `not-reserved:	TERMINAL` in gcos.conf.
But then, when compiling a program that uses the `TERMINAL` system name, I would get this error :
```
unknown device 'TERMINAL'; it may exist in another dialect
```
Not sure why. Does `not-reserved` remove it both from the system names and the reserved names ?

PR #31, opened 2022-03-28
[CANCELLED] Allow paragraph names to redefine field and section names

ddeclerck:
This allows a paragraph to have the same name as a previous field or section.

It is not clear whether this is actually allowed or not by the standard or the GCOS dialect, however our customer does have Cobol programs where some paragraphs have the same name as fields or sections. Further discussion is probably required.

Below are the relevant excerpt from the standards for reference.

Both the COBOL 85 standard and GCOS say :
> Within a given source program, but excluding any contained program, the user-defined words are grouped into the following disjoint sets:
> { ..., data-names, ..., paragraph-names, ..., section-names, ... }
> All user-defined words, except segment-numbers and level-numbers, can belong to one and only one of these disjoint sets. Further, all user-defined words within a given disjoint set must be unique, except as specified in the rules for uniqueness of reference.

(nb: these rules say nothing about paragraphs)

The 2014 standard is a bit different :
> Within a source element, a given user-defined word may be used as only one type of user-defined word with the
> following exceptions:
> 1) a compilation-variable-name may be the same as any other type of user-defined word
> 2) a level-number may be the same as a paragraph-name or a section-name
> 3) the same name may be used as any of the following types of user-defined words:
> - constant-name
> - data-name
> - property-name
> - record-key-name
> - record-name

The wording in item 3 is rather ambiguous...

### DIFF
--- a/cobc/config.def
+++ b/cobc/config.def
@@ -442,3 +442,7 @@ CB_CONFIG_SUPPORT (cb_record_contains_depending_clause, "record-contains-dependi
 
 CB_CONFIG_SUPPORT (cb_picture_l, "picture-l",
 	_("PICTURE string with 'L' character"))
+
+CB_CONFIG_SUPPORT (cb_non_unique_procedure_names, "non-unique-procedure-names",
+	_("allow paragraph names to redefine field and section names"))
+
--- a/cobc/error.c
+++ b/cobc/error.c
@@ -1074,6 +1074,24 @@ cb_verify (const enum cb_support tag, const char *feature)
 	return cb_verify_x (&loc, tag, feature);
 }
 
+/**
+ * tells whether the given compiler option is supported by the current std/configuration
+ * \return	1 = ok/warning/obsolete, 0 = skip/ignore/error/unconformable
+ */
+unsigned int
+cb_is_supported (const enum cb_support tag)
+{
+	switch (tag) {
+	case CB_OK:
+	case CB_WARNING:
+	case CB_ARCHAIC:
+	case CB_OBSOLETE:
+		return 1;
+	default:;
+	}
+	return 0;
+}
+
 enum cb_warn_val
 redefinition_error (cb_tree x)
 {
--- a/cobc/typeck.c
+++ b/cobc/typeck.c
@@ -2123,8 +2123,15 @@ cb_build_section_name (cb_tree name, const int sect_or_para)
 		if (!CB_LABEL_P (x) || sect_or_para == 0 ||
 		    (sect_or_para && CB_LABEL_P (x) &&
 		    CB_LABEL (x)->flag_section)) {
-			redefinition_error (name);
-			return cb_error_node;
+			if (cb_is_supported(cb_non_unique_procedure_names) &&
+			    (CB_FIELD_P (x) ||
+			     (CB_LABEL_P (x) && CB_LABEL (x)->flag_section))) {
+				/* To display the usual warning messages, if any */
+				cb_verify(cb_non_unique_procedure_names, _("non-unique section or paragraph name"));
+			} else {
+				redefinition_error (name);
+				return cb_error_node;
+			}
 		}
 	}
 
--- a/config/acu-strict.conf
+++ b/config/acu-strict.conf
@@ -285,6 +285,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				" "
 picture-l:				unconformable
+non-unique-procedure-names:		ok
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		ACU
--- a/config/bs2000-strict.conf
+++ b/config/bs2000-strict.conf
@@ -282,6 +282,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0		# not verified yet, but likely to be as IBM
 picture-l:				unconformable
+non-unique-procedure-names:		error
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		BS2000
--- a/config/ibm-strict.conf
+++ b/config/ibm-strict.conf
@@ -280,6 +280,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0
 picture-l:				unconformable
+non-unique-procedure-names:		error
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		IBM
--- a/config/mf-strict.conf
+++ b/config/mf-strict.conf
@@ -284,6 +284,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				" "
 picture-l:				unconformable
+non-unique-procedure-names:		error
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		MF
--- a/config/mvs-strict.conf
+++ b/config/mvs-strict.conf
@@ -280,6 +280,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0		# not verified yet, but likely to be as IBM
 picture-l:				unconformable
+non-unique-procedure-names:		error
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		MVS
--- a/config/realia-strict.conf
+++ b/config/realia-strict.conf
@@ -285,6 +285,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0		# not verified, but likely like IBM
 picture-l:				unconformable
+non-unique-procedure-names:		error
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		realia
--- a/config/rm-strict.conf
+++ b/config/rm-strict.conf
@@ -288,6 +288,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				" "		# not verified, but possibly like ACU/MF
 picture-l:				unconformable
+non-unique-procedure-names:		ok
 
 # obsolete in COBOL85 and currently not available as dialect features:
 # 1: All literal with numeric or numeric edited item
--- a/tests/testsuite.src/syn_definition.at
+++ b/tests/testsuite.src/syn_definition.at
@@ -650,7 +650,10 @@ AT_DATA([prog.cob], [
            .
 ])
 
-AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+AT_CHECK([$COMPILE_ONLY prog.cob], [0], [],
+[prog.cob:10: warning: non-unique section or paragraph name used
+])
+AT_CHECK([$COMPILE_ONLY -fnon-unique-procedure-names=error prog.cob], [1], [],
 [prog.cob:10: error: redefinition of 'prog'
 prog.cob:7: note: 'prog' previously defined here
 ])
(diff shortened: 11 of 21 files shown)


GitMensch:
It seems this is https://sourceforge.net/p/gnucobol/feature-requests/260 - in this case please check the discussion there and let us discuss the issue itself there (patches and changes here are still ok).

GitMensch:
You've likely done that - but the actual syntax test is missing; it should cover:

* data name and paragraph name identical
* data name and section name identical

is there more? I think this can go "upstream" very soon, I guess you want to commit a GCOS base commit with NEWS entry and `-std` (so conf and words) first.

config/realia-strict.conf
```diff
@@ -155,6 +155,9 @@ console-is-crt:			yes
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	yes
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
GitMensch, line 159: no with comment

config/rm-strict.conf
```diff
@@ -162,6 +162,9 @@ console-is-crt:			no
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	no
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
GitMensch, line 166: was just reported to be "yes"

GitMensch:
The test from the FR may be used:

```cobol
       identification division.
       program-id. WORD.
       data division.
       working-storage section.
      *-----------------------------------------------------------------
       77 word  pic 9.
       77 word2 pic x.
      *-----------------------------------------------------------------
       PROCEDURE DIVISION.
       main section.
      *
           move 0 to word
           perform word
      *
           exit program returning word.
      *-----------------------------------------------------------------
       word section.
      *
           add 1 to word
           if word = 2 goto word2.
           add 1 to word.
      *
       word2.
      *
           continue.
      *-----------------------------------------------------------------
```

cobc/config.def
```diff
@@ -385,3 +385,5 @@ CB_CONFIG_SUPPORT (cb_vsam_status, "vsam-status",
 CB_CONFIG_SUPPORT (cb_self_call_recursive, "self-call-recursive",
 	_("CALL to own PROGRAM-ID implies RECURSIVE attribute"))
 
+CB_CONFIG_BOOLEAN (cb_relax_paragraph_name_uniqueness, "relax-paragraph-name-uniqueness",
```
GitMensch, line 388: change to either `relax-procedure-name-uniqueness` or possibly `procedure-name-with-data-name`, please

GitMensch (changes requested):
Some minor adjustments needed, but nearly finished - the SF feature-request is now assigned to you, too :-)

config/bs2000-strict.conf
```diff
@@ -157,6 +157,9 @@ console-is-crt:			no
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	no
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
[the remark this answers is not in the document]
ddeclerck: Should the option thus be defined using `CB_CONFIG_SUPPORT` instead of `CB_CONFIG_BOOLEAN` ? (so as to be able to use "warning")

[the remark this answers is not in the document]
GitMensch: Sounds good (possibly will have an effect on the name chosen, too)

config/rm-strict.conf
ddeclerck: Done.

config/mf-strict.conf
```diff
@@ -158,6 +158,9 @@ console-is-crt:			no
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	yes
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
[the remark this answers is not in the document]
ddeclerck: Noted.

config/mvs-strict.conf
```diff
@@ -155,6 +155,9 @@ console-is-crt:			no
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	yes
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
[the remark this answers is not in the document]
ddeclerck: Done.

config/ibm-strict.conf
```diff
@@ -155,6 +155,9 @@ console-is-crt:			no
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	yes
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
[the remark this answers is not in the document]
ddeclerck: Done.

config/bs2000-strict.conf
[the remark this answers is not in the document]
ddeclerck: Done.

config/acu-strict.conf
```diff
@@ -155,6 +155,9 @@ console-is-crt:			yes
 # use in a prototype-format CALL/CANCEL statement.
 program-name-redefinition:	yes
 
+# If yes, paragraph names may redefine field and section names (GCOS extension)
+relax-paragraph-name-uniqueness: no
```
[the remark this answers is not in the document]
ddeclerck: Done.

cobc/typeck.c
```diff
@@ -1938,8 +1938,15 @@ cb_build_section_name (cb_tree name, const int sect_or_para)
 		if (!CB_LABEL_P (x) || sect_or_para == 0 ||
 		    (sect_or_para && CB_LABEL_P (x) &&
 		    CB_LABEL (x)->flag_section)) {
-			redefinition_error (name);
-			return cb_error_node;
+			if (cb_is_supported(cb_non_unique_procedure_names) &&
```
GitMensch, line 1941: Shouldn't this be something like:

```c
			if ((CB_FIELD_P (x) ||
			     (CB_LABEL_P (x) && CB_LABEL (x)->flag_section))) {
			        cb_verify (cb_non_unique_procedure_names, _("non-unique procedure-name"));
			} else {
				redefinition_error (name);
				return cb_error_node;
			}
```

tests/testsuite.src/syn_definition.at
```diff
@@ -953,7 +960,11 @@ AT_DATA([prog.cob], [
...
 
-AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+AT_CHECK([$COMPILE_ONLY prog.cob], [0], [],
+[prog.cob: in section 'L':
+prog.cob:6: warning: non-unique section or paragraph name used
+])
+AT_CHECK([$COMPILE_ONLY -fnon-unique-procedure-names=error prog.cob], [1], [],
 [prog.cob: in section 'L':
 prog.cob:6: error: redefinition of 'L'
 prog.cob:5: note: 'L' previously defined here
```
GitMensch, line 970: the FIXME below can be removed

tests/testsuite.src/syn_definition.at
```diff
@@ -928,7 +931,11 @@ AT_DATA([prog.cob], [
...
```
GitMensch, line 941: the FIXME below can be removed

tests/testsuite.src/syn_definition.at
GitMensch: Is everything of `https://github.com/OCamlPro/gnucobol/pull/31#issuecomment-1137626598` already covered? If not then please add it as new testcase compiling once without std (expected warning with zero status) and with `-std-cobol2002` (expecting error -> status 1)

GitMensch (changes requested):
in general this could be either committed before or after the GCOS "base commit"

cobc/error.c
```diff
@@ -848,6 +848,31 @@ cb_verify (const enum cb_support tag, const char *feature)
...
+	case CB_OBSOLETE:
+		return 1;
+	case CB_SKIP:
+	case CB_IGNORE:
+	case CB_ERROR:
+	case CB_UNCONFORMABLE:
+	default:
+		return 0;
+	}
+#ifndef _MSC_VER
```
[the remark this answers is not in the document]
ddeclerck: It's just an occurrence of careless copy-paste (I made this function by copying `cb_verify_x`). Indeed the #ifndef should not be necessary.
I'm not aware of compilers specifically warning for falling through "into" the default case, though some compilers do have warnings (usually disabled) for fall-through in general.
Anyways, I'll just remove the default case and let the function return 0.

tests/testsuite.src/syn_definition.at
nberth: I've added the new test; for now it does not pass and reports an ambiguity. We are currently checking whether this one should indeed pass on GCOS.

On that note, the FR links to a former implementation are all broken. Is there another place where that previous attempt can be investigated?

nberth: Ah yes indeed; somehow I had forgotten that was a former name for gnucobol.

Anyways, as far as our tests go, GCOS appears to subject uses of identical names for sections/paragraphs and data to syntactic scoping rules (as mentioned by David in the FR). This is the behavior of the current implementation.
For this reason we should expect the test cases imported from the FR to fail.

GitMensch: If there's a section and a paragraph defined and you have `PERFORM same-name` in GCOS that is auto-scoped?
So if you do it from `main section` it will call the section with that name, but if you call it from `other-section` that has the `same-name` paragraph defined it uses that?
I'm a bit skeptic.

GitMensch: > If there's a section and a paragraph defined and you have `PERFORM same-name` in GCOS that is auto-scoped?
> So if you do it from `main section` it will call the section with that name, but if you call it from `other-section` that has the `same-name` paragraph defined it uses that?
> I'm a bit skeptic.

The part "what does GCOS do if there is both a section and a paragraph name with the same word" is the most important one not answered (or - totally possible: answered but I've overlooked that).

Would it be useful to go with a lesser version _first_ that matches the old FR? This would be `procedure-shares-data-name` or similar.

GitMensch:
> Would it be useful to go with a lesser version first that matches the old FR? This would be procedure-shares-data-name or similar.

friendly ping

nberth:
> friendly ping

Our migration project does not appear to require this anymore. As this may still be quite useful for other cases, I'll try to see if I can do something quick after a rebase/merge on `gcos4gnucobol-3.x` in the coming days; if not I'm afraid this will need to wait as I do not have much time to allocate on that in the short term.

GitMensch:
friendly ping to recheck if this can be rebased and possibly "slimmed down" a bit for now

> Would it be useful to go with a lesser version first that matches the old FR? This would be `procedure-shares-data-name` or similar [hopefully with a better name, because once found we should not change it in any 3.x].

... actually it would then likely be more reasonable to start in a new (recent) branch and "only" add the "slimmed down" code there, handling the rebase here after we merged that new "slim" version

GitMensch:
While that PR is a cancelled draft... I've just checked IBM behavior, it is as setup in the config here:

~~~
 ==000005==> IGYPS2003-E "L" was defined as both a paragraph-name and a section-name.  This
                         definition was discarded.
~~~

lefessan:
> While that PR is a cancelled draft... I've just checked IBM behavior, it is as setup in the config here:
> 
> ```
>  ==000005==> IGYPS2003-E "L" was defined as both a paragraph-name and a section-name.  This
>                          definition was discarded.
> ```

So, it's not really an error, one of the definitions is just ignored... So, we should add an ibm-like behavior that would not allow duplicated names, but still not error on them...

( I cancelled many PRs that I think will not be finished, and at the same, are interesting to keep in case somebody wants to restart them... )

GitMensch:
> So, it's not really an error, one of the definitions is just ignored... So, we should add an ibm-like behavior that would not allow duplicated names, but still not error on them...

Nope, that's a general thing of IBM and also MF: you can
* define on which severity you don't want to compile/link the parsed result -> defaults to error (I guess Severe may never be compilable)
* adjust the level of the messages (so for example downgrade that to a warning)

The "ignored" mostly means "it is skipped" for further parsing (and even if you downgrade it, it won't be taken into account -> is not added to the AST - or deleted). We already do that for several places with cb_verify - that just would need to be done here as well (depending on its return code: remove the label).

But as likely noted in the review: this is only a partial implementation of the original request, the main point (variables and labels and ids with same name, I guess) was not done, therefore thee very special case implemented here may come _after_ the "main part" is done.

PR #32, opened 2022-03-29
[GCOS] Support literal operands in partial replacing phrases

nberth:
Also factorize code for unquotation of alphanumeric literals in the
preprocessor


nberth:
GCOS also allows a `SPACE` or `SPACES` figurative constant as second operand for partial replacing phrases (cf https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN23-mar2005/doc/05ula04.pdf , pp 534, 538, 135). I'm currently adding this feature.

nberth:
Note this may be related to https://sourceforge.net/p/gnucobol/feature-requests/137/ (except ACU is mentioned there).

ChangeLog
```diff
@@ -1,3 +1,16 @@
+2022-03-29  Nicolas Berthier <nicolas.berther@ocamlpro.com>
```
GitMensch, line 1: Wrong file - you have it in the correct one (cobc), too.

tests/testsuite.src/syn_copy.at
```diff
@@ -577,3 +651,33 @@ AT_CHECK([$COMPILE -free prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [OK], [])
 
 AT_CLEANUP
+
+
+AT_SETUP([REPLACE: partial replacement with SPACES figurative constant])
+AT_KEYWORDS([replace gcos])
```
GitMensch, line 657: add "acu" too, please :-)

GitMensch: That's good, I've meant "because it is in acu conf, use `AT_KEYWORDS([replace gcos acu])`" :-)

cobc/ChangeLog
```diff
@@ -1,18 +1,21 @@
 
-2022-05-31  Simon Sobisch <simonsobisch@gnu.org>
```
GitMensch, line 2: looks like something was lost :-)

cobc/ppparse.y
```diff
@@ -41,6 +41,7 @@
 #ifdef	HAVE_STRINGS_H
 #include <strings.h>
 #endif
+#include <assert.h>
```
GitMensch, line 44: we currently don't use assert.h and may change that in GnuCOBOL 4.x

cobc/ppparse.y
```diff
@@ -72,15 +73,36 @@ static enum cb_directive_action		current_cmd = PLEX_ACT_IF;
 /* Local functions */
 
 static char *
-fix_filename (char *name)
+unquote_alnum (char *name)
 {
 	/* remove quotation from alphanumeric literals */
 	if (name[0] == '\'' || name[0] == '\"') {
```
GitMensch, line 79: it would be reasonable to just check `name[0] == name [size - 1]` and only do the unquote in this case, otherwise leave as-is

cobc/ppparse.y
```diff
@@ -72,15 +73,36 @@ static enum cb_directive_action		current_cmd = PLEX_ACT_IF;
...
-		name[strlen (name) - 1] = 0;
 	}
 	return name;
 }
+#define fix_filename(filename) unquote_alnum (filename)
+
+static char *
+literal_token (char *t, int allow_spaces)
+{
+	if (t[0] == '\'' || t[0] == '\"') {
```
GitMensch, line 92: I _guess_ that should be `t[0] == '"'`?

cobc/ppparse.y
```diff
@@ -1725,6 +1726,23 @@ lead_trail:
   }
 ;
 
+unquoted_literal:
```
GitMensch, line 1729: I suggest that you directly commit the "unquoted literal" part, along with unquote_alnum and correct whitespace trimming with its Changelog entry to svn, leaving the _actual_ change which this PR is about in a more clean state.

GitMensch (changes requested):
This looks good in most places. As noted: the cleanup part of it can go directly to svn.
Question: Do you have a chance to look at some dialect docs to do the verification for the config entries?
Doing so is not mandatory to commit this as it didn't worked before, likely resulting in a preprarser error message and now don't work with a cleaner error message, but it would be "nice".

cobc/ppparse.y
nberth: I am currently splitting this PR into two commits for the purpose of separating code factorization from the support for literal operands.
About the latter, as of now I failed to locate relevant documentations for Realia, BS2000 and RM-COBOL. For the others I could only trust https://www.microfocus.com/documentation/visual-cobol/vc60/VS2019/HRLHLHCOMP02.html .

GitMensch: > About the latter, as of now I failed to locate relevant documentations for Realia, BS2000 and RM-COBOL. For the others I could only trust https://www.microfocus.com/documentation/visual-cobol/vc60/VS2019/HRLHLHCOMP02.html .

That's fine, thank you for taking it that far.

> I am currently splitting this PR into two commits for the purpose of separating code factorization from the support for literal operands.

Thank you, as noted: that first commit could go to svn directly when you consider it clear (or I could have a look at both commits here and you check it in "upstream" with two commits when this PR is review-approved).

nberth: Ok I've just pushed the two commits here.

GitMensch:
Both commits look good - as long as they pass `make checkall` - please commit (also in svn as 2 commits, please).
Well done.

PR #33, opened 2022-03-29
[GCOS] Add support for CONTROL DIVISION (with SUBSTITUTION SECTION only)

nberth:


cobc/ppparse.y
```diff
@@ -721,13 +724,55 @@ ppparse_clear_vars (const struct cb_define_struct *p)
...
+
+control_division_with_replace:
+ /* The second period could be optional. */
+  control_division_prefix SUBSTITUTION_SECTION DOT replace_statement
+;
+
+control_division_prefix:
+  CONTROL_DIVISION DOT
+  {
+	  if (!ppparse_verify (cb_control_division, _("Control division"))) {
```
GitMensch, line 755: no need for a translation here, but all upper-case --> `ppparse_verify (cb_control_division, "CONTROL DIVISION")`

config/default.conf
```diff
@@ -204,6 +204,7 @@ dpc-in-data:			xml
 
 alter-statement:			obsolete
 comment-paragraphs:			obsolete
+control-division:			obsolete
```
GitMensch, line 207: as it is very old and kind of "exotic" - use "unconformable" here and "+obsolete" in lax.inc

cobc/ppparse.y
GitMensch: As this cannot be part of any other construct just drop `YYERROR` here, which then should still have the possible error from cb_verify depending on the configuration, but still consume the tokens so not fail parsing later (which is currently seen in the test).

nberth: Ok thanks for the hint!

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2022-01-27  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```
GitMensch, line 1: That's not what we use the ChangeLog file on the root for.
If it would be more important it should go to the NEWS file, but for now only the new dialect should go in there "in general".

cobc/pplex.l
```diff
@@ -444,7 +446,41 @@ DEFNUM_LITERAL	[+-]?[0-9]*[\.]*[0-9]+
 	return CONTROL_STATEMENT;
 }
 
-("ID"|"IDENTIFICATION")[ ,;\n]+"DIVISION" {
+"CONTROL"[ ,;\n]+"DIVISION" {
```
GitMensch, line 449: Wouldn't it be reasonable to do the validation of the dialect option here?

cobc/ppparse.y
```diff
@@ -721,13 +724,53 @@ ppparse_clear_vars (const struct cb_define_struct *p)
 
 %%
 
+program_structure:
+  program_with_control_division
```
GitMensch, line 728: can you move the `control_division_prefix` here before `` program_with_control_division`?

tests/ChangeLog
```diff
@@ -1,3 +1,7 @@
+2022-01-27  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* testsuite.src/syn_control.at: new testsuite for CONTROL DIVISION
```
GitMensch, line 3: Let's move this to an existing syn group, if we have support for many `CONTROL DIVISION` entries we can still move it out to a new one.

config/xopen.conf
```diff
@@ -205,6 +205,7 @@ dpc-in-data:			xml
 
 alter-statement:			warning		# should not be used ...
 comment-paragraphs:			warning		# should not be used ...
+control-division:			unconformable	# not verified yet
```
GitMensch, line 208: definitely unconformable; if possible have a look at the dialect's documentation if new entries are supported somewhere; it _think_ in this case it isn't.

GitMensch (changes requested):
I guess that push was an update from upstream, correct?
Are you currently working on those review comments?

nberth:
@GitMensch I think that one can be reviewed.

GitMensch (approved):
Approved "in general" - can go upstream when these review-comments were handled (either direct or with a short ping and a re-check - as you like).

GitMensch:
Did you spot any `CONTROL DIVISION` entries other than `SUBSTITUTION SECTION` that could be added (or should at least be syntax-parsed)?

nberth:
Indeed, it turns out by inspecting some GCOS programs I've seen an additional `DEFAULT SECTION` as part of `CONTROL DIVISION`.
The documentation lists several statements available in there, yet I opted to parse the only two I've actually witnessed (`DISPLAY` and `ACCEPT`); they are anyways quite exotic so I guess only parsing them and  raising "not implemented" warnings is ok for now.

tests/testsuite.src/syn_misc.at
```diff
@@ -8453,8 +8453,71 @@ AT_DATA([prog.cob], [
...
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       01 VAR PIC X(2) VALUE "OK".
+       PROCEDURE        DIVISION.
+           DISPLAY VAR NO ADVANCING
+           END-DISPLAY.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE -fcontrol-division=ok prog.cob], [0], [], [])
```
GitMensch, line 8480: please use a single testcase for this wit prog.ob and prog2.cob (or empty.cob)

GitMensch:
I've just rechecked the changes and the docs.
The change _looks good_ and the `DEFAULT SECTION` should **definitely** be added, ... but actually I think that's the wrong place.

Reasoning: the `DEFAULT SECTION` is actually what the standard 2002 added with the `OPTIONS` paragraph and it should be matched against reserved words.
To do so: `ppecho()` the `CONTROL DIVISION` (no need to do that for `SUBSTITUTION SECTION`) to get it to the scanner. There match it returning the same token as it is in ppparse.y (add it to parser.y where it is an optional addition before the `IDENTIFICATION DIVISION` and only contains the optional `DEFAULT SECTION` (likely no need to add that or the contents into scanner.l at all, no need to keep that in pplex.l or pparse.y).

This also helps to keep the preprocessor "cleaner" and actually allows us later to implement part of the options we may see there - in a context where reserved-words are used for the tokenization and we have all the other nice features at hand.

Note: you may consider to move the cb_verify for `CONTROL DIVISION` into parser.y, too.

GitMensch:
You may want to consider splitting this - so checking in the original point of the PR without the `DEFAULT SECTION` parts; then with a secondary PR add the additional `ppecho()` to bring the `CONTROL DIVISION` into the parser and add `DEFAULT SECTION` there.

nberth:
> I've just rechecked the changes and the docs. The change _looks good_ and the `DEFAULT SECTION` should **definitely** be added, ... but actually I think that's the wrong place.
> 
> Reasoning: the `DEFAULT SECTION` is actually what the standard 2002 added with the `OPTIONS` paragraph and it should be matched against reserved words. To do so: `ppecho()` the `CONTROL DIVISION` (no need to do that for `SUBSTITUTION SECTION`) to get it to the scanner. There match it returning the same token as it is in ppparse.y (add it to parser.y where it is an optional addition before the `IDENTIFICATION DIVISION` and only contains the optional `DEFAULT SECTION` (likely no need to add that or the contents into scanner.l at all, no need to keep that in pplex.l or pparse.y).
> 
> This also helps to keep the preprocessor "cleaner" and actually allows us later to implement part of the options we may see there - in a context where reserved-words are used for the tokenization and we have all the other nice features at hand.

When I saw the menmonics and reserved words in statements of the `DEFAULT SECTION`, I indeed thought its place would need to end up in the actual parser. Yet, I wanted to just "parse/ignore" the section and make that transition later as I was unaware this corresponded to a dedicated paragraph in the 2002 standard.

> Note: you may consider to move the cb_verify for `CONTROL DIVISION` into parser.y, too.

I think it is cleaner if I do that move in the PR where the `DEFAULT SECTION` is introduced, so the current one does not change `parse.y` at all.

GitMensch:
> I think it is cleaner if I do that move in the PR where the `DEFAULT SECTION` is introduced, so the current one does not change `parser.y` at all.

Sure. So what do you think about checking this in without any of the the `DEFAULT SECTION`, then do a second PR that introduces the `DEFAULT SECTION` - completely in parser.y?

cobc/pplex.l
```diff
@@ -1317,10 +1354,19 @@ ppcopy (const char *name, const char *lib, struct cb_replace_list *replace_list)
 	return -1;
 }
 
+/* `newline_count` is one line ahead when the two functions below are called
+   (from `ppparse.y`). */
+
+unsigned int
+ppparse_verify (const enum cb_support tag, const char *feature)
```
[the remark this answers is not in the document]
GitMensch: Still unused, so please remove this and the changelog entry for it.

GitMensch (approved):
Approved (with the unused function removed).

GitMensch:
> Approved (with the unused function removed).

... somehow I've got an old version shown of these files , but not of the rest ?!?

Going to files: everything is clean, please commit. I'd like to see a follow-up PR for the `DEFAULT SECTION` :-)

nberth:
Now upstream.

PR #34, opened 2022-03-29
[GCOS] Allow the WITH CONVERSION clause right after DISPLAY

ddeclerck:


GitMensch:
I suggest to implement this completely by creating an intermediate field for this according to the GCOS rules, then do an implicit move and then display the temporary field.
... and obviously "pretty-display" should be set to "false" in gcos-strict.cfg (haven't checked its current value).

ddeclerck:
I'm not sure how quickly I could implement this *correctly* (haven't dug too far in GnuCobol's source code yet), considering our time/budget constraints. For now, our customer just needs this to parse, even if the output is not exactly as expected.

GitMensch:
Then please use CB_PENDING for this now.

cobc/parser.y
```diff
@@ -12101,6 +12101,11 @@ display_body:
 	cb_emit_command_line ($1);
   }
 | screen_or_device_display _common_exception_phrases
+| _with CONVERSION screen_or_device_display _common_exception_phrases
+  {
+	check_repeated ("CONVERSION", SYN_CLAUSE_8, &check_duplicate);
```
GitMensch, line 12106: There is no way to have CONVERSION repeated here, is it? If it isn't drop the check.
Explanation: it is used in cases where a single statement can have multiple modifiers - to ensure no modifier is used more than once and to ensure no conflicting modifier is used.

ddeclerck: Indeed. Fixed. ;)

GitMensch:
To finish this PR please add the missing syntax check to ensure that `WITH CONVERSION` can actually be parsed (I see no reason that it doesn't, but the added test ensures the test coverage does not go down and that the syntax will be supported in the future, too.

... and you may find the test not working in which case you'd have to adjust cobc/reserved.c and add `CB_CS_DISPLAY` to the context check.

cobc/parser.y
```diff
@@ -12101,6 +12101,10 @@ display_body:
 	cb_emit_command_line ($1);
   }
 | screen_or_device_display _common_exception_phrases
+| _with CONVERSION screen_or_device_display _common_exception_phrases
+  {
+	CB_PENDING ("WITH CONVERSION");
```
GitMensch, line 12106: Now after seeing the testcase output (which showed itself to be reasonable by highlighting the c/s issue): 

Please change that to `CB_PENDING ("DISPLAY WITH CONVERSION");` and the testcase accordingly (as the current one looks a bit ugly).

This PR can be seen as "solved" then (sorry for not posting the complete relevant comments altogether, but I hope there was a bit of learning involved in which case the time was not lost).

ddeclerck: > Please change that to `CB_PENDING ("DISPLAY WITH CONVERSION");` and the testcase accordingly (as the current one looks a bit ugly).

Done. However I'm not sure what you mean about the test case looking "a bit ugly" : could you clarify, so that I can make better testcases in the future ?

> sorry for not posting the complete relevant comments altogether, but I hope there was a bit of learning involved

Yes there was ;)

ddeclerck:
Alright, thanks.

nberth:

GitMensch:
Feel free to commit this as-is to svn at any time.

PR #35, opened 2022-03-29
Area A enforcement

nberth:
In fixed form programs, section and paragraph headers must start in
area A, and in the GCOS dialect this form is exploited to allow
paragraphs that do not terminate with a period in procedure divisions.
Up to now, GNUCobol discarded any information about the use of area A,
since every supported dialect requires paragraphs to terminate with a
period.

To support the GCOS dialect, the suggested changes modify the
pre-processor to insert special markers `#area_a` before every word
that starts in area A when a new option `rely-on-area-a` is set.  The
parser can now make use of these tokens to detect section and
paragraph terminations within procedure divisions.


nberth:
With this change, passing `-frely-on-area-a` may allow proper parsing of https://sourceforge.net/p/gnucobol/feature-requests/309 and https://sourceforge.net/p/gnucobol/feature-requests/365/ (without warnings though).

GitMensch:
The general idea is good and this is indeed solving https://sourceforge.net/p/gnucobol/feature-requests/309.
The main requested change would be the following:
* always output area a tokens in non-free mode (not depending on dialect), renaming them to `TOK_AREA_A`
* use area a tokens in the parser as optional after `TOK_DOT`, otherwise for alternate paths
* in the alternate paths check for relaxed-syntax-checks and depending on this raise a warning or an error (which still would provide the currently missing error recovery)
* use of this token would be in the places already tackled in this PR + variable definitions

GitMensch:
In any case we should add area a/b in a separate PR beforehand, which would also fix https://sourceforge.net/p/gnucobol/feature-requests/360/ and https://sourceforge.net/p/gnucobol/bugs/836/.

GitMensch:
> However, in several cases some lines start with a period in Area A. I'm wondering whether this should indeed be allowed.

Can you point out where this is, please?

BTW: if that is based on current 3.x: CI fails (2 times, funnily related to GCOS [but that's just a co-incidence]), I have a working version locally and try to fix something that was designed but never worked, will try to commit in the next 30 minutes.

nberth:
For instance in the `syn_definition.at`, line 2446 (or thereabout), test `APPLY COMMIT clause`.

nberth:
> Does GCOS explicit forbid that?

I should have a test result for that whenever one of my colleagues can have access to a GCOS system.

For the approach, a filtering based on the contents of the area is indeed what I was attempting. I wanted to check before going further as I find both implementing the changes quite tricky and fixing the test cases quite tedious.

GitMensch:
Just a note to

> BTW: if that is based on current 3.x: CI fails (2 times, funnily related to GCOS [but that's just a co-incidence]), I have a working version locally and try to fix something that was designed but never worked, will try to commit in the next 30 minutes.

I've fixed the fixing fix, and am finished to break the CI now... back to my original work: numeric performance optimization tests, then commit that.

nberth:
> Does GCOS explicit forbid that?

As it appears yes in the `PROCEDURE DIVISION`. A dot in area A gives:
```
         7         .
                   1
###  1  1-23  This word in Area A is not a user-defined word.
```
In other parts, like `CONTROL DIVISION`, they seem to be allowed.

GitMensch:
Thank you for the test - so what about implementing that as a rule "if there is an area a/b (and not for terminal mode), then tokens in area a must be a user-defined word?
To not possibly break a bunch of existing programs this means that `-std=default` will stay with COBOL2002+ -  without explicit area a/b. Nice side effect: all existing test cases will still pass unchanged (and hooray for the testsuite to have shown this before releasing that change).

nberth:
Margin B is needed to detect the end of comment paragraphs, so I think it cannot be dropped completely from the formats.
This could be an additional "Area A/B-sensitivity" property for non-free formats.

nberth:
BTW the full test suite seems to pass now, with only some parts that required some adjustment in spacing.
Do you think changing the "default" behavior to be Area-sensitive could break many existing programs?
If so, maybe a more suitable approach is to have a dedicated flag and make it so the default only raises a new warning.

GitMensch:
> Do you think changing the "default" behavior to be Area-sensitive could break many existing programs?

Yes - as many people write in COBOL2002+ fixed-form reference-format, which is what -std=default effectively had done since the beginning.

> If so, maybe a more suitable approach is to have a dedicated flag and make it so the default only raises a new warning.

I think neither default nor cobol2002+ should raise a warning; gcos+cobol85 should raise an error, mf+acu a warning (not sure for the others).

Lexing with all those dialects and differences _is_ hard (I'd not have tackled that myself). When people complain about compiler xyz has this different / better they commonly forgot that those compilers only have a limited set while GnuCOBOL supports more dialects than anyone else... (in this case: only COBOL2002+ complete, the rest partial but commonly enough, with the potential downside of "needs adjustments that will work both in the old compiler and GnuCOBOL").

nberth:
@GitMensch So here's what I've come up to.

I've tried to extend the approach to deal with missing periods to the `DATA DIVISION`, but it turns out even the more "manageable" sub-sections (like `WORKING-STORAGE` or `REPORT`) in the COBOL85 NIST tests have sentences that spill in Area A. In the end I'd rather leave that division as is and avoid messing with the grammar too much in there.

tests/testsuite.src/syn_misc.at
```diff
@@ -4542,6 +4542,204 @@ prog.cob:12: warning: ignoring redundant .
...
+       COPY "sec-2.inc".
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+[sec-1.inc:2: error: 'SEC-1' is not defined
+sec-1.inc:2: error: syntax error, unexpected SECTION
+sec-2.inc:2: error: 'SEC-2' is not defined
+sec-2.inc:2: error: syntax error, unexpected SECTION
+])
+AT_CHECK([$COMPILE -floose-area-a=skip -frelax-syntax-checks prog.cob], [0], [],
```
GitMensch, line 4643: looks like we always test both switches together - but one should be enough for the test, no?

tests/testsuite.src/run_extensions.at
```diff
@@ -1629,47 +1629,47 @@ AT_DATA([prog.cob], [
        PROCEDURE        DIVISION.
 *     *    MOVE NON-INTEGER TO ALPHA-NUMERIC --> ignore Decimal Point!
        S-01.
-          MOVE SPACES TO ALPHA-FIELD.
-          MOVE INTEGER TO ALPHA-FIELD.
-          DISPLAY  ALPHA-FIELD NO ADVANCING
-          END-DISPLAY.
+           MOVE SPACES TO ALPHA-FIELD.
```
GitMensch, line 1632: please ensure one of those "code in column 10" is in the syntax tests below, compiled with -std=default/cobol2002/mf (should compile without warnings) and cobol85/gcos (should likely fail compilation) - for the later MF (with dialect ANS85) errors with

```
Word starts or is continued in wrong area of source line
```

Something similar would be nice for area-a enforcement

GitMensch (commented):
some more things to do (sadly: unlikely to be reviewed in depth very soon, but you may want to have a look at these comments and possibly adjust already)

In any case this changes the behavior for all dialects but default/cobol2002/cobol2014/mf - so needs a note in the NEWS entry.

cobc/flag.def
```diff
@@ -141,6 +141,10 @@ CB_FLAG (cb_flag_traceall, 1, "traceall",
 CB_FLAG (cb_flag_syntax_only, 1, "syntax-only",
 	_("  -fsyntax-only         syntax error checking only; don't emit any output"))
 
+CB_FLAG (cb_flag_preproc_area_a, 1, "preproc-area-a",
```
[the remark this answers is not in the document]
nberth: Do you mean we should always output the `#` whenever not in free format?

cobc/config.def
```diff
@@ -369,6 +369,9 @@ CB_CONFIG_SUPPORT (cb_record_delim_with_fixed_recs, "record-delim-with-fixed-rec
 CB_CONFIG_SUPPORT (cb_missing_statement, "missing-statement",
 	_("missing statement (e.g. empty IF / PERFORM)"))
 
+CB_CONFIG_SUPPORT (cb_loose_area_a, "loose-area-a",
```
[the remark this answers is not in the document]
nberth: Maybe I should go straight for `non-word-in-area-a` or something similar then.

cobc/scanner.l
```diff
@@ -305,6 +307,26 @@ static void *	copy_literal (cb_tree l);
...
+	/* <*>\n"#AREA_A"[[:space:]]*[ ]/[^[:alnum:][:space:]] { */
+	count_lines (yytext + 9); /* skip "\n#area_a\n" */
+	if (cobc_in_procedure) {
+		(void) cb_verify (cb_loose_area_a, _("punctuation in Area A"));
+	}
+}
+
+<*>{AREA_A} {
+	if (cobc_in_procedure) {
+		last_token_is_dot = 1; /* always assume we've just seen a dot. */
```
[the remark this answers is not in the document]
nberth: Well… as of now any area A should either follow a dot or a "missing dot". `last_token_is_dot` is used to classify integer literals as they may be section or paragraph names; so maybe it may just need to be renamed?

tests/testsuite.src/syn_misc.at
```diff
@@ -4542,6 +4542,204 @@ prog.cob:12: warning: ignoring redundant .
...
+	   END-DISPLAY.
+           GOBACK.
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+[prog.cob:10: error: 'SEC-1' is not defined
+prog.cob:10: error: syntax error, unexpected SECTION
+])
+AT_CHECK([$COMPILE -floose-area-a=skip -frelax-syntax-checks prog.cob], [0], [],
+[prog.cob:10: warning: missing '.' above this line
```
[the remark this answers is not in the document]
nberth: I though the relaxed syntax checks would be enough to filter. A `missing-period` config flag is probably in order then.

cobc/config.def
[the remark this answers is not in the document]
GitMensch: That is something I have an idea of.

cobc/flag.def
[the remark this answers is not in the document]
GitMensch: No, depending on the -fformat (with / without area a enforcement) this should be inserted and later on checked.
Note: if we need another -fformat option to distinguish cobol85 and fixed - then it would be good to add that soon, even if they currently are identical (documenting the first as "with area a enforcement [currently not done]", or similar).

cobc/scanner.l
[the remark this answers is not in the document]
GitMensch: Renaming that (before the actual PR) would be possible, bot "period, then number" is also a level number. Not sure about the name yet.

tests/testsuite.src/syn_misc.at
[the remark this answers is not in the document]
GitMensch: Yes, CB_SUPPORT missing-period sounds good.
Please try to split the test cases (or at least the test runs) between "loose" and "period".

cobc/parser.y
```diff
@@ -10346,19 +10373,21 @@ procedure:
 	/* check_unreached = 0; */
 	cb_end_statement();
   }
+  _dot_or_else_area_a
```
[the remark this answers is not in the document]
nberth: I've combined this with introduction of COBOL85 format in #56 .

cobc/scanner.l
```diff
@@ -304,6 +310,21 @@ static void *	copy_literal (cb_tree l);
 	}
 }
 
+<*>^{AREA_A}[ ]*/"." {
+	count_lines (yytext + 9); /* skip "\n#area_a\n" */
+	if (cobc_in_procedure) {
+		(void) cb_verify (cb_non_word_in_area_a,
```
GitMensch, line 316: This is definitely mis-leading. From the standard:

> 5.2.7    Punctuation
The separators comma and semicolon may be used anywhere the separator space is used in general formats and other  syntactic specifications. In the compilation group, these separators are interchangeable.
The separator period, when specified in a general format, is required when that format is used.

Therefore we don't use "punctuation" in any msgid.
From the definitions in `PROCEDURE DIVISION`:

> A paragraph consists of a paragraph-name followed by a separator period and by zero, one, or more successive sentences or, if the paragraph-name is omitted, one or more successive sentences following the procedure division header or a section 
header.

So when we want to reference a _new_ procedure-name with a missing separator period (not here) before or after, we can name it `procedure-name` or, if we know what we have "paragraph header" / "section header".

The main question: Is there any use of explicit raising an error / warning with a separator period here?

cobc/config.def
```diff
@@ -383,6 +383,12 @@ CB_CONFIG_SUPPORT (cb_record_delim_with_fixed_recs, "record-delim-with-fixed-rec
 CB_CONFIG_SUPPORT (cb_missing_statement, "missing-statement",
 	_("missing statement (e.g. empty IF / PERFORM)"))
 
+CB_CONFIG_SUPPORT (cb_missing_period, "missing-period",
+	_("missing period in PROCEDURE DIVISION (when reference format supports Area A enforcement)"))
+
+CB_CONFIG_SUPPORT (cb_non_word_in_area_a, "non-word-in-area-a",
```
GitMensch, line 389: What is a "non user-defined word"? Is this is about "statements" or "reserved words" or ???

GitMensch (commented):
I think this is likely near to "as clean as possible".
Somehow I am concerned about performance implications of the scanning/parsing part; I hope to find the time soon to do an actual check using the testsuite.

Please see the comments for some things I've stumbled over.


The entry from the Fujitsu docs on Area A and Area B seem to be correct for COBOL85 fixed-form reference-format.
Does this PR have checks for all of those? If not, which are missing?

Words that must start in Area A:
* all procedure-names / headers (including DIVISIONs, SECTIONs, paragraphs)
* level indicators (FD, SD or RD)
* level 1 + 77 (+ as an extension, level 78)
* [`END`] `DECLARATIVES`
* `END PROGRAM` header

Words that must start in Area B:
* all statements but `COPY` and `REPLACE` (which can start in both areas)
* comment entries (in `IDENTIFICATION DIVISION`)
* (level numbers other than 1/77/78 _can_ start in both areas)

GitMensch:
As it looks like this should be only a minor missing piece - could you please add the `AREACHECK` directive (=wouldn't it be enough to have that toggle the dialect option)?

tests/testsuite.src/syn_misc.at
```diff
@@ -4549,6 +4549,250 @@ prog.cob:12: warning: ignoring redundant .
...
+prog.cob: in section 'SEC-1':
+prog.cob:17: error: statement in Area A used
+prog.cob:18: error: punctuation in Area A used
+])
+
+# That error is hard to bypass: -fformat=fixed is needed (and loss of
+# support for missing periods).
+AT_CHECK([$COMPILE_ONLY -std=cobol85 -fnon-word-in-area-a=ok prog.cob], [1], [],
+[prog.cob: in paragraph 'MAIN':
+prog.cob:13: error: syntax error, unexpected END-IF
```
nberth, line 4722: @GitMensch In my mind this right here is currently the main issue with this PR, and I think it is hard to overcome with the current approach. Essentially, inserting AREA_A tokens to be able to implement the checks makes us unable to deal with such tokens appearing in the middle of statements, in which case we end up with syntax errors.
But maybe you're ok with this limitation…

nberth (commented):
A request for comment before pursuing on this PR…

nberth:
@GitMensch This may now be getting somewhere. I'll see to add the `AREACHECK` directive, and area A enforcement should be done for `PROCEDURE DIVISION`.

tests/testsuite.src/syn_misc.at
```diff
@@ -5542,6 +5542,7 @@ prog.cob:6: error: WITH ... LINKAGE does not conform to COBOL 85
 prog.cob:8: error: CALL-/ENTRY-CONVENTION does not conform to COBOL 85
 prog.cob:10: error: WITH ... LINKAGE does not conform to COBOL 85
 prog.cob:10: error: CALL-CONVENTION and WITH LINKAGE are mutually exclusive
+prog.cob:12: warning: 'GOBACK' should be in Area A
```
GitMensch, line 5545: Wow, many people would not have recognized that this is a paragraph - another good reason for the warning!

cobc/pplex.l
```diff
@@ -260,82 +260,105 @@ MAYBE_AREA_A	[ ]?#?
...
 }
 
-^[ ]*">>"[ ]*\n		{
+^{MAYBE_AREA_A}[ ]*">>"[ ]?"AREACHECK"	|
+^{MAYBE_AREA_A}[ ]*">>"[ ]?"AREA-CHECK"	{
+	if (cobc_has_areacheck ()) {
+		fprintf (ppout, "#AREACHECK\n");
+	} else {
+		cb_plex_warning (COBC_WARN_FILLER, newline_count,
+				 _("ignoring AREACHECK directive"));
```
GitMensch, line 337: to make this more useful for the users and nicer for the translators lets switch here and below to something like

```c
	cb_plex_warning (COBC_WARN_FILLER, newline_count,
				 _("ignoring %s directive because of %s"), "AREACHECK", "-fformat=fixed/free");
```

--> a bit more detail and still one msgid added less

tests/testsuite.src/syn_misc.at
```diff
@@ -9075,3 +9075,60 @@ AT_CHECK([$COMPILE_ONLY -fpicture-l=warning prog.cob], [0], [],
 ])
 
 AT_CLEANUP
+
+# " (close unclosed quotation above)
```
GitMensch, line 9079: Where's this quotation? We have that in other places, maybe copy+past issue?

tests/testsuite.src/syn_misc.at
```diff
@@ -9075,3 +9075,60 @@ AT_CHECK([$COMPILE_ONLY -fpicture-l=warning prog.cob], [0], [],
...
+AT_SETUP([AREACHECK / NOAREACHECK directives])
+AT_KEYWORDS([areacheck])
```
GitMensch, line 9081: this duplicates the keyword that gets in from `AT_SETUP`, but "directive" (= singular) would be useful.

GitMensch (changes requested):
Thanks for following the route down and for these nice changes! The directive is actually wrong, but should be easy to fix.

(I hope to find some time tomorrow to check the collation/ebcdic issue in exchange)

NEWS
```diff
@@ -53,7 +53,8 @@ NEWS - user visible changes				-*- outline -*-
    These formats come with Area A enforcement in the PROCEDURE
    DIVISION, where paragraph and section names should start in Area A
    (i.e, before margin B).  The underlying checks can be enabled or
-   disabled using a new flag `areacheck`.
+   disabled using a new flag `areacheck`, or with `>>AREACHECK` and
```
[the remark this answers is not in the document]
nberth: The reference format is already a dialect option, and `-f(no-)areacheck` has an effect only based on the format. Are there dialects with format not in {free,fixed}, and for which area A enforcement should not be enabled by default?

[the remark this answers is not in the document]
nberth: Hum I see: you'd want the `areacheck` dialect option to be about support for `>>AREACHECK` directives and the likes? This may prevent one from being able to completely disable such checks without manually adding the directives, for instance on `cobol85` source, right? (or is this what `-fareacheck=ignore` would be about?).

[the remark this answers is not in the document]
GitMensch: > Hum I see: you'd want the areacheck dialect option to be about support for `>>AREACHECK` directives and the likes? 

No!

Dialect comes first, then command line options, then directives (`$SET [NO]AREACHECK` being ignored when there is no area as you've implemented is, is of course correct).

The point of a dialect option would be to have it default to "on" for some of dialects (ibm and mvs, possibly gcos?).
As you've noted a boolean dialect option would be reasonable.

cobc/parser.y
```diff
@@ -13032,13 +13033,20 @@ enable_statement:
 
 entry_statement:
   ENTRY
+  {
```
GitMensch, line 13036: These are actually two different statements with the second being highly unportable and not recommend for common use; I'd prefer those to be kept separate.

nberth: The issue here is we need to have a call to `check_non_area_a` right after reading `STOP`, without any lookahead. I admit this is kind of a fragile approach, but I did not find much better (yet).
Another means might be to find a way to attach an `in_area_a` to the `STOP` token… not sure this is straightforward.

GitMensch: Hm, that's still the lookahead you want to prevent, isn't it?

nberth: Yes I meant `ENTRY` here (but `STOP` is a similar statement that needed rework to avoid the lookahead).
In the case above we still have the lookahead. As an alternative I'll try with an `entry_word` non-terminal…

nberth: Using dedicated non-terminals seem to circumvent the lookahead issue in a "cleaner" way. I'll push some changes in that direction.

nberth:
@GitMensch I think with all periods that have now become optional, this PR starts to cover a bit more than just area A enforcement. Maybe it's time to stop and consider performance implications…

GitMensch:
The optional dots need some test cases.

In the meanwhile I'll (after lunch) drop a note how you can test performance implications on cobc (via `perf`).

GitMensch:
I'll do the performance check with comparing the gcos4gnucobol branch with the optinal-dots branch and report back.

GitMensch:
Tested in the following way ()

```sh
cd /tmp
git clone https://github.com/OCamlPro/gnucobol.git --branch gcos4gnucobol-3.x gcos4gnucobol-3.x
gcos4gnucobol-3.x/autogen.sh
mkdir build_ocamlpro
pushd build_ocamlpro
mkdir gcos
../../gcos4gnucobol-3.x/configure --enable-cobc-internal-checks --enable-debug & make -j8
# minimal verification:  no new warnings during build

vim tests/atlocal   #  within the  "PERFSUFFIX" test: only enable PERF and COMPILE_ONLY, then write and exit
make PERFSUFFIX=old check TESTSUITEFLAGS="--jobs=12"
 
popd

git clone https://github.com/nberth/gnucobol.git --branch optinal-dots gnucobol-3x-optinal-dots
gnucobol-3x-optinal-dots/autogen.sh
pushd build_ocamlpro
mkdir dots
../../gnucobol-3x-optinal-dots/configure --enable-cobc-internal-checks --enable-debug & make -j8
# minimal verification:  no new warnings during build

vim tests/atlocal   #  within the  "PERFSUFFIX" test: only enable PERF and COMPILE_ONLY, then write and exit
make PERFSUFFIX=new check TESTSUITEFLAGS="--jobs=12"
```

Then inspect perf_old.log and perf_new.log for differences (after aggregation):

~~~
Aggragation for perf_old.log:

        instructions    39.637.065.951

seconds time elapsed           17.6608
        seconds user           16.9826
         seconds sys            2.5485


Aggragation for perf_new.log:

        instructions    40.225.748.835

seconds time elapsed           18.0532
        seconds user           17.1179
         seconds sys            2.8909
~~~

The "new" has some minor tests more, but in general there is no "big issue" to see - so performance for this change (which does bring additional features and better error handling) is fine.

Where do you see the state of this PR (feature, documentation, and test-wise)?

nberth:
Well, I'm relieved to see the performances are still quite okay.

Making some of the dots optional can have bad consequences on some syntax error messages, in the sense that expected tokens could be not so useful, and that the parser can now try do progress more but output completely irrelevant messages some lines below missing dots. To correct this, some more test files with "relevant" missing dots need to be written, and decisions be made w.r.t what `cobc` reports. However, I'm not in a position to judge what periods are "more important" than others in COBOL.

About documentation, for now only the `NEWS` file mentions Area A enforcement I think. Maybe I can add some notes on that into `gnucobol.texi` as well.

NEWS
```diff
@@ -55,6 +55,43 @@ NEWS - user visible changes				-*- outline -*-
    If LS_NULLS is active and invalid data (bad encoded or missing encoding)
    is found io status 71 is returned.
 
+** new -fformat dialect option, and extended SOURCE FORMAT directives, with
```
GitMensch, line 58: this is currently duplicate (merge-conflict, I guess), it belongs below "change to options", not "to change to existing programs"

nberth: Ouch! Thanks for catching that!

NEWS
```diff
@@ -76,13 +76,38 @@ NEWS - user visible changes				-*- outline -*-
...
+   `>>AREACHECK` and `>>NOAREACHECK` compiler directives.
+
+   Area A enforcement additionally enables detection of missing
+   periods before paragraph and section names, configurable with
+   option `missing-period`.
+
+   As a result of reference format being a dialect option, Area A
+   enforcement is available for the following dialects:
+
+   BS2000 COBOL
```
GitMensch, line 101: BS2000 explicit lists the areacheck (without option to disable it) in its docs; I've found nothing about Realia (I don't have a Language Reference and the User Reference have no entry for reference-format or rules for defining sections); RM-COBOL Language Reference also list this explicit check, so at least RM and BS2000 should have that enabled by default.

I'd actually go the easy way and say it is enabled by default (of course also adjust in all three of these missing) and only note the dialects, comma-separated (maybe even as their "std" abbreviations, but this may look bad, not tested).
If someone reports that this was an error for Realia we can adjust this later.

nberth:
@GitMensch I'm not sure I'll be able to do much more in the coming days, so I'd consider this done.

GitMensch (approved):
Please doe those minor additions - and push that upstream.
Ideally also drop a note in the FR and change its status via Edit to 3.2 + closed (I'm assigning you to the ticket-mod group so you have that option).

cobc/parser.y
```diff
@@ -3201,6 +3233,9 @@ set_record_size (cb_tree min, cb_tree max)
 %token YYYYMMDD
 %token ZERO
 
+%token LEVEL_NUMBER_IN_AREA_A	"level-number (in Area A when relevant)"
```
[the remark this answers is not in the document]
nberth: I did that but bison complains about duplicate identifier strings. Also I noted recent versions of  bison have some internationalization support (but alas, versions not even available in the CI here).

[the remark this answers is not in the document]
nberth: It's only warnings though…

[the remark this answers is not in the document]
nberth: … but then it uses `LEVEL_NUMBER_IN_AREA_A` in error messages.
All this could be improved by redefining the syntax error printing function to intercept those tokens, but that's a bit too much work for a short-term upstreaming.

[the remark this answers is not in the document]
GitMensch: Agreed, so let's go with the "pure" name + " (Area A)".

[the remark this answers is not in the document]
nberth: Ok, better keep it short indeed.

nberth:
I'll close this now. Note I set milestones to `3.x` on SF as I did not find a way to add `3.2`…

cobc/pplex.l
```diff
@@ -2337,6 +2430,17 @@ start:
...
 		}
+		if (in_area_a) {
+			/* (*area-a*): prepend special marker `#', indicating
+			   the current non-continuation line starts in area A
+			   (see "AREA_A" and "MAYBE_AREA_A"... in lexer rules
+			   above).
+
+			   Assumes `buff' is large enough for one extra char. */
+			memmove (buff + 1, buff, gotcr + 1);
+			buff[1] = '#';
```
GitMensch, line 2441: That is fishy. We move the buffer one byte to the right, then overwrite the first of the moved data by `#`.
This way we effectively insert `#` into the second position. That won't work if we have any code on the first column (like in terminal format).

I also _think_ that this code must be moved _before_ we add the \n in front of the buffer- otherwise the area-a token is not before the code it marks but after the previous code (which makes a difference for where some line numbers go).

I've adjusted both and the testsuite still passes (and my additional tests, too).
@nberth: Do you remember if this may raise issues we miss testcases for? Otherwise I'll commit these changes in about 10 hours.

nberth: Yes that was actually designed at a time when we only had FIXED/COBOL85 format for which the area A check was relevant, and in these cases these issues don't apply.
Issues I can think of and that might not have been tested are about:
- first program line starting in area A (in terminal format indeed — unlike most tests in the testsuite);
- pseudo-text in which the first token starts in area A: these cases are always tricky to handle.

GitMensch:
Coming back to this as I've now recognized that we miss have quite a lot code for this in parser.y and to make this complete need even more.

Currently we only check at the start of statements if we are not in area A, but likely should do so either in general (I'm not sure what _should_ happen with `ADD A TO B`  if there is a line break after `ADD` and the next line starts with `A` in column 8) or at least with every reserved word.

Question: Wouldn't it be reasonable to move most (or even all?) of the checks to the scanner?
I did add a `check_non_area_a_of` function and added it in a bunch of places (for example `END-PERFORM`) but am not sure if I should commit that or if that should be reworked to be handled by the scanner.

nberth:
For sure the current approach is incomplete and does not cover many cases as you noticed (I think I mentioned this limitation a while ago, but maybe I was not explaining this clearly enough…).

I think working that around properly and in full generality is very tricky with the current pull parser. One approach may be to switch to a push parser whose state may be checked in the scanner when issuing tokens; I tried this at some point, but hit some other problems as this messes up parts of the state that is shared between the scanner and the parser. Going the *ad hoc* route and adding more manual checks in some places could work, but only in the short term.

Note that moving area A detection to the scanner may require moving the detection of missing periods there as well…

GitMensch:
Agreed, so I'm going with the adhoc solution.
Would you mind creating an issue here for the possible later move of area-a and missing period check to the scanner, so we have the information needed gathered there?

PR #36, opened 2022-03-29
[GCOS] Add STOP ERROR statement

ddeclerck:


cobc/parser.y
```diff
@@ -15391,6 +15391,15 @@ stop_statement:
 	check_unreached = 1;
 	cobc_cs_check = 0;
   }
+| STOP ERROR /* GCOS */
+  {
+	begin_statement ("STOP ERROR", 0);
+	cb_verify (cb_stop_error, "STOP ERROR");
+	cobc_cs_check = CB_CS_STOP;
+	cb_emit_stop_run (cb_int1);
```
GitMensch, line 15399: The GCOS docs say

> The ERROR phrase has been implemented for debugging purpose. If used, the
execution of the run unit aborts and control is transferred to the operating system. It
is the way, in case of malfunction, to get a dump where all elements (hardware
registers, etc.) remain unchanged.

So instead of a "normal" `STOP RUN RETURNING 1` it should likely trigger the dump feature, shouldn't it?
That would be triggered by any exception, we could add a new fatal one `EC-PROGRAM-STOP-ERROR` or `EC-IMP-STOP-ERROR` and let it be triggered there. This would automatically return with 1 but before trigger exception handlers (both possibly registered ones and the internal one which triggers the dump for all modules compiled with that feature).

ddeclerck: Trying to add a `EC-IMP-STOP-ERROR` exception (set to fatal in exception.def).
Not sure to know how to actually raise it. I use `cob_set_exception (COB_EC_IMP_STOP_ERROR)`, but this only sets the last exception.

Would just using a simple `cob_runtime_error ("STOP ERROR")` instead of using exceptions do it ?

GitMensch: I agree that adding the exception alone won't help. The most reasonable thing to do is to emit `cob_runtime_error ("STOP ERROR");` and `cob_stop_run(1);`.

Note: when there are no more "context sensitive" tokens to follow on (which does not seem the case here) you don't need to set the context-sensitive words check `cobc_cs_check`, so that line can likely be dropped.

GitMensch (approved):
There's only one minor thing to handle, then this PR is "finished":

please move `stop-error-statement` in config.def (and therefore also in testsuite.src/used_binaries.at) and config/* directly after `stop-identifier-statement`.

Well done.

ddeclerck:
> please move `stop-error-statement` in config.def (and therefore also in testsuite.src/used_binaries.at) and config/* directly after `stop-identifier-statement`.

Done. By the way, should the `.conf` files follow this order too ?

GitMensch:
They can use a different order than config.def and may do; but this one should be moved.

> config/* directly after `stop-identifier-statement`

:-)

cobc/config.def
```diff
@@ -222,6 +222,9 @@ CB_CONFIG_SUPPORT (cb_stop_literal_statement, "stop-literal-statement",
 CB_CONFIG_SUPPORT (cb_stop_identifier_statement, "stop-identifier-statement",
 	_("STOP-identifier statement"))
 
+CB_CONFIG_SUPPORT (cb_stop_error, "stop-error",
```
GitMensch, line 225: looks like it should have been "stop-error-statement" :-/

_runs away_

ddeclerck: Indeed. Fixed ;)

nberth:

GitMensch:
@lefessan: What do you think of adding this to gnucobol-3.x in sourceforge after adjusting the last missing piece?
Note: The actual coredump request noted in the GCOS manual for this is configurable, soon - runtime errors will then raise an abort which is passed to the OS which will then coredump.

libcob/common.c
```diff
@@ -2294,6 +2294,13 @@ cob_stop_run (const int status)
 	exit (status);
 }
 
+void
+cob_stop_error (void)
+{
+	cob_runtime_error ("STOP ERROR");
+	cob_stop_run(1);
```
GitMensch, line 2301: should use `cob_hard_failure ()` now instead of `cob_stop_run`

ddeclerck: Done (by Nicolas)

GitMensch (approved):
Can be checked-in directly after the GCOS "base commit".

nberth:
@ddeclerck could you update this PR based on https://github.com/nberth/gnucobol/tree/stop_error so that the diff can be extracted directly for upstreaming?

nberth:
Hi @GitMensch ,
This PR has just been rebased against the upstream versions of GCOS config files. I can commit to the svn if you think it's ok.

GitMensch:
reviewed and ready to be checked in upstream

nberth:
Should be upstream now.

PR #39, opened 2022-05-19
[WIP] Compare Gcos4gnucobol 3.x with Gnucobol-3.x

lefessan:


cobc/field.c
```diff
@@ -584,15 +584,15 @@ cb_build_full_field_reference (struct cb_field* field)
...
 	while (field) {
-		if (!field->flag_filler) {
-			cb_tree rchain = cb_build_reference (field->name);
-			if (ref) {
-				CB_REFERENCE (ref)->chain = rchain;
-			} else {
-				ret = rchain;
-			}
-			ref = rchain;
+
```
GitMensch, line 587: The original issue with the loop was fixed in the meantime, is there still a reason for this change? If yes: should definitely get a testsuite entry.

cobc/ppparse.y
```diff
@@ -1536,7 +1538,7 @@ copy_source:
   TOKEN
   {
 	$$ = fix_filename ($1);
-	if (cb_fold_copy == COB_FOLD_LOWER) {
+	if (cb_fold_copy == COB_FOLD_LOWER || cb_fold_copy == COB_FOLD_BOTH) {
```
GitMensch, line 1541: if both is added it should uppercase first, then lowercase (and actually this is what the commit log, says :-).
...

Is the folding both a GCOS related issue? Please elaborate.

cobc/warning.def
```diff
@@ -114,5 +114,8 @@ CB_ONWARNDEF (cb_warn_dialect, "dialect",
 CB_NOWARNDEF (cb_warn_source_after_code, "dangling-text",
 	_("  -Wdangling-text       warn about source text after program-area"))
 
+CB_ONWARNDEF (cb_missing_newline, "missing-newline",
```
GitMensch, line 117: This one (the separate warning flag) can directly go into 3.x. As copyright assignments are done, you could do it directly, too (I just need your SF user to provide write access).

GitMensch (changes requested):
for common patches please leave the msg catalogues out, they pot file is commonly recreated directly before a release and the updates to the language ones comes from the translation project.

GitMensch:
For 5cc12e16b0dbfe8117332eee790f3866f6bb6fe6 I _think_ we discussed it would be useful to run `make distcheck` to see if it still works and to ideally also provide the resulting tarball as artifact. Did we?

nberth:
@GitMensch This PR is mostly to help us track the current diff between the branch with our pending contributions and the `gnucobol-3.x` branch.
About the CI, I am about to setup a new PR that makes the necessary changes. As there are some issues still with some commits above (fold-copy, already fixed issue, etc), I am also thinking of (force) rebasing the branch `gcos4gnucobol-3.x` to have the initial CI setup and GCOS config files only.

PR #42, opened 2022-05-23
[GCOS] Add support for `L' character in picture strings

nberth:
For now, only runtime support for DISPLAY and MOVE is included.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -785,6 +785,8 @@ chk_field_variable_address (struct cb_field *fld)
 					return 1;
 				}
 #endif
+				/* Skip redefines as they should not impact addresses: */
+				if (p->redefines) continue;
 				if (p->depending || chk_field_variable_size (p)) {
 #if 0	/* only useful with the code above */
 					/* as we have a variable address, all sisters will also;
@@ -1043,7 +1045,6 @@ static void
 output_base (struct cb_field *f, const cob_u32_t no_output)
 {
 	struct cb_field		*f01;
-	struct cb_field		*p;
 	struct base_list	*bl;
 
 	/* LCOV_EXCL_START */
@@ -1111,31 +1112,14 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
 	}
 
 	if (!gen_init_working
+	 && cb_odoslide
 	 && chk_field_variable_address (f)) {
 		if (f01->level == 0
 		 && f01->sister
 		 && strstr (f01->name, " Record")) {	/* Skip to First 01 within FD */
 			f01 = f01->sister;
 		}
-		if (cb_odoslide) {
-			out_odoslide_offset (f01, f);
-		} else {
-			struct cb_field		*v;
-			for (p = f->parent; p; f = f->parent, p = f->parent) {
-				for (p = p->children; p != f; p = p->sister) {
-					v = chk_field_variable_size (p);
-					if (v) {
-						output (" + %d + ", v->offset - p->offset);
-						if (v->size != 1) {
-							output ("%d * ", v->size);
-						}
-						output_integer (v->depending);
-					} else {
-						output (" + %d", p->size * p->occurs_max);
-					}
-				}
-			}
-		}
+		out_odoslide_offset (f01, f);
 	} else if (f->offset > 0) {
 		output (" + %d", f->offset);
 	}
@@ -1621,10 +1605,9 @@ output_attr (const cb_tree x)
 			case COB_TYPE_GROUP:
 			case COB_TYPE_ALPHANUMERIC:
 				if (f->flag_justified) {
-					id = lookup_attr (type, 0, 0, COB_FLAG_JUSTIFIED, NULL, 0);
-				} else {
-					id = lookup_attr (type, 0, 0, 0, NULL, 0);
+					flags |= COB_FLAG_JUSTIFIED;
 				}
+				id = lookup_attr (type, 0, 0, flags, NULL, 0);
 				break;
 			default:
 				if (f->pic->have_sign) {
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -787,12 +787,10 @@ chk_field_variable_address (struct cb_field *fld)
 					return 1;
 				}
 #endif
-				/* Skip redefines as they should not impact addresses: */
-				if (p->redefines) continue;
 				/* Skip PIC L fields as their representation
 				   have constant length */
-				if (p->flag_picture_l) continue;
-				if (p->depending || chk_field_variable_size (p)) {
+				if (p->depending ||
+				    (!p->flag_picture_l && chk_field_variable_size (p))) {
 #if 0	/* only useful with the code above */
 					/* as we have a variable address, all sisters will also;
 					   store this for next check */
@@ -1119,7 +1117,6 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
 	/* CHECKME: May varying addresses occur only with sliging ODO?  If yes
 	   we can jump straight into the else branche when !cb_odoslide. */
 	if (!gen_init_working
-	 && cb_odoslide
 	 && chk_field_variable_address (f)) {
 		if (f01->level == 0
 		 && f01->sister
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -74,7 +74,7 @@
 /* Type of initialization to be done */
 enum cobc_init_type {
 	INITIALIZE_NONE = 0,	/* no init (beause of FILLER, REDEFINES, ...) */
-	INITIALIZE_ONE,		/* initialize a single varialbe */
+	INITIALIZE_ONE,		/* initialize a single variable */
 	INITIALIZE_COMPOUND,	/* init structure */
 	INITIALIZE_DEFAULT	/* init to default-byte value / PIC (USAGE) */
 };
@@ -751,6 +751,8 @@ chk_field_variable_size (struct cb_field *f)
 			if (fc->depending) {
 				f->vsize = fc;
 				break;
+			} else if (fc->flag_picture_l) {
+				continue;
 			} else if ((p = chk_field_variable_size (fc)) != NULL) {
 				f->vsize = p;
 				break;
@@ -787,6 +789,9 @@ chk_field_variable_address (struct cb_field *fld)
 #endif
 				/* Skip redefines as they should not impact addresses: */
 				if (p->redefines) continue;
+				/* Skip PIC L fields as their representation
+				   have constant length */
+				if (p->flag_picture_l) continue;
 				if (p->depending || chk_field_variable_size (p)) {
 #if 0	/* only useful with the code above */
 					/* as we have a variable address, all sisters will also;
@@ -829,7 +834,7 @@ out_odoslide_fld_offset (struct cb_field *p, struct cb_field *fld)
 	if (p == fld) 	/* Single field */
 		return 1;
 
-	if (p->children) {
+	if (p->children && !p->flag_picture_l) {
 		if (out_odoslide_grp_offset (p, fld))
 			return 1;
 	} else {
@@ -4851,17 +4856,14 @@ output_initialize_uniform (cb_tree x, const int c, const int size)
 	} else {
 		output ("memset (");
 		output_data (x);
-		if (size <= 0) {
-			output (", %d, ", c);
-			output_size (x);
-			output (");");
-		} else if (CB_REFERENCE_P(x) && CB_REFERENCE(x)->length) {
+		if (size <= 0 ||
+		    (CB_REFERENCE_P(x) && CB_REFERENCE(x)->length)) {
 			output (", %d, ", c);
 			output_size (x);
 			output (");");
 		} else {
 			struct cb_field		*v = NULL;
-			if (!gen_init_working 
+			if (!gen_init_working
 			 && (f->flag_unbounded || cb_odoslide)) {
 				v = chk_field_variable_size (f);
 			}
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -1116,6 +1116,8 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
 		output ("%s%d", CB_PREFIX_BASE, f01->id);
 	}
 
+	/* CHECKME: May varying addresses occur only with sliging ODO?  If yes
+	   we can jump straight into the else branche when !cb_odoslide. */
 	if (!gen_init_working
 	 && cb_odoslide
 	 && chk_field_variable_address (f)) {
@@ -1124,7 +1126,25 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
 		 && strstr (f01->name, " Record")) {	/* Skip to First 01 within FD */
 			f01 = f01->sister;
 		}
-		out_odoslide_offset (f01, f);
+		if (cb_odoslide) {
+			out_odoslide_offset (f01, f);
+		} else {
+			struct cb_field		*v, *p;
+			for (p = f->parent; p; f = f->parent, p = f->parent) {
+				for (p = p->children; p != f; p = p->sister) {
+					v = chk_field_variable_size (p);
+					if (v) {
+						output (" + %d + ", v->offset - p->offset);
+						if (v->size != 1) {
+							output ("%d * ", v->size);
+						}
+						output_integer (v->depending);
+					} else {
+						output (" + %d", p->size * p->occurs_max);
+					}
+				}
+			}
+		}
 	} else if (f->offset > 0) {
 		output (" + %d", f->offset);
 	}
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -1114,8 +1114,6 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
 		output ("%s%d", CB_PREFIX_BASE, f01->id);
 	}
 
-	/* CHECKME: May varying addresses occur only with sliging ODO?  If yes
-	   we can jump straight into the else branche when !cb_odoslide. */
 	if (!gen_init_working
 	 && chk_field_variable_address (f)) {
 		if (f01->level == 0
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -1252,6 +1252,7 @@ validate_occurs (const struct cb_field * const f)
 		/* Cache field for later checking */
 		cb_depend_check = cb_list_add (cb_depend_check, x);
 
+		/* CHECKME: && !f->flag_induce_complex_odo */
 		if (!cb_odoslide && !cb_complex_odo) {
 			/* The data item that contains a OCCURS DEPENDING clause shall not
 			   be subordinate to a data item that has an OCCURS clause */
@@ -1290,10 +1291,12 @@ validate_redefines (const struct cb_field * const f)
 	}
 
 	/* Check variable occurrence */
-	if (f->depending || cb_field_variable_size (f)) {
+	if (f->depending || (!f->flag_induce_complex_odo &&
+			     cb_field_variable_size (f))) {
 		cb_error_x (x, _("'%s' cannot be variable length"), f->name);
 	}
-	if (cb_field_variable_size (f->redefines)) {
+	if (!f->redefines->flag_induce_complex_odo &&
+	    cb_field_variable_size (f->redefines)) {
 		cb_error_x (x, _("the original definition '%s' cannot be variable length"),
 			    f->redefines->name);
 	}
@@ -2161,6 +2164,7 @@ validate_field_1 (struct cb_field *f)
 		/* END: Not validation */
 		validate_occurs (f);
 	}
+
 
 	if (f->level == 66) {
 		/* no check for redefines here */
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -2889,10 +2889,13 @@ validate_field_value (struct cb_field *f)
 	if (f->values) {
 		if (f->flag_picture_l) {
 			cb_error_x (CB_TREE (f),
-				    _("field with PICTURE L may not have "
-				      "a VALUE clause"));
+				    _("%s and %s are mutually exclusive"),
+				    "PICTURE L", "VALUE");
+			cobc_parse_free (f->values);
+			f->values = NULL;
+		} else {
+			validate_move (CB_VALUE (f->values), CB_TREE (f), 1, NULL);
 		}
-		validate_move (CB_VALUE (f->values), CB_TREE (f), 1, NULL);
 	}
 
 	if (f->children) {
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -1252,7 +1252,7 @@ validate_occurs (const struct cb_field * const f)
 		/* Cache field for later checking */
 		cb_depend_check = cb_list_add (cb_depend_check, x);
 
-		/* CHECKME: && !f->flag_induce_complex_odo */
+		/* CHECKME: && !f->flag_picture_l */
 		if (!cb_odoslide && !cb_complex_odo) {
 			/* The data item that contains a OCCURS DEPENDING clause shall not
 			   be subordinate to a data item that has an OCCURS clause */
@@ -1291,12 +1291,11 @@ validate_redefines (const struct cb_field * const f)
 	}
 
 	/* Check variable occurrence */
-	if (f->depending || (!f->flag_induce_complex_odo &&
-			     cb_field_variable_size (f))) {
+	if (f->depending ||
+	    (!f->flag_picture_l && cb_field_variable_size (f))) {
 		cb_error_x (x, _("'%s' cannot be variable length"), f->name);
 	}
-	if (!f->redefines->flag_induce_complex_odo &&
-	    cb_field_variable_size (f->redefines)) {
+	if (!f->redefines->flag_picture_l && cb_field_variable_size (f->redefines)) {
 		cb_error_x (x, _("the original definition '%s' cannot be variable length"),
 			    f->redefines->name);
 	}
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -1312,7 +1312,11 @@ validate_group (struct cb_field *f)
 		group_error (x, "PICTURE");
 	}
 	if (f->flag_justified) {
-		group_error (x, "JUSTIFIED RIGHT");
+		if (!f->flag_picture_l)
+			group_error (x, "JUSTIFIED RIGHT");
+		else
+			cb_error_x (x, _("'%s' cannot have JUSTIFIED RIGHT clause"),
+				    cb_name (x));
 	}
 	if (f->flag_blank_zero) {
 		group_error (x, "BLANK WHEN ZERO");
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -1252,7 +1252,6 @@ validate_occurs (const struct cb_field * const f)
 		/* Cache field for later checking */
 		cb_depend_check = cb_list_add (cb_depend_check, x);
 
-		/* CHECKME: && !f->flag_picture_l */
 		if (!cb_odoslide && !cb_complex_odo) {
 			/* The data item that contains a OCCURS DEPENDING clause shall not
 			   be subordinate to a data item that has an OCCURS clause */
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -2893,7 +2893,7 @@ validate_field_value (struct cb_field *f)
 		if (f->flag_picture_l) {
 			cb_error_x (CB_TREE (f),
 				    _("%s and %s are mutually exclusive"),
-				    "PICTURE L", "VALUE");
+				    _("variable-length PICTURE"), "VALUE");
 			f->values = NULL;
 		} else {
 			validate_move (CB_VALUE (f->values), CB_TREE (f), 1, NULL);
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -2887,6 +2887,11 @@ static int
 validate_field_value (struct cb_field *f)
 {
 	if (f->values) {
+		if (f->flag_picture_l) {
+			cb_error_x (CB_TREE (f),
+				    _("field with PICTURE L may not have "
+				      "a VALUE clause"));
+		}
 		validate_move (CB_VALUE (f->values), CB_TREE (f), 1, NULL);
 	}
 
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -2891,7 +2891,6 @@ validate_field_value (struct cb_field *f)
 			cb_error_x (CB_TREE (f),
 				    _("%s and %s are mutually exclusive"),
 				    "PICTURE L", "VALUE");
-			cobc_parse_free (f->values);
 			f->values = NULL;
 		} else {
 			validate_move (CB_VALUE (f->values), CB_TREE (f), 1, NULL);
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -866,7 +866,7 @@ copy_into_field (struct cb_field *source, struct cb_field *target)
 	} else {
 		struct cb_picture* new_pic = NULL;
 		int modifier = cb_get_int (target->like_modifier);
-		if (modifier) {			
+		if (modifier) {
 			switch (target->usage) {
 
 			case CB_USAGE_COMP_X:
@@ -919,7 +919,7 @@ copy_into_field (struct cb_field *source, struct cb_field *target)
 					"LIKE", cb_get_usage_string (target->usage));
 				target->flag_invalid = 1;
 			}
- 
+
 #if 0		/* TODO, also syntax-check for usage here */
 			if (target->cat is_numeric) {
 				sprintf (pic, "9(%d)", size_implied);
@@ -1121,7 +1121,7 @@ create_implicit_picture (struct cb_field *f)
 			ret = 1;
 		}
 	}
-	
+
 	/* Checkme: should we raise an error for !cb_relaxed_syntax_checks? */
 	if (!ret) {
 		cb_warning_x (cb_warn_additional, x, _("defining implicit picture size %d for '%s'"),
@@ -1319,7 +1319,11 @@ validate_group (struct cb_field *f)
 				    cb_name (x));
 	}
 	if (f->flag_blank_zero) {
-		group_error (x, "BLANK WHEN ZERO");
+		if (!f->flag_picture_l)
+			group_error (x, "BLANK WHEN ZERO");
+		else
+			cb_error_x (x, _("'%s' cannot have BLANK WHEN ZERO clause"),
+				    cb_name (x));
 	}
 
 	if (f->storage == CB_STORAGE_SCREEN &&
@@ -2690,7 +2694,7 @@ unbounded_again:
 		}
 
 		/* Ensure items within OCCURS are aligned correctly. */
-		if (f->occurs_max > 1 
+		if (f->occurs_max > 1
 		 && occur_align_size > 1
 		 && (size_check % occur_align_size) != 0) {
 			pad = occur_align_size - (size_check % occur_align_size);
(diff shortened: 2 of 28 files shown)


cobc/tree.c
```diff
@@ -2883,43 +2888,44 @@ emit_precedence_error (const int preceding_idx, const int following_idx)
...
+		{ 1, 1, 0, 0, 1, 0, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 1, 0, 0, 0, 0 },
+		{ 1, 1, 0, 0, 1, 0, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 0, 1, 0, 1, 0, 0, 0, 0 },
+		{ 0, 0, 0, 0, 1, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 1, 0, 0, 0 },
+		{ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0 },
+		{ 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0 },
+		{ 1, 1, 1, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
 	};
-	int		error_emitted[24][24] = {{ 0 }};
-	int		chars_seen[24] = { 0 };
+	int		error_emitted[25][25] = {{ 0 }};
```
GitMensch, line 2927: The 25/24 should have been moved out to a constant defined in tree.c next to its use - can you add that in this PR, too?

cobc/tree.c
```diff
@@ -2883,43 +2888,44 @@ emit_precedence_error (const int preceding_idx, const int following_idx)
...
 		  Refer to the standard's PICTURE clause precedence rules for
 		  complete explanation.
 		*/
 		/*
-		  B  ,  .  +  +  + CR cs cs  Z  Z  +  + cs cs  9  A  S  V  P  P  1  N  E
+		  B  ,  .  +  +  + CR cs cs  Z  Z  +  + cs cs  9  A  L  S  V  P  P  1  N  E
 		  0           -  - DB        *  *  -  -           X
 		  /
 		*/
-		{ 1, 1, 1, 0, 1, 0, 0, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 0, 1, 0, 1, 0 },
```
GitMensch, line 2896: Which part of the GCOS documentation did you found the precedence for `L` in?

GitMensch:
@nberth: I understand the idea of this change, but the runtime part (and therefore the codegen) is a design break (not necessarily a show-stopper but a reason to tripple-check).

Would it be possible to do it similar to other `OCCURS DEPENDING` items?

```cobol
01 tab.
   05 tab-row occurs 1 to 50 depending on some-var pic x.
```

In this case the _codegen_ sets the field_size from some-var - the runtime doesn't need to know about its varying length.

Note: this _only_ works if a `MOVE some-string TO lvar` does not change its size; something we actually want to add later (ISO COBOL DYNAMIC LENGTH fields - which has no explicit length set, but resets its length and storage depending on what is moved to it) - but I _guess_ that L-PICs always have the length based on its depending fields - in which case its handling be made according to other `OCCURS DEPENDING` would be most reasonable.

cobc/tree.c
nberth: Sure!

cobc/tree.c
nberth: The precedence rules for character L that can be found on page 287 at https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN23-mar2005/doc/05ula04.pdf
The rules for the other character types are as before the change.

nberth:
Indeed I designed the current tentative draft with more generality than needed. It may be possible to reuse the `OCCURS DEPENDING` mechanics.

GitMensch:
Did you have a look at using the `OCCURS DEPENDING` mechanics?

Note: I _think_ there should be an option to the right to define this PS as work-in-progress/draft. Using this also helps to directly see which PRs may be inspected.

GitMensch:
you may want to change this and other WIP but the OCamlPro-GCOS-Full one [to a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft).
This will show in the UI, so easier recognizable and won't trigger CI if not explicit requested via action.

nberth:
@GitMensch So in the end I think I've managed to use the `OCCURS DEPENDING` mechanics.

Supporting the possible combinations of variable-length fields (`PIC L…`) and `REDEFINES` led me to implement a simplification in `codegen.c (output_base)` that may be tricky to validate; however all the tests seem to pass for now.

tests/testsuite.src/run_extensions.at
```diff
@@ -5601,3 +5600,143 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+AT_CHECK([$COBCRUN_DIRECT ./redefines], [0],
+[0123456789*01234*01234*
+012..*012*
+])
+
+AT_CHECK([$COMPILE -fodoslide redefines.cob], [1], [],
+[redefines.cob:9: error: when ODOSLIDE is set, REDEFINES with PICTURE string with 'L' character are not yet supported
+])
+
+AT_DATA([prog_extraneous_depending.cob], [
```
GitMensch, line 5690: The test following are good, but should be moved to syn_misc.at, using `COMPILE_ONLY`.

There also should be a dialect option for `PIC L` (set to unconformable in all strict dialects, +warning in lax-inc, and ok for gcos and default).

tests/testsuite.src/run_extensions.at
```diff
@@ -5601,3 +5600,143 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+])
+
+AT_CHECK([$COMPILE prog_errs.cob], [1], [],
+[prog_errs.cob:10: error: DEPENDING clause must only be given for alphabetic or alphanumeric PICTURE strings
+prog_errs.cob:11: error: L must be at start of PICTURE string
+prog_errs.cob:11: error: L cannot follow A or X
+prog_errs.cob:11: error: DEPENDING clause must only be given for alphabetic or alphanumeric PICTURE strings
+])
+
+AT_CLEANUP
```
GitMensch, line 5742: Please add missing runtime tests (per "Data Division - Clauses
47 A2 05UL Rev04 9-35") for:

* depending-on-var <= 1 >= max (not sure what should happen then, what happens with GCOS per default / if runtime checks are enabled/disabled ?)
* `INITIALIZE group-containing-lvar` (should always ignore the size, and likely does already)
* `INITIALIZE lvar` (should always ignore the size, I'm not sure it does)
* MOVE CORRESPONDING when l-var is a sending field (should be only copied from with its current size)
* MOVE CORRESPONDING when l-var is a receiving field (should be copied to with its max size)

tests/testsuite.src/run_extensions.at
```diff
@@ -5601,3 +5600,143 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+          2 T-DAT       REDEFINES Z-DAT PICTURE LX(7) DEPENDING ON LGT.
+          2 UNUSED-3    PIC X VALUE ".".
+       1 W-LGS.
+          2 LGX         PIC 99 VALUE 10.
+          2 LGY         PIC 9 VALUE 5.
+          2 LGZ         PIC 9 VALUE 3.
+          2 LGT         PIC 9 VALUE 5.
+       PROCEDURE        DIVISION.
+       MAIN.
+           MOVE "0123456789" TO X-DAT
```
GitMensch, line 5673: Can you please adjust the test to only DISPLAY when there is an unexpected result? Most of the tests were changed into this direction and we want to keep that.

GitMensch (changes requested):
Thank you for the adjustments, I think that's much cleaner than the previous approach.

I'm not sure yet if the results are correct in all cases, but the requested additional tests will show that and also provide something to work on if they aren't.

I think we can simplify it a bit more when doing the suggested parser changes.

cobc/parser.y
```diff
@@ -7302,23 +7302,66 @@ volatile_clause:
...
+			chld->storage = current_field->storage;
+			chld->depending = current_field->depending;
+			chld->flag_occurs = 1;
+			chld->occurs_min = 1;
+			chld->occurs_max = current_field->pic->size - 1;
+			chld->parent = current_field;
+			current_field->children = chld;
+			current_field->pic = NULL;
+			current_field->depending = NULL;
+			if (cb_odoslide && !odoslide_check && current_field->redefines) {
```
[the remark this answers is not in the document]
nberth: I believe using the `OCCURS` mechanics for `PIC L` requires enabling complex ODO checks for the relevant fields, while GCOS rejects the ODO cases those check allow (like for instance the `complex-odo` checks in `syn_occurs.at`).
So my comment when setting `flag_induce_complex_odo` was not correct; I'll change it to be more specific.

tests/testsuite.src/run_extensions.at
nberth: Thanks for noting the missing tests. Dealing with `INITIALIZE` and `MOVE CORRESPONDING` was tricky ;-)
I am still having some difficulties grasping what `flag_relative_odo` is meant for…

cobc/parser.y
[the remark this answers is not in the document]
nberth: I actually renamed this flag `flag_picture_l` as unfortunately I feel I need some bits of special handling all the way into `typeck.c` and `codegen.c`.

tests/testsuite.src/run_extensions.at
GitMensch: I don't find any `relative_odo` in the sources.

cobc/parser.y
```diff
@@ -7301,86 +7301,41 @@ volatile_clause:
 
 picture_clause:
   PICTURE	/* token from scanner, includes full picture definition */
-  _pic_locale_format
-  _pic_depending_on
   {
 	check_repeated ("PICTURE", SYN_CLAUSE_4, &check_pic_duplicate);
 	current_field->pic = CB_PICTURE ($1);
+  }
+  _pic_locale_format
```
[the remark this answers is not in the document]
nberth: I was not sure about that: combining the two clauses is not part or any standard per se, but may be doable with the `default` dialect. Of course `PIC L` is a rather exotic extension, so I can understand why it makes sense to have those clauses mutually excluded.

tests/testsuite.src/run_extensions.at
nberth: Yes I cannot perform the runtime checks for now, so that's postponed.

cobc/parser.y
```diff
@@ -7393,6 +7348,56 @@ locale_name:
   }
 ;
 
+_pic_depending_on:
+  _occurs_depending
+  {
+	cb_tree depending = CB_TREE ($1);
+	if (CB_VALID_TREE (depending) && !current_field->flag_occurs) {
```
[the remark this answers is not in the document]
GitMensch: OK, then let's add a testcase for `PIC LAAAA DEPENDING ON LSZ`, at least one that has a parent with `OCCURS` and once verified with GCOS possibly also one that has both clauses.

Actually when checking this in the docs, I've found two other rules to test:

> The `JUSTIFIED` clause must not be specified for a variable-length data item

--> should be a simple test

> if a group item which contains a variable-length data item is referenced, all
character positions in the variable-length data item participate in the operation;
i.e., the variability of the valid contents of the data item(s) is ignored

--> test with `LENGTH OF` and `FUNCTION LENGTH` (first in GCOS, then in GnuCOBOL)


(and yes: there were no results about `OCCURS` and even `OCCURS DEPENDING ON` not being allowed)

cobc/parser.y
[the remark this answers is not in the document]
nberth: Apologies: do you mean you prefer mutual exclusion, or leaving as is?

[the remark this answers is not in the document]
GitMensch: mutual exclusion

cobc/field.c
```diff
@@ -2891,6 +2891,11 @@ static int
 validate_field_value (struct cb_field *f)
 {
 	if (f->values) {
+		if (f->flag_picture_l) {
+			cb_error_x (CB_TREE (f),
+				    _("field with PICTURE L may not have "
+				      "a VALUE clause"));
+		}
```
GitMensch, line 2898: I'd suggest to set `f->values = NULL` in this case and have the validation done in the else part.
In any case: `("%s and %s are mutually exclusive"), "PICTURE L", "VALUE")`.

cobc/field.c
```diff
@@ -2893,10 +2893,13 @@ validate_field_value (struct cb_field *f)
 	if (f->values) {
 		if (f->flag_picture_l) {
 			cb_error_x (CB_TREE (f),
-				    _("field with PICTURE L may not have "
-				      "a VALUE clause"));
+				    _("%s and %s are mutually exclusive"),
+				    "PICTURE L", "VALUE");
+			cobc_parse_free (f->values);
```
GitMensch, line 2898: The idea is good and reasonable. Some people will say "that's not relevant" because at the end of the parsing the values will be freed in any case (that's the purpose of the parse malloc); but _if_ you opt do do it then that needs to be done completely (the f->values is always a list with at least (in this case only) one contained cb_tree - only freeing the list is not good.
I think I'd personally leave that to the auto-cleanup at the end of parsing...

nberth: Ok (I though the free was greedier, but I should have checked).

nberth:
@GitMensch I'm getting unexpected results on GCOS, notably for the `INITIALIZE` statement (that seems to obey the `DEPENDING`, contrary to what the documentation seems to indicate). So I'll do some more tests to check.

nberth:
Otherwise, when the `DEPENDING` var <= 1 >= max, we get errors at runtime by default (we don't really know if/how runtime check may be disabled).

GitMensch:
> Otherwise, when the `DEPENDING` var <= 1 >= max, we get errors at runtime by default (we don't really know if/how runtime check may be disabled).

I _guess_ there is a runtime error with GnuCOBOL already when using `--debug` - right? If missing please add a testcase for this.

cobc/codegen.c
```diff
@@ -1072,31 +1078,14 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
...
+	 && cb_odoslide
 	 && chk_field_variable_address (f)) {
 		if (f01->level == 0
 		 && f01->sister
 		 && strstr (f01->name, " Record")) {	/* Skip to First 01 within FD */
 			f01 = f01->sister;
 		}
-		if (cb_odoslide) {
-			out_odoslide_offset (f01, f);
-		} else {
```
GitMensch, line 1083: Where is that "else" branch gone? Wasn't it executed before?

GitMensch (changes requested):
Apart of the (relative minor - compared to this huge PR) change requests and comments below, please take care to document differences in behaviour vs. documentation in both the code parts - possibly leaving the "according to docs" code in as `#if 0` - and the tests.
This helps a _lot_ for later work, we've seen code be changed before to match the docs because of those missing comments (and sometimes it is found that the behaviour is depending on the version of the other environment - in this case the fun starts...).

cobc/parser.y
```diff
@@ -7800,10 +7807,8 @@ _occurs_integer_to:
 ;
 
 _occurs_depending:
-| DEPENDING _on reference
-  {
-	current_field->depending = $3;
-  }
+  /* empty */			{ $$ = NULL; }
+| DEPENDING _on reference	{ $$ = $3; }
```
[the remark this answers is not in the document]
nberth: Ok no problem

cobc/codegen.c
nberth: While investigating that code, I had the impression that at that point the `offset` field was already properly set for every field that does not have a varying address, and that the latter case may only occur with `ODOSLIDE` (but still there may be some features I've overlooked).
Furthermore, if I've missed cases that require this specific else branch, then they are not exercised by the test suite: what I suggest is restore the else branch as before and add a comment about this.

cobc/typeck.c
```diff
@@ -4376,14 +4378,15 @@ cb_validate_program_data (struct cb_program *prog)
...
+					 && !q->parent->flag_picture_l
 					 && x != xerr) {
 						xerr = x;
 						cb_error_x (x,
 							_ ("'%s' cannot have OCCURS DEPENDING because of '%s'"),
 							cb_name (x), p->sister->name);
 						break;
 					}
-					p->flag_odo_relative = 1;
+					p->flag_odo_relative = /* CHECKME: !q->parent->flag_picture_l */1;
```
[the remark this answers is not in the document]
nberth: Well… I'm still unsure about the meaning of `flag_odo_relative`. I've left the `CHECKME` as I'd like a clarification on its semantics.

tests/testsuite.src/run_extensions.at
```diff
@@ -5601,3 +5600,364 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+# MOVE CORRESPONDING
+
+AT_DATA([corr.cob], [
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      corr.
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       1 W-DATA.
+         2 GA.
+           3 X1         PICTURE LX(5) DEPENDING ON LGX.
```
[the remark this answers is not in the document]
nberth: Thanks for the suggestion. Indeed that was not tested.

cobc/typeck.c
```diff
@@ -6938,19 +6941,24 @@ emit_move_corresponding (cb_tree x1, cb_tree x2)
...
+				t1 = cb_build_field_reference (f1, x1);
+				t2 = cb_build_field_reference (f2, x2);
+				if (f1->flag_picture_l) {
+					CB_REFERENCE (t1)->length = cb_int (f1->size);
+				}
+				if (f2->flag_picture_l) {
+					CB_REFERENCE (t2)->length = cb_int (f2->size);
+				}
+				if (f1->children && !f1->flag_picture_l &&
+				    f2->children && !f1->flag_picture_l) {
```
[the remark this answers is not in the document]
nberth: Typo indeed: thanks! I've added a comment about non-conformity of GCOS 7 w.r.t its docs.

cobc/typeck.c
```diff
@@ -4353,7 +4354,8 @@ cb_validate_program_data (struct cb_program *prog)
 		for (p = q; ; p = p->parent) {
 			if (p->depending) {
 				if (odo_level > 0
-				 && !cb_odoslide) {
+				 && !cb_odoslide
+				 && !q->parent->flag_picture_l) {
```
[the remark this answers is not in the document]
nberth: Did not know we could have a FILLER at toplevel…

cobc/typeck.c
```diff
@@ -2254,6 +2254,7 @@ cb_build_identifier (cb_tree x, const int subchk)
 		if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT) && f->odo_level != 0) {
 			for (p = f; p; p = p->children) {
 				if (p->depending && p->depending != cb_error_node
+				 && (p == f || p->parent == f || !p->parent->flag_picture_l)
```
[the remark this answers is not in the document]
nberth: Here the runtime check is disabled for any *implicit* access to a PIC L field (*i.e.* via an enclosing group), as normally those should disregard the `DEPENDING`. I've added a small comment on that.

cobc/tree.c
```diff
@@ -3803,6 +3837,8 @@ cb_field_variable_address (const struct cb_field *fld)
 	f = fld;
 	for (p = f->parent; p; f = f->parent, p = f->parent) {
 		for (p = p->children; p != f; p = p->sister) {
+			/* Skip redefines as they should not impact addresses */
```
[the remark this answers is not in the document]
nberth: I need to check that indeed (especially as larger redefines seem to be allowed on GCOS).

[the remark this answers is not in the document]
nberth: BTW should this pass without `odoslide`?

cobc/typeck.c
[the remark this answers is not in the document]
GitMensch: yes, both a filler (standard) and a (nonstandard) level 01 OCCURS.

cobc/tree.c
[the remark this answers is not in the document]
GitMensch: something after an ODO item is non-standard and not identical between compilers, I think GnuCOBOL forbids that and only allows with with `-fcomplex-odo` (in this case fld is always at offset 13) or with `-fodoslide` (in this case the offset is variable).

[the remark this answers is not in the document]
nberth: As of now in GNUCobol it seems `REDEFINES` cannot be used in combination with a `OCCURS DEPENDING` even with complex ODO or ODOSLIDE (cf. `validate_redefines`).

cobc/typeck.c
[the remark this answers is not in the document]
GitMensch: I find the comment quite clear, but that was added by me so...

> complex-odo: item address depends on size of a different (ODO) item

`flag_odo_relative` therefore (from a quick look at the code) means that the field's address is variable.

This should only (but always) apply to a PIC L item if it either itself or a parent has a a previous sibling with ODO.

cobc/codegen.c
```diff
@@ -1110,6 +1114,8 @@ output_base (struct cb_field *f, const cob_u32_t no_output)
 		output ("%s%d", CB_PREFIX_BASE, f01->id);
 	}
 
+	/* CHECKME: May varying addresses occur only with sliging ODO?  If yes
```
GitMensch, line 1117: What is "sliging"? Maybe "sliding"?
In general every BASED item and item in LINKAGE may have a variable address, too - but I'm not sure that this part is about that.

cobc/typeck.c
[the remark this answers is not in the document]
GitMensch: That's still unclear to me, what I've understood would be `&& (!p->parent || !p->parent->flag_picture_l)`

cobc/typeck.c
```diff
@@ -4594,14 +4601,15 @@ cb_validate_program_data (struct cb_program *prog)
...
+					 && !parent_is_pic_l
 					 && x != xerr) {
 						xerr = x;
 						cb_error_x (x,
 							_("'%s' cannot have OCCURS DEPENDING because of '%s'"),
 							cb_name (x), p->sister->name);
 						break;
 					}
-					p->flag_odo_relative = 1;
+					p->flag_odo_relative = /* CHECKME: !parent_is_pic_l */1;
```
GitMensch, line 4612: As noted before this is not directly related.

cobc/typeck.c
```diff
@@ -7156,19 +7164,28 @@ emit_move_corresponding (cb_tree x1, cb_tree x2)
...
-						}
-					}
+		if (f1->redefines || f1->flag_occurs) continue;
+		for (f2 = CB_FIELD_PTR (x2)->children; f2; f2 = f2->sister) {
+			if (f2->redefines || f2->flag_occurs) continue;
+			if (strcmp (f1->name, f2->name) == 0) {
+				t1 = cb_build_field_reference (f1, x1);
+				t2 = cb_build_field_reference (f2, x2);
+				/* GCOS 7: Contrary to the documentation,
+				   handling of PIL L fields in MOVE
```
GitMensch, line 7174: `s/PIL/PIC/`

cobc/typeck.c
```diff
@@ -9415,6 +9432,9 @@ cb_emit_initialize (cb_tree vars, cb_tree fillinit, cb_tree value,
 		 && CB_REFERENCE_P (x)
 		 && CB_REFERENCE   (x)->subs == NULL
 		 && CB_REFERENCE   (x)->length == NULL) {
+			/* GCOS 7: Contrary to what the documentation states,
```
GitMensch, line 9435: You've likely have that checked in the testsuite and on GCOS - Does this apply to both the "initial state" `VAR PIC LX(5) DEPENDING ON LEN VALUE ALL 'A'` with `LEN PIC 9 VALUE 3` and with `MOVE 4 TO LEN. INITIALIZE VAR`?

tests/testsuite.src/run_extensions.at
GitMensch: I guess those checks were finished now?

GitMensch (commented):
only minor adjustments - I think in general the code changes are fine and there is enough NEWS entry already (you may or may not want to add "including PICTURE L" to the GCOS news entry, I'm fine with both).

There are some discussions to be marked as resolved and CHECKME removed.

cobc/typeck.c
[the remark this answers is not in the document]
nberth: Ok noted.
What I may have missed is: this flag is only relevant when `complex-odo` holds, but it is always set anyways. From the two times it is actually read (in `cb_build_length_1` and `output_size`), I guess it does not impact generated code when `odoslide` is set. However if the latter is not set, how may item addresses be not fixed?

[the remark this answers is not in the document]
nberth: Ok I think you've answered this question in another comment (`LINKAGE`, etc).

tests/testsuite.src/run_extensions.at
nberth: Yes unless we have missed some instructions. The tests for `PICTURE L` currently included in `run_extensions.at` are as expected on a GCOS 7 platform.

cobc/typeck.c
nberth: As per GCOS 7 `PIC L` fields may not have a `VALUE` clause. This is checked and rejected somewhere in this PR.

nberth:
I've updated the NEWS file to mention the specific support for `PICTURE L`, and took the occasion to mention the `CONTROL DIVISION` that, albeit being quite obsolete, seems very specific to GCOS and may still be found in various code bases.

cobc/typeck.c
GitMensch: I see, this would only leave a test like:

```cobol
MOVE 5 TO LEN
MOVE ALL 'A' TO VAR
MOVE 3 TO LEN
INITIALIZE VAR
MOVE 5 TO LEN
IF VAR = "   AA" ...
```

tests/testsuite.src/syn_misc.at
```diff
@@ -8592,3 +8592,93 @@ prog.cob:9: warning: DISPLAY statement in DEFAULT SECTION is not implemented
...
+       1 L-DAT		PICTURE LX(10) DEPENDING ON LGS.
+       PROCEDURE        DIVISION.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -fpicture-l=warning prog.cob], [0], [],
+[prog.cob:7: warning: PICTURE string with 'L' character used
+])
+
+AT_CLEANUP
```
GitMensch, line 8684: Test missing: PIC L  may not have VALUE clause (and other stuff)

nberth: Actually the message might not be clear enough: it's tested with `prog_value.cob` a few lines above that. The error is
```
prog_value.cob:7: error: PICTURE L and VALUE are mutually exclusive
```

GitMensch: That's fine. I _thought_ this and other added syntax rules were missing from the tests, just give it a quick check and when you think you have those covered then that's fine.

cobc/parser.y
```diff
@@ -7305,43 +7305,88 @@ volatile_clause:
...
-		}
-	}
+  }
+  _pic_locale_format_or_depending_on
+  {
+	  if ((!current_field->pic || current_field->pic->variable_length) &&
+	      !current_field->flag_picture_l) {
+		  /* Current field with PIC L was not translated */
+		  cb_error_x (CB_TREE (current_field->pic),
+			      _("missing DEPENDING clause for PICTURE string with "
```
GitMensch, line 7318: Checking that message... we seem to have no matching msgid we can use, here but what do you think about `_("variable-length PICTURE requires DEPENDING clause")` (not much shorter, just more similar to existing msgids)?

cobc/parser.y
```diff
@@ -7305,43 +7305,88 @@ volatile_clause:
...
 
-_pic_locale_format:
+_pic_locale_format_or_depending_on:
   /* empty */
-  { $$ = NULL; }
-| LOCALE _is_locale_name SIZE _is integer  
+| LOCALE _is_locale_name SIZE _is integer
+  {
+	  /* $2 -> optional locale-name to be used */
+	  if (CB_VALID_TREE ($5)) {
```
GitMensch, line 7329: The valid_tree check for 5 (the integer) can be dropped (that complete if here); using $1 (the LOCALE token) below for PENDING seems to be troublesome, maybe drop the `_X ($1, `?

cobc/parser.y
```diff
@@ -7305,43 +7305,88 @@ volatile_clause:
...
+			  /* TODO: check that not we're not within a CONSTANT RECORD */
+			  CB_PENDING_X ($1, "locale-format PICTURE");
+		  }
+	  }
+  }
+| DEPENDING _on reference
   {
-	/* $2 -> optional locale-name to be used */
-	$$ = $5;
+	  cb_tree depending = $3;
```
GitMensch, line 7343: Most of this code and above is not plain parsing based on tokens but validation - is there any option to move code out of it to `cb_validate_field`?

cobc/typeck.c
```diff
@@ -2420,6 +2420,14 @@ cb_build_identifier (cb_tree x, const int subchk)
...
 			for (p = f; p; p = p->children) {
 				if (p->depending && p->depending != cb_error_node
+				    /* Do not check length for implicit access
+				       to a PIC L field (i.e, via enclosing
+				       group), as those disregard the DEPENDING.
+				       However, assuming the filler is never
+				       explicitly accessed (p == f), check is
+				       still done for explicit access to PIC L
+				       field (p->parent == f). */
+				 && (p->parent == f || !p->parent->flag_picture_l)
```
GitMensch, line 2430: Sorry to be such a pain, but as we're shortly before a RC (currently planned end of August) this looks too suspicious to me:

`&& (p->parent == f || !p->parent->flag_picture_l)`
is - for existing code - equal to
`&& (p->parent == f)`
and as the loop always iterates through he grand-children this seems to contradict the loop (now is only active if the direct parent is still `f`).

This is currently the main reason for "not merge ready" - I _think_ this removes the odo-check for grand-children.

tests/testsuite.src/syn_misc.at
```diff
@@ -8592,3 +8592,93 @@ prog.cob:9: warning: DISPLAY statement in DEFAULT SECTION is not implemented
...
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       1 L		COMP-1 VALUE 10.
+       1 F-DAT		PICTURE LX(10) DEPENDING ON L
+                        VALUE "0123456789".
+       PROCEDURE        DIVISION.
+           STOP RUN.
+])
+AT_CHECK([$COMPILE_ONLY prog_value.cob], [1], [],
+[prog_value.cob:7: error: PICTURE L and VALUE are mutually exclusive
```
GitMensch, line 8644: I suggest to only reference `L` during parsing of the `PICTURE` string. Let's stay with "variable-length PICTURE" - in this case as `_("variable-length PICTURE")` in all other places (I may have not spot all of those yet).

tests/testsuite.src/syn_misc.at
```diff
@@ -8592,3 +8592,93 @@ prog.cob:9: warning: DISPLAY statement in DEFAULT SECTION is not implemented
...
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       1 L		COMP-1.
+       1 F-DAT		PICTURE X(10) DEPENDING ON L.
+       PROCEDURE        DIVISION.
+           MOVE SPACES TO F-DAT
+           STOP RUN.
+])
+AT_CHECK([$COMPILE_ONLY prog_extraneous_depending.cob], [1], [],
+[prog_extraneous_depending.cob:7: error: DEPENDING clause must either come with an OCCURS clause or a PICTURE string with 'L' character
```
GitMensch, line 8612: see below:
"DEPENDING clause needs either an OCCURS clause or a variable-length PICTURE"

GitMensch (changes requested):
The odo-check adjustment looks wrong to me - please have _another_ look at it.

Please do some msgid adjustments as outlined, both to keep the existing style in general and to ease translator work.

If possible: move some validation out of the parser.


And in any case: thank you for your patience and re-iteration work!

tests/testsuite.src/syn_misc.at
nberth: I checked the manual and found `BLANK WHEN ZERO` was (obviously) not allows for `PIC L` fields. This was already rejected but I clarified the message.
I found no other clauses that explicitly conflicted with such fields.
Just tell me if you think of now other clause, and if that can go upstream.

cobc/typeck.c
nberth: I think for existing code `(p->parent == f || !p->parent->flag_picture_l)` is equal to true (as no field had `flag_picture_l` true).

cobc/parser.y
nberth: I would have liked, but we may reuse the `depending` field later on for `… VAR PIC L… DEPENDING … OCCURS … DEPENDING`. To move this for later validation, an alternative `depending` field may need to be added.

cobc/typeck.c
GitMensch: ... that's of course correct and I was wrong.
* p _must_ have a parent here, because its a (grand-)child of `f`
* those never have this attribute set

This leaves us with "the main issue" resolved :-)

cobc/parser.y
GitMensch: I see, so also resolved.

GitMensch:
Congrats - this is nearly finished. Please handle #55 -> upstrean, then update from upstream, adjust the msgids and we're good.

cobc/parser.y
```diff
@@ -7326,33 +7326,31 @@ _pic_locale_format_or_depending_on:
...
 | DEPENDING _on reference
   {
 	  cb_tree depending = $3;
 	  if (!current_field->pic->variable_length) {
-		  cb_error_x ($3, _("DEPENDING clause must either come with "
-				    "an OCCURS clause or a PICTURE string "
-				    "with 'L' character"));
+		  cb_error_x ($3, _("DEPENDING clause needs either an "
+				    "OCCURS clause or a %s"),
+			      _("variable-length PICTURE"));
```
GitMensch, line 7345: NEARLY - this one and the one below are complete sentences with msgids we won't use anywhere else - so "variable-length PICTURE" should be part of these msgids

GitMensch (approved):
minor msgid use (testsuite will be unchanged) - otherwise ready to commit upstream (no need to update this branch from upstream before as you'll close it afterwards in any case)

nberth:
now  upstream

cobc/typeck.c
GitMensch: Nope, I wasn't wrong initially, only when marking that as resolved :-)

I've compiled a bigger codebase with the current nightly which made cobc SIGSEGV here.
Reason: `OCCURS DEPENDING ON` with a non-standard level 01 - goes in here, `f = p`, so `p` has no parent.

I'll add the testcase and the fix to svn in the next hour.

nberth: Argh… obviously I must also have convinced myself that `p` always had a parent here. Thanks for fixing that then.

PR #43, opened 2022-05-25
Add support for GCOS dialect

nberth:
This is a follow up on #26


ChangeLog
```diff
@@ -1,4 +1,8 @@
 
+2022-05-25  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* NEWS: mention new -std options for GCOS 7 dialect
```
GitMensch, line 4: nah, can be dropped (but better consider it first)

nberth: So it's only mentioned in NEWS, right?

config/gcos.words
```diff
@@ -0,0 +1,673 @@
+# List of GCOS reserved words
+
+# NOTE: the mapping for COMP* usages does not exactly match the GCOS
+# specification, due to the former missing some of the representations
+# requiret.  The mapping below associates each COMP* with the closest
```
GitMensch, line 5: Good comment on the right place, just one typo "required".

ChangeLog
GitMensch: yes (and cobc/Changelog as you would put that in cobc/help.c, too ;-)

config/gcos.words
```diff
@@ -0,0 +1,478 @@
...
+reserved:	LEFT
+reserved:	LENGTH
+reserved:	LESS
+reserved:	LIMIT
+reserved:	LIMITS
+reserved:	LINAGE
+reserved:	LINAGE-COUNTER
+reserved:	LINE
+reserved:	LINE-COUNTER
+reserved:	LINE-SEQUENTIAL*	# GNUCobol
```
GitMensch, line 248: If this isn't supported in GCOS just drop it (the lax config will additionally use the default words).

config/gcos.words
```diff
@@ -0,0 +1,478 @@
...
+reserved:	EXTEND
+reserved:	EXTERNAL
+reserved:	FALSE
+reserved:	FD
+reserved:	FILE
+reserved:	FILE-CONTROL
+reserved:	FILLER
+reserved:	FINAL
+reserved:	FIRST
+reserved:	FLOAT-LONG		# COBOL 2002
```
GitMensch, line 198: If this isn't supported in GCOS, then please drop it here.
If it is drop the COBOL 2002 comment, also in the other places.

config/default.conf
```diff
@@ -25,12 +25,15 @@ name: "GnuCOBOL"
...
 standard-define			0
+# NOTE: see enum cb_std_def, defined in cobc/cobc.h.
 #        CB_STD_GC = 0,
 #        CB_STD_MF,
 #        CB_STD_IBM,
 #        CB_STD_MVS,
 #        CB_STD_BS2000,
 #        CB_STD_ACU,
+#        CB_STD_RM,
+#        CB_STD_GCOS,
```
GitMensch, line 36: Seems like `CB_STD_GCOS` is gone.

NEWS
```diff
@@ -53,6 +53,11 @@ NEWS - user visible changes				-*- outline -*-
 
 * Changes to the COBOL compiler (cobc) options:
 
+** New -std options:
+
+   gcos            GCOS compatibility
+   gcos-strict     GCOS compatibility                - strict
+
```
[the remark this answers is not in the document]
nberth: Ok I've moved the entry and added a note.

NEWS
```diff
@@ -8,6 +8,8 @@ NEWS - user visible changes				-*- outline -*-
 
   BIT-OF, BIT-TO-CHAR, HEX-OF, HEX-TO-CHAR
 
+** Support of GCOS 7 (Bull) dialect
+
```
GitMensch, line 12: New entries belong to the end of the section - in this case before "TODO - more to add".

GitMensch (changes requested):
Nearly finished!

config/gcos.words
nberth: Ok I dropped this.

config/gcos.words
nberth: Got rid of the coments.

NEWS
```diff
@@ -80,6 +82,15 @@ NEWS - user visible changes				-*- outline -*-
...
    and fatality `cobc --list-exceptions`
 
+** New -std options:
+
+   gcos            GCOS compatibility
+   gcos-strict     GCOS compatibility                - strict
+
+   We define the GCOS dialect based on the COBOL85 standard, with new
+   dialect configuration options accompanying each specificity
+   introduced by the dialect
```
GitMensch, line 92: minor typo: either end with a full-stop or start lower case, otherwise this file is "done"

config/gcos.words
```diff
@@ -0,0 +1,477 @@
...
+reserved:	EXTEND
+reserved:	EXTERNAL
+reserved:	FALSE
+reserved:	FD
+reserved:	FILE
+reserved:	FILE-CONTROL
+reserved:	FILLER
+reserved:	FINAL
+reserved:	FIRST
+reserved:	FLOAT-LONG
```
GitMensch, line 198: that doesn't match https://support.bull.com/ols/product/system/gcos7/gcos7-com/g7-dps7000/doc-com/docf/g/47X257TN23-mar2005/doc/05ula04.pdf A1 - do we currently need to have FLOAT-LONG in here to alias COMP-10? Otherwise this ad other COBOL2002 words should be removed.

GitMensch (changes requested):
finish line: reserved word list

config/gcos.words
nberth: Indeed some of the "COBOL 2002" words are not in the list from the GCOS documentation, so I dropped them.

About `FLOAT-LONG` and the likes, however, the purpose of their appearance in the list is for the definition of aliases. For instance `FLOAT-LONG` is used as it matches the definition of `COMP-10`, and the definition of the alias `COMP-10=FLOAT-LONG` requires `FLOAT-LONG` to be included in the list. I can see where this may be an issue. Another alternative would be to add a dedicated entry for `COMP-10` and `COMP-15` in `cobc/reserved.c` as "GCOS extension"s. Would you prefer the latter approach? Or another I am currently failing to identify?

nberth: For the latter, would an entry with
```c
  { "COMP-10",		0, 0, FLOAT_LONG,		/* GCOS extension */
				0, 0
  },
```
be enough?

GitMensch: Adjusting reserved.c sounds good, your proposed change would be correct, you'd only need the alias shortform<->longform `COMPUTATIONAL` in gcos.words then and have the cobol2002-usages be `not-reserved:` in gcos.conf after enabling everything.
This has also the benefit that you'll likely add a syntax check with `-std=gcos` that compiles `01 FLOATS. 05 FLOAT-LONG USAGE COMP-10 VALUE 1.2456.  05 FLOAT-SHORT ....`.

GitMensch: ... and this has the benefit that `-std=default` will also compile those `USAGE`s.

nberth: Ok I can try to do this. I'm not sure to exactly follow what you mean by: "then and have the cobol2002-usages be not-reserved: in gcos.conf after enabling everything.". 

The additional issue I have with the approach is that I need to define `COMP-1` to be equivalent to `BINARY-SHORT` without enabling the latter as a reserved word, yet `COMP-1` is already defined in `reserved.c`.

GitMensch: I haven't checked this in detail, but are you sure, that this doesn't work?
If I do `cobc -std=rm-strict -freserved=BANANA:APPLE --list-reserved`  I get a  "BANANA not implemented", if I do `cobc -std=rm-strict -freserved=BANANA:FLOAT-LONG --list-reserved` I get a "BANANA implemented", which also should work that way.
If it works then it should also work to have `reserved: COMP-1=BINARY-SHORT COMPUTATIONAL-1=BINARY-SHORT`.
I meant to have the part

```
# not-reserved:
# Value: Word to be taken out of the reserved words list
# -fill upon request-"
```

after `include "lax.conf-inc"` (should be there in any case) then have `not-reserved: FLOAT-LONG` and friends - but on second thought: if not using the strict dialect, then there are word conflicts - that's the way it is, we can leave it that way.

Still wondering about "doesn't it work to just have `reserved: COMP-1=BINARY-SHORT`? Is there an error message, does it just "don't work"?

nberth: > If it works then it should also work to have `reserved: COMP-1=BINARY-SHORT COMPUTATIONAL-1=BINARY-SHORT`.

Indeed I did not check the limitation myself before, and in the end you are right, it turns out we can approach the problem as suggested.

config/gcos.words
```diff
@@ -0,0 +1,470 @@
...
+reserved:	COMP-3
+reserved:	COMP-5=PACKED-DECIMAL
+reserved:	COMP-8=PACKED-DECIMAL
+reserved:	COMP-9			# single precision
+reserved:	COMP-10			# double precision
+reserved:	COMP-15			# quadruple precision
+reserved:	COMPUTATIONAL=COMP-3
+reserved:	COMPUTATIONAL-1=BINARY-SHORT
+reserved:	COMPUTATIONAL-2=BINARY-LONG
+reserved:	COMPUTATIONAL-3
```
GitMensch, line 101: Doesn't this need the `=COMP-3`, too?

GitMensch (changes requested):
It would be good to have at least one testcase for the new words, otherwise it is likely they will drop out or not work at some time in the future.
I suggest something that uses
```cobol
01 FLOATS.
  03 FLOAT-BINARY-32  USAGE COMPUTATIONAL-9.
  03 FLOAT-BINARY-64  USAGE COMPUTATIONAL-10.
  03 FLOAT-BINARY-128 USAGE COMPUTATIONAL-11.
```
with a `VALUE` clause and some `ADD`; if you want a runtime check then displaying them, but I'd be fine with a plain syntax check, too.

config/gcos.words
nberth: In `reserved.c`, I found `COMP-3` and `COMPUTATIONAL-3` are both defined as `COMP_3` already. And with the current words list `cobc -std=gcos-strict --list-reserved` gives me:
```
COMP-3                          Yes (aliased with COMP, COMPUTATIONAL, COMPUTATIONAL-3)
COMPUTATIONAL-3                 Yes (aliased with COMP, COMP-3, COMPUTATIONAL)
```

GitMensch: The list in reserved.c is only active with `-std=default` and with the lax include (additional to the dialect specific word list).
It just looked strange to me that `COMP-5` and friends do have the alias given, while this one doesn't.

nberth: I have added a simple test for `COMP-{9,10,15}`, but I limited the computations to a minimum as the proper keywords `FLOAT_BINARY_{32,64,128}` appear to trigger internal errors (some missing entries in `cobc/field.c (setup_parameters)`, at least). There's now an entry for that in `TODO`, in case (but I guess there's probably a pending issue about that already).

nberth:
Ok that's done. Obviously I forgot to `svn add` the new files after the patch, so I needed a second commit afterwards: the switch from git to svn requires some adaptation…

GitMensch:
No problem, similar to git you may want to use `svn status` to get some information, untracked files not in svn-ignore will be shown along modified and added files.

CI on appveyor is running now (does the tests [found the missing addition in the first place]), I assume it to pass as it passes locally.
* https://ci.appveyor.com/project/GitMensch/gnucobol-3-x/builds/43709122
* https://ci.appveyor.com/project/GitMensch/gnucobol-3-x-win32-posix/builds/43709123 (waits for the one before as it does not create the maintainer files on its own but uses the last working version from the "posix" one.

So time to close this one and go on with the next :-)

nberth:
Closed as now on svn side.

PR #44, opened 2022-06-01
[WIP] GNUCobol 3.x with OCamlPro contributions for GCOS

nberth:
The purpose of this branch is to track pending changes, before they are upstreamed


GitMensch:
I'm still wondering where the deletion of ar-lib and recreation comes from...

BTW: as you've seen - the usptream changes for GCOS support are now also in trunk/master.

GitMensch:
Sadly the CI build does not ship the NIST tests on failure (easiest: just tar the complete tests/cobol85 folder, better: drop all files that have no "log" "out" "txt" "err" extension before) - so one would have to build it locally to check the reason for the DB compile error (a compile error is often easily fixable).

GitMensch:
Hm, where does this go to? Should the branch be updated or this PR closed (and the branch possibly deleted)?

lefessan:
Indeed, this PR was for archival, since some of the work hasn't been upstreamed, but might be useful in the future.

GitMensch:
@nberth @lefessan Is there still any content in that which "hasn't been upstreamed, but might be useful in the future"?
 If not let's delete the branch.

PR #45, opened 2022-06-17
[GCOS] Add mnemonics TERMINAL, ALTERNATE-CONSOLE, ALTERNATE CONSOLE

nberth:
This follows from and replaces #30.


nberth:
Now upstream.

PR #46, opened 2022-06-24
add -ffold-copy=BOTH

lefessan:
Add flag value `BOTH` to `-ffold-copy`. It can be used to match filenames of copies independantly of the case (before, it was either `UPPER` or `LOWER`)

### DIFF
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -3581,11 +3581,13 @@ process_command_line (const int argc, char **argv)
 			break;
 
 		case 4:
-			/* -ffold-copy=<UPPER/LOWER> : COPY fold case */
+			/* -ffold-copy=<UPPER/LOWER/BOTH> : COPY fold case */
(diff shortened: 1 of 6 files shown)


GitMensch:
Fabrice, can you please share a bit about the use case? In which case would you need "upper first, then lower"?
I'm not against this _addition_ but we should know where it is necessary / useful.

lefessan:
My use-case for this option was that we have to work with a codebase where the names of copies have not been "standardized", so some of  them are uppercase and some of them are lowercase, or a mix of them, and we are not allowed to change these names... So, the `BOTH` option allows us not to care about the upper/lower case status of the filenames... @GitMensch

GitMensch:
The "both" will not help with mixed upper/lowercase, will it?

lefessan:
Yes, you are right, `BOTH` will only match filenames if they are either full uppercase or full lowercase...

PR #47, opened 2022-06-28
[GCOS] Add support for DEFAULT SECTION

nberth:
Followup of #33

These changes will be rebased on top of `gnucobol-3.x` once the branch is up-to-date, and then edited to move parsing of `DEFAULT SECTION` into the main scanner.

### DIFF
--- a/cobc/parser.y
+++ b/cobc/parser.y
@@ -3342,6 +3342,7 @@ nested_list:
 	depth = 0;
 	setup_from_identification = 0;
   }
+  _control_division		/* GCOS extension */
   source_element_list
 ;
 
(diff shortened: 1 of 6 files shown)


GitMensch:
Just a note: had the option to check the language references of Fujitsu BS2000 (both COBOL 85 and COBOL 2000) and RM/COBOL, they don't support `CONTROL DIVISION` so you may drop the comment in the bs2000-strict.conf and rm-strict.conf files with the follow-up commit.

cobc/parser.y
```diff
@@ -3529,6 +3530,51 @@ _prototype_procedure_division_header:
...
+  _default_display_clause
+  TOK_DOT
+;
+
+_default_accept_clause:
+  /* empty */
+| ACCEPT _is word_or_terminal
+  {
+	  CB_PENDING (_("ACCEPT statement in DEFAULT SECTION"));
+	  /* (void) cb_build_accept_name ($3); */
```
GitMensch, line 3565: This one won't build anything but either change a flag in current_program or a global one.
Please also change to non-translated strings (dropping the `_()`, we don't want the translators to have to msgids for this quite rare feature)

GitMensch (approved):
As soon as those minor parts are changed: good to commit, easy to get another one out and actually helping for the GCOS support mentioned in NEWS already.

GitMensch (approved):
ready for svn when you think it is and passes `make checkall`

Issue #51, opened 2022-07-11
GitHub branch configuration


GitMensch:
I suggest to do the following:
* [change default branch](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/changing-the-default-branch) to gcos4gnucobol-3.x
* [add branch protection](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule#creating-a-branch-protection-rule) to master and gnucobol-3.x

nberth:
I agree the purpose of this repository is mainly about branch `gcos4gnucobol-3.x`.
@lefessan do you think protecting branches `master` and `gnucobol-3.x` would break your mechanics for importing upstream branches?

lefessan:
Unfortunately, I don't think I can protect the branches as the sync process pushes directly into these branches.

I am going to push a base branch (maybe with another name than `gcos4gnucobol`) with the git-related commits.

GitMensch:
Would it be useful for the gnucobol3 repo to re-import (if not possible otherwise) with enabled committer translation? Just have a look the history nbert is not nbert :-)

lefessan:
It's a bit weird, I didn't do anything different from the previous `gnucobol` repo, committer translation was better done the first time. Ok, I will need to investigate what changed and how it can be fixed...

GitMensch:
Ping @lefessan the current mirroring seem to not translate (any more?)
Neither sf-mensch nor ddeclerck are transformed in the git branches

Note: if you haven't the translation from "chaat" to @chuck-haatvedt you may want to add it.

GitMensch:
@lefessan the repo and branches are fine now, only open point is the commiter-id part (and possibly importing old commits/tags, maybe even including the "old" history - but anything from the single upstream svn would already be useful for git blame).

GitMensch:
I'm closing this as I plan moving upstream to git and codeberg in 2026 - this will possibly have committer translation enabled. Afterwards the gitside branches may be moved there as well or a direct mirror be setup...

PR #52, opened 2022-07-11
adding gitpod configuration

GitMensch:
this file alone only "prepares" for use of GitPod, but I consider this useful, too

I suggest to additional do the following, after #51:
* possibly register for GitPod, if not done already
* [apply for "Open Source contribution"](https://www.gitpod.io/for/opensource) (effectively upgrading the gratis account to a professional one without costs)
* login and install the app here, see https://github.com/apps/gitpod-io
* configuring [pre-builds](https://www.gitpod.io/docs/prebuilds)

but as noted, that's "optional"; some benefits of GitPod: nice performance, running without costs, _full_ coding and building possible in the browser from anywhere, when pre-builds are active the development can start nearly in an instant, in theory also easy for people to start coding on the project

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,60 @@
+
+image: gitpod/workspace-c
+
+tasks:
+
+- name: setup coding environment on Ubuntu 20.04
+  before: |
+    # note: sadly we need this to be done every time as only /workspace is kept, but linked
+    #       against those dependencies; and also we do want to recompile after adjustments
+    #       this can all be dropped as soon as we would use a prepared docker
+    sudo apt update && sudo apt upgrade -y
+    sudo apt install -y build-essential libgmp-dev libdb-dev libjson-c-dev ncurses-dev libxml2-dev \
+         automake libtool flex bison help2man texinfo \
+         lcov
+    # sudo apt install gettext texlive-base  # for make dist (po/*, doc/gnucobol.pdf)
+    gp sync-done system-prepare
+
+- name: building GnuCOBOL
+  init: |
+    mkdir -p $GITPOD_REPO_ROOTS/_build
+    cd $GITPOD_REPO_ROOTS/_build
+    gp sync-await system-prepare
+    ../autogen.sh
+    ../configure --enable-cobc-internal-checks --enable-debug --enable-code-coverage
+    make --jobs=$(nproc)
+  command: |
+    cd $GITPOD_REPO_ROOTS/_build
+    gp sync-done build-finish
+
+- name: running GnuCOBOL tests with coverage
+  command: |
+    gp sync-await build-finish
+    cd $GITPOD_REPO_ROOTS/_build
+    half_jobs=$(( $(nproc) / 2 ))
+    nice make --jobs=${half_jobs} check-code-coverage TESTSUITEFLAGS="--jobs=${half_jobs}"
+
+- name: running NIST85 tests
+  command: |
+    gp sync-await build-finish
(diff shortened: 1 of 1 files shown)


GitMensch:
@lefessan @nberth What do you think of this?

A testing _should_ be possible from this URL: https://gitpod.io/#https://github.com/GitMensch/gnucobol-1

(hm, and authorize gitpod, I see that this can be an issue, but maybe you still try that)

GitMensch:
rebased and rechecked, seems fine to me (one _could_ use the vnc base workspace definition to allow running the manual screen tests; but that can be done, later, too

GitMensch:
Juist found out that I've set that to draft months ago... @ddeclerck do you have suggestions compared to the common CI?

ddeclerck:
> Juist found out that I've set that to draft months ago... @ddeclerck do you have suggestions compared to the common CI?

I'm clueless about GitPod...
The proposed configuration seems to "do the job".
Don't know if GitPod allows to upload artifacts (eg. testsuite log) ?

> Would it be reasonable to use the docker you seem to build on/for gitlab? Does it have all the tools installed and is public accessible?

Which docker image are you referring to ?

GitMensch:
> I'm clueless about GitPod...
The proposed configuration seems to "do the job".

With gitpod you don't have any artifact as it opens an online vscode instance connected to a running docker - you have *all* the files ready at hand with the current revision used.

Still - if you have any suggestions then feel free to drop them :-)

> Which docker image are you referring to ?

I've stumbled yesterday over
https://github.com/OCamlPro/gnucobol/blob/21b5d516ffd33e2e25b319680bdcf668b42ff935/.gitlab-ci.yml#L3

Maybe that's something reasonable to use in general, and comes with all the dependencies included?

ddeclerck:
> > I'm clueless about GitPod...
> > The proposed configuration seems to "do the job".
> 
> With gitpod you don't have any artifact as it opens an online vscode instance connected to a running docker - you have _all_ the files ready at hand with the current revision used.
> 
> Still - if you have any suggestions then feel free to drop them :-)

Looks interesting ; I shall give it a try if I find a moment.

> > Which docker image are you referring to ?
> 
> I've stumbled yesterday over
> 
> https://github.com/OCamlPro/gnucobol/blob/21b5d516ffd33e2e25b319680bdcf668b42ff935/.gitlab-ci.yml#L3
> 
> Maybe that's something reasonable to use in general, and comes with all the dependencies included?
> 
> ... self answer when checking further down: the dependencies are installed as pre-script, so reworded question: "Shouldn't those be part of a new image that possibly is based on the other one and includes all dependencies - along with those in the gitpod file here and the ones used in the github ci?"

I wasn't even aware of that GitLab CI 😅. Apparently that was added during early experiments on our internal GitLab instance, before we created the GitHub mirror. The docker image you mention has been made by another team at OCamlPro (I believe it is automatically generated and refreshed from time to time) and is intended for OCaml development. I'd have to check with that team if we can come up with something to use for GnuCOBOL on GitPod.

GitMensch:
@ddeclerck  @lefessan This PR should work as-is already and improve the use of the repo on GitPod. Can you please pull that into the gcos3 and gcos4 branches?

We can come back to adjust this later to care for a possible new base image.
I'll check for the gcos4 one later on as both the CI and the GitPod defintiion should be extended to cater for the multi-io feature.

ddeclerck:
Hi @GitMensch ,
I manually added the configration file to the `gcos4gnucobol-3.x` and `gc4` branches.

PR #54, opened 2022-07-20
Fix handling of quotes in testsuite artifact name

nberth:


GitMensch:
Looks good to me. Just a note: One could also get it directly out of configure.ac (note: the code below is not necessary portable as it uses a GNU grep extension `-A`, your version isn't portable because of the used cpp flags either) - the biggest benefit is that this can be done before running configure.
Working since GnuCOBOL 2.2rc (not with earlier versions, but every one after, too):

`grep "AC_INIT(" -A1 configure.ac | grep -v AC_INIT | sed -e 's/.*\[\(.*\)\].*/\1/'`

nberth:
Indeed. Or even with one less call to `sed` :-)
```sh
grep "AC_INIT(" -A1 configure.ac | sed -e '/AC_INIT/ d; s/.*\[\([^]]*\)\].*/\1/'
```
Anyways, we may assume portability is less of an issue on this CI.

GitMensch:
Using modifiers also allows a match that works also back to OpenCOBOL (needing both GNU grep and GNU sed):

`grep "AC_INIT(" -A1 configure.ac | sed -e ':nl;$!N; s/.*\[\([0-9][^]]\+\)\].*/\1/'`

:-)

nberth:
Oh full GNU sed is available, we can probably even avoid `grep` ;-)
```sh
sed -e '/AC_INIT(/ {z; N; s/.*\[\([^]]*\)\].*/\1/; t}; d' configure.ac
```

PR #55, opened 2022-07-27
Avoid SIGSEGV on SORT of toplevel field

nberth:
Yet, `SORT` on a `01 table` might not be allowed for some dialects. In any case, until now this raised a segfault.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,8 @@
 
+2022-07-27  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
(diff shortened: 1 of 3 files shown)


GitMensch (changes requested):
Good catch - I wanted to write "please commit upstream asap" - but actually the test would fail in EBCDIC machines.

As we not want to check the different sort orders here I suggest to change only use numbers or to check for both

```cobol
IF  VAL NOT = "13abc"  *> ASCII
AND VAL NOT = "abc13"  *> EBCDIC
    DISPLAY VAL.
```

No need to commit that here - you can directly commit it with a fixed test upstream.

nberth:
now upstream

PR #57, opened 2022-08-25
[CANCELLED] Add support for GCOS-specific SELECT

nberth:


cobc/scanner.l
```diff
@@ -776,6 +776,46 @@ H#[0-9A-Za-z]+ {
 	RETURN_TOK (DATA);
 }
 
+"WITH"[ ,;\n]+"ASA"/[ .,;\n] {
```
GitMensch, line 779: Did you get compiler conflicts you could not resolve when adding WITH in front of the options in `with_clause`? Otherwise I wouldn't understand the use of the scanner change.

GitMensch:
Which pages should we look at to check the docs?

cobc/parser.y
```diff
@@ -5318,6 +5338,15 @@ select_clause:
 */
 ;
 
+with_clause:
```
GitMensch, line 5341: Shouldn't that do anything? Here's parsing only which means it possibly needs a PENDING above if that clause is enabled.

GitMensch:
Whenever this is tackled again we should know where the docs for that are exactly and most important need actual tests for this feature.

lefessan:
I just pushed a rebased version of the PR, there is still a lot of work to be done before it is ready, as there are also tests in the current testsuite that are failing with it.

GitMensch:
When/if you feel a need for that change - feel free to provide an updated PR :-)
Now: if possible then optional words should always be handled in the parser.

PR #58, opened 2022-08-25
[CANCELLED] Extended SELECT clauses & ASSIGN statement for GCOS

nberth:
(Tests are still missing, awaiting validation on a GCOS platform)


GitMensch:
Can you please summarize what this statement is used for?
Would it be necessary to have #57 be integrated first or would it work "on its own", too? That definitely would ease the review.

GitMensch:
Ping @nberth (please rebase in any case).

PR #60, opened 2022-09-26
Strict Area A enforcement, and self-dispatching syntax messages

nberth:


cobc/error.c
```diff
@@ -603,21 +634,18 @@ configuration_error (const char *fname, const int line,
 }
 
 /* Generic warning/error routines */
-enum cb_warn_val
-cb_warning_x (const enum cb_warn_opt opt, cb_tree x, const char *fmt, ...)
+static enum cb_warn_val
+cb_warning_x_ (const enum cb_warn_opt opt, cb_tree x, const char *fmt, va_list ap)
 {
-	va_list ap;
 	const enum cb_warn_val pref = cb_warn_opt_val[opt];
```
GitMensch, line 640: It seems that this is duplicate, I suggest to have that code and the "disabled check" only here, not in the caller.

cobc/parser.y
```diff
@@ -314,24 +314,20 @@ static COB_INLINE void
...
 		if (stmt)
-			cb_warning_x (COBC_WARN_FILLER, stmt,
-				      _("'%s' should start in Area A"),
-				      CB_NAME (stmt));
+			(void) cb_syntax_check_x (stmt, _("'%s' should start in Area A"),
+						  CB_NAME (stmt));
 		else
-			cb_warning (COBC_WARN_FILLER,
-				    _("statement should start in Area A"));
+			(void) cb_syntax_check (_("statement should start in Area A"));
```
GitMensch, line 320: I've wondered before: is this really a `stmt`->`statement` or something else (paragraph/procedure?)

tests/testsuite.src/syn_misc.at
```diff
@@ -9077,8 +9077,9 @@ AT_CHECK([$COMPILE_ONLY -fpicture-l=warning prog.cob], [0], [],
 
 AT_CLEANUP
 
+
 AT_SETUP([AREACHECK / NOAREACHECK directives])
-AT_KEYWORDS([directive])
+AT_KEYWORDS([misc directive])
 
 AT_DATA([prog.cob], [
        IDENTIFICATION  DIVISION.
```
GitMensch, line 9085: Please let's drop that test, along with pplex.l scanning for >> AREACHECK - this would be a GnuCOBOL only extension where we support an existing extension for the same purpose.
Sorry for not having catched that (or be clear enough about that) in the originating PR.

tests/testsuite.src/syn_misc.at
```diff
@@ -9220,3 +9226,44 @@ cobol85.cob:14: warning: optional period used
...
+       INPUT-OUTPUT SECTION.
+       FILE-CONTROL.
+       SELECT FILE01 ASSIGN "file01.dat".
+       DATA DIVISION.
+       FILE SECTION.
+       FD FILE01.
+       01 RECORD-FILE01 PIC X(50).
+       WORKING-STORAGE SECTION.
+       01 W-DATA01.
+      * 02 items may still lie in Area A for now:
```
GitMensch, line 9247: The purpose is to adjust this and warn/error there, too; correct?

cobc/error.c
```diff
@@ -780,6 +822,61 @@ cb_error_x (cb_tree x, const char *fmt, ...)
...
+
+	return ret;
+}
+
+/**
+ * dispatches the given message as a warning if cb_relaxed_syntax_checks holds,
+ * as an error otherwise
+ */
+enum cb_warn_val
+cb_syntax_check (const char *fmt, ...)
```
GitMensch, line 849: please use this at other places where `cb_flag_relaxed_syntax_checks` is currently tested - filed.c `duplicate_clause_message` , parser.y and typeck.c on several places.

GitMensch (changes requested):
I'm fine with the wrapped warning/error, but following the other places we should make the internal explicit `xyz_` -> `xyz_internal` (the trailing underscore is also a known naming convention, but so far no/less used in GnuCOBOL).
The new syntax check function should be used for existing cases, too, to stay consistent.

cobc/parser.y
```diff
@@ -314,24 +314,20 @@ static COB_INLINE void
 check_area_a (cb_tree stmt) {
 	if (!cobc_in_area_a && cobc_areacheck) {
```
GitMensch, line 315: Should we still use `cobc_areacheck` here? I think the correct way is to check the format, no?

tests/testsuite.src/syn_misc.at
nberth: We first need to identify dialects where these non-toplevel numbers *must not* be within Area A. If there are some, then a dedicated dialect option is probably in order to refine those checks.

cobc/parser.y
nberth: `cobc_areacheck` itself is set based on the format: in `cobc_set_source_format` (`pplex.l`), or else in the scanner when `#(NO)AREACHECK` is seen, filtered using `cobc_has_areacheck_directive` (`pplex.l` too).

cobc/parser.y
nberth: Hum indeed… in general that can either be a word (section/paragraph name), or NULL (in which case it's typically a statement or a data division entry). So we may get "statement should start in Area A" for data items: for those we may duplicate `check_area_a` and use a specialized message.

cobc/parser.y
GitMensch: Hm, so long as  `cobc -fformat=cobol85 prog.cob` raises the messages (=-fformat also sets the areacheck value) that's fine; this would allow both `cobc -std=cobol85` and `cobc -fformat=cobol85` raising these expected messages and both of these with `-fno-areacheck` to not raise them.

But then `cobc -fareacheck` would do nothing... also fine as long as this is clear from the docs.

cobc/parser.y
```diff
@@ -750,12 +738,10 @@ setup_occurs_min_max (cb_tree occurs_min, cb_tree occurs_max)
 		if (occurs_max != cb_int0) {
 			current_field->occurs_max = cb_get_int (occurs_max);
 			if (!current_field->depending) {
-				if (cb_relaxed_syntax_checks) {
-					cb_warning (COBC_WARN_FILLER, _("TO phrase without DEPENDING phrase"));
-					cb_warning (COBC_WARN_FILLER, _("maximum number of occurrences assumed to be exact number"));
+				if (cb_syntax_check (_("TO phrase without DEPENDING phrase"))) {
+					cb_warning (COBC_WARN_FILLER,
+						    _("maximum number of occurrences assumed to be exact number"));
```
GitMensch, line 743: As note is used in the case above we should use it here, too.

cobc/parser.y
nberth: May renaming the `areacheck` option into `default-areacheck` or `areacheck-by-default` be clearer?

nberth: I mean, as the effect of this option is both: dependent on the format, and can be overridden by `$SET` directives.

GitMensch: > can be overridden by `$SET` directives.

Last time I've checked they were only active if the format has the area a; which is reasonable.

> May renaming the `areacheck` option [...] would be clearer?

possibly yes, not sure about a reasonable name, though (note: we do not necessarily name that "areacheck" at all.

GitMensch:
Rechecked: I've found no single spec that enforces level 78 to be in area a, this only happens with 01 + 77 and _very rarely_ (we can ignore that) for level 66.
That wrong information came from me in the beginning - sorry for not checking it in depth - please fix that within this PR.

cobc/parser.y
nberth: Using `areacheck` in the name of the option has the advantage that it easily relates to the `AREACHECK` directive.

nberth: In any case, I've extended its documentation to state what this entails. Maybe too much for a compiler flag though.

GitMensch (approved):

cobc/parser.y
GitMensch: I think that's OK (I've pushed a minor edit because of gnucobol.pdf). I'm ok with using that name (there are some days left, in case we come up with a better name) ;-)

cobc/parser.y
```diff
@@ -6176,7 +6161,7 @@ _data_division_header:
 | data_division_header
 ;
 
-data: DATA { check_area_a ($1); };
+data: DATA { check_area_a_of ("DATA"); };
```
GitMensch, line 6164: Please use DATA DIVISION here and similar `FILE SECTION`, `COMMUNICATION SECTION`, ...

GitMensch (approved):
The last missing piece is the adjustment of "full" tokens in `check_area_a_of` - otherwise that's fine to commit and close the re-opened issue (you should be able to do that via "Edit").

Thanks again!

nberth:
Now upstream (although. it's a bit late after along day, so I managed to mess up the commit message and mention the FR instead of the commit…)

GitMensch:
:-) I've post-adjusted the commit message. Thanks again for taking care!

PR #61, opened 2022-09-29
Handle many more missing periods before PROCEDURE DIVISION

nberth:
This starts to become more and more *ad hoc*. However, that allows us to deal with much more cases of missing dots than before.

### DIFF
--- a/tests/testsuite.src/syn_misc.at
+++ b/tests/testsuite.src/syn_misc.at
@@ -6195,7 +6195,7 @@ prog.cob:10: error: syntax error, unexpected Identifier
 AT_CHECK([$COMPILE_ONLY -std=mf-strict -freserved=CONSTANT prog.cob], [1], [],
 [prog.cob:6: error: 01 CONSTANT does not conform to Micro Focus COBOL
 prog.cob:7: error: 01 CONSTANT does not conform to Micro Focus COBOL
-prog.cob:7: error: syntax error, unexpected Identifier, expecting . or level-number (Area A)
+prog.cob:7: error: syntax error, unexpected Identifier
 ])
 
 AT_CHECK([$COMPILE_ONLY -std=mf prog.cob], [1], [],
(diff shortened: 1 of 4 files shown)


tests/testsuite.src/syn_misc.at
```diff
@@ -9227,6 +9227,43 @@ cobol85.cob:14: warning: optional period used
...
+       77 PMODE   PIC X
+       PROCEDURE DIVISION.
+          STOP RUN.
+])
+
+# The syntax error below is hard to avoid, and can be missleading.
+# The error is actually on the `77`, which is only parsed as a
+# LEVEL_NUMBER if the preceding token is a `.`.
+#
+# TODO: To avoid this, we need to unput the WORD/Literal when dealing
```
GitMensch, line 9247: Not sure if/what this chagnes, but after the section headers before procedure we could directly parse the error token as an alternative to `_dot` - `_dot_or_error - is the result better?

nberth: That may work fine if we did not have the mechanics to recover using Area A. Otherwise, this work-around needs to be implemented for each kind of section. Still, it may allow the parser to pursue its work a little.

GitMensch: never mind, that was just an idea.

GitMensch:
we've started to go down that road, ... so feel free to commit that now (I have the feeling that there will be "something" to adjust in that area after 3.2-rc1 in any case...)

nberth:
> we've started to go down that road, ... so feel free to commit that now (I have the feeling that there will be "something" to adjust in that area after 3.2-rc1 in any case...)

Yes there's a need for some returns from users having actually encountered related issues to refine the approach. If we observe the needs for *ad hoc* adjustments like this skyrockets, then we may be better off dropping recovery from missing periods via areacheck, in exchange for a more maintainable parser.

PR #63, opened 2022-09-30
New module `cobc/cconv` for EBCDIC & ASCII conversion tables

nberth:


cobc/codegen.c
```diff
@@ -2567,6 +2567,22 @@ output_literals_figuratives_and_constants (void)
 
 /* Collating tables */
 
+#include "cconv.h"
+
+/* Outputs identity conversion if `table == NULL`. */
```
GitMensch, line 2572: please adjust that, something like "from specified table or, when omitted as native output"

GitMensch (changes requested):
Please add tables from https://www.ibm.com/docs/en/iis/11.3?topic=tables-ascii-ebcdic https://www.ibm.com/docs/en/iis/11.3?topic=tables-ebcdic-ascii, all that are currently used, and the gcos one., too.

Question: Will you tackle `-febcdic-table` (= possible selection) in this PR, too?

cobc/cconv.c
```diff
@@ -223,3 +262,19 @@ const cob_u8_t cob_ascii_alt_ebcdic[256] = {
 	0xF0, 0xF1, 0xF2, 0xF3, 0xF4, 0xF5, 0xF6, 0xF7,
 	0xF8, 0xF9, 0xFA, 0xFB, 0xFC, 0xFD, 0xFE, 0xFF
 };
+
+/* Decipher character conversion table names */
+int cobc_deciph_ebcdic_table_name (enum cb_ebcdic_table * const t,
```
GitMensch, line 344: I suggest to return the enum value (and a new `CB_EBCDIC_UNDEFINED` in the error case) here.

cobc/ChangeLog
```diff
@@ -3,6 +3,8 @@
 
 	* codegen.c, cconv.h, cconv.c: extract EBCDIC & ASCII conversion tables
 	  to a dedicated module
+	* flag.def: add -febcdic-table flag to customize EBCDIC translation;
+	  drop -falernate-ebcdic (replaced by -febcdic-table=restricted-gc)
```
GitMensch, line 7: As this is (likely) in use we don't drop that. Th following in cobc.c  (around line 610) should work (possibly with some casting):

```c
	/* alias for backwards-compatibility, removed with 4.x:
	   possibly add to help.c, possibly only output there
	   if verbose mode is used:
	  "alternate-ebcdic use restricted ASCII to EBCDIC translate" */
	{"falternate-ebcdic",		CB_NO_ARG, &cb_ebcdic_table, CB_EBCDIC_RESTRICTED_GC},
	{"fno-alternate-ebcdic",	CB_NO_ARG, &cb_ebcdic_table, CB_EBCDIC_DEFAULT},
```

cobc/cconv.c
```diff
@@ -20,6 +20,9 @@
 */
 #include "cconv.h"
 
+#include <stdio.h>		/* for FILE (FIXME: move to cobc.h?) */
```
GitMensch, line 23: here's that comment again :-)

OK, let's include that in cobc.h, but or now without adjusting other files.

cobc/cobc.c
```diff
@@ -267,6 +268,8 @@ int			cb_flag_dump = 0;
 int			cobc_seen_stdin = 0;
 int			cb_unix_lf = 0;
 
+enum cb_ebcdic_table	cb_ebcdic_table = CB_EBCDIC_DEFAULT;
```
GitMensch, line 271: please do not name an enum identical to a variable and consider to move the variable to cconv.c, external declaration to cconv.h

nberth:
@GitMensch I can anticipate dealing with cases where `COB_EBCDIC_MACHINE` holds being a bit of a challenge. Unless we clearly specify what this actually means in terms of the native encoding… (also I don't think I/we have any simple way to test that).

GitMensch (changes requested):
Looks good to me otherwise.

You may consider adding the `INSPECT CONVERTING alphabet` test in run_misc.at to validate the results of the different tables.

cobc/cobc.c
nberth: So here we go for actual modularization contrary to almost everything else in `cobc`? Not that I'm not ok with that, just that it's a bit of a change ;-)

GitMensch:
> I can anticipate dealing with cases where `COB_EBCDIC_MACHINE` holds being a bit of a challenge. Unless we clearly specify what this actually means in terms of the native encoding… (also I don't think I/we have any simple way to test that).

As noted before that only means that the space is not hex 20 which means "any EBCDIC". I actually expect to find some issues via testcase when running 3.2-rc1 on those machines which I will then fix with either rc2 or 3.2 final.
As long as you don't see that something must be completely broken there - feel free to check that in.

cobc/cconv.c
nberth: Having an explicit undefined in the enumeration goes all the way to `codegen`, where specific error cases are then required on each test of a conversion table. Is this really what we want?

cobc/cobc.c
GitMensch: That was one of the purposes of cconv* in general, so: yes.

This will also be an ongoing work for future changes (and in 4.x we already have split out our biggest player libcob/fileio.c); including the testsuite sources (the next point there is to split run_fileio.at and move out indexed tests to a separate file - because with 4.x where we can have multiple indexed handlers we actually want to run those tests with a single run under each indexed handler that is active.

cobc/cconv.c
GitMensch:
```c
		case 14: {
			/* -febcdic-table=<cconv-table> */
			const enum ebcdic_table cb_ebcdic_table = cobc_deciph_ebcdic_table_name (cob_optarg);
			if (ebcdic_table == SOME_UNDEFINED) {
				cobc_err_exit (COBC_INV_PAR, "-febcdic-table");
			}
			break;
		}
```

:-)

nberth: I mean, every check on `cb_ebcdic_table` in `codegen` need to deal with `UNDEFINED` if that is part of the enumeration type, not just the code that checks the option argument.

nberth: Anyways, with `cb_ebcdic_table` encapsulated in `cconv` that aspect is moot ;-)

nberth: Sure the C way of handling enumerations is quite poor. But I usually operate under the assumption that, in a program where the only possible assignments to an enum variable satisfy some strict discipline (like only using well-declared named constants), then uses of that variable may only need to consider declared cases.

However, that's considerations from a dev that is more used to much higher-level (functional) languages ;-)

GitMensch (approved):
As noted adding to run_misc.at `INSPECT CONVERTING alphabet` to actually verify that the tables are used would be nice, but I'd also say this PR is fine and not only ready for review but for commit (OK, an entry in NEWS about that new option would be good, but if necessary that and test case could be post-added [just makes your commit history less clean as it was before :-) ).

nberth:
now upstream

GitMensch:
Just a note: when checking for copyrights of that module I've found the place where the "alternative" came from: that's https://sourceforge.net/p/gnucobol/history/2540#diff-4 (but sadly neither the VCS log comment nor the Changelog explain much here on that).

GitMensch:
The inclusion of stdio in cobc.h broke VS builds, see https://ci.appveyor.com/project/GitMensch/gnucobol-3-x-vs/builds/44939738 (it may be a good idea to include those in the local CI action ;-)

For now I've reverted this part of the change, which solved that https://ci.appveyor.com/project/GitMensch/gnucobol-3-x-vs/builds/44939945 - we'll have a look at it when inspecting cobc for more separation of modules.

GitMensch:
Just a note: shortly after this was implemented the question came up "what about other EBCDIC encodings" - and the idea to handle the option to `-febcdic-table` - if it doesn't match to one of the default names (or if it includes a period) as a file name; checking if it exists in `COB_CONFIG_DIR` and if it does if the file contains exactly 512 bytes - when yes set the `cb_ebcdic_table` to a new entry `GC_EBCDIC_USER` and store its content as two static variables in ccovnc - to return those for conversions.

I wanted to at least drop that idea here.

PR #66, opened 2022-11-08
DEPENDING var for PIC L data-items can come after parent groups

nberth:
As it turns out, this is supported on GCOS.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,9 @@
 
+2022-11-08  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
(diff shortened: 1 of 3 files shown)


GitMensch:
Does it need to be the last item like ODO in the standard?

GitMensch (approved):
> Does it need to be the last item like ODO in the standard?

To answer myself, looks so, according to the test cases. If those are fine, then I'm with the change, can go into 3.2.dev.

I haven't rechecked, this may be a good time to inspect the syntax checks we have in the testsuite for `PIC L`.

Just "likely again": if you `DISPLAY` the parent then you always see the full contained data, right?

PR #68, opened 2022-11-14
"FREE" source format means X/Open free-form under MF

nberth:
I've recently come across some COBOL source files *apparently* targeting MicroFocus, where comments could be seen as lines starting with a `*` after a `$ SET SOURCEFORMAT"FREE"` directive. This appears to be backed up by https://www.microfocus.com/documentation/visual-cobol/vc50pu7/VS2019/HRLHLHINTR01U008.html .
This PR is a suggestion to handle `*`|`/`-starting lines as comments when the `-fmfcomment` is given, even in FREE reference source format.

The space between `$` and `SET` in the directive is neither documented not supported by GnuCOBOL. I'm not sure whether this is actually allowed.

### DIFF
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -2050,7 +2050,7 @@ clean_up_intermediates (struct filename *fn, const int status)
 		return;
 	}
 	if (fn->need_preprocess
-	 && (status 
+	 && (status
 		||  cb_compile_level > CB_LEVEL_PREPROCESS
 		|| (cb_compile_level == CB_LEVEL_PREPROCESS && save_temps))) {
 		cobc_check_action (fn->preprocess);
@@ -6428,6 +6428,11 @@ print_line (struct list_files *cfile, char *line, int line_num, int in_copy)
 		cfile->listing_on = on_off;
 		/* always print the directive itself */
 		do_print = 1;
+	} else if (cb_flag_mfcomment && CB_SF_FIXED (cfile->source_format) && \
+		   line[0] == '*') {
+		/* When MFCOMMENT holds, asterisk in column 1 means comment line
+		   with listing suppression in fixed format. */
+		do_print = 0;
 	} else if (line_has_page_eject (line, cfile->source_format)) {
 		force_new_page_for_next_line ();
 	} else if (line_has_listing_statement (line, cfile->source_format)) {
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -6428,8 +6428,9 @@ print_line (struct list_files *cfile, char *line, int line_num, int in_copy)
 		cfile->listing_on = on_off;
 		/* always print the directive itself */
 		do_print = 1;
-	} else if (cb_flag_mfcomment && CB_SF_FIXED (cfile->source_format) && \
-		   line[0] == '*') {
+	} else if (line[0] == '*'
+		&& cb_flag_mfcomment
+		&& CB_MFCOMMENT_APPLIES (cfile->source_format)) {
 		/* When MFCOMMENT holds, asterisk in column 1 means comment line
 		   with listing suppression in fixed format. */
 		do_print = 0;
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -3275,12 +3275,12 @@ process_command_line (const int argc, char **argv)
 
 		case 'F':
 			/* --free, alias of `-fformat=free` */
-			cobc_set_source_format (CB_FORMAT_FREE);
+			(void) cobc_deciph_source_format ("FREE");
 			break;
 
 		case 'f':
 			/* --fixed, alias of `-fformat=fixed` */
-			cobc_set_source_format (CB_FORMAT_FIXED);
+			(void) cobc_deciph_source_format ("FIXED");
 			break;
 
 		case 'q':
--- a/cobc/flag.def
+++ b/cobc/flag.def
@@ -169,7 +169,7 @@ CB_FLAG (cb_flag_write_after, 1, "write-after",
 	  "                        * default: BEFORE 1"))
 
 CB_FLAG (cb_flag_mfcomment, 1, "mfcomment",
-	_("  -fmfcomment           '*' or '/' in column 1 treated as comment\n"
+	_("  -fmfcomment           '*' or '/' in column 1 treated as comment with listing suppression\n"
 	  "                        * FIXED format only"))
 
 CB_FLAG (cb_flag_acucomment, 1, "acucomment",
--- a/cobc/flag.def
+++ b/cobc/flag.def
@@ -169,8 +169,8 @@ CB_FLAG (cb_flag_write_after, 1, "write-after",
 	  "                        * default: BEFORE 1"))
 
 CB_FLAG (cb_flag_mfcomment, 1, "mfcomment",
-	_("  -fmfcomment           '*' or '/' in column 1 treated as comment with listing suppression\n"
-	  "                        * FIXED format only"))
+	_("  -fmfcomment           '*' in column 1 treated as comment with listing suppression\n"
+	  "                        * FIXED/COBOL85/VARIABLE format only"))
 
 CB_FLAG (cb_flag_acucomment, 1, "acucomment",
 	_("  -facucomment          '$' in indicator area treated as '*',\n"
--- a/doc/gnucobol.texi
+++ b/doc/gnucobol.texi
@@ -445,26 +445,26 @@ Micro Focus' Variable format.  Identical to the fixed format above
 except for the program-text area which extends up to column 250 instead
 of 72.
 
-@item -fformat=xopen
-X/Open Free-form format.  The program-text area may start in column 1
-unless an indicator is present, and lines may contain up to 80
-characters.  Indicator for debugging lines is @code{D } instead of
-@code{D} or @code{d}.
-
 @item -fformat=xcard
 ICOBOL xCard format.  Variable format with right margin set at column
 255 instead of 250.
 
+@item -fformat=xopen
+X/Open Free-form format.  The program-text area may start in column 1
+unless an indicator is present, and lines may contain up to 255
+characters.  Indicator for debugging lines is ``@code{D }'' instead of
+``@code{D}'' or ``@code{d}''.
+
 @item -fformat=crt
 ICOBOL Free-form format (CRT).  Similar to the X/Open format above, with
 lines containing up to 320 characters and single-character debugging
-line indicators (@code{D} or @code{d}).
+line indicators (``@code{D}'' or ``@code{d}'').
 
 @item -fformat=terminal
 ACUCOBOL-GT Terminal format.  Similar to the CRT format above, with
-indicator for debugging lines being @code{\D} instead of @code{D} or
-@code{d}.  This format is mostly compatible with VAX COBOL terminal
-source format.
+indicator for debugging lines being ``@code{\D}'' instead of
+``@code{D}'' or ``@code{d}''.  This format is mostly compatible with VAX
+COBOL terminal source format.
 
 @item -fformat=cobolx
 COBOLX format.  This format is similar to the CRT format above, except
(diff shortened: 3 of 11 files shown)


nberth:
> `-fmf-comment` adjusts fixed-form only (an alternative would be a `CB_FORMAT_MF_FIXED` which implies this).
> `CB_FORMAT_XOPEN_FFF` does handle "*" and "/" and "D " correctly without additional flags, doesn't it? If not then this should definitely be fixed.

Ok I see. The files I've seen always start with the `SOURCEFORMAT"FREE"` directive, so resorting to X/Open, or adding other formats with a different name, might not solve the issue completely. The alternative is to change the interpretation of `FREE` using as specific flag… not sure this is very clean though. In addition, as the suggested change did not break any test, that might indicate the current tests for COBOL free-form reference-format might not be incomplete (e.g., a `*` at the beginning of a line that need to be interpreted as an arithmetic operation is probably not tested). 

The space-after-`$` issue seems rather easy to fix at first glance; I'll see to prepare a new PR for this. Does this work on MF for any directive or just `SET`?

GitMensch:
> Ok I see. The files I've seen always start with the `SOURCEFORMAT"FREE"` directive, so resorting to X/Open, or adding other formats with a different name, might not solve the issue completely.

Good point, that means that reading the SOURCEFORMAT directive with `-std=mf` (check the dialect flag) should set the `CB_FORMAT_XOPEN_FFF`, similar - if a fixed format for MF is added, which _seems_ reasonable - this should also be selected automatically for `$ SET SOURCEFORMAT FREE` (actually set or implied).
... So that means that `-fixed` / `-free` should also be dialect aware.

nberth:
> Good point, that means that reading the SOURCEFORMAT directive with `-std=mf` (check the dialect flag) should set the `CB_FORMAT_XOPEN_FFF`, similar - if a fixed format for MF is added, which _seems_ reasonable - this should also be selected automatically for `$ SET SOURCEFORMAT FREE` (actually set or implied).
> ... So that means that `-fixed` / `-free` should also be dialect aware.

Poking about the docs I think we need another format to handle this, like a `CB_FORMAT_MF_FREE`, that is very similar to X/Open free-form format but allows lines of 256 single-byte characters (instead of 80 "character positions" for X/Open FFF). As per the MF doc on "FREE" format: "There is no fixed right margin, though for practical purposes this implementation restricts the maximum length of a source line to 256 bytes of source when represented in UTF8 encoding."

Then, to handle `$ SET SOURCEFORMAT"FREE"` properly, maybe it's a matter of defining dialect-specific dictionaries of source format names, with a default and an MF-specific one that remaps "FREE". If this dictionary can then be selected using a dialect option like `source-format-dict=[mf|default]`, we'd need a way to ensure that `-fformat=free -fsource-format-dict=mf` and `-fsource-format-dict=mf -fformat=free` behave the same. I'm not sure whether this is easy to do with the current way configuration options are handled…
A more naive but simpler approach would be to treat `$ SET SOURCEFORMAT"FREE"`, which is an MF extension, differently than the `>>SOURCE` equivalent (with possibly for the MF one a configurable note or warning about the name mismatch).

nberth:
> Nice, works on all I've tested, which includes conditional compilation and SET - just go with all.

One last detail: is any number of space allowed, or a single one?

GitMensch:
> > Nice, works on all I've tested, which includes conditional compilation and SET - just go with all.
> 
> One last detail: is any number of space allowed, or a single one?

Rechecked: both "any number of space" and/or any number of tabs.

GitMensch:
> Poking about the docs I think we need another format to handle this, like a `CB_FORMAT_MF_FREE`, that is very similar to X/Open free-form format but allows lines of 256 single-byte characters (instead of 80 "character positions" for X/Open FFF). As per the MF doc on "FREE" format: "There is no fixed right margin, though for practical purposes this implementation restricts the maximum length of a source line to 256 bytes of source when represented in UTF8 encoding."

As we don't explicit handle "character positions" we should go with "possible bytes per character" - and this would mean 1-4 for UTF8 so I'd say 256 "practical limit" is fine.

> to handle `$ SET SOURCEFORMAT"FREE"` properly, maybe it's a matter of defining dialect-specific dictionaries of source format names

Nah, just check for `cb_std_def` (`CB_STD_MF`) when parsing that, and possibly add to `cb_std_def` as needed.

nberth:
> As we don't explicit handle "character positions" we should go with "possible bytes per character" - and this would mean 1-4 for UTF8 so I'd say 256 "practical limit" is fine.

The current implementation restricts X/Open free-form format to 80 columns. So I was probably wrong when setting that limit, and should have selected 320 instead as this is a count of bytes and not characters.

nberth:
Current changes also start to remove some "free-from" formats from those eligible for Area A enforcement: for some reason I included some terminal-like formats in this set originally. I'm mostly sure about X/Open free-format that mentions it has no Area A. I'll need to check the others (CRT, TERMINAL, and XCARD).

GitMensch (approved):
Looks fine to me. I'm a bit confused about the texi adjustment - please comment on that, but feel free to commit in any case.

GitMensch:
... and I've just recognized that we don't handle MFCOMMENT correctly either - if active it does not only set that in fixed format an asterisk at column 1 means "comment line" - it actually means "comment line with listing suppression".

... not sure if you get to that (all in cobc/cobc.c) part, but wanted to drop it as "related". Could be done here or with a follow-up PR.

GitMensch (approved):
still approved, but we may should support [`$SET DIALECT`](https://www.microfocus.com/documentation/visual-cobol/vc70/DevHub/HRCDRHCDIR2P.html) to do the right based on the sources only, too

doc/gnucobol.texi
```diff
@@ -445,26 +445,26 @@ Micro Focus' Variable format.  Identical to the fixed format above
...
+X/Open Free-form format.  The program-text area may start in column 1
+unless an indicator is present, and lines may contain up to 255
+characters.  Indicator for debugging lines is ``@code{D }'' instead of
+``@code{D}'' or ``@code{d}''.
+
 @item -fformat=crt
 ICOBOL Free-form format (CRT).  Similar to the X/Open format above, with
 lines containing up to 320 characters and single-character debugging
-line indicators (@code{D} or @code{d}).
+line indicators (``@code{D}'' or ``@code{d}'').
```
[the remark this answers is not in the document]
nberth: I added the quotes as when I looked at what `@code{D }` (D followed by a space) rendered in the PDF I found it was not clear enough. So that was an attempt to clarify. The other bits of code do not use the quotes: an alternative would be to just put them for the case with space followed by the clarification in parenthesis.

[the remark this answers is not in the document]
GitMensch: rechecked in the gdb manual: they use [`@samp{D }`](https://www.nongnu.org/texi2latex/texinfo/samp.html) in this case. Let's do the same.

If you find the time to do so an overall change would obviously be better than only doing it here ;-) but I understand if this comes later.

nberth:
> Thank you. Ready for svn if you don't want to add $SET DIALECT before :-)

Just at attempt at better honoring MFCOMMENT w.r.t listings. Please check the tests, as I'm not sure this is indeed the expected behavior.

cobc/cobc.c
```diff
@@ -6428,6 +6428,11 @@ print_line (struct list_files *cfile, char *line, int line_num, int in_copy)
 		cfile->listing_on = on_off;
 		/* always print the directive itself */
 		do_print = 1;
+	} else if (cb_flag_mfcomment && CB_SF_FIXED (cfile->source_format) && \
```
GitMensch, line 6431: note: as the asterisk is most unlikely I'd check that first, then go to the flag and last to the "most complex" check of the source format; also - just tested - cb_flag_mfcomment  is __to be__ applied in all of CB_FORMAT_FIXED, CB_FORMAT_COBOL85, CB_FORMAT_VARIABLE cases, in all other cases it has to be ignored.

[even explicit setting MFCOMMENT after the SOURCEFORMAT"FREE" does not re-enable it]

style only issue:  `\` only for preprocessor lines, and lines starting with tab + space indent + operator

```c
	} else if (line[0] == '*'
	        && cb_flag_mfcomment
	        && CB_SF_FIXED (cfile->source_format)) {
```

Instead of `CB_SF_FIXED (cfile->source_format)` maybe use an `cb_mfcomment_applies` or similar function?

cobc/flag.def
```diff
@@ -169,7 +169,7 @@ CB_FLAG (cb_flag_write_after, 1, "write-after",
 	  "                        * default: BEFORE 1"))
 
 CB_FLAG (cb_flag_mfcomment, 1, "mfcomment",
-	_("  -fmfcomment           '*' or '/' in column 1 treated as comment\n"
+	_("  -fmfcomment           '*' or '/' in column 1 treated as comment with listing suppression\n"
```
GitMensch, line 172: Very good that you've remembered to adjust that, please change the one below to FIXED/VARIABLE format only.

tests/testsuite.src/listings.at
```diff
@@ -3566,6 +3566,73 @@ AT_CHECK([diff prog20.lst prog.lis], [0], [], [])
...
+AT_KEYWORDS([listing])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       PROCEDURE        DIVISION.
+      *    DISPLAY 'COMMENTASTERISK'
+      /    DISPLAY 'COMMENTSLASH'
+*          DISPLAY 'MFCOMMENTASTERISK'
+/          DISPLAY 'MFCOMMENTSLASH'
```
GitMensch, line 3579: There is no MFCOMMENTSLASH thing, neither in MF's old or new docs; and it also is only seen as sequence-number (fixed/variable) or as normal print after line-break indicator in xopen_free.
Please drop that from _all_ tests (I've added it back then because I've seen it in the code - I should have checked the docs and then asked for people with MF access... - instead; but that's now done).

GitMensch (changes requested):
Please adjust pplex,l more:

to handle the code identical to printing and according to the docs

```diff
 	if (source_format == CB_FORMAT_FREE) {
 		bp = buff;
 	} else {
 		if (n <= indicator_column) {
 			/* Line too short */
 			newline_count++;
 			goto start;
 		}

-		if (cb_flag_mfcomment) {
+		if (buff[0] == '*' && cb_flag_mfcomment && ...) {  // same as in listing code
-			if (buff[0] == '*' || buff[0] == '/') {
				newline_count++;
				goto start;
-			}
		}
```

Ideally would be to additional:
auto-enable mfcomment and acucomment according to std; as moving it to dialect would mean to include two items in every file and have it only enabled in one file each maybe just set add a switch in `cb_load_std` and check for the STD after loading the file, then have special cases like this there (or would you prefer the dialect to have all of those at a single place? That _may_ be more clear and would then also list the options in the dialect help; I'm not sure, so if you tackle this: both would be fine with me),
But the "ideal" is no per-requisite for this PR to be ready.

tests/testsuite.src/listings.at
```diff
@@ -3566,6 +3566,73 @@ AT_CHECK([diff prog20.lst prog.lis], [0], [], [])
...
+
+000001
+000002         IDENTIFICATION   DIVISION.
+000003         PROGRAM-ID.      prog.
+000004         PROCEDURE        DIVISION.
+000005        *    DISPLAY 'COMMENTASTERISK'
+
+LINE    PG/LN  A...B............................................................
+
+000006        /    DISPLAY 'COMMENTSLASH'
```
GitMensch, line 3617: and "yes" this is the expected result - here's my test with a level 05 var being hidden:

```
     9 01  T01-ALL.
    11        10 T01-SOMEBER                PIC X(500) VALUES SPACES.
```

The line is just "missing", so that's fine.

GitMensch (approved):
LGTM

nberth:
I think I'll leave auto-enabling `mf/acucomment` for later. Not fan of putting an ad-hoc piece of code in `cb_load_std`, although it's tempting as it is indeed the simplest approach…

nberth:
Now upstream :tada:

PR #69, opened 2022-11-15
Allow spaces in MF-style directives

nberth:
Followup from discussion in #68


GitMensch (approved):
as soon as CI tests are passing - ready

Thanks!

PR #70, opened 2022-12-05
"Safe" partial replacement à la GCOS

nberth:
Partial replacement with literals only apply when it does not lead to an empty word, meaning that, e.g, `REPLACE LEADING "X" BY ""` (or `SPACES`) will apply on ` XX `, but not on ` X `. `REPLACE LEADING "X" BY "Y"` still applies as usual.

In addition, right operand of such a partial replacing must not contain any space or separator.


nberth:
> It is quite confusing to have a mixed merge of a PR and "upstream changes"; which are the commits to inspect for this PR?

Apologies for the mishandling of branches. Github seemed a bit lost when I tried to update the PR against the proper base.

GitMensch:
That's currently marked as draft, what, if any is "open" from your side?

GitMensch:
> These added tests fail similarly with and without the changes of this PR.

Thank you for checking.

> If so, is there a pending patch somewhere about these changes, so I check whether it still applies after this PR?

That would be https://sourceforge.net/p/gnucobol/patches/54/.

nberth:
> > If so, is there a pending patch somewhere about these changes, so I check whether it still applies after this PR?
> 
> That would be https://sourceforge.net/p/gnucobol/patches/54/.

I've just checked: `patch_replacement.patch` still applies fine after this PR, and all tests (including those added by the patch `tests.patch` mentioned above) pass.

GitMensch:
Checked: ACUCOBOL around 1997 does not handle those literals, the version around 2001 does, erroring out (after adding a mandatory section or paragraph) with:

~~~
prog.cob, line 18: Undefined data item: NORM
prog.cob, line 18: Verb expected, NO found
~~~

the current version of cobc says:

~~~
prog.cob: in paragraph 'main':
prog.cob:18: error: 'NORM' is not defined
~~~

So I _guess_ the patch would fix GnuCOBOL -std=gcos and break -std=acu?

cobc/ChangeLog
```diff
@@ -1,4 +1,14 @@
 
+2022-12-05  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* cobc.h (struct cb_replace_src, struct list_replace): adapt internal
+	  structures for strict partial text replacement
+	* cobc.c (print_replace_text, print_replace_main),
+	  pplex.l (set_print_replace_list, ppecho),
+	  ppparse.y (ppp_replace_src, ppp_replace_list_add): handle strict
+	  partial text replacement
+	* cobc.c: replace some incorrect '%d' formats indicators with '%lu'
```
GitMensch, line 10: the format fixing could be checked in directly to minimize the diff, if you want to; this also applies to moving the conversion table in the "if" from https://github.com/OCamlPro/gnucobol/pull/71/files#diff-e2d9f5867efd7382ed31747ef1d5c8c495c20ea34c4e2b58abc7b019e7407f13 - feel free to commit one or both directly

nberth:
> ... but the most important part is: that seems to break the "non-strict" behaviour of ACUCOBOL, no?

Indeed the behavior w.r.t strictness should be different with -std=acu. Instead of adding an additional dialect option that would only be relevant when `-fpartial-replacing-with-literal` is not `unconformable`, maybe a quick but good-enough fix would be to just check for `CB_STD_ACU` before setting a partial match as strict. What do you think?

GitMensch:
> maybe a quick but good-enough fix would be to just check for `CB_STD_ACU` before setting a partial match as strict. What do you think?

seems like something that could be applied soon enough TM, the default strictness should be acu as that came earlier so check for `CB_STD_GCOS` instead.
I suggest to make that adjustment along with the move to tree.h, then either re-request a review or directly commit.

... and ideally directly commit the other partial replace fix from the referenced ticket shortly afterwards.

Thank you very much!

nberth:
> the default strictness should be acu as that came earlier so check for CB_STD_GCOS instead. 

That's what I was thinking. But we did not add `CB_STD_GCOS` to avoid those kinds of hard-coded checks… maybe we need to to that in the end.

GitMensch:
> > the default strictness should be acu as that came earlier so check for CB_STD_GCOS instead.
> 
> That's what I was thinking. But we did not add `CB_STD_GCOS` to avoid those kinds of hard-coded checks… maybe we need to to that in the end.

or, as a hack which is done similar in another case: give `skip` a special meaning like "support, but skip of it isn't partial...

nberth:
> or, as a hack which is done similar in another case: give `skip` a special meaning like "support, but skip if it isn't partial...

Do you mean `-fpartial-replacing-with-literal=skip` would be ACU's behavior of replacing complete matches (i.e. treating the REPLACING source literal as if it was pseudo-text)? This would make GCOS's behavior the default unless that is given `skip`, right?

GitMensch:
> > or, as a hack which is done similar in another case: give `skip` a special meaning like "support, but skip if it isn't partial...
> 
> Do you mean `-fpartial-replacing-with-literal=skip` would be ACU's behavior of replacing complete matches (i.e. treating the REPLACING source literal as if it was pseudo-text)? This would make GCOS's behavior the default unless that is given `skip`, right?

Nearly, more like -fpartial-replacing-with-literal=skip` would be GCOS's behavior of skipping the replacement if it isn't a full match ;-)

nberth:
> Nearly, more like -fpartial-replacing-with-literal=skip` would be GCOS's behavior of skipping the replacement if it isn't a full match ;-)

So the option should rather be named something like, e.g, `strict-partial-replacing-when-literal-src`…

nberth:
> > Nearly, more like -fpartial-replacing-with-literal=skip` would be GCOS's behavior of skipping the replacement if it isn't a full match ;-)
> 
> So the option should rather be named something like, e.g, `full-partial-replacing-when-literal-src`… (but then values like `unconformable` and `obsolete` might not be that clear: the underlying feature is about support of literals, not the semantics on replacement)

nberth:
Turns out I've misread the documentation. Replacement happens "only if it results in a legal word" means the behavior differs depending on whether the right operand is SPACES or just spaces and separators, or whether it is also a valid prefix.

nberth:
@GitMensch I've updated the PR description to better match the issue. In any case, tests on a GCOS appear to show that the right operand of `REPLACE LEADING/TRAILING` must not contain spaces (and maybe separators),

GitMensch:
Just tested with ACU 2001: "Partial replacement with literals" is compiled as-is.
The other test case is totally broken as it does remove the X:
~~~
       IDENTIFICATION   DIVISION.
       PROGRAM-ID.      prog.
       DATA             DIVISION.
       WORKING-STORAGE  SECTION.
       COPY "copy.inc"
            REPLACING LEADING "TEST" BY "FIRST"
                      LEADING "NORM" BY "SECOND".
                      LEADING "NORM" BY "SECOND"
                      LEADING "X" BY "".
       COPY "copy2.inc"
            REPLACING TRAILING "FIRST" BY "VAR1"
                      TRAILING "SECOND" BY "VAR2".
                      TRAILING "SECND" BY "VAR2"
                      TRAILING "Y" BY "".
       PROCEDURE        DIVISION.
           DISPLAY FIRST-VAR NO ADVANCING
           END-DISPLAY.
           DISPLAY SECOND-VAR NO ADVANCING
           END-DISPLAY.
           DISPLAY TEST-VAR1 NO ADVANCING
           END-DISPLAY.
           DISPLAY TEST-VAR2 NO ADVANCING
           END-DISPLAY.
           DISPLAY FIRST-VAR  NO ADVANCING END-DISPLAY.
           DISPLAY SECOND-VAR NO ADVANCING END-DISPLAY.
           DISPLAY X          NO ADVANCING END-DISPLAY.
           DISPLAY TEST-VAR1  NO ADVANCING END-DISPLAY.
           DISPLAY TEST-VAR2  NO ADVANCING END-DISPLAY.
           DISPLAY Y          NO ADVANCING END-DISPLAY.
           STOP RUN.
~~~

Does this mean we do need the unconformable/ok/skip, and does it mean it works "as expected"?

Do we have an option to shorted the dialect option name?


Just out of interest: did you use the `DEBUG_REPLACE`?

cobc/tree.h
```diff
@@ -1590,7 +1590,87 @@ extern void		ppparse_clear_vars (const struct cb_define_struct *);
 extern struct cb_define_struct *ppp_search_lists (const char *name);
 
 extern void		plex_action_directive (const enum cb_directive_action,
-	const unsigned int);
+					       const unsigned int);
+
+/* COPY/REPLACING/REPLACE */
+
```
GitMensch, line 1596: please move the defines to tree.h, too, and use for all lead_trail

```c
/* COPY extended syntax defines */
enum cb_partial_replace {
	CB_REPLACE_COMPLETE	= 0,  /// open for better name 
	CB_REPLACE_LEADING		= 1,
	CB_REPLACE_TRAILING		= 2
};
```

cobc/ppparse.y
```diff
@@ -132,19 +134,53 @@ fold_upper (char *name)
...
 	p = cobc_plex_malloc (sizeof (struct cb_replace_list));
 	p->line_num = cb_source_line;
-	p->old_text = old_text;
+	src->lead_trail = lead_or_trail;
+	if (!lead_or_trail) {
+		/* Strictness flag is irrelevant for non-LEADING nor TRAILING
+		   replacements */
+		src->strict = 0;
+	} else {
+		/* Use replacement text to decide stricness of partial match */
```
GitMensch, line 168: typo: strictness

GitMensch (approved):
approved. please move the define out to an enum in the correct header and consider shortening the dialog option name; feel free to commit afterwards

nberth:
> Does this mean we do need the unconformable/ok/skip, and does it mean it works "as expected"?

No I think `ok` is fine for ACU then.

> Do we have an option to shorted the dialect option name?

Good question. I've yet to find a good name for that…

> Just out of interest: did you use the `DEBUG_REPLACE`?

Yes I've checked this still works.

cobc/tree.h
```diff
@@ -1590,7 +1590,94 @@ extern void		ppparse_clear_vars (const struct cb_define_struct *);
...
+	CB_REPLACE_ALL		= 0,
+	CB_REPLACE_LEADING	= 1,
+	CB_REPLACE_TRAILING	= 2,
+};
+
+/* Strict/loose source text replacement structure */
+struct cb_replace_src {
+	const struct cb_text_list	*text_list; /* single-cell when strict */
+	enum cb_replace			lead_trail;
+	unsigned int			strict: 1;
```
GitMensch, line 1608: Is there any use in packing a single attribute?

nberth: Just declaring that a single bit is used, in case other flags need to be added later on. So let's say this acts like a comment.

cobc/config.def
```diff
@@ -206,9 +206,10 @@ CB_CONFIG_SUPPORT (cb_comment_paragraphs, "comment-paragraphs",
 CB_CONFIG_SUPPORT (cb_control_division, "control-division",
 	_("CONTROL DIVISION"))
 
-CB_CONFIG_SUPPORT (cb_partial_replacing_with_literal,
-		   "partial-replacing-with-literal",
-		   _("partial replacing with literal"))
+CB_CONFIG_SUPPORT (cb_partial_replace_when_literal_src,	/* Enables literals as */
+		   "partial-replace-when-literal-src", /* replacement operands. */
+	_("apply partial replacing with literal source operand even when it replaces with spaces only;\n"
+	  "                        * \"skip\" prevents such replacements."))
```
GitMensch, line 212: I'm fine with the shortened name. The comment is nice - please add that directly in the complete line before the entry.

```
/* Enables literals as replacement operands. */
CB_CONFIG_SUPPORT (cb_partial_replace_when_literal_src, "partial-replace-when-literal-src",
	_("apply partial replacing with literal source operand even when it replaces with spaces only;\n"
	  "                        * \"skip\" prevents such replacements"))
```


the trailing period here and on line 198 is wrong, please drop that

GitMensch:
@nberth I guess that will be checked in soon?

nberth:
Now upstream

PR #72, opened 2022-12-09
New default-colseq flag to configure program's default COLLATING SEQUENCE

nberth:
In the end this relies more on "alphabets" than "collating sequences", so I'm not sure yet whether the approach is appropriate, or if the option is properly named.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,12 @@
 
+2022-12-09  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* flag.def, cobc.c (process_command_line): add new default-colseq flag
+	* tree.h, typeck.c (cb_deciph_default_colseq_name): new helper function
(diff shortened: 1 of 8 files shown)


GitMensch:
I'd say the main question is: should this not be applied in parser.y to base the alphanumeric_collation / default on it?

Or would you in general split file collating sequences (then we'd need a different name)?

And Changelog should likely reference the SF issue.

GitMensch (approved):
LGTM

nberth:
Now upstream.

@GitMensch I've had two failures for the following two tests (not related to this PR):
- System routines for directories (1)
- System routines for files - filename mapping

Failures appear to be random; one of them also happened on https://ci.appveyor.com/project/GitMensch/gnucobol-3-x/builds/45653125/job/7h019vour18kj9mb

GitMensch:
Thanks for the notice, I've just checked in the fix.

PR #73, opened 2022-12-14
Fix REDEFINES GLOBAL

nberth:
To assess whether the test is appropriately placed if into `run_extensions.at`. It's not just syntax-related, so I'm not sure it'd fit into `syn_redefines.at`.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -847,6 +847,11 @@
 	* tree.c: fix bug #772 check for report field missing PIC
 	  when using SUM clause
 
+2021-12-21  Samuel Belondrade <samuel.belondrade@atos.net>
+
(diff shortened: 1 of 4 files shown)


GitMensch (changes requested):
That's no extension but perfectly valid COBOL - see the final DIS 2022 https://isotc.iso.org/livelink/livelink?func=ll&objId=22377830&objAction=Open (If you don't have that I highly suggest to do so).

Place it in either run_fundamental.at or run_misc.at, whatever matches better. I'd likely grep for `AT_SETUP.*GLOBAL` an then check what is most similar.

GitMensch (approved):
LGTM

nberth:
Now upstream

PR #74, opened 2022-12-15
Exhibit issue with `memset` used to assign values that do not fit the picture string

nberth:
Note failure almost surely depends on the (version of the) C compiler.


GitMensch:
> Note failure almost surely depends on the (version of the) C compiler.

Yes, it does not fail on the CI or on gitpod (Ubuntu 20.04, default GCC) either.
It may also depend on the warning options used.

Do you know if the same memset issue exists with 3.1.2 (because we changed code for that in between)?
I'd guess that the issue in each case is in codegen.c, output_initialize_to_value, where the fields's size is not handled.

... ok, I've spot it in the code and will commit the fix along with your testcase.

PR #75, opened 2023-01-15
[WPI] Fix COPY REPLACING and REPLACE

lefessan:
Current implementation does not conform to standard (COPY REPLACING and REPLACE are supposed to be in two successive passes instead of applied together), and has various bugs.

This version is probably less efficient, but better conforms to the standard.

Limitations:

* This first version only modify the replacements during preprocessing, not the ones during the listing printing.

* Since REPLACE are interpreted *before* the COPY REPLACING, they cannot be modified by them. A conforming implementation would interpret REPLACE strictly after COPY REPLACING.


GitMensch:
Note: I guess WPI is "WIP", depending on if you see it ready for GC yourself you may request a review; if you see it as a draft in general, please set it as draft (on the right side on the Conversation page of the PR).

In any case: please keep the ci definitions as extra PR :-)

lefessan:
Yes, it was WIP. I don't know why there is no "draft" button on my interface.

I am playing with the CI at this moment, but yet, I will definitively put the git related stuff in another PR.

lefessan:
I used the CI to check that everything was running correctly. It passes all the tests and I removed the CI commit.

cobc/replace.c
```diff
@@ -0,0 +1,687 @@
...
+/* Whether a word matches the definition of WORD in pplex.l */
+static
+int is_word (WITH_DEPTH const char* s){
+	int i;
+	int len = strlen (s);
+
+
+	for( i = 0; i<len ; i++){
+		char c = s[i];
+		if( c == '_'
```
GitMensch, line 471: Shouldn't this just check for `c != ',' &&  c != ';' && !isspace(c)`?

lefessan: This function is used to merge multiple WORD tokens into a single WORD token. So, I followed the definition of WORD here: https://github.com/OCamlPro/gnucobol/blob/516b813b3ca05ecbc732efdf359ca77db89af449/cobc/pplex.l#L191

GitMensch:
Just a short update: the testcase is fine (I'll add that as expected failure with the next commit) and passes with other compilers.

It would likely be most useful to have this also fixed in the listing code. Maybe that's a similar fix or the new replace.c could be used from both places?
To reach a "conforming" implementation: Would you consider adjusting pplex? Would this mean to iterate twice over the file, with skipping everything but COPY REPLACING to the second iteration?

Concerning the draft question: You should be able to setup it via the UI in the following place:
gh-ui-draft

GitMensch:
When checking the root issue it presented itself as follows:

* `ppecho_replace` is called when any "replace" is active during `ppecho`
* this matches as necessary
* in the case of the given example, where the `REPLACE` code can only be applied after the `COPY REPLACING` code was handled it does not work because:
   * main issue: a match is directly output by `ppecho_direct`; instead it has to "wait" until there is no outer (`r->next`) entry
   * even if it would the outer replace would possibly - as in this case - see only a part of the word `COLON` (because the original `:TEST:` leads to an extra token), while it needs the _resulting_ token
   * as the content can be really long and can't be directly replaced anywhere this may in general only be solved by giving the replaced content back to the lexer - to do so a loop of `unput`, then leaving the functions (and the outer replace, so that it may be triggered in the next run) is to be done; but to catch the sample case here the `unput` would have to be done until last space (separators , and ; are already unput as space here) / line break

lefessan:
I tried to understand how to use this work with the listings, but my main problem is that I don't understand what is expected as the result of listing. Indeed, COPY-REPLACING and REPLACE can completely change the source, including removing complete parts of the code.

If the goal is to have an equivalent source to the original one, but with COPY/REPLACE applied, and then well-formatted, then I would push to use the generated <file.i> and format it back in a fixed format, while re-including comments from the original file when possible. 

If the goal is different, then I don't really understand what it can be, and so how to reach it.

(note that I am not familiar with IBM/MF tooling, so maybe the goal is obvious for people used to them if they have a listing mode...)

GitMensch:
The goal is to have a complete code as it is processed, in the "original style", with COPY and REPLACE applied and all comments, intermixed with diagnostics, applied "print formatting" (pages + headers) and possible suppression (+extra stuff like symbol listing, xref, diagnostic summary, ...).

The _goal_ is to drop nearly all of the listing code in cobc.c, moving some parts (possibly to a new file, possibly to the existing ones) and to do everything during the "pre-processing" itself. Note that we do want to have that available there later also for display errors in their context (instead of "position only, add the context in the text").

... there's just not someone that took on the work to do so...

GitMensch:
That PR is not in draft state, so (while I don't expect this to be available upstream soon) would you mind rebasing it to the current version?

cobc/cobc.h
```diff
@@ -506,6 +508,9 @@ extern void			cobc_parse_free (void *);
 
 extern void			*cobc_plex_malloc (const size_t);
 extern void			*cobc_plex_strdup (const char *);
+extern void			*cobc_plex_strsub (const char *, const int);
```
GitMensch, line 511: `Those and the on above should go to tree.h

cobc/cobc.c
```diff
@@ -985,8 +985,8 @@ cobc_strdup (const char *dupstr)
 	return p;
 }
 
-#if	defined (_WIN32) || defined (__CYGWIN__)
-static char *
+#if        defined (_WIN32) || defined (__CYGWIN__)
```
GitMensch, line 988: this is currently just defined for WIN32 so should not be unconditionally declared in the header (not sure if we need that anywhere outside of this file...)

cobc/cobc.c
```diff
@@ -1315,6 +1315,59 @@ cobc_plex_strdup (const char *dupstr)
...
+#endif
+
+	p = cobc_plex_malloc (len + 1);
+	memcpy (p, s, len);
+	return p;
+}
+
 /* variant of strcpy which copies max 'max_size' bytes from 'src' to 'dest',
    if the size of 'src' is too long only its last/last bytes are copied and an
    eliding "..." is placed in front or at end depending on 'elide_at_end' */
```
GitMensch, line 1373: Should we move out all those memory-only related functions to a separate file?

cobc/pplex.l
```diff
@@ -106,6 +107,9 @@ static int ppwrap (void) {
 
 #define PLEX_COND_DEPTH		16
 
+// replace yytext by some simplified text in ppecho()
```
GitMensch, line 110: please change that to c89 comment and explain why we do that - it replaces all texts below ?!?

cobc/pplex.l
```diff
@@ -2586,204 +2554,11 @@ pp_text_list_add (struct cb_text_list *list, const char *text,
...
-			}
-			check_partial_match = 0;
-			text_queue = queue;
-			continue;
-		}
-	}
-	return (save_ptr ? 1 : 0);
+	/* This new versoin does not use `alt_space`, so the output
+	   will be different from the previous one. */
+	ppecho_copy( text );
```
GitMensch, line 2561: has unused variable warning because now only text is used?

GitMensch (changes requested):
currently I'm finding it hard to follow the changes

Could you please split it into two commits?

First: moves the relevant code from pplex.l to replace.c without any logic change (so the code will also be changed only minimal) and include a new test to listings.at which has a result that will be changed by the second one;
Then have a second commit which does the "actual" change in the already existing files and listing test from the first commit.

GitMensch:
As noted above: whenever this is tackled it should be split into two commits (those may both be in this single PR).

lefessan:
I modified this PR to have 3 commits:
* First one provides a replacement for the alt_space argument, that is cleaner (former version would only use the alt_space if there was no possible replacement, whereas in this version, it is passed around, making the code more uniform)
* Second commit moves the code to replace.h/replace.c
* Third commit implements a completely different algorithm for copy-replacing/replace using two phases instead of one (and using the same code for both phases)

cobc/pplex.l
```diff
@@ -2613,33 +2649,16 @@ ppecho (const char *text, const cob_u32_t alt_space, const int textlen)
...
 	}
 	if (!current_replace_list && !base_replace_list) {
 		/* Output queue */
 		for (; text_queue; text_queue = text_queue->next) {
-			fputs (text_queue->text, ppout);
-		}
-		fwrite (text, (size_t)textlen, (size_t)1, ppout);
-		if (cb_listing_file) {
-			check_listing (s, 0);
+			ppecho_direct(text_queue->text, text_queue->token);
```
GitMensch: replacing that looks correct; but shockingly we seem to have no single testcase for this - can you please duplicate the test below in syn_copy.at for "preprocess listing" (different expected result), so that we see it is not changed unexpected?

cobc/replace.c
```diff
@@ -0,0 +1,313 @@
+/*
+   Copyright (C) 2003-2022 Free Software Foundation, Inc.
+   Written by Fabrice Le Fessant
```
GitMensch, line 3: the date is fine, but the written by must include the original authors mentioned in the original file plus yours

cobc/replace.c
```diff
@@ -0,0 +1,313 @@
...
+static int      ppecho_replace          (struct cb_replace_list *);
+
+static struct cb_token_list *
+pp_token_list_add (struct cb_token_list *list,
+		      const char *text,
+		      const char *token)
+{
+	struct cb_token_list	*p;
+	void			*tp;
+	int text_size = strlen (text);
```
GitMensch: the suggested use of `cobc_plex_strdup` mentioned in the review for the previous commit has the benefit that we don't have the length fields at all (which otherwise should be changed to its real type `size_t`)

cobc/replace.c
GitMensch, line 1: Moving this code out is fine! Please consider use of `cobc_plex_strdup`.

cobc/cobc.c
```diff
@@ -1316,6 +1316,61 @@ cobc_plex_strdup (const char *dupstr)
...
 }
 
+/* Return a newly allocated string with only the first len chars of the first argument */
+void *
+cobc_plex_strsub (const char *s, const int len)
+{
+	void	*p;
+	int	n;
+
+#ifdef	COB_TREE_DEBUG
```
GitMensch, line 1326: the `!s` checks here and in `cobc_plex_stradd` can be removed as we use the parameter(s) to call `strlen` - in the case of `NULL` we'll get a SIGSEGV that will also create a nice coredump or allows us to stop there

cobc/cobc.c
```diff
@@ -1316,6 +1316,61 @@ cobc_plex_strdup (const char *dupstr)
...
+		cobc_abort_terminate (1);
+	}
+	/* LCOV_EXCL_STOP */
+#endif
+	n = strlen (s);
+
+#ifdef	COB_TREE_DEBUG
+	/* LCOV_EXCL_START */
+	if ( len>n ) {
+		cobc_err_msg (_("call to %s with bad argument len=%d>%d=strlen(s)"),
```
GitMensch, line 1340: please ensure that these internal-only messages are normally _not_ gettextized, because a user should never see them and translators also tend to don't want to translate those (and if they are translated and we get bug reports, then we often don't understand it either and need to reverse-lookup the msgid); there are exceptions like the one above - but this is also used _very_ often and is an existing old msgid

tests/testsuite.src/syn_copy.at
```diff
@@ -921,7 +921,7 @@ AT_CLEANUP
 
 
 AT_SETUP([COPY and REPLACE in same file])
-AT_KEYWORDS([copy])
+AT_KEYWORDS([replacing preprocess])
```
GitMensch, line 924: that's better, thanks

tests/testsuite.src/syn_copy.at
```diff
@@ -921,7 +921,7 @@ AT_CLEANUP
...
 
 # see Bug #868
 # the issue with this example is that the outer REPLACE
```
GitMensch, line 927: please adjust that comment

tests/testsuite.src/syn_copy.at
```diff
@@ -951,8 +951,6 @@ AT_DATA([prog.cob], [
        REPLACE OFF.
 ])
 
-AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
-[prog.cob:10: error: 'VAR-COMMA' is not defined
-])
+AT_CHECK([$COMPILE_ONLY prog.cob], [0], [], [])
 
 AT_CLEANUP
```
GitMensch, line 956: as mentioned in the first commit's review: please duplicate that for "in preprocess listing" (or consider including that in the same test, after all it just writes to an _additional_ file which you can have a reference here, then just add `-P preproc.list` and compare the file afterwards.

GitMensch (changes requested):
The Changelog entries for the last two commit are missing. I _guess_ this will also contain a reference to _at least_ one upstream bug issue (Bug 868).

There were some _minor_ change requests, but that's "far enough" to be finished soon and then integrated into RC3.

GitMensch:
Just to check: do you have an ETA?

cobc/pplex.l
```diff
@@ -2592,8 +2599,38 @@ pp_text_list_add (struct cb_text_list *list, const char *text,
...
+	void			*tp;
+	int text_size = strlen (text);
+
+	p = cobc_plex_malloc (sizeof (struct cb_token_list));
+	if (token == NULL) {
+		tp = cobc_plex_malloc (text_size + 1);
+		p->token = NULL;
+	} else {
+		int token_size = strlen (token);
+		tp = cobc_plex_malloc( text_size + token_size + 2);
```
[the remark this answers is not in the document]
lefessan: My understanding is that everything allocated during the preprocessing phase is freed in a single pass once the `.i` file has been generated, and nothing survives, so both data have the same lifetime expectancy.

Since there was a comment that a lot of time was spent in this part of the compiler, I tried to optimize the code with this single allocation and one removing one copy, though this code is probably almost very rarely executed...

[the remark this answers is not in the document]
GitMensch: I understand that idea, but yes, that should be used rarely in any case so going for "more clear" is better.
And yes, that memory is freed all at once.

cobc/pplex.l
```diff
@@ -161,7 +161,14 @@ static struct cb_replace_list	*current_replace_list = NULL;
 static struct cb_replace_list	*save_current_replace = NULL;
 static struct cb_replace_list	*base_replace_list = NULL;
 
-static struct cb_text_list	*text_queue = NULL;
+struct cb_token_list {
+	struct cb_token_list	*next;			/* next pointer */
+	struct cb_token_list	*last;
+	const char		*text;
+	const char		*token;
```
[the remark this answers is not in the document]
lefessan: done

cobc/replace.c
lefessan: done

cobc/replace.c
```diff
@@ -0,0 +1,313 @@
...
+		return p;
+	}
+	list->last->next = p;
+	list->last = p;
+	return list;
+}
+
+void cb_free_replace( void )
+{
+               current_replace_list = NULL;
```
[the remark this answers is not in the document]
lefessan: done

cobc/pplex.l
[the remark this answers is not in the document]
lefessan: ok, I used `cobc_plex_strdup` in the latest version.

cobc/cobc.c

lefessan:
In this new version, I think I took most of your comments into account. Just two things:

* In fact, after rereading bug 868, it looks like it is not actually a bug, as a single copy-replacing should not perform multipass (its copy-replacing and replace that have to be in two different passes). So, actually, this PR is not fixing 868, but a part of bug 831 (the part that is not related to listing).

* About the test with `-P`, I added it, together with tests with `-t` and `-T`, but I am not sure at all that the outputs are the ones expected. I need to read more doc about `-P` to understand what it should actually perform.

lefessan:
I also added a new flag `-fno-ttime`, to suppress time information from listings (i.e. version and date fields). The purpose is to be able to easily compare listings, for example in the testsuite (instead of using UNIFY_LISTING).

lefessan:
> * About the test with `-P`, I added it, together with tests with `-t` and `-T`, but I am not sure at all that the outputs are the ones expected. I need to read more doc about `-P` to understand what it should actually perform.

Looking at the code, I don't really understand what `-P` is expected to do, actually. The generated file does not currently contain the text of the file. All tokens are passed to `check_listing()`, but this function does not output the token, unless `COB_INTERNAL_XREF` is defined. Also, there is a variable `requires_new_line` that is never set in any part of the code.

lefessan:
The current test with `-P` gives the following result:
```
AT_CHECK([$COMPILE_ONLY -P=preproc.lst prog.cob], [0], [], [])
AT_CHECK([cat preproc.lst], [0],
[     1      2       3       4       5       6       7       8      9      10     11      12          13          14          15  
],
[])
```

It does not pass on Macos/Windows, probably because of CRLF problems. I am removing it for now.

GitMensch:
> I also added a new flag `-fno-ttime`, to suppress time information from listings (i.e. version and date fields). The purpose is to be able to easily compare listings, for example in the testsuite (instead of using UNIFY_LISTING).

That is a reasonable approach and helps a lot to keep the testsuite clean, but that should be kept out of this PR; please move that to a separate PR and adjust the implementation:

* `-fno-ttimestamp` (minor rename, but with getopt a user can still abbreviate that) will "suppress timestamp in listing headers"; note: this may provide more space for the title (not sure if we'd use it anywhere)
* new option, which also fixes the TODO from `set_standard_title()`:
```c
CB_FLAG_RQ (cb_listing_with_title, 1, "ttitle",
	_("  -fttitle=<title>      set listing title;\n"
	  "                        defaults to package name and version"))
```
* adjust the testsuite - only use UNIFY_LISTING in one place, in all the others set `-fttitle="GnuCOBOL V.R.P" -fno-ttimestamp` (the best option is possibly to add a new $COMPILE_LISTING to atlocal which uses `$COMPILE_ONLY -fttitle="GnuCOBOL V.R.P" -fno-ttimestamp -tlines=0` an use that.

It may also contain the open points of[ FR #294](https://sourceforge.net/p/gnucobol/feature-requests/294/) (where I've just added the spec above, so you can reference this):
* in the listing header: show only the filename of the processed source, even when it is specified as path (reason: when a full path is specified the source file name is often truncated completely)
* after the first listing header: show the complete name for the source (as specified on command line), with line continuation if necessary

GitMensch:
> The current test with `-P` [...] does not pass on Macos/Windows, probably because of CRLF problems. I am removing it for now.

Please don't remove that but comment the `AT_CHECK` out with a TODO note. Hint: https://github.com/OCamlPro/gnucobol/compare/669403e2f5cc561885be70d1ce1af9ebbdfb69f5..d627bdae090c590a0a851d96739cd34b291e5c91 shows that preproc2 was also dropped

GitMensch:
> Also, there is a variable `requires_new_line` that is never set in any part of the code.

This came in with the OC2.0 "commit" that added Roger's change (rev. 68); this already has this never set, so please drop all its uses (potentially with the listing commit, but that is fine with any commit)

GitMensch:
> > Also, there is a variable `requires_new_line` that is never set in any part of the code.
> 
> This came in with the OC2.0 "commit" that added Roger's change (rev. 68); this already has this never set, so please drop all its uses (potentially with the listing commit, but that is fine with any commit)

Don't mind, I'll do that just now along with fixing the code that code was broken by [@nberth's change in r4644](https://github.com/OCamlPro/gnucobol/commit/9693f670f18b428f3cefb4b5f9c3a3652e19f98b#diff-45eef141c84fba6c7efc1d684a8f8f93603be89315a6ff4af203e904d37edfce) for "pplex.l (check_listing): do not output sequence number of short lines"

lefessan:
Is there anything more to do on this PR ? The `-P` option seems to be working correctly now.

cobc/ChangeLog
```diff
@@ -1,4 +1,22 @@
 
+2023-07-05  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* replace.c: rewrite the code for preprocessing with a two-phase
+	algorithm. The first phase performs COPY REPLACING on the stream
```
GitMensch, line 5: two space indent ;-)

cobc/replace.c
```diff
@@ -0,0 +1,841 @@
...
+#define MAX_TEXT_LIST_STRING 10000
+char text_list_string[MAX_TEXT_LIST_STRING];
+
+static
+char * string_of_list(const struct cb_text_list *list)
+{
+	int pos = 1;
+	text_list_string[0] = '[';
+
+	for(; list != NULL; list = list->next){
```
GitMensch, line 177: please add a comment to this code and/or function doc, from looking at it I have no clue what this may does

cobc/replace.c
```diff
@@ -0,0 +1,841 @@
...
+}
+
+static
+int is_leading_or_trailing (WITH_DEPTH int leading,
+			    const char* src_text,
+			    const char* text,
+			    int strict)
+{
+
+	size_t src_len = strlen (src_text);
```
GitMensch, line 319: The length  variables can be const, no? Do we need to strlen on each access of these or do we have the size in the tructs from the caller?

cobc/replace.c
```diff
@@ -0,0 +1,841 @@
...
+#endif
+#include <ctype.h>
+#include <time.h>
+#include <limits.h>
+
+#include "cobc.h"
+#include "tree.h"
+#include "replace.h"
+
+#define TRUE  1
```
GitMensch, line 43: Please don't use TRUE/FALSE pseudo-boolean in 3.x; just go with 0 and 1 in the code for now and in most cases instead of storing the result to 0/1 you can directly `return 1`.

tests/testsuite.src/syn_copy.at
```diff
@@ -954,6 +946,51 @@ AT_DATA([prog.cob], [
        REPLACE OFF.
 ])
 
-AT_CHECK([$COMPILE_ONLY prog.cob], [0], [], [])
+AT_CHECK([$COMPILE_ONLY -P=preproc.lst prog.cob])
```
GitMensch, line 949: Not sure - can we use` `-P-` as with the listings? If not it likely won't be much work to copy that over...

GitMensch (commented):
minor things to consider, but overall LGTM

cobc/ChangeLog
```diff
@@ -1,4 +1,22 @@
...
+	of tokens, while the second phase perform REPLACE on the resulting
+	stream of tokens. This rewriting is closer to the COBOL standard
+	and fixes bug #831 partially.
```
GitMensch, line 8: Hm, why only partially? []

cobc/replace.h
```diff
@@ -0,0 +1,45 @@
...
+extern void     cb_free_replace (void);
+
+/* For COPY-REPLACING */
+extern void cb_set_copy_replacing_list (struct cb_replace_list *list);
+extern struct cb_replace_list * cb_get_copy_replacing_list (void);
+
+extern void
+cb_set_print_replace_list	(struct cb_replace_list *);
+
+static COB_INLINE COB_A_INLINE int
```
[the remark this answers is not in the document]
lefessan: I removed `replace.h`. To be honest, I am not very comfortable with having these huge header files, I usually prefer to have every `FILE.c` source file should have its `FILE.h` header file.

cobc/ChangeLog
lefessan: done

cobc/replace.c
lefessan: It prints a list of tokens/texts into a simple string, using always the same statically allocated string... but only when the DEBUG macro is set for debugging.

cobc/replace.c
lefessan: Why not ? I am not sure to understand why these (local) macros are a problem in 3.x

lefessan: (anyway, I removed them in the last commit)

tests/testsuite.src/syn_copy.at
lefessan: I added a commit for that.

cobc/ChangeLog
lefessan: #831 actually describes several bugs. One is in listing mode (and is addressed in PR #92 ) and another is in source mode (the one addressed in this PR). I was more concerned by the second part, as it prevents correct code from being compiled, while the first part is only a problem when using the listing mode (we advised our devs to look at the .i files instead of the listing files to exactly know the result of preprocessing).

GitMensch:
That is an option, but should be tackled "in general" and it's more a 4.x thing.

GitMensch:
Only because this should be consistent between these files and 4.x should use boolean types ("backported" via gnulib for old compilers).

cobc/ChangeLog
GitMensch: Only because this should be consistent between these files and 4.x should use boolean types ("backported" via gnulib for old compilers).

cobc/replace.h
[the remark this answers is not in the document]
GitMensch: That is an option, but should be tackled "in general" and it's more a 4.x thing.

cobc/ChangeLog
```diff
@@ -97,8 +97,8 @@
 	* typeck.c (cb_build_expr): remove cb_ prefix from static functions
 	  and comment algorithm
 
-	Bugs #875 V IS ZERO AND, #880 error compiling abbreviated conditions
-	and #887 error compiling parenthesized relation
+	  Bugs #875 V IS ZERO AND, #880 error compiling abbreviated conditions
```
GitMensch, line 100: this is actually a "topic note", those are recognizable because they don't include the leading spaces... (the others are fine now)

GitMensch:
Apart from the missing Changelog entry for -P (this may even be NEWS worthy) and a NEWS entry about "cobc now uses a two-pass algorithm..." that is ready for SVN.

cobc/replace.c
GitMensch: Only because this should be consistent between these files and 4.x should use boolean types ("backported" via gnulib for old compilers).

cobc/ChangeLog
GitMensch: Would https://github.com/OCamlPro/gnucobol/pull/92 solve the listing part already?

GitMensch (approved):
good to go, listing may also be fixed after RC3

cobc/replace.c
```diff
@@ -0,0 +1,852 @@
...
+				const char* text = texts->text;
+				texts = texts->next;
+				if ( is_space_or_nl(text[0]) ){
+					/* skip spaces in texts */
+					check_replace_all (MORE_DEPTH repls,
+							   new_text,
+							   texts, src,
+							   replace_list);
+				} else {
+					if (!strcasecmp(src_text,text)){
```
GitMensch, line 530: @lefessan We do have the text len(yylen) in the scanner, wouldn't it make sense to pass this and include it in the struct?
Then we can compare the length itself and don't need to use strcasecmp for all content (which is done as soon as at least one REPLACE[ING] is active, no?

lefessan: It's not obvious if the speedup would be worth the work. It's likely that, for most words, `strcasecmp(...)` immediately returns after checking only one or two chars. So, on one hand, testing the size would save some tests when the sizes are different, but on the other hand, if the sizes are equal, it adds one test that was not done before.

cobc/ChangeLog
lefessan: I haven't found the time to check yet...

lefessan:
Merged, thanks !

cobc/replace.c
GitMensch: Good point!

GitMensch:
Would you mind fixing the vs builds? The new file has to be added to all files under build_windows that reference help.c.

GitMensch:
Does this also fix https://sourceforge.net/p/gnucobol/bugs/23? If yes then it is likely good to check and possibly reference it.

GitMensch:
> Would you mind fixing the vs builds? The new file has to be added to all files under build_windows that reference help.c.

Done upstream.

GitMensch:
Primarily a note: it seems that this PR has increased memory usage of cobc during preparing a lot (testing with sources that have ~500000 LOC and no REPLACE or REPLACING at all).

hotspots

The main cpu instructions and memory are to be seen in replace.c (which is logically, we also moved old memory allocations there).

Biggest hotspot:
add_text_to_replace

Questions: Could you recheck for possible optimizations, especially reducing memory when no REPLACE is active?

Note: cpu-time wise that's no problem, with this big source there is a cpu time of 1.2 seconds for the preparsing, which is ok. But memory-wise - I'm not sure.

GitMensch:
Friendly ping @lefessan on the memory usage/performance, especially when there's no REPLACE active in the whole file.

GitMensch:
@lefessan you may have missed that: this PR has increased memory usage of cobc during preparing **a lot**; can you please reinspectif you find a better way to handle that?

GitMensch:
Here some numbers from a huge (generated) COBOL file, from valgrinds recorded heap usage

total heap usage | allocs | frees | bytes allocated
--------------------|---------|-------|------------------
GC 3.1.2                   | 5,526 |  5,526 | 20,443,319
GC 3.2 (after this PR) | 21,677,268 | 21,677,268 | 893,378,306

PR #77, opened 2023-01-20
New options to deal with character encodings

ddeclerck:
This PR adds some support for character encodings in different places, using iconv to build conversion tables.

Note to @GitMensch : I still have to add some tests, but you can start reviewing.

It introduces the following new options :
-fnative-charset
-ftarget-charset
-fcodeset-ascii
-fcodeset-ebcdic
-fconvert-ascii
-fconvert-ebcdic

The -fnative-charset and -ftarget-charset allow to specify the encodings used in the execution environment (native) and in the environment we are trying to simulate (target). When they are both set, the collating tables (for the purpose of sorting only) are built by combining these two encodings and the "vendor" EBCDIC/ASCII table.

Our typical use case is the following. We simulate a GCOS environment with EBCDIC-500 encoding (our target). We run on a standard Linux PC, where we decided that all the data we process would use the ISO-8859-1 encoding (we call this "native"). We have a file with some records we'd like to sort according to the EBCDIC collating sequence (the default one on this platform), so that our program sorts in the exact same order it would if it were run on our GCOS environment. If we use the "vendor" EBCDIC/ASCII table (the one from the GCOS documentation), this will fail: this table does not define an EBCDIC-500 <-> ISO-8859-1 translation (in particular, extended characters are just given consecutive values).

That's where the -fnative-charset and -ftarget-charset are useful. Here, we set -fnative-charset to ISO-8859-1 and -ftarget-charset to IBM500 (iconv's name for EBCDIC 500). As they are both set, we will call iconv to produce an ISO-8859-1 to EBCDIC-500 translation table (cob_colseq_native_ebcdic) and give that one instead of the vendor table to the sort function.

Now, what if the SORT instruction contained a COLLATING SEQUENCE IS ASCII clause ? Here, we would want to sort as our GCOS system would, hence using the vendor table. But iconv does not know about the specific ASCII encoding used by GCOS. What we want to do is thus to convert from ISO-8859-1 to EBCDIC-500 and then use the vendor table to translate that EBCDIC to the GCOS-specific ASCII. To avoid performing two conversions, we simply build a new conversion table (cob_colseq_native_ascii) that combines the two steps.

The implementation has been made generic enough so that it works on any combination of native/target encodings among the EBCDIC and ASCII encodings ; in particular, it will use the vendor table in one direction or the other depending on the actual target encoding (i.e. if the target encoding is EBCDIC, it will add a vendor EBCDIC -> ASCII translation when requesting an ASCII collating sequence, and symmetrically, if the target encoding is ASCII, it will add a vendor ASCII -> EBCDIC translation when requesting an EBCDIC collating sequence).

That's for the main stuff.

Now, since I had started digging about encodings, I thought it could also be useful to be able to specify the ASCII and EBCDIC encodings used in the CODE-SET and CONVERTING/REPLACING clauses. That's the purpose of the -fcodeset-ascii, -fcodeset-ebcdic, -fconvert-ascii and -fconvert-ebcdic options.

The -fcodeset-ascii and -fcodeset-ebcdic work together with the -fnative-charset option. When specified, and when a CODE-SET clause is active on a file, the necessary conversions will be performed by iconv using the specified encodings instead of instead of the vendor table. For instance, reading from a file that has CODE-SET IS EBCDIC will perform a translation from the encoding specified by the -fcodeset-ebcdic option to the encoding specified by the -fnative-charset option. Could be convenient if migrating an application from a mainframe while keeping data files in an EBCDIC encoding.

As for the -fconvert-ascii and -fconvert-ebcdic options, they allow to specify the conversions to perform in INSPECT CONVERTING / REPLACING and TRANSFORM CONVERTING.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -213,6 +213,14 @@ static unsigned int		need_save_exception = 0;
 static unsigned int		gen_nested_tab = 0;
 static unsigned int		gen_ascii_ebcdic = 0;
 static unsigned int		gen_ebcdic_ascii = 0;
+static unsigned int		gen_colseq_native_ebcdic = 0;
+static unsigned int		gen_colseq_native_ascii = 0;
+static unsigned int		gen_convert_ebcdic_ascii = 0;
+static unsigned int		gen_convert_ascii_ebcdic = 0;
+static unsigned int		gen_codeset_ascii_native = 0;
+static unsigned int		gen_codeset_native_ascii = 0;
+static unsigned int		gen_codeset_ebcdic_native = 0;
+static unsigned int		gen_codeset_native_ebcdic = 0;
 static unsigned int		gen_native = 0;
 static unsigned int		gen_custom = 0;
 static unsigned int		gen_figurative = 0;
@@ -2540,6 +2548,60 @@ output_literals_figuratives_and_constants (void)
 
 /* Collating tables */
 
+static int
+use_colseq_charsets(void)
+{
+	if (cb_native_charset != NULL && cb_target_charset != NULL &&
+	    strcasecmp(cb_native_charset, cb_target_charset) != 0) {
+		return 1;
+	} else {
+		return 0;
+	}
+}
+
+static int
+use_convert_charsets(void)
+{
+	if (cb_convert_ascii != NULL && cb_convert_ebcdic != NULL) {
+		return 1;
+	} else {
+		return 0;
+	}
+}
+
+static int
+use_codeset_charsets(int alphabet_type)
+{
+	if (cb_native_charset != NULL &&
+	    ((alphabet_type == CB_ALPHABET_ASCII && cb_codeset_ascii != NULL &&
+	      strcasecmp(cb_native_charset, cb_codeset_ascii) != 0) ||
+	     (alphabet_type == CB_ALPHABET_EBCDIC && cb_codeset_ebcdic != NULL &&
+	      strcasecmp(cb_native_charset, cb_codeset_ebcdic) != 0))) {
+		return 1;
+	} else {
+		return 0;
+	}
+}
+
+static int
+need_codeset_conversion(int alphabet_type)
+{
+	if (use_codeset_charsets(alphabet_type)) {
+		return 1;
+	} else {
+		switch (alphabet_type) {
+#ifdef	COB_EBCDIC_MACHINE
+		case CB_ALPHABET_ASCII:
+#else
+		case CB_ALPHABET_EBCDIC:
+#endif
+			return 1;
+		default:
+			return 0;
+		}
+	}
+}
+
 /* Outputs conversion from given table, or a native conversion (identity) when
    omitted (if table == NULL). */
 static void
@@ -2547,13 +2609,25 @@ output_colseq_table (const char * const table_name,
 		     const cob_u8_t table[256])
 {
 	int i;
-	output_storage ("static const unsigned char\t%s[256] = {", table_name);
+	output_storage ("static const cob_u8_t\t%s[256] = {", table_name);
 	for (i = 0; i < 256; i++) {
 		output_storage ("%s%#04x,",
 				i % 8 == 0 ? "\n\t" : " ",
 				table ? table[i] : i);
 	}
 	output_storage ("\n};\n");
+}
+
+static void
+output_colseq_table_empty (const char * const table_name)
+{
+	output_storage ("static cob_u8_t\t%s[256];\n", table_name);
+}
+
+static void
+output_colseq_table_ptr (const char * const table_name)
+{
+	output_storage ("static const cob_u8_t *\t%s = NULL;\n", table_name);
 }
 
 static void
@@ -2567,6 +2641,7 @@ output_colseq_table_field (const char * field_name, const char * table_name)
 static void
 output_collating_tables (void)
 {
+	/* Generic conversion tables */
 	if (gen_native) {
 		output_storage ("\n/* NATIVE table */\n");
 		output_colseq_table ("cob_native", NULL);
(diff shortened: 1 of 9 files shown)


GitMensch (commented):
The general idea is good and I've not seen something "too strange" in the changes (other than for this to be usable iconv have to be checked in configure and the "not available" case be handled).
Just by looking "from the distance" we have the following scenarios:

* collation (where we have the system specific "native" collation, which we may want to override, and the actual relation of a referenced "name" (like `EBCDIC`) to a "value")
    * comparison at compile-time to resolve constant expressions (one could apply a hack and always postpone this to runtime if a collation would apply to this comparison)
    * comparison at runtime in general (specified with the `COLLATING SEQUENCE`) - this is on all conditions (including the implicit comparisons done by the `CONTROL` clause in report writer)
    * comparison during `SORT` (both file sort and table sort) (and `MERGE`), which allows an _override_ of the one above
    * file collation on `INDEXED` files (which possibly needs support in the external API [I _do_ think that BDB for example allows to specify this] by  `COLLATING SEQUENCE` in the file control entry
    * post-added when checking the standard: for intrinsic functions `LOCALE-COMPARE` or `STANDARD-COMPARE`
* conversion:
    * on a file to convert between "external" data and "processed data" (the `CODESET` clause)
    * usable with `INSPECT`
    * between alphanumeric (which may be any single-byte character set) and national (which is either UTF-16 [assumed by GnuCOBOL] or UTF-32 / UCS-4) data; between these two (three) and UTF-8; for example with the intrinsic functions `NATIONAL-OF`, `DISPLAY-OF`, `CONVERT` function and for UTF-8 <-> others during `MOVE`
    * in the compiler when reading source files in an encoding other than the system's current one (for example reading source files that were binary transferred from big iron, or converting literals stored in UTF-8 as two bytes in the source to alphanumeric data items as one byte, or converting extended/national character data into UTF-8 for either setting or compare with utf-8 data)
* possible conversions would be needed for
    * `ACCEPT`/ `DISPLAY` (if the terminal uses a different encoding than the program)
    * possibly for `LINKAGE` between programs that do not share the same collation (this is likely no issue if the internal data is always "native")
* extra: validation in class conditions "identifier IS NOT EBCDIC"  and  "identifier is ALPHABETIC-LOWER"

Likely these are not all handled with this PR ;-)

We should split the support for handling these scenarios (did I forget something?) into multiple PRs as long as each one:
* does not seem to make the later harder (a rough plan for all of those up-front would help)
* has a practical value to be available before the others
* comes with its own Changelog entries and tests

When writing this I _feel_ that the most beneficial changes we can do first, which benefit the users (a different amount and set of users, but still) are the following - which all needs a conversion at runtime:
1. implementing of the intrinsic conversion functions
2. specified `CODESET` conversion for files
3. handling specified collation
4. support to convert on the fly during `DISPLAY` / `ACCEPT` (between alphanumeric/national "native"/"specified" and the terminal), this would possibly need (additional to conversion) wide character input/output in screenio.c
5. support for UTF-8 per IBM (that is one of the bigger things, but _likely_ would benefit most users)

The question is: Can any of the above be at least improved without adding the iconv dependency? Because adding a new dependency make it definitely a GnuCOBOL 4 feature (at least "first", and only for the case that we may create a GnuCOBOL 3.3, which - if it does - won't happen before the first GC4 preview release, and possibly not before a proper V-ISAM release, too).

cobc/codegen.c
```diff
@@ -2575,40 +2650,118 @@ output_collating_tables (void)
...
 			output_line("f_ebcdic_ascii.data = (cob_u8_ptr)cob_ebcdic_ascii;");
 		}
+		if (gen_ascii_ebcdic > 1) {
+			output_line("f_ascii_ebcdic.data = (cob_u8_ptr)cob_ascii_ebcdic;");
+		}
+	}
+
+	/* Charset-aware conversion tables for COLLATING SEQUENCE */
+	if (gen_colseq_native_ebcdic > 0 || gen_colseq_native_ascii > 0) {
+		output_line ("cob_build_collation(\"%s\", \"%s\", \"%s\", %s, %s);",
```
GitMensch, line 2738: The return code of this function is ignored - should this be done? Similar below.

ddeclerck:
> The question is: Can any of the above be at least improved without adding the iconv dependency?

Well, if we restrict to EBCDIC <-> 8-bit ASCII translations (mostly ISO-8859-*) (and possibly UTF-* / UCS-*), it guess it would be acceptable to just copy the tables we need from iconv (the lib itself is LGPL, as GnuCobol runtime). Otherwise, if we'd like to support all kind of encodings, then we would end up just reimplementing iconv :p.

Anyways, the problem right now is that our client only needs these tables to get the SORT working as on his mainframe (I added support for CODE-SET and CONVERTING as a "bonus"), so we can hardly spend extra time implementing all the scenarios (if we had unlimited funding, I'd be happy to investigate all of these). Maybe we can think of a way to implement this quickly without impairing our ability to implement the remaining scenarios ?

GitMensch:
Then what about only adding the necessary tables hard-wired for now? This could get into 3.2 final if it is quick and clean, too.

ddeclerck:
> Then what about only adding the necessary tables hard-wired for now? This could get into 3.2 final if it is quick and clean, too, and provide others with a hint how to add their own translations.

So you mean, adding another translation table use the current (-febcdic-table) option ?

GitMensch:
Yes, that's what I can offer for 3.2 inclusion. Please add a comment on how you created that table.

... thinking again: even better would be to allow -febcdic-table to read in a file (checking just the length would be fine) and store it in the module.

ddeclerck:
> Yes, that's what I can offer for 3.2 inclusion. Please add a comment on how you created that table. ... thinking again: even better would be to allow -febcdic-table to read in a file (checking just the length would be fine) and store it in the module.

That would probably solve the problem in the easiest possible way. I'll do that in another PR.

GitMensch:
Given  `-fdefault-colseq` and #78  - where does this lead us concerning the issues above?

ddeclerck:
Well, #78 and `-fdefault-colseq` do solve our current issue, yet the points raised here remain valid. Let's keep that here as a draft for reference for now.

GitMensch:
@ddeclerck  Can you please rebase this and then review what's missing feature-wise so far?

Ideally you could add / adjust some testcases, allowing everyone to grasp more easy what the new options are to be used for.

ddeclerck:
Hi @GitMensch,

Actually this PR is rather obsolete ; the new option `-fdefault-colseq` (#72 #101) and the modifications to allow using custom translation tables in `-febcdic-table` (#78) are as useful and simpler to use (and easier to implement).

GitMensch:
Its your PR, so if all things from the PR's initial post are covered otherwise, then feel free to close this (maybe adding a new label obsolete or similar, so its easier to distinguish from the closed "merged upstream" ones).

One thing that I've seen which was the reason to check the PRs: I'm not sure if `-fdefault-colseq` is applied correctly for `SPECIAL-NAMES. ALPHABET nat IS NATIVE.` - it would be helpful if you could check if that's the case and if we have a testcase for that.

ddeclerck:
> One thing that I've seen which was the reason to check the PRs: I'm not sure if `-fdefault-colseq` is applied correctly for `SPECIAL-NAMES. ALPHABET nat IS NATIVE.` - it would be helpful if you could check if that's the case and if we have a testcase for that.

To my understanding, `-fdefault-colseq` is like an implicit `PROGRAM COLLATING SEQUENCE` clause ; so should it really affect the `NATIVE` alphabet definition ?

GitMensch:
Hm, MF has a NATIVE directive which changes the default collating sequence, but that doesn't change the default alphabet either.
To change that you'd use the CHARSET directive (wich also changes DEFAULTBYTE, SIGN, NATIVE),

I thionk we don't have an option to effectively change the native alphabet, do we?

ddeclerck:
> Hm, MF has a NATIVE directive which changes the default collating sequence, but that doesn't change the default alphabet either. To change that you'd use the CHARSET directive (wich also changes DEFAULTBYTE, SIGN, NATIVE),
> 
> I thionk we don't have an option to effectively change the native alphabet, do we?

Not really, though that might have some use (we talked to someone who claimed he needed the internal, in-memory data representation to be in EBCDIC while running on an ASCII machine).

PR #78, opened 2023-01-23
Allow specifying a file as argument in -febcdic-table

ddeclerck:
TODO: add a test

@GitMensch Would something like this work for you ?

Alternatively, I could make a single function combining `cob_load_collation` and `cob_get_collation_by_name`, but that could cause confusion since the tables would be dynamically allocated when loading from a file but not when copying pointers to builtin tables.

Note that, even though the external table is built into the generated program, I made the function that reads the table available in libcob, so that external tools (GCSORT in my case) can use that too.

P.S: Should I add my EBCDIC500 to Latin-1 table somewhere in the repo ?


nberth:
> P.S: Should I add my EBCDIC500 to Latin-1 table somewhere in the repo ?

Does this one correspond to the IBM table without irregularities (documented in a comment somewhere in `cconv`)?

ddeclerck:
> > P.S: Should I add my EBCDIC500 to Latin-1 table somewhere in the repo ?
> 
> Does this one correspond to the IBM table without irregularities (documented in a comment somewhere in `cconv`)?

No, not really. The conversion table in cconv (and in the IBM documentation) actually maps EBCDIC 500 (yes, 500, not 37 nor 1047) characters to 7-bit ASCII. However, if we only consider the first 128 characters of Latin-1, then the table are almost identical, the only exception being EBCDIC 6A (¦) remapped to ASCII 7C (|), since ¦ does not exist in 7-bit ASCII, and as a consequence EBCDIC BB (|) replaced by the ASCII substitution character 1A (I wonder why they simply did not directly replace EBCDIC 6A (¦) with a substitution character...).

[the following is a bit off-topic, but out of curiosity]

Interestingly, the documentation about conversion irregularities (https://www.ibm.com/docs/en/iis/11.3?topic=tables-conversion-table-irregularities) seems to be using the EBCDIC 1047 representation of characters in the first column (i.e characters [ and ] at positions AD and BD). Yet it maps EBCDIC 4A and 5A to ASCII 5B and 5D ([ and ]), which suggests they intended to map EBCDIC 500 characters.

libcob/cconv.c
```diff
@@ -471,3 +471,84 @@ cob_get_collation_by_name (const char *col_name,
...
+			return -1;
+		}
+	}
+
+	fclose(f);
+
+	for (i = 0; i < 256; ++i) {
+		n = cob_convert_hex_byte(&hex[2 * i]);
+		if (n < 0) {
+			cob_runtime_error (_("Invalid hex byte in translation table '%s' at index %d: '%c%c'"), col_name, i, hex[2 * i], hex[2 * i + 1]);
```
GitMensch, line 540: Not a request, mainly wondering:
Should we do this like for the configuration?

```c
if (col_name != last_col_name) {
   cob_runtime_error (_("Errors in translation table '%s':"), col_name);
   last_col_name = col_name;
}
```
-> one message with the filename, then N detail messages without the filename?

libcob/cconv.c
```diff
@@ -471,3 +471,84 @@ cob_get_collation_by_name (const char *col_name,
...
+
+	f = fopen(col_name, "rb");
+	if (f == NULL) {
+		cob_runtime_error (_("Can't open translation table '%s'"), col_name);
+		return -1;
+	}
+
+	n = fread(hex, 1, 512, f);
+	if (n != 512) {
+		if (ferror(f)) {
```
GitMensch, line 513: please add testcases for all of these

libcob/cconv.c
```diff
@@ -471,3 +471,84 @@ cob_get_collation_by_name (const char *col_name,
...
+int
+cob_load_collation (const char *col_name,
+		    cob_u8_t *ebcdic_to_ascii,
+		    cob_u8_t *ascii_to_ebcdic)
+{
+	char hex[512];
+	int n, i;
+	FILE *f;
+
+	f = fopen(col_name, "rb");
```
GitMensch, line 505: file opening should be done as with configuration files (which I think is prefixing COB_CONFIG_DIR if not absolute [possibly that needs refactoring to be usable, possibly this can just be called)

libcob/cconv.c
```diff
@@ -471,3 +471,84 @@ cob_get_collation_by_name (const char *col_name,
...
+	if (h >= '0' && h <= '9') return h - '0';
+	else if (h >= 'A' && h <= 'F') return 10 + h - 'A';
+	else if (h >= 'a' && h <= 'f') return 10 + h - 'a';
+	else return -1;
+}
+
+static int
+cob_convert_hex_byte(const char *h)
+{
+	int d1 = cob_convert_hex_digit(h[0]);
```
GitMensch, line 487: minor (general) issue: please add a space before each parameter, so

```diff
 static int
-cob_convert_hex_byte(const char *h)
+cob_convert_hex_byte (const char *h)
 {
-	int d1 = cob_convert_hex_digit(h[0]);
+	int d1 = cob_convert_hex_digit (h[0]);
```

For GC4 there is an indent configuration (unfinished, but "that's to come").

cobc/flag.def
```diff
@@ -93,11 +93,12 @@ CB_FLAG_NQ (1, "callfh", 9,
 	  "                        as external provided EXTFH interface module"))
 
 CB_FLAG_RQ (cb_ebcdic_table, 1, "ebcdic-table", 0, 14, /* cf cconv.h for all available tables */
-	_("  -febcdic-table=[DEFAULT|RESTRICTED-GC|IBM|GCOS]\tdefine EBCDIC translation table:\n"
+	_("  -febcdic-table=[DEFAULT|RESTRICTED-GC|IBM|GCOS|<filename>]\tdefine EBCDIC translation table:\n"
 	  "                        * default: translation to extended ASCII as per MF\n"
```
GitMensch, line 97: I haven't catched that before - either the values should be in all upper-case as in the parameter, or lower-case there (I _think_ the former is the way to go). Please adjust.

cobc/flag.def
```diff
@@ -93,11 +93,12 @@ CB_FLAG_NQ (1, "callfh", 9,
...
 	  "                        * restricted-gc: translation from restricted ASCII only\n"
 	  "                        * ibm: translation to restricted ASCII as per IBM\n"
-	  "                        * gcos: translation to extended ASCII as per GCOS7"))
```
GitMensch, line 100: @nberth: Do you see a reason to not move ibm and gcos out to a file, too? If we do we could just name them the same placing in COB_CONFIG_DIR, possibly always using an extension like ".ttbl" (for translation table)? if the filename does not start with a perioid, slash (additional for `_WIN32`: with a `:` at the second place)

GitMensch (changes requested):
> Note that, even though the external table is built into the generated program, I made the function that reads the table available in libcob, so that external tools (GCSORT in my case) can use that too.

I've _thought_ this to be the case, please add this as note to libcob/ChangeLog and the function definition in libcob/cconv.c.

> P.S: Should I add my EBCDIC500 to Latin-1 table somewhere in the repo ?

Yes, it should be placed in "config" (don't forget to add that in the Makefile.am there). See also the detail comments.

cobc/flag.def
nberth: I see no reason not to extract those from the source code indeed.

@ddeclerck Now I see the ibm translation I added was for restricted ASCII — did not recall that was the case.

GitMensch: I'd place the tables as $COB_CONFIG_DIR/NAME.ttbl (so the original parameter matches) and instead of having ibm and gcos7 in there add a hint after <filename> "e. g. IBM, GCOS"

ddeclerck: Alright.

Examining these tables, it looks like the IBM table is a "generic" EBCDIC500 to 7-bit ASCII table. Would it make sense to have that table in a file called "ebcdic500_ascii7bit" and symlink "IBM" to that file ?
A similar remark can be made about the GCOS table (though it maps to an 8-bit ASCII where extended characters are just given consecutively).

ddeclerck: Also: the collation ids (CB_EBCDIC_IBM, etc...) seems no longer necessary ; should we get rid of this notion ?

GitMensch: > Also: the collation ids (CB_EBCDIC_IBM, etc...) seems no longer necessary ; should we get rid of this notion ?

yes

GitMensch: > Examining these tables, it looks like the IBM table is a "generic" EBCDIC500 to 7-bit ASCII table. Would it make sense to have that table in a file called "ebcdic500_ascii7bit" and symlink "IBM" to that file ? A similar remark can be made about the GCOS table (though it maps to an 8-bit ASCII where extended characters are just given consecutively).

Feel free to drop GCOS and IBM completely, giving them reasonable names in all lower-case.

ddeclerck: Also, what should be do about the `cob_get_collation_by_name` that we introduced a few weeks ago ? The new function `cob_load_collation` assumes it is given pointers to actual arrays, while the former expects pointers to pointers.

cobc/cobc.c
```diff
@@ -3658,7 +3661,12 @@ process_command_line (const int argc, char **argv)
 			/* -febcdic-table=<cconv-table> */
```
GitMensch, line 3661: please adjust that, too, guess that would be something like:

```diff
-		/* -febcdic-table=<cconv-table> */
+		/* -febcdic-table=<cconv-table>/<file> */
```

cobc/codegen.c
```diff
@@ -2598,6 +2608,10 @@ output_collating_tables (void)
 static void
 output_init_collating_tables (void)
 {
+	if (use_external_tables) {
+		return;
```
GitMensch, line 2612: This return looks strange, either the variables below are set - then we need those or they aren't.
If they are set but we don't need them, well, that's a different error, but should be fixed, too.

ddeclerck: Actually this is going to change since we're making all tables external.

We needed that early return because for external tables we embed the table in the header file (at compilation), while for builtin tables we just store a pointer (at runtime).

config/ebcdic500_latin1.ttbl
```diff
@@ -0,0 +1 @@
+000102039C09867F978D8E0B0C0D0E0F101112139D8508871819928F1C1D1E1F80818283840A171B88898A8B8C050607909116939495960498999A9B14159E1A20A0E2E4E0E1E3E5E7F15B2E3C282B2126E9EAEBE8EDEEEFECDF5D242A293B5E2D2FC2C4C0C1C3C5C7D1A62C255F3E3FF8C9CACBC8CDCECFCC603A2340273D22D8616263646566676869ABBBF0FDFEB1B06A6B6C6D6E6F707172AABAE6B8C6A4B57E737475767778797AA1BFD0DDDEAEA2A3A5B7A9A7B6BCBDBEAC7CAFA8B4D77B414243444546474849ADF4F6F2F3F57D4A4B4C4D4E4F505152B9FBFCF9FAFF5CF7535455565758595AB2D4D6D2D3D530313233343536373839B3DBDCD9DA9F
```
GitMensch, line 1: as this is all two-byte hex as string please adjust it to use 513 bytes, switching to the translation part at first line break (it is just much easier to match that way) and move the gcos translation table to a file on the way

libcob/cconv.c
```diff
@@ -471,3 +472,111 @@ cob_get_collation_by_name (const char *col_name,
...
+		/* If it's a path, use it as-is */
+		snprintf (filename, (size_t)COB_FILE_MAX, "%s", col_name);
+	} else {
+		/* Otherwise, prepend the config dir and append the .ttbl suffix */
+		config_dir = getenv ("COB_CONFIG_DIR");
+		snprintf (filename, (size_t)COB_FILE_MAX, "%s%c%s.ttbl",
+			  config_dir ? config_dir : COB_CONFIG_DIR, SLASH_CHAR, col_name);
+	}
+
+	f = fopen(filename, "rb");
```
GitMensch, line 526: Is there a reason to read the file in binary mode? Otherwise please drop the "b".

config/ebcdic500_latin1.ttbl
ddeclerck: Should I keep my code that allows any trailing blanks after the table ?

libcob/cconv.c
```diff
@@ -471,3 +472,111 @@ cob_get_collation_by_name (const char *col_name,
...
+	fclose(f);
+
+	for (i = 0; i < 256; ++i) {
+		n = cob_convert_hex_byte (&hex[2 * i]);
+		if (n < 0) {
+			if (col_name != last_col_name) {
+				cob_runtime_error (_("Errors in translation table '%s':"), col_name);
+				last_col_name = col_name;
+			}
+			cob_runtime_error (_("Invalid hex byte at index %d: '%c%c'"), i, hex[2 * i], hex[2 * i + 1]);
```
GitMensch, line 565: please start the error messages with a lowercase letter

config/ebcdic500_latin1.ttbl
GitMensch: > Should I keep my code that allows any trailing blanks after the table ?

Let's adjust that to just skip any space (also in between) and read until file end. Decide for yourself if the match needs to be "complete" (256 translations) or if it makes more reason to have "missing entries are filled up using the last one" (I don't know).

In any case docs/gnucobol.texi should have a note on this format.

cobc/flag.def
GitMensch: It is a new function only available in the preview and rc-1, feel free to decide on the API, but ensure that the decision is final ;-)

config/ebcdic500_latin1.ttbl
GitMensch: ... and possibly skip the line if it starts with a comment `#` or is empty?

ddeclerck: Alright, I changed to a more flexible format, allowing comment lines starting with # and allowing blank spaces everywhere (except between the two digits of a single byte).

And now I'm just realizing the translations are not necessarily symmetric (specifically when using substitution characters).
What do you think if the table file contained not 256 bytes but 512, the first half for EBCDIC to ASCII and the other half for ASCII to EBCDIC ? Also, to ease input of symmetric translations, we could allow for files containing only 256 bytes and in this case build the complement automatically (as I'm already doing right now).

GitMensch: Make this depending on "4 data row" or "2 data rows" and that's fine.

GitMensch:
For the changelogs - and because we won't do things like that between rc-1 and rc-2/final otherwise: https://sourceforge.net/p/gnucobol/feature-requests/429/

config/ebcdic500_latin1.ttbl
ddeclerck: Sorry I'm not sure what you mean about that.

For info, currently the format allows to write the table like this :
```
# EBCDIC 500 to Latin-1 translation table

00 01 02 03 9C 09 86 7F 97 8D 8E 0B 0C 0D 0E 0F
10 11 12 13 9D 85 08 87 18 19 92 8F 1C 1D 1E 1F
80 81 82 83 84 0A 17 1B 88 89 8A 8B 8C 05 06 07
```
(line breaks can be anywhere)

GitMensch: OK, I see. I've thought that each non-comment, non-empty line is a "data set", but those can span over multiple lines - which is likely much easier to read.

So each data set ends after 256 characters now? [one data set row is likely easier to implement and _maybe_ easier to understand when looking at the file]

Then the file either needs 2 data sets (conversion is done from the first to the second and vice versa) or 4 data sets (conversion is done from 1 to 2 in the one direction and from 3 to 4 for the other direction).

if this doesn't makes sense then I've not understood something, please elaborate in this case.

GitMensch: Correct. So your file should either have one or two data sets, if only one is given the other one is implicit (it is just important to note that in the format documentation in gnucobol.texi; ideally we have at least two ttbl we ship and can reference two of those with one being symmetric, the other one not.

ddeclerck: Yes, the old "built in" tables are most likely not symmetric, and the new EBCDIC 500 to Latin-1 is, so there will be examples of both cases.

GitMensch:
Feel free to drop a note as soon as you consider this PR ready and I'll have a look.

ddeclerck:
@GitMensch not done yet

config/ebcdic500_latin1.ttbl
ddeclerck: Almost done, just this documentation to add. In which section should I add the format description ?

GitMensch: > In which section should I add the format description ?

Please add a short description on `-febcdic-table` as new paragraph at the end of "2.1.5 Configuration options". We could document the format in default.ttbl (likely a good idea in general) and generate an appendix from it, similar as it is done for other files (I could help with the generation part if necessary) - but this may be overkill; or just reference the file within the explanation "as seen in default.ttbl in the configuration directory (see `cobc --info`), ...".

When on it: please drop `-fmfcomment`, `-acucoment` and `-fsign` from 2.1.8 where they don't belong any more.

Additional it would be useful to add the new external functions as 2.3.7 (its enough to give a short blurb and the function signature); but "missing" that won't hinder the PR to get merged. People should not need to inspect libcob's source code to know about these functions and where they may be useful and how they should be called.

config/ChangeLog
```diff
@@ -1,4 +1,16 @@
...
+2023-01-23 David Declerck <david.declerck@ocamlpro.com>
+
+	* ebcdic500_latin1: new translation table from EBCDIC 500 to Latin-1
+	* ebcdic500_ascii7bit: translation table from EBCDIC 500 to 7-bit ASCII
+	  as per IBM, extracted from libcob/cconv.c (cob_ibmebcdic_ascii)
+	* ebcdic500_ascii8bit: translation table from EBCDIC 500 to 7-bit ASCII
+	  as per GCOS7, extracted from libcob/cconv.c (cob_gcos7ebcdic_ascii)
+	* default: the original GnuCobol translation table,
+	  extracted from libcob/cconv.c (cob_ebcdic_ascii)
+	* alternate: the original GnuCobol "alternate" translation table,
```
GitMensch, line 11: `s/GnuCobol/GnuCOBOL/g`

libcob/cconv.c
```diff
@@ -20,454 +20,142 @@
...
+	char filename[COB_FILE_BUFF];
+	const char *last_col_name = NULL;
+
+	if (col_name[0] == '.' || col_name[0] == SLASH_CHAR
+#ifdef _WIN32
+ || col_name[0] != '\0' && col_name[1] == ':'
+#endif
+	   ) {
+		/* If it's a path, use it as-is */
+		snprintf (filename, (size_t)COB_FILE_MAX, "%s", col_name);
```
GitMensch, line 87: please `strlen` + check sizeof (filename) + memcpy instead

cobc/cobc.c
```diff
@@ -3905,6 +3906,17 @@ process_command_line (const int argc, char **argv)
 		cobc_early_exit (EXIT_SUCCESS);
 	}
 
+	/* Load the collating tables */
+	if (cb_ebcdic_table == NULL) {
+		cb_ebcdic_table = cobc_main_strdup (
+			cb_flag_alt_ebcdic ? "alternate" : "default");
+	}
+	if (cob_load_collation (cb_ebcdic_table,
```
GitMensch, line 3914: Please postpone this call until we know we need it, when `cb_ebcdic_table == NULL`.

libcob/cconv.c
ddeclerck: Alright. Is there a specific error to generate to indicate the string would overflow ? Or should I just truncate ?

cobc/cobc.c
ddeclerck: You mean, calling cob_load_collation in `codegen.c:output_collating_tables` ?

libcob/cconv.c
ddeclerck: Alright. And for the second case (`snprintf` with `COB_CONFIG_DIR` and so on), should I switch to memcpy ? Or just checking the length is enough ?

cobc/cobc.c
GitMensch: Yes, at least if there was no explicit request up-front. Otherwise we'd read in these files on _any_ call of cobc.

libcob/cconv.c
GitMensch: length check is enough

ddeclerck:
@GitMensch I think everything has been addressed.

GitMensch (commented):
he who changes the test data...

GitMensch:
After handling some relative minor issues, I consider the PR ready for check in to 3.x branch now, reference the commit in the related FR and close that.

GitMensch (commented):
...

GitMensch (approved):

ddeclerck:
@GitMensch Merged on SVN. I'm now going to fix GCSORT to use these changes.

GitMensch:
Note: because you made me check VS2008 for GCSORT I also did so for 3.x - `isblank` is not available there, I've replaced it by the more matching `isspace`.

Again: Thanks for the changes (and also thanks for caring for GCSORT).
It is also good to know that the collation issues work for your conversion project now.

PR #76, opened 2023-01-16
[WIP] Add .github and .gitignore for better Github support

lefessan:

PR #79, opened 2023-01-23
Add working Github action files, except for Windows

lefessan:
This PR is just here to merge into `gcos4gnucobol-3.x` working files for Macos (with testsuite) and Windows (without testsuite).

The goal is to keep PR #76 opened until the Windows workflow is completely fixed (especially as some interesting comments are there).


GitMensch:
As noted in the referenced PR, I'd just ignore the Windows "make check" result for now, but keep the check run.

PR #80, opened 2023-01-23
Allow use of external preprocessor

lefessan:
* New option --pp <cobc_pp> to call an external cobc-aware preprocessor
* New env variable COB_PP to call an external cobc-aware preprocessor
* New env variable COB_SAVE_TEMPS to save temporary files

The external preprocessor has to be cobc-aware, i.e. add #area_a directives and remove some paragraphs from the IDENTIFICATION DIVISION.

The basic idea here is that GnuCOBOL preprocessor is not fully compliant to the standard, and making it compliant would require completely rewriting that part. Instead, we propose to give the user the possibility to use another external preprocessor in such cases.

### DIFF
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -344,6 +344,7 @@ static struct cobc_mem_struct	*cobc_parsemem_base = NULL;
 static struct cobc_mem_struct	*cobc_plexmem_base = NULL;
 
 static const char	*cobc_cc;		/* C compiler */
+static const char	*cobc_pp;               /* COBOL preprocessor */
 static char		*cobc_cflags;		/* C compiler flags */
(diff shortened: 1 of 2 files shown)


GitMensch:
As long as there's no free "cobc aware COBOL preprocessor" this Feature Request (FRs should be always created in the GC issue tracker) won't be considered.

And in general it would be more useful to make the preparser working completely instead.

GitMensch:
Note: especially with make files it is easy to first call the own preprocessor (writing a _p.cob or similar, or even do cobc-intended bigger pre-processing creating a  .i file) and then cobc.

Also note that `make` works quite nice, especially when you include the dependency chain, which is possible via `-MT -MD`.

PR #81, opened 2023-01-24
Auto-detect free format from wrong indicator on first line

lefessan:
Simpler version than what ws proposed in https://sourceforge.net/p/gnucobol/feature-requests/45/

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -208,6 +208,11 @@ NEWS - user visible changes				-*- outline -*-
    be reported and recovered from.  This is configurable with option
    `missing-period`.
 
+   If not specified, the compiler tries to automatically recognize the format,
+   using either fixed or free, depending on column 7 in the first non-space
+   line. This feature can be disabled by setting the format manually
+   with `-free`, `-fixed` or `-fformat`.
+
 ** the new -febcdic-table option enables one to specify the
    translation table used when dealing with EBCDIC codeset;
    these translation tables are stored as new configuration files with
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -4,6 +4,15 @@
 	* codeoptim.c: adjusted inline functions (cob_cmp_packed_int,
 	  cob_get_packed_int) to use register types and less intermediate values
 
+2023-01-28  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c (main): initialize cb_config_text_column to 72 to avoid a
+	  race condition in config files.
+	* pplex.l (ppopen): try to autodetect the format of a file after
+	  checking the BOM of a file
+	* pplex.l (cobc_set_source_format): add new format name "AUTO" to
+	  activate auto-detection of file format
+
 2023-01-26  Simon Sobisch <simonsobisch@gnu.org>
 
 	* codegen.c: generate field offsets and indexes as long long to
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -5009,6 +5009,7 @@ preprocess (struct filename *fn)
 	} else {
 		sourcename = fn->source;
 	}
+	save_source_format = cobc_get_source_format ();
 	if (ppopen (sourcename, NULL) != 0) {
 		cobc_terminate (sourcename);
 	}
@@ -5037,7 +5038,6 @@ preprocess (struct filename *fn)
 
 	/* Save default exceptions and flags in case program directives change them */
 	memcpy (save_exception_table, cb_exception_table, exception_table_size);
-	save_source_format = cobc_get_source_format ();
 	save_fold_copy = cb_fold_copy;
 	save_fold_call = cb_fold_call;
 	save_ref_mod_zero_length = cb_ref_mod_zero_length;
@@ -5124,8 +5124,8 @@ preprocess (struct filename *fn)
 /* Routines to generate program listing */
 
 
-static void
-set_listing_header_code (void)
+void
+cobc_set_listing_header_code (void)
 {
 	strcpy (cb_listing_header, "LINE    ");
 	if (! CB_SF_FREE (cb_listing_file_struct->source_format)) {
@@ -8874,7 +8874,6 @@ process_file (struct filename *fn, int status)
 		cb_listing_page = 0;
 		cobc_elided_strcpy (cb_listing_filename, fn->source,
 			sizeof (cb_listing_filename), 0);
-		set_listing_header_code ();
 	}
 
 	if (cb_compile_level >= CB_LEVEL_PREPROCESS
@@ -8887,7 +8886,10 @@ process_file (struct filename *fn, int status)
 			cb_flag_syntax_only = 1;
 			cb_flag_fast_compare = 0;
 		}
-	}
+	} else
+		if (cb_src_list_file) {
+			cobc_set_listing_header_code ();
+		}
 
 	if (cobc_list_file) {
 		putc ('\n', cb_listing_file);
@@ -8977,6 +8979,12 @@ main (int argc, char **argv)
 
 	cb_saveargc = argc;
 	cb_saveargv = argv;
+
+	/* This value is used when setting the format in FIXED mode in
+	   pplex, but it might happen that the value has not yet been 
+	   initialized in the configuration file.
+	*/
+	cb_config_text_column = 72;
 
 	/* Process command line arguments */
 	iargs = process_command_line (argc, argv);
--- a/cobc/cobc.h
+++ b/cobc/cobc.h
@@ -105,6 +105,7 @@ enum cb_format {
 	CB_FORMAT_ICOBOL_CRT,	/* ICOBOL Free-form format (CRT) */
 	CB_FORMAT_ACUTERM,	/* ACU Terminal format, named "TERMINAL" */
 	CB_FORMAT_COBOLX,	/* GCOS's COBOLX */
+	CB_FORMAT_AUTO, 	/* Auto-detect format */
 };
 #define CB_SF_FREE(sf) (sf == CB_FORMAT_FREE)
 #define CB_SF_FIXED(sf) (sf == CB_FORMAT_FIXED || sf == CB_FORMAT_COBOL85)
@@ -650,6 +651,8 @@ extern char *		cb_get_strerror (void);
 extern void		cb_add_error_to_listing (const char *, int, const char *, char *);
 DECLNORET extern void		flex_fatal_error (const char *, const char *,
 					 const int) COB_A_NORETURN;
+
+extern void		cobc_set_listing_header_code ();
 
 /* reserved.c */
 extern struct reserved_word_list	*cobc_user_res_list;
--- a/cobc/pplex.l
+++ b/cobc/pplex.l
@@ -145,7 +145,7 @@ static int			quotation_mark = 0;
 static int			listing_line = 0;
 static int			requires_listing_line;
 static int			requires_new_line = 0;
-static enum cb_format		source_format;
+static enum cb_format		source_format = CB_FORMAT_AUTO;
 static int	indicator_column = 7;
 static int	text_column = 72;	   /* end of area B (in single-byte
 					      characters) */
@@ -1148,6 +1148,23 @@ pp_set_replace_list (struct cb_replace_list *list, const cob_u32_t is_pushpop)
 	}
 }
 
+static int
+is_fixed_indicator (char c){
+	switch (c){ /* same indicators as in ppinput() */
+	case ' ':
+	case '-':
+	case 'd':
+	case 'D':
+	case '*':
+	case '/':
+	case '\\':
+	case '$':
+		return 1;
+	default:
+		return 0;
+	}
+}
+
 /* open file (source or coypbook) for further processing */
 int
 ppopen (const char *name, struct cb_replace_list *replacing_list)
@@ -1208,19 +1225,47 @@ ppopen (const char *name, struct cb_replace_list *replacing_list)
 	/* Check for BOM - *not* for input from stdin as rewind() clears the input
 	  buffer if used on stdin and output in console has normally no BOM at all */
 	if (ppin && strcmp (name, COB_DASH) != 0) {
-		unsigned char		bom[4];
-		if (fread (bom, 3, 1, ppin) == 1) {
-			if (bom[0] != 0xEF || bom[1] != 0xBB || bom[2] != 0xBF) {
-				rewind (ppin);
-			}
-		} else {
-			rewind (ppin);
+		int fseek_to = 0 ;
+#define COBC_LOOKAHEAD 20
+		unsigned char buffer[COBC_LOOKAHEAD];
+		int nread = fread (buffer, 1, COBC_LOOKAHEAD, ppin);
+		int pos = 0;
+		if (nread >= 3 && buffer[0] == 0xEF && buffer[1] == 0xBB && buffer[2] == 0xBF) {
+			fseek_to = 3;
+			pos = 3;
 		}
+		if (source_format == CB_FORMAT_AUTO){
+			/* If indicator is wrong on first line, switch to free format */
+			/* skip empty lines */
+			int amount_of_0a_seen = 0;
+			while (nread-pos > 7 && (buffer[pos] == '\r' || buffer[pos] == '\n')){
+				if (buffer[pos] == '\n') amount_of_0a_seen++;
+				pos++;
+			}
+			/* check tab or indicator */
+			if ( nread-pos > 7 && buffer[pos] != '\t' && !is_fixed_indicator (buffer[pos+6]) ){
+				struct cb_tree_common	loc;
+				loc.source_file = name;
+				loc.source_line = 1 + amount_of_0a_seen;
+				loc.source_column = 7;
+				cb_note_x (COB_WARNOPT_NONE, &loc, _("free format detected"));
+				(void) cobc_deciph_source_format ("FREE");
+			}
+		}
+		fseek(ppin, fseek_to, SEEK_SET);
 	}
 
-	/* Save name for listing */
-	if (cb_current_file && !cb_current_file->name) {
-		cb_current_file->name = cobc_strdup (name);
+	if (source_format==CB_FORMAT_AUTO)
+		cobc_set_source_format (CB_FORMAT_FIXED);
+
+	if (cb_current_file){
+		if (cb_current_file->source_format==CB_FORMAT_AUTO)
+			cb_current_file->source_format=cobc_get_source_format ();
+		/* This is delayed until after format detection */
+		cobc_set_listing_header_code ();
+		/* Save name for listing */
+		if (!cb_current_file->name)
+			cb_current_file->name = cobc_strdup (name);
 	}
 
 	/* Add to dependency list */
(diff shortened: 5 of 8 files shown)


cobc/cobc.c
```diff
@@ -3341,6 +3343,7 @@ process_command_line (const int argc, char **argv)
 
 		case 'F':
 			/* --free, alias of `-fformat=free` */
+			cobc_free_format = 1 ;
 			(void) cobc_deciph_source_format ("FREE");
```
GitMensch, line 3347: Let's move the check function to pplex.l, this way we cover `-fformat=free`/`variable`..., too.
If not done already the default should be changed from "fixed" to "unknown", so that the check is only done if there wasn't no explicit format set.

cobc/cobc.c
```diff
@@ -4946,6 +4949,22 @@ force_new_page_for_next_line (void)
...
+	case 'd':
+	case 'D':
+	case '*':
+	case '/':
+	case '\\':
+		return 1;
+	default:
+		return 0;
+	}
+}
```
GitMensch, line 4966: we should add `$` here (and move that to pplex.l)

cobc/cobc.c
```diff
@@ -4983,6 +5002,20 @@ preprocess (struct filename *fn)
 		sourcename = COB_DASH;
 	} else {
 		sourcename = fn->source;
+		if( !cobc_free_format ){
+			/* If indicator is wrong on first line, switch to
+			   free format */
+			FILE* fd = fopen (sourcename, "r");
+			char buffer[7];
+			if (fd != NULL){
```
GitMensch, line 5010: mind the potential BOM

GitMensch (changes requested):
This approach is actually much better than the original suggested one.
Please finish that PR:

* moving the check to pplex.l [may still be called from cobc.c], only if the format is undefined
* add an entry in cobc/Changelog
* add a test for that in in tests/testsuite.src and change the default flags in atlocal.in to contain --fixed (so that we don't execute that new internal check in every testsuite compile)

GitMensch:
Note: the implicit change should also only be done "per file":

`cobc -x main.cob other.cob` - if main.cob is "implied fixed" then other.cob may be "implied free", to do so: issue the check in cobc.c (preprocess) after `save_source_format = cobc_get_source_format ();` or - possibly better - in `ppparse()` directly.

cobc/cobc.c
GitMensch: When adjusting `enum cb_format` to include `CB_FORMAT_UNKNOWN` as first entry (no need for the zero as we don't "serialize" that as int somewhere) then no `cobc_autodetect_source_format` is needed.

GitMensch:
Note: the NEWS entry for `-fformat` should be adjusted to say something like "if not specified, then the format is tried to be automatically recognized as either fixed or free, depending on column 7 in the first non-space line".

lefessan:
I will modify the `Changelog` and `NEWS` files when leaving the Draft mode.

@GitMensch What's the difference between `make check` and `make distcheck` ? On my computer, `make check` prints no errors on the modified testsuite, whereas `make distcheck` displays the same errors as the CI. Any idea why ?

Also, what is the canonical way to run only a subset of tests ? I currently use `../tests/testsuite -k cobc`, is there something better ?

GitMensch:
> I will modify the `Changelog` and `NEWS` files when leaving the Draft mode.

+1

> What's the difference between `make check` and `make distcheck` ? On my computer, `make check` prints no errors on the modified testsuite, whereas `make distcheck` displays the same errors as the CI. Any idea why ?

`make check` (when executed on the buildroot) runs `make all` in each directory, then builds the testsuite script if it isn't up-to date (actually part of `make all` in "tests"), then executes it.
`make distcheck` packages the source tree, then tries to build from there, then runs `make check` in that. In 98% failures here are related to `Makefile.am` 's not containing newly added source files. Less likely, but also possible is that somewhere paths are newly hardcoded where they shouldn't, another reason that happens from time to time is that configure does not work as expected (only relevant when changing one of the autotools).

> Also, what is the canonical way to run only a subset of tests ? I currently use `../tests/testsuite -k cobc`, is there something better ?

Sure:  `make check TESTSUITEFLAGS="--jobs=6 -k cobc"` (or `--list` or numbers). When doing changes I commonly run `make --jobs=4 check TESTSUITEFLAGS="--jobs=6"`, then see only failures in the internal suite, then doing ``make check TESTSUITEFLAGS="--recheck --verbose"`.

GitMensch:
As seen in testsuite.log:

~~~
| FLAGS="--fixed -debug -Wall ${COBOL_FLAGS} -fno-diagnostics-show-option"
....
../../../tests/syn_misc.at:4747: $COMPILE_ONLY -fformat=auto -std=cobol85 prog.cob
~~~

the testsuite is correctly used "as expected".

If I understood you correctly you get the same result locally  - `make check` works, but `make distcheck` fails - correct?

BTW: Do we need an adjustment in syn_misc.at - if yes, why?

cobc/cobc.c
GitMensch: if we postpone this to pplex.l (ppopen) then we can do this directly after the BOM check (only if source format is unknown).

lefessan:
My understanding for the fixes is that I added `--fixed` to `atlocal.in`, so that all tests are called with that flag, which causes errors on a few tests. So, I had to add `-fformat=auto` to force auto-detection of format.

GitMensch:
Looks like the tests pass now.

NEWS
```diff
@@ -208,6 +208,10 @@ NEWS - user visible changes				-*- outline -*-
    be reported and recovered from.  This is configurable with option
    `missing-period`.
 
+   If not specified, the compiler tries automatically recognize the format,
+   using either fixed or free, depending on column 7 in the first non-space
+   line. This feature can be disabled using `-fformat=noauto`.
```
GitMensch, line 213: Do we need noauto? If in question someone would explicit specify -fformat=fixed (for the old behavior), no? In this case we can drop the noauto completely.

cobc/ChangeLog
```diff
@@ -1,5 +1,28 @@
 
-2023-01-23 David Declerck <david.declerck@ocamlpro.com>
+2023-01-28  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c (process_command_line): disable auto-detection of format
+	  if -f or -F options are provided.
+	* cobc.c (main): initialize cb_config_text_column to 72 to avoid a
+	  race condition in config files.
+	* cobc.c (preprocess): try to autodetect the format of a file
```
GitMensch, line 8: As mentioned before: please move that call to ppopen, this way it is done one each "main source" again, and we cover the BOM. We can also get rid of the extern autoformat, by just setting the default format to "unknown" and do the check in ppopen if the format is "still unknown".

NEWS
lefessan: Indeed, it is equivalent to `fixed`, I will remove it.

cobc/ChangeLog
lefessan: What do you mean by BOM ?

`ppopen` is also used to read copybooks, so I will have to change its interface to add an additional parameter to tell it whether autodetection should be performed or not.

GitMensch: > What do you mean by BOM ?

The byte-order-mark which ppopen already handles.

> `ppopen` is also used to read copybooks, so I will have to change its interface to add an additional parameter to tell it whether autodetection should be performed or not.

I _think_ it should be able to just check "is the format still undefined" - if yes then we are at a main program.
If it isn't, then it likely should be that way ;-) Otherwise `copy_stack == NULL` would do the trick, but I'd prefer the former.

GitMensch:
Note: I try to get to rc-2 state until tomorrow, so that would be the last possible time for this PR (which is in a "general good state") to be included in 3.2.

cobc/cobc.c
GitMensch: ping @lefessan: Using a default of `source_format = CB_FORMAT_UNKNOWN;` we neither need `cobc_autodetect_source_format` nor a new `main_source` parameter, do we?

lefessan: I set it to `CB_FORMAT_AUTO` by default and got rid of `cobc_autodetect_source_format` to get the same behavior.

tests/testsuite.src/run_fundamental.at
```diff
@@ -6178,3 +6178,41 @@ prog.cob:7: note: value size is 3
...
+  WORKING-STORAGE  SECTION.
+  PROCEDURE        DIVISION.
+     DISPLAY "OK" NO ADVANCING
+     END-DISPLAY.
+     DISPLAY "KO" NO ADVANCING
+     END-DISPLAY.
+     STOP RUN.
+])
+
+AT_CHECK([$COMPILE -fformat=auto prog.cob], [0], [],
```
GitMensch, line 6198: Please adjust to 

```m4
AT_CHECK([$COMPILE prog.cob], [0], [],
[prog.cob:2: note: automatically switching to free format
])
```

-->  include source reference, setting the info as note, use lower-case

Please move the test to syn_...something, change to COMPILE_ONLY and drop the run.

One check missing (can be done at the end of this test): AT_DATA ([prog2.cob] which is fixed format, then compile both together (and have the message only for the first one.

libcob/common.h
```diff
@@ -108,6 +108,11 @@ typedef __mpz_struct mpz_t[1];
 #define	cob_screen_ptr		cob_screen *
 #define	cob_file_key_ptr	cob_file_key *
 
+/* For use as booleans */
```
GitMensch, line 111: bool is something for 4.x with stdbool, but we don't need that for this PR, so please drop

cobc/pplex.l
```diff
@@ -1208,16 +1225,33 @@ ppopen (const char *name, struct cb_replace_list *replacing_list)
 	/* Check for BOM - *not* for input from stdin as rewind() clears the input
 	  buffer if used on stdin and output in console has normally no BOM at all */
 	if (ppin && strcmp (name, COB_DASH) != 0) {
-		unsigned char		bom[4];
-		if (fread (bom, 3, 1, ppin) == 1) {
-			if (bom[0] != 0xEF || bom[1] != 0xBB || bom[2] != 0xBF) {
-				rewind (ppin);
+		int fseek_to = 0 ;
+#define COBC_LOOKAHEAD 20
+		char buffer[COBC_LOOKAHEAD];
```
GitMensch, line 1230: somehow I tend to

```c
		char pre_buffer[20];
		int nread = fread (pre_buffer, 1, sizeof(pre_buffer), ppin);
```

... but I can't explain why so I'm fine either way

cobc/ChangeLog
```diff
@@ -4,6 +4,17 @@
 	* codeoptim.c: adjusted inline functions (cob_cmp_packed_int,
 	  cob_get_packed_int) to use register types and less intermediate values
 
+2023-01-28  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c (main): initialize cb_config_text_column to 72 to avoid a
+	  race condition in config files.
+	* pplex.l (ppopen): try to autodetect the format of a file before
```
GitMensch, line 11: actually it is still after the BOM check and that is mentioned directly two four lines below !?!

cobc/cobc.h
```diff
@@ -105,9 +105,10 @@ enum cb_format {
 	CB_FORMAT_ICOBOL_CRT,	/* ICOBOL Free-form format (CRT) */
 	CB_FORMAT_ACUTERM,	/* ACU Terminal format, named "TERMINAL" */
 	CB_FORMAT_COBOLX,	/* GCOS's COBOLX */
+	CB_FORMAT_AUTO, 	/* Auto-detect format */
 };
 #define CB_SF_FREE(sf) (sf == CB_FORMAT_FREE)
-#define CB_SF_FIXED(sf) (sf == CB_FORMAT_FIXED || sf == CB_FORMAT_COBOL85)
+#define CB_SF_FIXED(sf) (sf == CB_FORMAT_FIXED || sf == CB_FORMAT_COBOL85 || sf == CB_FORMAT_AUTO)
```
GitMensch, line 111: We shouldn't reach this macro when the SF is still auto, should we?

cobc/cobc.c
```diff
@@ -8978,6 +8978,12 @@ main (int argc, char **argv)
 	cb_saveargc = argc;
 	cb_saveargv = argv;
 
+	/* This value is used when setting the format in FIXED mode in
+	   pplex, but it might happen that the value has not yet been 
+	   initialized in the configuration file.
+	*/
+	cb_config_text_column = 72;
```
GitMensch, line 8985: Just to double-check: this still happens before `cobc_set_source_format` is called, which sets it?

GitMensch (changes requested):

GitMensch:
Still a draft?

cobc/cobc.c

cobc/cobc.c
lefessan: In `pplex.y`, `set_source_format` performs:
```
	case CB_FORMAT_FIXED:
	case CB_FORMAT_AUTO:
	case CB_FORMAT_COBOL85:
		text_column = cb_config_text_column; /* 72 by default */
```
So, before this change, if a configuration file did not set the `text_column` parameter, the `format=fixed` parameter would set `text_column` to 0, instead of 72.

cobc/cobc.h
lefessan: Actually, we reach it: if you compile in listing mode, autodetection will detect that the file is fixed, but then it will switch it back to `CB_FORMAT_AUTO` at the end of preprocessing, so that, when you reach the listing phase, either you have to redo autodetection or just assume it was fixed.

tests/Makefile.am
```diff
@@ -45,6 +45,7 @@ testsuite_sources = \
 	testsuite.src/syn_set.at \
 	testsuite.src/syn_subscripts.at \
 	testsuite.src/syn_value.at \
+	testsuite.src/syn_format.at \
```
GitMensch, line 48: A new file is fine, but if we do this, then the existing tests from syn_misc.at and configuration.at (the later wasn't a very good place for it in any case) should be moved there, too (just `grep -R tests/testuite.src "fformat"` - most of those should be moved).

As an alternative if you feel that's too much for the available time: stick the new test to the ~~trash~~ catch-all syn_misc.at, please.

cobc/ChangeLog
lefessan: done.

cobc/cobc.c
GitMensch: I see, then the missing piece is to set `source_format` in the `struct list_files` after ppopen.

tests/testsuite.src/run_fundamental.at
lefessan: I created a file `syn_format.at`, as `syn_misc.at` seems to be already too long.
I added more examples, including with BOM (it seems it was not tested before).

cobc/cobc.h
GitMensch: I see, then the missing piece is to set `source_format` in the `struct list_files` after ppopen.

libcob/common.h
lefessan: done

cobc/ChangeLog
lefessan: Oups, actually, I did it completely wrong the first time, I wanted to do the two tests independantly, but it's just not possible because you have to do the autodetect after the BOM test, but at the end, you still need to be careful when rewinding not to go back before the BOM...

GitMensch: yes, the current code is fine now (better would be only if we would not peek but read and unput, this way it could also be done for code from stdin; but that could be done later, too)
I'll check the ChangeLog again once this leaves draft state.

tests/Makefile.am
lefessan: Ok, indeed, I moved the test back to `syn_misc.at`, given the lack of time.

I modified the `texi` documentation too.

And I don't know why, but I cannot find a button to switch the PR back from the draft mode... Maybe it's because you have requested changes ?

tests/testsuite.src/run_fundamental.at
lefessan: So, I moved the test back to `syn_misc.at` :-)

tests/Makefile.am
GitMensch: Hm, so you don't have that at the PRs bottom?

PR state

cobc/cobc.c
lefessan: Maybe we should create a ticket and postpone a complete fix for later, as this problem is unrelated to this PR (though I found the bug by testing it).

tests/testsuite.src/run_fundamental.at
GitMensch: ... one more to add to https://sourceforge.net/p/gnucobol/feature-requests/search/?q=testsuite - I've just did that with https://sourceforge.net/p/gnucobol/feature-requests/430/.

cobc/pplex.l
```diff
@@ -1208,16 +1225,33 @@ ppopen (const char *name, struct cb_replace_list *replacing_list)
...
+		}
+		if (source_format == CB_FORMAT_AUTO){
+			/* If indicator is wrong on first line, switch to free format */
+			/* skip empty lines */
+			while( nread-pos > 7 && ( buffer[pos] == '\r' || buffer[pos] == '\n' )){
+				pos++;
+			}
+			/* check tab or indicator */
+			if ( nread-pos > 7 && buffer[pos] != '\t' && !is_fixed_indicator (buffer[pos+6]) ){
+				fprintf(stderr, "Automatically switching to free format\n");
```
GitMensch, line 1245: that still needs to be a standard "note" with the current position (and  starting in lowercase)

lefessan: I added a warning that displays `free format detected` on the first line of the file.

GitMensch: I'm sorry! I should have been more clear...

What about using 
```c
				struct cb_tree_common	loc;
    
				loc.source_file = name;
				loc.source_line = 1 + amount_of_0a_seen;
				loc.source_column = 7;
				cb_note_x (COB_WARNOPT_NONE, &loc, _("free format detected"));
```


and dropping the warning? Note: the test actually shows that the current line calculation is always 1, not the actual line.

lefessan: Ok, I dropped the warning, and added the note with correct line calculation. It's a bit a pity not to be able to disable the warning, though.

GitMensch: > Ok, I dropped the warning, and added the note with correct line calculation.

Thank you.

>  It's a bit a pity not to be able to disable the warning, though.

There would be a way if you want to do that: add a new `-fformat=default` (which will be set in config/default.conf); this will be duplicating the current "auto" and then "auto" will not raise that note (only default will).

cobc/cobc.h
lefessan: Indeed, I didn't notice it was saved there. I modified `ppopen` to update it and display the header afterwards.

cobc/cobc.c
GitMensch: I'll handle the issue creation and removing of `cb_config_text_column` completely (because it just does not work as it did, so better "break hard").

GitMensch (approved):

GitMensch:
Feel free to check that into gc3x branch.

Afterwards I'll drop the text-column (and maybe reformat the patch a bit), so that will be solved, too.

lefessan:
I am going to try to do it right now, I have half an hour and it's the first time I will commit in the SVN. FWIW, I am traveling all the week, so I am not very responsive.

lefessan:
Yeah, I managed to do it ! Just hoping the SVN message is ok.

GitMensch:
Thanks, I'll post adjust it a bit, but that's fine in general.

lefessan:
To be honest, @nberth sent me a small « cheatsheet » on how to do it, a few days ago. It helped me a lot :-)

PR #82, opened 2023-01-25
homogenization of testsuite syntax

lefessan:
The syntax of the testsuite is not homogeneous, so it makes it difficult to use other tools than `autoconf` to work with it.

The goal of this PR is to make it more homogeneous, and test/dev some other tooling to simplify running the testsuite and fixing issues (while still using `autoconf` testsuite for mainstream use).

### DIFF
--- a/tests/testsuite.src/run_file.at
+++ b/tests/testsuite.src/run_file.at
@@ -251,7 +251,8 @@ AT_DATA([prog.cob], [
 
 AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
-[echo "Test" > ./nosubhere]
+AT_DATA([nosubhere], [Test
+])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
 
 AT_CLEANUP
@@ -435,7 +436,8 @@ AT_DATA([prog.cob], [
 
 AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
-[echo "Test" > ./nosubhere]
+AT_DATA([nosubhere], [Test
+])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
 
 AT_CLEANUP
--- a/tests/testsuite.src/syn_misc.at
+++ b/tests/testsuite.src/syn_misc.at
@@ -9605,7 +9605,7 @@ AT_DATA([fixed2.cob], [
 ])
 
 # Generate source files with DOM
-AT_CHECK([printf '\xEF\xBB\xBF' > header.dom])
+AT_CHECK([printf '\357\273\277' > header.dom])
 AT_CHECK([cat header.dom free.cob >> domfree.cob])
 AT_CHECK([cat header.dom fixed1.cob >> domfixed.cob])
 
--- a/tests/testsuite.src/configuration.at
+++ b/tests/testsuite.src/configuration.at
@@ -647,11 +647,11 @@ AT_SETUP([runtime configuration: entries])
 AT_KEYWORDS([configuration misc])
 
 AT_CHECK([echo "$PATHSEP"], [0], [;
-], [],
+], [], [
 
 # Previous test "failed" --> PATHSEP isn't ;
 
-AT_DATA([defunc.cfg], [
+  AT_DATA([defunc.cfg], [
 novar
 physical_cancel notwithme
 load_case  insensitive
@@ -664,7 +664,7 @@ trace_file /tmp:/temp
 ])
 
 # conf entries must be clean
-AT_CHECK([$COBCRUN -c defunc.cfg --runtime-conf], [1], [],
+  AT_CHECK([$COBCRUN -c defunc.cfg --runtime-conf], [1], [],
 [configuration error:
 defunc.cfg:2: unknown configuration tag 'novar'
 defunc.cfg:3: invalid value 'notwithme' for configuration tag 'physical_cancel';
@@ -682,12 +682,12 @@ defunc.cfg:9: WARNING - 'sort_chunk' without a value - ignored!
 defunc.cfg:10: invalid value '/tmp:/temp' for configuration tag 'trace_file';
 	should not contain ':'
 ])
-
+]
 ,
-
+[
 # Previous test "passed" --> PATHSEP is ;
 
-AT_DATA([defunc.cfg], [
+  AT_DATA([defunc.cfg], [
 novar
 physical_cancel notwithme
 load_case  insensitive
@@ -700,7 +700,7 @@ trace_file C:\tmp;C:\temp
 ])
 
 # conf entries must be clean
-AT_CHECK([$COBCRUN -c defunc.cfg --runtime-conf], [1], [],
+  AT_CHECK([$COBCRUN -c defunc.cfg --runtime-conf], [1], [],
 [configuration error:
 defunc.cfg:2: unknown configuration tag 'novar'
 defunc.cfg:3: invalid value 'notwithme' for configuration tag 'physical_cancel';
@@ -718,7 +718,7 @@ defunc.cfg:9: WARNING - 'sort_chunk' without a value - ignored!
 defunc.cfg:10: invalid value 'C:\tmp;C:\temp' for configuration tag 'trace_file';
 	should not contain ';'
 ])
-
+]
 )
 
 AT_CLEANUP


tests/testsuite.src/data_pointer.at
```diff
@@ -44,18 +44,18 @@ AT_CHECK([test "$COB_HAS_64_BIT_POINTER" = "yes"], [0], [], [],
 
 # Previous test "failed" --> 32 bit
 
-[AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [0x00000000
```
GitMensch, line 47: This is using a format we commonly don't use, a conditional when a test "passed"/"failed". The [] scope these parts, so they should be kept in..

tests/testsuite.src/listings.at
```diff
@@ -2558,7 +2558,7 @@ SIZE  TYPE           LVL  NAME                           PICTURE
 0 errors in compilation group
 ])
 
-[AT_CHECK([diff prog17-32.lst prog.lis], [0], [], [])]
+AT_CHECK([diff prog17-32.lst prog.lis], [0], [], [])
```
GitMensch, line 2561: same as above

tests/testsuite.src/run_file.at
```diff
@@ -251,7 +251,7 @@ AT_DATA([prog.cob], [
 
 AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
-[echo "Test" > ./nosubhere]
+echo "Test" > ./nosubhere
```
GitMensch, line 254: We may use `AT_DATA` here, what do you think?

tests/testsuite.src/run_file.at
```diff
@@ -2880,10 +2880,10 @@ rewriting key: 98651234
 
 AT_CHECK([test "$COB_HAS_ISAM" = "db"], [0], [], [],
 # Previous test "failed" --> other ISAM, no need for more checks
-[AT_CHECK([true], [0])],
+AT_CHECK([true], [0]),
```
GitMensch, line 2883: same as above

tests/testsuite.src/run_fundamental.at
```diff
@@ -826,7 +826,7 @@ AT_DATA([prog.cob], [
            ADD  1.0       TO DST2.
            IF DST1 not = DST2 THEN
                DISPLAY 'move/compare of very near numbers failed (not id
-      -                'entical) [added]: ' DST1 " - " DST2
+      -                'entical) [[added]]: ' DST1 " - " DST2
```
GitMensch, line 829: I've reworded that

tests/testsuite.src/run_file.at
```diff
@@ -695,7 +695,7 @@ AT_DATA([prog.cob], [
 ])
 
 AT_CHECK([$COMPILE -fassign-clause=external prog.cob], [0], [], [])
-AT_CAPTURE_FILE(./whatever)
+AT_CAPTURE_FILE([./whatever])
```
GitMensch, line 698: Those captures are a good catch. I've adjusted them (no need for the dot either):
AT_CAPTURE_FILE(./something) ->AT_CAPTURE_FILE([something])
AT_CAPTURE_FILE(SOME) -> AT_CAPTURE_FILE([SOME])

tests/testsuite.src/run_manual_screen.at
```diff
@@ -24,7 +24,7 @@
 AT_SETUP([LINE])
 AT_KEYWORDS([screen])
 
-AT_SKIP_IF(test "$COB_HAS_CURSES" != "yes")
+AT_SKIP_IF([test "$COB_HAS_CURSES" != "yes"])
```
GitMensch, line 27: good catch, I've did the same

tests/testsuite.src/run_misc.at
```diff
@@ -2819,7 +2819,7 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
 AT_CLEANUP
 
 
-AT_SETUP([STRING / UNSTRING [NOT] ON OVERFLOW])
+AT_SETUP([STRING / UNSTRING [[NOT]] ON OVERFLOW])
```
GitMensch, line 2822: taken

tests/testsuite.src/syn_file.at
```diff
@@ -1128,7 +1128,7 @@ AT_DATA([prog.cob], [
        01  i-rec PIC X(10).
 ])
 
-/* FIXME: this test will false-positive fail on EBCDIC machines */
+# FIXME: this test will false-positive fail on EBCDIC machines
```
GitMensch, line 1131: interesting that this ever worked, thanks for catching it

tests/testsuite.src/syn_misc.at
```diff
@@ -1112,22 +1112,22 @@ AT_DATA([prog.cob], [
...
 
 # Previous test "failed" --> 32 bit
-[AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
 [prog.cob:6: error: 'SINGLE-ITEM' cannot be larger than 268435456 bytes
 prog.cob:8: error: 'FILLER 1' cannot be larger than 268435456 bytes
 prog.cob:7: error: 'GROUP-ITEM1' cannot be larger than 268435456 bytes
 prog.cob:9: error: 'GROUP-ITEM2' cannot be larger than 268435456 bytes
-])]
+])
```
GitMensch, line 1120: same as above

tests/testsuite.src/syn_multiply.at
```diff
@@ -47,7 +47,7 @@ AT_DATA([prog.cob], [
            STOP RUN.
 ])
 
-AT_CHECK([$COMPILE_ONLY prog.cob], [1], ,
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
```
GitMensch, line 50: already did that :-)

tests/testsuite.src/used_binaries.at
```diff
@@ -323,19 +323,19 @@ AT_CHECK([test -f prog.s], [0], [], [],
 # Previous test "failed" --> prog.s not available --> likely a VS build
 # only check for file as cl.exe cannot create executables from self-created
 # assembler sources
-[AT_CHECK([test -f prog.asm], [0], [], [])],
+AT_CHECK([test -f prog.asm], [0], [], []),
```
GitMensch, line 326: same as above

GitMensch (commented):
Thank you for taking the time to review the testsuite syntax.
I'm currently working on testsuite adjustments (because of adjusted/dropped msgids) and did part of that, but missed a lot of the others, which I've now included (the commit likely will need at least one day as I'm not finished with the msgid check).

Open points:
* I'm not sure yet about possibly using `AT_DATA` instead of `echo` - that would be reasonable, no?
* the seldom used format  "do something on failure/success of the previous test" likely should keep []

Just FYI:
* https://www.gnu.org/software/autoconf/manual/autoconf-2.71/html_node/Using-Autotest.html#Using-Autotest with included notes on writing and using the testsuite
* https://sourceforge.net/p/gnucobol/contrib/HEAD/tree/trunk/at_test_win/ for a 2017 approach to run the testsuite directly from its source via Java (which did work quite well back then)
* https://sourceforge.net/p/gnucobol/contrib/HEAD/tree/trunk/tools/gctest/ this is what Ron commonly uses to add new testcases (it may be useful to check if it outputs some "non homogenious parts")

GitMensch:
Because of other pending commits I've checked in all of the testsuite changes that were related and done, please update the mirror and branch, enabling us to see what's "missing"; do you have any thoughts on the open points outlined above?

lefessan:
I will update this PR on top of your commits, but here is a few points:

* I have a small prototype of tool that is able to run a large part of the tests, it's currently very basic, but I hope it will soon have some nice "modern" features. Given my agenda, I am only working on it from time to time, but it is on my top list when I have time for GnuCOBOL, as I am struggling more and more with the testsuite;

* About the current syntax: if we want tools to be able to use the test files, we should use as much M4 macros as possible, instead of using shell commands (that are extremely hard to parse...).

   * So, it is always better to use `AT_DATA` instead of `echo` (my tool for example can only handle single-line `echo` )
   * It would probably be better not to use `[ ... ]` to scope macros inside *run-if-fail* or *run-if-pass*,  because `[...]` is usually used to store *unparsed* string. Yet, it is still possible to recall the parser on these strings later, so it is probably still possible to make that work.
   * I don't know yet what to do with `if`s, but there are not so many of them...

GitMensch:
Most of the conditional "if failing, then ..." are about 32/64 bit, you may come up with an
GC_TEST32
GC_TEST64
macro (ideally defined in testsuite.at), the ABI that does not match would then be ignored, the other one executed. This should reduce the issues and create not so many conflicts (writing and using more macros needs to be postponed until everything is merged to trunk).

GitMensch:
Hm, I'll have a look at your suggestion next week after the update and push.

lefessan:
I still need also to review it, in particular the replacements of `if` by `AT_CHECK`.

lefessan:
For information, the program I wrote is here: https://github.com/ocamlpro/autofonce
It's main features:
* running the testsuite in parallel with minimal output
* creating a test by capturing a command execution

It requires this PR to be merge to fully work on gnucobol (once the workflows are fixed)

lefessan:
I have a new version of the tool that is much more tolerant to the syntax, I am going to submit an updated version of this PR with much fewer changes.

tests/testsuite.src/run_file.at
```diff
@@ -251,7 +251,7 @@ AT_DATA([prog.cob], [
 
 AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [], [])
-[echo "Test" > ./nosubhere]
+AT_CHECK([echo "Test" > ./nosubhere])
```
GitMensch, line 254: Please use AT_DATA here and below

GitMensch (approved):
the current version - 3 files only - can go to SVN any time, please switch the 2 echo to AT_DATA though.

tests/testsuite.src/data_pointer.at
lefessan: Now, these `[...]` scopes are mandatory for my tool, so I actually had to add them in a few places.

GitMensch:
feel free to check in

lefessan:
Merged !

PR #85, opened 2023-02-16
Add -fdiagnostics-show-caret to display the context of an error

lefessan:
When a warning or error is printed, prints the source context with the line of the error, plus 2 lines before and after.
COBC_PREVIEW_ERROR can be used to set the option globally.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -321,7 +321,7 @@ NEWS - user visible changes				-*- outline -*-
    original version; note: their use will be adjusted where they don't match
    GCC's same options in later versions, including addition of -M and -MD
 
-** New -std options:
+** new -std options:
 
    gcos            GCOS compatibility
    gcos-strict     GCOS compatibility                - strict
@@ -329,6 +329,14 @@ NEWS - user visible changes				-*- outline -*-
    We define the GCOS dialect based on the COBOL85 standard, with new
    dialect configuration options accompanying each specificity
    introduced by the dialect.
+
+** new diagnostic format for errors: the diagnostics now print the source
+   code context with a left margin showing line numbers, configurable with
+   -fno-diagnostics-show-line-numbers, and possible to disable completely
+   with -fno-diagnostics-show-caret;
+
+   the option -fdiagnostics-plain-output was added to request that diagnostic
+   output look as plain as possible and stay more stable over time
 
 * Important Bugfixes:
 
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -589,6 +589,7 @@ static const struct option long_options[] = {
 	{"P",			CB_OP_ARG, NULL, 'P'},
 	{"Xref",		CB_NO_ARG, NULL, 'X'},
 	{"use-extfh",		CB_RQ_ARG, NULL, 9},	/* this is used by COBOL-IT; Same is -fcallfh= */
+	{"fdiagnostics-plain-output",	CB_NO_ARG, NULL, '/'},
 	{"Wall",		CB_NO_ARG, NULL, 'W'},
 	{"Wextra",		CB_NO_ARG, NULL, 'Y'},		/* this option used to be called -W */
 #if 1
@@ -3457,6 +3458,12 @@ process_command_line (const int argc, char **argv)
 						cb_lines_per_page, 20);
 				cb_lines_per_page = 20;
 			}
+			break;
+
+		case '/':
+			/* -fdiagnostics-plain-output */
+			cb_diagnostics_show_caret = 0 ;
+			cb_diagnostics_show_line_numbers = 0;
 			break;
 
 		case 'P':
--- a/cobc/error.c
+++ b/cobc/error.c
@@ -32,6 +32,13 @@
 #include "cobc.h"
 #include "tree.h"
 
+enum cb_error_kind {
+	CB_KIND_ERROR,
+	CB_KIND_WARNING,
+	CB_KIND_NOTE,
+	CB_KIND_GENERAL
+};
+
 static char		*errnamebuff = NULL;
 static struct cb_label	*last_section = NULL;
 static struct cb_label	*last_paragraph = NULL;
@@ -63,10 +70,80 @@ print_error_prefix (const char *file, int line, const char *prefix)
 	}
 }
 
+/* Display a context around the location of the error/warning, only if
+ * cb_diagnostics_show_caret is true
+
+   Only display two lines before and after. No caret yet for the column as
+   we only have the line. Since we directly use the file, source is printed
+   before any REPLACE.
+ */
+
+#define CARET_MAX_COLS 73
 static void
-print_error (const char *file, int line, const char *prefix,
+diagnostics_show_caret (const char *file, int line)
+{
+	FILE* fd = fopen(file, "r");
+	if (fd == NULL) return;
+	char buffer[ CARET_MAX_COLS+1 ];
+	int line_pos = 1;
+	int char_pos = 0;
+	int printed = 0; /* nothing printed */
+	while(1){
+		int c = fgetc (fd);
+		if ( c == EOF ){
+			if (printed){
+				fprintf(stderr, "\n");
+			}
+			fclose(fd);
+			return ;
+		}
+		buffer[char_pos] = c ;
+		if (c == '\n' || char_pos == CARET_MAX_COLS){
+			buffer[char_pos] = 0;
+			if (line_pos > line-3 && line_pos < line+3){
+				if (line_pos == line-2) fprintf(stderr, "\n");
+				printed = 1;
+				fprintf (stderr, "  ");
+				if (cb_diagnostics_show_line_numbers){
+					fprintf (stderr, "%04d ",
+						 line_pos);
+				}
+				fprintf (stderr, "%c %s%s\n",
+					 line == line_pos ? '>' : ' ',
+					 c == '\n' ? "" : ".." ,
+					 buffer);
+				if (line_pos == line+2){
+					fprintf(stderr, "\n");
+					fclose(fd);
+					return;
+				}
+			}
+			while (c != '\n'){
+				/* skip end of line too long */
+				c = fgetc (fd);
+				if( c == EOF ) { fclose(fd); return ; }
+			}
+			line_pos++;
+			char_pos=0;
+		} else {
+			char_pos++;
+		}
+	}
+}
+
+static void
+print_error (const char *file, int line, enum cb_error_kind kind,
 	     const char *fmt, va_list ap, const char *diagnostic_option)
 {
+	const char *prefix;
+
+	switch( kind ){
+	case CB_KIND_ERROR: prefix = _("error: "); break;
+	case CB_KIND_WARNING: prefix = _("warning: "); break;
+	case CB_KIND_NOTE: prefix = _("note: "); break;
+	case CB_KIND_GENERAL: prefix = ""; break;
+	}
+
 	if (!file) {
 		file = cb_source_file;
 	}
@@ -119,12 +196,26 @@ print_error (const char *file, int line, const char *prefix,
 		}
 		cb_add_error_to_listing (file, line, prefix, errmsg);
 	}
+
+	static const char* last_caret_file = NULL ;
+	static int last_caret_line = -1 ;
+	if (cb_diagnostics_show_caret
+	    && file != NULL
+	    && strcmp (file, COB_DASH) != 0
+	    && line
+	    && (last_caret_file != file || last_caret_line != line)
+		){
+		/* remember last printed location to avoid reprinting it */
+		last_caret_file = file;
+		last_caret_line = line;
+		diagnostics_show_caret (file, line);
+	}
 }
 
 static void
 cobc_too_many_errors (void)
 {
-	if (!cb_diagnostic_show_option) {
+	if (!cb_diagnostics_show_option) {
 		fprintf (stderr, "cobc: %s\n",
 			_("too many errors"));
 	} else
@@ -306,7 +397,7 @@ static char *warning_option_text (const enum cb_warn_opt opt, const enum cb_warn
 {
 	const char *opt_name;
 
-	if (!cb_diagnostic_show_option) {
+	if (!cb_diagnostics_show_option) {
 		return NULL;
 	}
 	switch (opt) {
@@ -343,9 +434,9 @@ cb_warning_internal (const enum cb_warn_opt opt, const char *fmt, va_list ap)
 	}
 
 	if (pref != COBC_WARN_AS_ERROR) {
-		print_error (NULL, 0, _("warning: "),  fmt, ap, warning_option_text (opt, pref));
+		print_error (NULL, 0, CB_KIND_WARNING,  fmt, ap, warning_option_text (opt, pref));
 	} else {
-		print_error (NULL, 0, _("error: "),  fmt, ap, warning_option_text (opt, pref));
+		print_error (NULL, 0, CB_KIND_ERROR,  fmt, ap, warning_option_text (opt, pref));
 	}
 
 	if (sav_lst_file) {
@@ -379,7 +470,7 @@ cb_error_always (const char *fmt, ...)
 
 	cobc_in_repository = 0;
 	va_start (ap, fmt);
-	print_error (NULL, 0, _("error: "), fmt, ap, NULL);
+	print_error (NULL, 0, CB_KIND_ERROR, fmt, ap, NULL);
 	va_end (ap);
 
 	if (sav_lst_file) {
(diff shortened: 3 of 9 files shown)


tests/testsuite.src/used_binaries.at
```diff
@@ -793,19 +793,40 @@ AT_CLEANUP
...
        PROGRAM-ID.      prog.
        PROCEDURE        DIVISION.
            DISPLAY "jobby"
            EXIT HERE.
 ])
+AT_CHECK([$COMPILE -j prog.cob], [1], [], [prog.cob:6: error: syntax error, unexpected Identifier
+])
+
+# check -fpreview-error prints error context
+AT_CHECK([$COMPILE -j -fpreview-error prog.cob], [1], [], [prog.cob:6: error: syntax error, unexpected Identifier
```
lefessan, line 808: Here is an example...

lefessan:
As an example:
```
prompt> $COMPILE -j -fpreview-error prog.cob
prog.cob:6: error: syntax error, unexpected Identifier

  0004          PROCEDURE        DIVISION.
  0005              DISPLAY "jobby"
  0006 >            EXIT HERE.

```

lefessan:
Note that the default is currently to disable this feature, but it could be activated by default (and disabled through `COBC_PREVIEW_ERROR=0` for the testsuites)

GitMensch:
Hm, the idea is good and the general point of reworking the diagnostic format is even on the GSOC idea list for GnuCOBOL (did you get it from there/the note on the dev list about that?).

But reading the source file again should be absolute last resort and will be very likely wrong for REPLACE/REPLACING and won't work for complies from stdin.

So... Could you please rework that to possibly save the context from the scanner via to a static buffer?

Also note that -fno-diagnostics-show-line-numbers should be added along with this change (which would drop the leading line numbers) and possibly -fdiagnostics-show-caret should be used to toggle this instead of -fpreview-error ... and ideally we should also use the caret to show where the error is.

As a general rule of thumb we'd like to reach https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Message-Formatting-Options.html ...
someday.

lefessan:
> Hm, the idea is good and the general point of reworking the diagnostic format is even on the GSOC idea list for GnuCOBOL (did you get it from there/the note on the dev list about that?). 

No, I didn't know about the GSOC idea. Is it discussed on the forum ?
I did that patch because I needed it myself... which is usually how I work, as I can only devote a small amount of time on the project from time to time.

> But reading the source file again should be absolute last resort and will be very likely wrong for REPLACE/REPLACING and won't work for complies from stdin. So... Could you please rework that to possibly save the context from the scanner via to a static buffer? 
> Also note that -fno-diagnostics-show-line-numbers should be added along with this change (which would drop the leading line numbers) and possibly -fdiagnostics-show-caret should be used to toggle this instead of -fpreview-error ... and ideally we should also use the caret to show where the error is. 

I won't be able to spend much more time on this patch. For now, I changed the name of the option to `-fdiagnostics-show-caret` and added another option to disable it for warnings (typically, we are working on a project where there are currently a lot of warnings, and we want to only focus on errors in a first step).

> As a general rule of thumb we'd like to reach https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Message-Formatting-Options.html ... someday.

Would it be ok to merge this work as it is, and wait for the GSOC to get it improved ? I will be working on adding a bit more GCOS support now, so I won't be able to improve this patch more before some time.

lefessan:
@GitMensch Hi, I say on the forum that you mentioned this patch. Do you think it would be ok to merge it in the current state ?

cobc/error.c
```diff
@@ -63,10 +70,73 @@ print_error_prefix (const char *file, int line, const char *prefix)
...
+   before any REPLACE.
+ */
+static void
+diagnostics_show_caret  (enum cb_error_kind kind, const char *file, int line)
+{
+	if ( cb_diagnostics_show_caret
+	     && file != NULL && line
+	     && ( cb_diagnostics_warning_show_caret || kind == CB_KIND_ERROR )
+		){
+		FILE* fd = fopen(file, "r");
```
GitMensch, line 88: needs to check `fd` after open and likely should check before and not do that if name matches `COB_DASH`

GitMensch (changes requested):
> @GitMensch Hi, I've seen on the forum that you mentioned this patch. Do you think it would be ok to merge it in the current state ?

Not in its current state :-/ But after reviewing the [initial GCC implementation](https://github.com/gcc-mirror/gcc/commit/9fec00429dd36ea51798afeaf5f6b9c1987b85cd) (also opened the file then read and print the lines), I'm now open to add it without the design change I previously thought of (GCC did that later on).

Things that are likely easy to do:

* drop `COBC_DIAGNOSTICS_SHOW_CARET` completely; for the testsuite it is enough to add `-fno-diagnostics-show-caret` in tests/atlocal.in, just as I did with `-fno-diagnostics-show-option`
* rebase
* add `-fno-diagnostics-show-line-numbers` (because it is just a single flag that only modifies the output in one place)
* add an entry in NEWS and doc/gnucobol.tex about the two new options

Then, as a follow up, we may [use source caches, like the current GCC implementation does](https://github.com/gcc-mirror/gcc/blob/5d85b5d649fff675ff00adcc99371bccf4ef5944/gcc/input.cc#L858).

But still you may want to have a look at the old diagnostic print function linked before your commit and look for good style; and have a look at [the current one](https://github.com/gcc-mirror/gcc/blob/5d85b5d649fff675ff00adcc99371bccf4ef5944/gcc/diagnostic-show-locus.cc#L2831)

cobc/cobc.c
```diff
@@ -8648,6 +8648,18 @@ set_cobc_defaults (void)
 		 *p == '1')) {
 		cb_unix_lf = 1;
 	}
+	p = cobc_getenv ("COBC_DIAGNOSTICS_SHOW_CARET");
```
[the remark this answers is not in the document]
lefessan: I want to argue to keep it: most of the options in the compiler do not need an env variable because they are project-dependent, i.e. the project developers should whether these options should be set or not for their project. On the contrary, the way errors are displayed is user-dependent, not project-dependent. As such, users may want to set or unset these options in their config files, so that they are used by `make` automatically. So the env variable is needed for that (unless we add a user config file for gnucobol).

cobc/flag.def
```diff
@@ -232,6 +232,16 @@ CB_FLAG (cb_listing_symbols, 1, "tsymbols",
...
+CB_FLAG_ON (cb_diagnostics_show_option, 1, "diagnostics-show-option",
 	_("  -fno-diagnostics-show-option\tsuppress output of option that directly\n"
 	  "                        controls the diagnostic"))
+
+CB_FLAG (cb_diagnostics_show_caret, 1, "diagnostics-show-caret",
+	_("  -fdiagnostics-show-caret     display source context on warning/error diagnostic\n"
+	  "                               (the source is displayed without any REPLACE performed."
+	))
+
+CB_FLAG_ON (cb_diagnostics_warning_show_caret, 1, "diagnostics-show-caret-on-warning",
```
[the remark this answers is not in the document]
lefessan: I removed it.

cobc/error.c
lefessan: I added checks for NULL and COB_DASH.

cobc/cobc.c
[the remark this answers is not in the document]
GitMensch: I see the point - but we shouldn't do this for a single option, a user may want to have the diagnostic name or not, he maybe even don't care about any warnings, ...
A project specific setup that handles this would set default flags in the makefile and passes additional flags from the user's environment or include file (depending on the setup either before or after the default makefile flags).

Actually the way projects can nicely be setup via Makefiles is the main reason for not having a global env var or a config file (other than dialect options). There's a FR about this on SF if you want to give that more thoughts - for this PR, please drop that.

NEWS
```diff
@@ -325,12 +325,19 @@ NEWS - user visible changes				-*- outline -*-
...
+  the error message.  With -fno-diagnostics-show-line-numbers, the line
+  numbers are not printed within the content.
+
 * Important Bugfixes:
 
 ** for dialects other than the GnuCOBOL default different reserved "alias" words
-   were not usable, for example SYNCHRONIZED or COMPUTATIONAL. This was fixed
-   and reserved words updated for the dialects "acu" (to ACUCOBOL-GT 10.4),
-   "ibm" (to Enterprise COBOL 6.3) and "mf" (to Micro Focus Visual COBOL 6.0)
+   were not usable, for example SYNCHRONIZED or COMPUTATIONAL. This
```
GitMensch, line 337: unrelated change?

cobc/error.c
```diff
@@ -63,10 +70,80 @@ print_error_prefix (const char *file, int line, const char *prefix)
...
+		}
+		buffer[char_pos] = c ;
+		if (c == '\n' || char_pos == CARET_MAX_COLS){
+			buffer[char_pos] = 0;
+			if (line_pos > line-3 && line_pos < line+3){
+				if (line_pos == line-2) fprintf(stderr, "\n");
+				printed = 1;
+				fprintf (stderr, "  ");
+				if (cb_diagnostics_show_line_numbers){
+					fprintf (stderr, "%04d ",
```
GitMensch, line 108: Currently that does not only skip line number but the whole printing, no?

tests/testsuite.src/used_binaries.at
```diff
@@ -918,3 +912,100 @@ AT_CHECK([cat prog.cob | $COMPILE_MODULE -j -], [0], [job], [],
...
+
+  0005          WORKING-STORAGE  SECTION.
+  0006          01 TEST-VAR PIC 9(2) VALUE 'A'.
+  0007 >        COPY 'CRUD.CPY'.
+  0008          PROCEDURE        DIVISION.
+  0009              DISPLAY TEST-VAR NO ADVANCING
+
+prog.cob:6: warning: numeric value is expected
+])
+AT_CHECK([$COMPILE -fdiagnostics-show-caret -fno-diagnostics-show-caret-on-warning -j prog.cob], [1], [], [prog.cob:7: error: CRUD.CPY: No such file or directory
```
GitMensch, line 999: `-fno-diagnostics-show-caret-on-warning` doesn't exists any more, please replace it with a test for `-fno-diagnostics-show-line-numbers`

GitMensch (changes requested):
Nearly ready now - see review details.

GitMensch:
freindly ping @lefessan

tests/testsuite.src/used_binaries.at
lefessan: done, replaced with a test on `-fno-diagnositcs-show-line-numbers`

cobc/error.c
lefessan: Wrong indentation, sorry...

lefessan:
I think all your requests have been fixed.

cobc/flag.def
```diff
@@ -232,6 +232,16 @@ CB_FLAG (cb_listing_symbols, 1, "tsymbols",
...
 	_("  -fno-diagnostics-show-option\tsuppress output of option that directly\n"
 	  "                        controls the diagnostic"))
+
+CB_FLAG (cb_diagnostics_show_caret, 1, "diagnostics-show-caret",
+	_("  -fdiagnostics-show-caret     display source context on warning/error diagnostic\n"
+	  "                               (the source is displayed without any REPLACE performed."
+	))
+
+CB_FLAG_ON (cb_diagnostics_show_line_numbers, 1, "diagnostics-show-line-numbers",
+	_("  -fno-diagnostics-show-line-numbers     suppress display of line numbers in diagnostics"
```
GitMensch, line 245: "strange" thing to do, but please start the second line on the same position that is used for `diagnostics-show`... and as the name is long delimit the first line with \t instead of spaces

tests/testsuite.src/used_binaries.at
```diff
@@ -794,23 +794,17 @@ AT_CLEANUP
 
 AT_SETUP([run job with unsuccessful compilation])
 AT_KEYWORDS([runmisc cobcrun])
-
-
-AT_DATA([prog.cob], [
+AT_DATA([prog.cob],[
```
GitMensch, line 797: please drop these and the following unrelated changes (enough during commit)

tests/testsuite.src/used_binaries.at
```diff
@@ -918,3 +912,62 @@ AT_CHECK([cat prog.cob | $COMPILE_MODULE -j -], [0], [job], [],
...
+       COPY 'CRUD.CPY'.
+       PROCEDURE        DIVISION.
+           DISPLAY TEST-VAR NO ADVANCING
+           END-DISPLAY
+           MOVE 12 TO TEST-VAR
+           DISPLAY TEST-VAR NO ADVANCING
+           END-DISPLAY
+           STOP RUN.
+])
+AT_CHECK([$COBC -Wall -fsyntax-only prog.cob], [1], [], [prog.cob:7: error: CRUD.CPY: No such file or directory
```
GitMensch, line 935: this line should be:

```
[prog.cob:7: error: CRUD.CPY: No such file or directory
```

the following test should be

```
AT_CHECK([$COMPILE_ONLY -fdiagnostics-show-caret -fdiagnostics-show-line-numbers prog.cob], [1], [],
[prog.cob:7: error: CRUD.CPY: No such file or directory
```

and the last (which may be switched with the second)

```
AT_CHECK([$COMPILE_ONLY -fdiagnostics-show-caret -prog.cob], [1], [],
[prog.cob:7: error: CRUD.CPY: No such file or directory
```

cobc/flag.def
```diff
@@ -232,6 +232,16 @@ CB_FLAG (cb_listing_symbols, 1, "tsymbols",
 CB_FLAG (cb_listing_cmd, 1, "tcmd",
 	_("  -ftcmd                specify command line in listing"))
 
-CB_FLAG_ON (cb_diagnostic_show_option, 1, "diagnostics-show-option",
+CB_FLAG_ON (cb_diagnostics_show_option, 1, "diagnostics-show-option",
 	_("  -fno-diagnostics-show-option\tsuppress output of option that directly\n"
 	  "                        controls the diagnostic"))
+
+CB_FLAG (cb_diagnostics_show_caret, 1, "diagnostics-show-caret",
```
GitMensch, line 239: If possible please change to `CB_FLAG_ON` and add the reverse to the flags in tests/atlocal.in

GitMensch (changes requested):
only minor changes needed (feel free to directly commit)

If you want to finish that completely, then add `-fdiagnostics-plain-output` to cobc.c (this is a "normal" option, no flag, it has no "no" and would set both new options to ´0`) and use this in tests/atlocal.in for the flags.

GitMensch:
Possibly use in the NEWS something similar to GCCs entry:

the diagnostics now print the source code context with a left margin showing line numbers, configurable with -fno-diagnostics-show-line-numbers, and possible to disable completely with -fno-diagnostics-show-caret;

the option -fdiagnostics-plain-output was added to request that diagnostic output look as plain as possible and stay more stable over time

cobc/error.c
```diff
@@ -63,10 +70,80 @@ print_error_prefix (const char *file, int line, const char *prefix)
 	}
 }
 
+/* Display a context around the location of the error/warning, only if
+ * cb_diagnostics_show_caret is true
+
+   Only display two lines before and after. No caret yet for the column as
```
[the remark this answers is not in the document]
lefessan: My intuition is that it's better to display the file as it is, so that it is easy for the developer to visually match the error location in a terminal with the one in an editor.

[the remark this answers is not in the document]
GitMensch: Totally reasonable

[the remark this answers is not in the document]
lefessan: My intuition is that it's better to display the file as it is, so that it is easy for the developer to visually match the error location in a terminal with the one in an editor.

GitMensch (commented):
the last commit pushed LGTM

GitMensch (approved):

lefessan:
I think I added all your suggestions, I will commit it to svn as soon as CI is passed. Thanks !

lefessan:
Merged in SVN

GitMensch:
Sadly that broke C89 support in cobc/error.c: https://ci.appveyor.com/api/buildjobs/vqa49l488as7peee/log

Please adjust to declare the variables at the beginning of the scope.

Apart from that: what do you think of adding a std=c89 to the CI here? That _should_ have likely called that.

lefessan:
Sorry, I didn't see your comment, I am going to fix this problem asap

GitMensch:
Please drop a note if you think you'll have moved around the declarations in the next two hours as I'm doing it otherwise afterwards.

GitMensch:
Fixing C89 and additional changes to the format and some of the details done "upstream" in [svn r5081](https://sourceforge.net/p/gnucobol/code/5081/).

CI PR already referenced.

GitMensch:
Can you have a look for some late-cleanup on this, moving ` -fdiagnostics-plain-output` from `COBC` to `COBOL_FLAGS`, please?
I'd like to keep the command line for NIST, which uses `COBC`, as small as possible.

GitMensch:
In the meantime I've did that adjustment upstream.

PR #86, opened 2023-02-16
Remove VALUES=VALUE in gcos.conf

lefessan:
Probably forgotten when VALUES was added.


GitMensch:
Good to be checked in to 3.x branch, please double check before if that needs to be done in any other word file

lefessan:
Merged upstream

GitMensch:
Thank you!

Maybe we should have one of the tests with VALUES ARE (a conditional one?) adjusted to be compiled with every std?

PR #87, opened 2023-02-19
Fix wrong invalid conditional expression

lefessan:
`IF V ZERO AND W EQUAL 1` is interpreted as invalid when it should be legal

The test added in the testsuite fails with previous versions.

### DIFF
--- a/cobc/tree.c
+++ b/cobc/tree.c
@@ -6048,12 +6048,12 @@ cb_build_binary_op (cb_tree x, const enum cb_binary_op_op op, cb_tree y)
 		rel_bin_op = 1;
 		if ((CB_REF_OR_FIELD_P (x))
 		 && CB_FIELD_PTR (x)->level == 88) {
-			cb_error_x (e, _("invalid expression"));
+			cb_error_x (e, _("invalid expression: conditional on the left of numeric operator"));
 			return cb_error_node;
 		}
 		if ((CB_REF_OR_FIELD_P (y))
 		 && CB_FIELD_PTR (y)->level == 88) {
-			cb_error_x (e, _("invalid expression"));
+			cb_error_x (e, _("invalid expression: conditional on the right of numeric operator"));
 			return cb_error_node;
 		}
 
@@ -6281,7 +6281,7 @@ cb_build_binary_op (cb_tree x, const enum cb_binary_op_op op, cb_tree y)
 				cb_error_x (e, _("invalid expression: %s %s %s"),
 					llit, explain_operator (op), rlit);
 			} else {
-				cb_error_x (e, _("invalid expression"));
+				cb_error_x (e, _("invalid expression: boolean expected with logical operator"));
 			}
 			return cb_error_node;
 		}
--- a/cobc/tree.c
+++ b/cobc/tree.c
@@ -6046,16 +6046,20 @@ cb_build_binary_op (cb_tree x, const enum cb_binary_op_op op, cb_tree y)
 	case ']':
 		/* Relational operators */
 		rel_bin_op = 1;
+#if 0	/* note: already tested in the parser with (check_not_88_level) */
 		if ((CB_REF_OR_FIELD_P (x))
 		 && CB_FIELD_PTR (x)->level == 88) {
-			cb_error_x (e, _("invalid expression: conditional on the left of numeric operator"));
+			/* because this code is not active and the translation would be new,
+			   we don't have that gettextized */
+			cb_error_x (e, "invalid expression: conditional on the left of numeric operator");
 			return cb_error_node;
 		}
 		if ((CB_REF_OR_FIELD_P (y))
 		 && CB_FIELD_PTR (y)->level == 88) {
-			cb_error_x (e, _("invalid expression: conditional on the right of numeric operator"));
+			cb_error_x (e, "invalid expression: conditional on the right of numeric operator");
 			return cb_error_node;
 		}
+#endif
 
 		if (x == cb_zero) {
 			xl = CB_LITERAL(cb_zero_lit);
--- a/cobc/typeck.c
+++ b/cobc/typeck.c
@@ -5801,22 +5801,23 @@ build_expr_finish (void)
 	/* Reduce all (prio of token 0 is smaller than all other ones) */
 	(void)build_expr_reduce (0);
 
+	expr_stack[TOPSTACK].value->source_file = cb_source_file;
+	expr_stack[TOPSTACK].value->source_line = cb_exp_line;
+
+	if (expr_index != TOPSTACK+1) {
+		cb_error_x (expr_stack[TOPSTACK].value, _("invalid expression: unfinished expression"));
+		return cb_error_node;
+	}
+
 	if (!expr_stack[TOPSTACK].value) {
 		/* TODO: Add test case for this to syn_misc.at invalid expression */
 		cb_error (_("invalid expression"));
 		return cb_error_node;
 	}
 
-	expr_stack[TOPSTACK].value->source_file = cb_source_file;
-	expr_stack[TOPSTACK].value->source_line = cb_exp_line;
-
-	if (expr_index != TOPSTACK+1) {
-		cb_error_x (expr_stack[TOPSTACK].value, _("invalid expression"));
-		return cb_error_node;
-	}
-
 	build_expr_expand (&expr_stack[TOPSTACK].value);
 	if (expr_stack[TOPSTACK].token != 'x') {
+		/* TODO: add a test case, for now, no idea how to reach this */
 		cb_error_x (expr_stack[TOPSTACK].value, _("invalid expression"));
 		return cb_error_node;
 	}
@@ -5883,8 +5884,22 @@ cb_build_expr (cb_tree list)
 			has_rel = 1;
 			break;
 		default:
+			if(TOKEN (-1) == '!'){
+/* switch `NOT` relation, e.g. the two expression tokens `NOT` and `>`
+ * are reduced to a single token `<=` */
+				switch(op){
+				case '=': TOKEN (-1) = '~'; continue;
+				case '>': TOKEN (-1) = '['; continue;
+				case '<': TOKEN (-1) = ']'; continue;
+				case ']': TOKEN (-1) = '<'; continue;
+				case '[': TOKEN (-1) = '>'; continue;
+				}
+			}
 			v = CB_VALUE (l);
 			if (op == 'x') {
+                                if( has_var && v == cb_zero ){
+                                        has_rel = 1;
+                                }
 				has_var = 1;
 				if (CB_TREE_TAG (v) == CB_TAG_BINARY_OP) {
 					has_rel = 1;
@@ -6269,7 +6284,7 @@ decimal_compute (const int op, cb_tree x, cb_tree y)
 /**
  * expand tree x to the previously allocated decimal tree d
  */
-static void
+static cb_tree
 decimal_expand (cb_tree d, cb_tree x)
 {
 	struct cb_literal	*l;
@@ -6280,14 +6295,14 @@ decimal_expand (cb_tree d, cb_tree x)
 	/* skip if the actual statement can't be generated any more
 	   to prevent multiple errors here */
 	if (error_statement == current_statement) {
-		return;
+		return d;
 	}
 	switch (CB_TREE_TAG (x)) {
 	case CB_TAG_CONST:
 		/* LCOV_EXCL_START */
 		if (x != cb_zero) {
 			cobc_err_msg (_("unexpected constant expansion"));
-			COBC_ABORT ();
+			return cb_error_node;
 		}
 		/* LCOV_EXCL_STOP */
 		dpush (CB_BUILD_FUNCALL_2 ("cob_decimal_set_llint", d, cb_int0));
@@ -6354,13 +6369,21 @@ decimal_expand (cb_tree d, cb_tree x)
 		 * Set t, Y
 		 * OP d, t */
 		p = CB_BINARY_OP (x);
-		if ((p->op == 'c' || p->op == 'd')		/* Circular Shift */
-		 && CB_REF_OR_FIELD_P (p->x)) {
-			sz_shift = cb_int (CB_FIELD_PTR (p->x)->size);
-		} else {
-			sz_shift = cb_int1;
+
+		switch (p->op){
+		case '=': case '~': case '<': case '>': case '[': case ']':
+		case '!': case '&': case '|':
+			cb_error_x (x, _("condition expression found where decimal expression was expected"));
+			return cb_error_node;
+		case 'c':
+		case 'd':
+			if (CB_REF_OR_FIELD_P (p->x)) {
+				sz_shift = cb_int (CB_FIELD_PTR (p->x)->size);
+				break;
+			}
+		default: sz_shift = cb_int1;
 		}
-		decimal_expand (d, p->x);
+		d = decimal_expand (d, p->x);
 
 		if (CB_TREE_TAG (p->y) == CB_TAG_LITERAL
 		 && CB_TREE_CATEGORY (p->y) == CB_CATEGORY_NUMERIC) {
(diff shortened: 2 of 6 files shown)


GitMensch:
Why should  `IF VAR ZERO` be valid?

lefessan:
> Why should `IF VAR ZERO` be valid?

It is already the case, GnuCOBOL allows `IF VAR IS ZERO` as a valid expression, but not `IF VAR IS ZERO AND ...`

Isn't it 8.8.4.7.2 of ISO 1989:2023 ?

GitMensch:
You confuse me, your previous version shows that after your change `IF VAR ZERO` would compile, but the operator is mandatory, so this is wrong and should raise an error.

`IF VAR IS ZERO AND VAR2 IS NOT ZERO` should of course work, and I believe it does already.

lefessan:
Actually, I wonder if the operator is mandatory: I made this patch because we have some code without the operator (that used to compile on GCOS) and ISO 1989:2013 does not underline the `IS`, so it should be optional, which is already the case in GnuCOBOL (I didn't modify the parser, and the `POSITIVE` case works without my patch).

>  `IF VAR IS ZERO AND VAR2 IS NOT ZERO` should of course work, and I believe it does already.

No, it complains about `invalid conditional expression`

lefessan:
I created a bug report: https://sourceforge.net/p/gnucobol/bugs/875/

lefessan:
I think there was a related bug on the sourceforge bugtracker, would it be possible to merge this PR too ?

GitMensch:
:-) this shows that there is no Changelog entry in this PR as this would have a reference to the bug issue :-)

Yes, there is - it is https://sourceforge.net/p/gnucobol/bugs/880/ and it actually should be fixed before 3.2 final, but it _may_ has to wait until after the RC3.

As far as I remember there was an issue with this change, I'd have to recheck.

Apart from the missing Changelog - is 880 the same as youz 875?

lefessan:
This new version should fix 875 and 880. 880 is a different bug, it is because `NOT EQUAL` was not translated to `~` in the parser, so it has to be done in `cb_build_expr`. When fixing that, I found another bug in the testsuite: PIC 99 >= 99 would print a warning that it may always be false, when it should only be printed for PIC 99 > 99. It is also fixed in the PR.

cobc/typeck.c
```diff
@@ -5404,12 +5412,16 @@ expr_chk_cond (cb_tree expr_1, cb_tree expr_2)
 	return is_ptr_1 ^ is_ptr_2;
 }
 
+/* Reduce expressions of the left of 'token' that have higher priority */
 static int
-expr_reduce (int token)
+cb_expr_reduce (int token)
```
GitMensch, line 5417: we are not 100% clean on this yet, but as a rule of thumb: `cb_` are externalized functions in cobc, `cob_` externalized in libcob, otherwise these prefixes should not be usd

lefessan: Indeed, I renamed two functions because they were the 2 only ones that didn't have this prefix in a clique of functions that were actually _all_ static... So, I should have done the contrary :-)

lefessan: I would suggest to merge this PR and fix the prefixes later :-)

GitMensch: I'm fine with fixing old prefixes later, but please revert the "bad" change here (or ensure later to not commit it) to not let the lines show up on git/svn blame for an unrelated change, then change back

lefessan: Ok, I have fixed it by changing the prefix from `cb_expr` to `build_expr` for all the functions in this clique.

lefessan:
@GitMensch  Do you think this PR is ready for merge ? (after fixing the ChangeLog conflict)

GitMensch:
I'll try to have a look at this soon (bit that may still take some time as I'm not at the PC).

tests/testsuite.src/syn_misc.at
```diff
@@ -3530,14 +3526,62 @@ AT_DATA([prog.cob], [
...
+prog.cob:16: error: invalid expression: boolean expected with logical operator
 prog.cob:28: error: wrong number of WHEN parameters
 prog.cob:31: error: wrong number of WHEN parameters
 ])
 
 AT_CLEANUP
 
 
+AT_SETUP([Condition IS ZERO AND])
+AT_KEYWORDS([condition zero])
```
GitMensch, line 3538: no need to re-iterate title keywords, either adjust the title or comment the keywords for now

tests/testsuite.src/syn_misc.at
```diff
@@ -2479,8 +2476,7 @@ prog.cob:55: error: HANDLE item not allowed here: 'mywindow'
 prog.cob:56: error: HANDLE item not allowed here: 'mydir'
 ])
 
-AT_CHECK([$COMPILE_ONLY -std=rm-strict prog.cob], [1], [],
-[prog.cob:12: error: 'handle' is not defined, but is a reserved word in another dialect
+AT_CHECK([$COMPILE_ONLY -std=rm-strict prog.cob], [1], [], [prog.cob:12: error: 'handle' is not defined, but is a reserved word in another dialect
```
GitMensch, line 2479: format change, please stay with the current one

cobc/typeck.c
```diff
@@ -5772,41 +5785,41 @@ expr_expand (cb_tree *x)
...
 	}
 
-	expr_stack[3].value->source_file = cb_source_file;
-	expr_stack[3].value->source_line = cb_exp_line;
-
-	if (expr_index != 4) {
+	if (!expr_stack[TOPSTACK].value) {
 		/* TODO: Add test case for this to syn_misc.at invalid expression */
-		cb_error_x (expr_stack[3].value, _("invalid expression"));
+		cb_error (_("invalid expression: final expression is not a value"));
```
GitMensch, line 5812: please add a testcase for that, then drop the TODO note (ideally also for the one below)

cobc/typeck.c
```diff
@@ -5772,41 +5785,41 @@ expr_expand (cb_tree *x)
...
-	(void)expr_reduce (0);
+	/* Reduce all (prio of token 0 is smaller than all other ones)*/
+	(void)build_expr_reduce (0);
+
+	expr_stack[TOPSTACK].value->source_file = cb_source_file;
+	expr_stack[TOPSTACK].value->source_line = cb_exp_line;
 
-	if (!expr_stack[3].value) {
+	if (expr_index != TOPSTACK+1) {
 		/* TODO: Add test case for this to syn_misc.at invalid expression */
```
GitMensch, line 5805: there is a testcase, so this TODO can be dropped

GitMensch (changes requested):
This effectively reverts a warning that was added on purpose in 3.1 or afterwards - this parts needs an adjustment.

To both make the svn diff more easy later on and this PR smaller: can you please
* commit to svn directly the renaming (with the Changelog entry) and the added comments
* rebase the target branch and this PR
* have this PR "only" include the functional changes (with the Changelog entry) and testsuite adjustments

GitMensch:
> When fixing that, I found another bug in the testsuite: PIC 99 >= 99 would print a warning that it may always be false, when it should only be printed for PIC 99 > 99. It is also fixed in the PR.

Actually this was _added_ (and therefore was **on purpose**) - we may could present a different message here, but the point is that  the ">" part can never be true, and this is an important information because it hints at possibly "unexpected" `PIC` or mistyped condition.

tests/testsuite.src/syn_misc.at
```diff
@@ -379,6 +379,7 @@ prog.cob:159: warning: 'PIC-9-SIGNED-DECIMAL' may not be GREATER THAN 99.99
 prog.cob:160: warning: 'PIC-9-SIGNED-DECIMAL' may not be LESS THAN -99.99
 prog.cob:161: warning: 'PIC-9-SIGNED-DECIMAL' may not be LESS THAN -99.99
 prog.cob:162: warning: literal '-99.991' has more decimals than 'PIC-9-SIGNED-DECIMAL'
+prog.cob:164: warning: 'XX' may not be GREATER THAN 99
```
GitMensch, line 382: I'll check later when you request the review, just wondering - why should this be (only) raised in one case but not in the other?

lefessan: I don't know, I just know that my fix replaces `NOT <` by `>=` and thus triggers the warning in one case.

cobc/typeck.c
lefessan: Hum... I don't really see when such a case can happen, my understanding of this code is still superficial...

cobc/typeck.c
```diff
@@ -7107,7 +7132,7 @@ cb_build_cond (cb_tree x)
 		if (x != cb_any && x != cb_true && x != cb_false) {
 			/* TODO: Add test case for this to syn_misc.at invalid expression */
```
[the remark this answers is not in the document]
lefessan: Same as above :-(

lefessan:
I think that, except the two test cases that I am unable to generate without deeper thinking, I addressed all your requests.

tests/testsuite.src/syn_misc.at
GitMensch: please leave a CHECKME comment in the testsuite source mentioning that we may be able to raise the same for `NOT a < b` as we do for `a NOT < b`

GitMensch (approved):
there are still minor things open, but the general patch is fine - feel free to directly commit adjusted

... I _guess_ you've rechecked that those expression work on GCOS?

GitMensch:
Please try to push that soon, as it should be part of RC3.

tests/testsuite.src/syn_misc.at
```diff
@@ -8200,3 +8200,82 @@ AT_CHECK([$COMPILE -freserved="XX*=BYTE-LENGTH" prog.cob], [0], [], [])
...
+           IF VAR1 = (VAR2 AND VAR3 AND VAR4)
+              DISPLAY "TRUE" UPON STDERR
+	   END-IF
+           IF VAR1 NOT = (VAR2 AND VAR3 AND VAR4)
+              DISPLAY "TRUE" UPON STDERR
+           END-IF
+           IF VAR1 NOT > (VAR2 AND VAR3 AND VAR4)
+              DISPLAY "TRUE" UPON STDERR
+           END-IF
+           GOBACK.
```
GitMensch, line 8277: Please adjust the test (which also gets rid of a leading tab):

```cob
           IF VAR1 = (VAR2 AND VAR3 AND VAR4)
              DISPLAY "TRUE 1".
           IF VAR1 NOT = (VAR2 AND VAR3 AND VAR4)
              DISPLAY "TRUE 2".
           IF VAR1 NOT > (VAR2 AND VAR3 AND VAR4)
              DISPLAY "TRUE 3".
```

then move both tests from syn_misc.at to run_fundamental.at and also check the second output, which - according to Micro Focus - should be 
> TRUE 2
> TRUE 3

GitMensch (changes requested):
With the CI working again (and actually catching a potential  - even if rare - issue) we can come back to this one; only the testcases need some adjustment, then this is good to go:

* move the two bug tests (with replacing keywords by comment and possibly use `IF` as keyword instead)
* check one additional result there
* ideally add a test case for `conditional on the left/right of numeric operator`

GitMensch:
Out of interest: does this bad program still compile?

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID.    CHECKCOND.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01  VAR1                    PIC X.
                88 VAR1-K VALUE 'K'
       01  VAR2                    PIC X.
                88 VAR2-K VALUE 'K'

       PROCEDURE DIVISION.
       MAIN-PROGRAM SECTION.
       BUG.
           IF VAR1-K AND NOT = VAR2-K
              DISPLAY "INVALID" UPON STDERR.
           GOBACK.
```

Please add this to the testsuite, either with not compiling or as expected failure.

cobc/tree.c
```diff
@@ -6048,12 +6048,12 @@ cb_build_binary_op (cb_tree x, const enum cb_binary_op_op op, cb_tree y)
 		rel_bin_op = 1;
 		if ((CB_REF_OR_FIELD_P (x))
 		 && CB_FIELD_PTR (x)->level == 88) {
-			cb_error_x (e, _("invalid expression"));
+			cb_error_x (e, _("invalid expression: conditional on the left of numeric operator"));
```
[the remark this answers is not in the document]
lefessan: So, I tried, but it does not look as easy as expected, as it triggers another error before this one (`condition-name not allowed here`). After a few tries, I found another error that segfaulted the compiler in `decimal_expand()`, hence another fix in this PR.

lefessan:
While trying to add another test case on conditions with numeric operators, I found a segfault in the compiler in `decimal_expand()`, hence another fix in this PR.

cobc/typeck.c
```diff
@@ -6280,14 +6295,14 @@ decimal_expand (cb_tree d, cb_tree x)
...
 	if (error_statement == current_statement) {
-		return;
+		return d;
 	}
 	switch (CB_TREE_TAG (x)) {
 	case CB_TAG_CONST:
 		/* LCOV_EXCL_START */
 		if (x != cb_zero) {
 			cobc_err_msg (_("unexpected constant expansion"));
-			COBC_ABORT ();
```
GitMensch, line 6290: That, as with (nearly?) any call of `cobc_err_msg()` should lead to an abort.
Did you got into this (the testcase don't contain an example)? If yes: what was the content of `x`?
If not: please revert that.

cobc/tree.c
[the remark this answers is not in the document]
GitMensch: resolved by disabling this check and drop the gettextize from the message that would otherwise result in a the new msgid; it does not look like we have a testcase for condition-name in expressions, so please add that to your test, then we can mark this as resolved

GitMensch (changes requested):
again "nearly finished", please review the review

cobc/typeck.c
```diff
@@ -6269,7 +6284,7 @@ decimal_compute (const int op, cb_tree x, cb_tree y)
 /**
  * expand tree x to the previously allocated decimal tree d
  */
-static void
+static cb_tree
```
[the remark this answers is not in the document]
lefessan: Actually, `decimal_expand` returns an error node if an error occurs, so it does not always return d/t. I think it is better to keep this approach than returning an integer that has to be tested on every call site to finally replace d/t by an error node at the caller.

cobc/typeck.c
lefessan: Indeed, no, I didn't get into this error, I just thought it was aborting because there was no way to return an error, but I can put back the previous behavior.

cobc/typeck.c
```diff
@@ -6354,21 +6369,29 @@ decimal_expand (cb_tree d, cb_tree x)
...
-		} else {
-			sz_shift = cb_int1;
+
+		switch (p->op){
+		case '=': case '~': case '<': case '>': case '[': case ']':
+		case '!': case '&': case '|':
+			cb_error_x (x, _("condition expression found where decimal expression was expected"));
+			return cb_error_node;
+		case 'c':
+		case 'd':
```
GitMensch: This is switching a `cb_binary_op_op` - it is likely more readable if not checking for the long names (and a lesser diff), I've just added a comment back and another in.

cobc/typeck.c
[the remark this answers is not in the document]
GitMensch: I see, there is now a single instance of this. If it is needed, then keep this (just mention the "returns x or cb_error_node" part to the function doc comment); I _think_ with this implementation we could get into the error case first, then get in to the CONST case then ABORT... if this is the case we'd need to add the check for d != error_node before the zero check.

I haven't checked - if we get away with just leaving d unchanged then this would be the easier solution and the lesser diff.

GitMensch:
Last thing to re-review is the new setting of cb_error_node in one case - I trust you make the right decision on how to tackle that - feel free to directly commit the result to svn.

cobc/typeck.c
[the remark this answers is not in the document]
lefessan: I updated the comment. Note that the modification I made in this function only applies in a case that had a completely undefined behavior, i.e. the code believed the expression was numeric when it was actually boolean. So, it is not clear what would happen by keeping d instead of the error node. To make the code a bit safer, I added a `error_statement = current_statement;` as it looks like some of the functions test these variables to avoid doing anything with the expressions.

GitMensch:
When everything passes - feel free to check the result in.

lefessan:
Merged in SVN, thanks for the reviews !

PR #90, opened 2023-02-21
Move syntax checks on literals to syn_literals

lefessan:
As discussed in #27


GitMensch (approved):
Thanks, that is good.

For SVN please don't create the file directly, instead use "svn copy syn_misc.at syn_literals.at" (you can then replace both with what you have here).

This way the svn history is correctly preserved.

GitMensch:
Done upstream https://sourceforge.net/p/gnucobol/code/4994. Thanks again for taking care!

PR #92, opened 2023-03-07
[CANCELLED] Fix for bug #831: COPY REPLACING in listing 

sbelondr:
Hi @GitMensch,

I propose the following fixes for the bugs :
- [#831](https://sourceforge.net/p/gnucobol/bugs/831/) (Problem for the copy replacing in listing mode) ;
- [#863](https://sourceforge.net/p/gnucobol/bugs/863/) (Variable is not referenced in report).

For the bug #863, I just add a loop to check if the variable exist (codegen.c). I added a test in run_reportwriter.at.

For the bug #831 I rewrite partialy the listing mode in the cobc.c file and I add 2 struct in tree.h for manage lines and replacements.

I added a test in the listings.at file (name : Multiple replacement) and I completed two tests (just add the expect.lst file) :
- COPY replacement with multiple partial matches ;
- COPY replacement with partial match.

But I didn't pass the huge REPLACE test because I add more space of the default implementation (8 precisely).

I would like to know if the implementation is right for you (I remove backtracking) ?


GitMensch:
Thanks for the work on this. Note that two PRs, one for each change are commonly more easy to review.
Let's start with the most easiest one: 863. Please adjust the check to be done in finalize_report, changing `Control field %s is not referenced in report` to a warning (the printf is wrong and codegen is too late in general).

cobc/cobc.c
```diff
@@ -6619,8 +6675,10 @@ print_line (struct list_files *cfile, char *line, int line_num, int in_copy)
 	struct list_skip	*skip;
 	int	do_print;
 	int	on_off;
+	char	pch;
```
GitMensch, line 6678: please leave local defintions in, we try to limit scope where possible

cobc/cobc.c
```diff
@@ -6689,10 +6747,14 @@ compare_prepare (char *cmp_line, char *pline[CB_READ_AHEAD],
 	int	in_string = 0;
 	int	last_col = cobc_get_text_column ();
 	int	last_nonspace;
+	const int	max_chars_on_line = cb_listing_wide ? 112 : 72;
 
 	cmp_line[0] = 0;
+	bzero (cmp_line, CB_LINE_LENGTH + 2);
```
GitMensch, line 6753: From its manpage:

> The bzero() function is deprecated (marked  as  LEGACY  in  POSIX.1-2001);  use  memset(3)  in  new  programs. POSIX.1-2008 removes the specification of bzero().

cobc/cobc.c
```diff
@@ -6838,526 +6856,329 @@ abort_if_too_many_continuation_lines (int pline_cnt, const char *filename, int l
...
 
-			/*
-			  Allocate a new out_line if we are on the last
-			  out_line.
-			*/
-			if (*out_line == *pline_cnt) {
-				make_new_continuation_line (cfile_name, pline,
-							    pline_cnt, line_num);
-			}
+	// calculation the len char
```
GitMensch, line 6915: Please use C89 comments here and in other places

```diff
-	// calculation the len char
+	/* calculation the len char */
```

cobc/cobc.c
```diff
@@ -7393,6 +7214,43 @@ cleanup_copybook_reference (struct list_files *cur)
 	cobc_free (cur);
 }
 
+static
+char *get_word (char *str)
```
GitMensch, line 7218: please move those two after `get_next_nonspace` to group those functions

cobc/cobc.c
```diff
@@ -6838,526 +6856,329 @@ abort_if_too_many_continuation_lines (int pline_cnt, const char *filename, int l
...
+	if is new string line replace don't skip the spaces
+*/
+static int
+strncpy_listing (char *dst, char *src, int n, int first, int in_copy, int is_new)
+{
+	int	i = 0;
+	const int	margin_a = cobc_get_margin_a (1);
+	int	cnt_space = 0;
+	int	j = 0;
+
```
GitMensch, line 6871: Looks like most of `DEBUG_REPLACE` is now out, was this explicit planned?
Did you use that to get an idea where the issue is?

GitMensch (changes requested):
Please split this PR into two ones, as mentioned I think the one for the report stuff # 863 should be quite easy to get through.

We can keep this PR for # 831; so far I did make some minor review comments and get a general overview (not there yet, need to combe back to this later...); I see that the issues we had - marked as AT_XFAIL_IF (true) are solved, as you've mentioned one of the older REPLACE tests now fail, mostly because the replacement is done by starting on the next line instead of the original one. Can you adjust that?

sbelondr:
Thank you for the feedback.
I'm working on it.

cobc/ChangeLog
```diff
@@ -190,6 +190,12 @@
 	* typeck.c (cb_build_move_literal): optimized output for literals to
 	  fields with BLANK WHEN ZERO and SIGN LEADING
 
+
```
GitMensch, line 193: meh, that was me (there's now an empty line too much)

cobc/tree.h
```diff
@@ -1656,6 +1656,17 @@ struct list_error {
 	char			*msg;		/* Error Message text */
 };
 
+typedef struct s_line_file {
```
GitMensch, line 1659: as this is _only_ used in cobc.c, it should be moved there

tests/ChangeLog
```diff
@@ -761,7 +767,7 @@
 	* testsuite.src/*.at: Added check for cobc's exit code and stderr
 		where missing
 
-2014-14-04  Philipp B�hme <phi.boehme@googlemail.com>
+2014-14-04  Philipp B�hme <phi.boehme@googlemail.com>
```
GitMensch, line 770: somethings broken here, maybe you can revert the line (otherwise I can do it after merging)

cobc/cobc.c
```diff
@@ -6533,16 +6481,103 @@ line_has_listing_statement (char *line, const enum cb_format source_format)
...
 
+static char *
+get_word (char *str)
+{
+	int i = 0;
+
+	if (!str) {
+		return NULL;
+	}
+	while (str[i] && isspace (str[i])) {
```
GitMensch, line 6492: What about separator commas and separator semicolons?
The old code used `isspace ((unsigned char)*bp) || *bp == ',' || *bp == '.' || *bp == ';'`.

cobc/cobc.c
```diff
@@ -6533,16 +6481,103 @@ line_has_listing_statement (char *line, const enum cb_format source_format)
...
+}
+
+static COB_INLINE COB_A_INLINE int
+is_comment_line (char *line, int fixed)
+{
+	if (line == NULL || line[0] == 0) {
+		return 0;
+	} else {
+		const int indicator = cobc_get_indicator ();
+		return fixed
```
GitMensch, line 6542: What happens here if the line contains less storage than the indicator points to - or can't that happen?

cobc/cobc.c
```diff
@@ -6533,16 +6481,103 @@ line_has_listing_statement (char *line, const enum cb_format source_format)
...
+			? line[indicator] == '*' || line[indicator] == '/'
+			: !strncmp (line, "*>", 2);
+	}
+}
+
+static void
```
GitMensch, line 6548: This should have already been `static COB_INLINE COB_A_INLINE void`, the new below should be, too - please adjust both.

cobc/cobc.c
```diff
@@ -6533,16 +6481,103 @@ line_has_listing_statement (char *line, const enum cb_format source_format)
...
 {
 	int		i;
-	int		len = strlen (line);
+	int		len;
 	const int	max_chars_on_line = cb_listing_wide ? 112 : 72;
 	const char	*format_str;
 	const int	indicator = cobc_get_indicator ();
 	const int	text_column = cobc_get_text_column ();
 
+	if (!line) {
```
GitMensch, line 6577: all callers have a guaranteed line, so that check should be dropped - please revert the change to `print_fixed_line` if that's correct

cobc/cobc.c
```diff
@@ -6565,6 +6600,145 @@ print_fixed_line (const int line_num, char pch, char *line)
 	}
 }
 
+static int
+is_quotes_char (char c, int quotes[2])
```
GitMensch, line 6604: please add a doc comment; I may be quite tired but I didn't understand what this is doing and how it sets the `quotes` around to 0/1

GitMensch (changes requested):
please check the first review parts and _consider_ handling the new `cb_partial_replace_when_literal_src` setting (otherwise thi can be included later)

please also fix the non-C89 definition - see the results of the new CI check (in detail here: https://github.com/OCamlPro/gnucobol/actions/runs/5392577360/jobs/9791122631#step:8:125)

I'll the review the result with a local checkout; the diff looks huge (and bigger as it is, while it is still big)

cobc/ChangeLog
sbelondr: Ok, I removed this line :)

cobc/tree.h
sbelondr: Ok, I placed that in cobc.c

tests/ChangeLog
sbelondr: I'll take care of that in the next commit

cobc/cobc.c
sbelondr: I also put that before but when I read the documentation I didn't understand why these characters is understood as a space and this not change the result of the test. I'll try other test for compare old replacing with the new replacing for check the differences

cobc/cobc.c
sbelondr: This cannot happen because the indicator can have a maximum size of 7 and the line has a size of 1024

cobc/cobc.c
sbelondr: Corrected but I don't understand when I should use this: COB_INLINE COB_A_INLINE ? Maybe when the function is often called ?

cobc/cobc.c
sbelondr: Yes, effectively it's useless. I removed that.

cobc/cobc.c
GitMensch: yes, when the function is often called and small

cobc/cobc.c
sbelondr: Indeed the implementation is not clean (quotes[0] = single quote, quotes[1] = double quote). I'll rework that for the next comit.

cobc/cobc.c
GitMensch: Note: it obviously would be good to have this in the testsuite, too.
In effect the point is that there are multiple "separators" - all "space characters", newline, separator semicolon and separator comma.
The last one is a bit tricky as, depending on `DECIMAL-POINT IS COMMA` this may or may not mean that "3,2" has to be separated (=2 "words"/tokens), while "3,a" would always mean two tokens. [note: the listing may "don't care" about that at all or always ignore the comma if `DECIMAL-POINT IS COMMA` is in effect].

cobc/cobc.c
GitMensch: ping @sbelondr

GitMensch:
@sbelondr: please adjust the testsuite changes - we moved from separate "expected.lst" to "list in stdout", so the changed expectation must be adjusted there (you currently have it just added as new AT_DATA)

cobc/cobc.c
```diff
@@ -6741,6 +6783,146 @@ print_fixed_line (const int line_num, char pch, char *line)
...
+	/* skip space */
+	while (isspace (line[i]))
+		++i;
+	start = i;
+	last_word = -1;
+
+	while (line[i])
+	{
+		/* loop space */
+		while (line[i] && isspace (line[i]) && !is_quotes_char (line[i], &simple_quote, &double_quote) && NOT_TOO_LONG (i, start)) {
```
GitMensch, line 6828: We have `isspace (line[i])´, so how should be ´(line[i])` be a quote?

cobc/cobc.c
```diff
@@ -6741,6 +6783,146 @@ print_fixed_line (const int line_num, char pch, char *line)
 	}
 }
 
+static int
```
GitMensch, line 6786: Please add a doc comment before the function so it is clear what is checks/changes.

Also: does the code handle an `"escaped "" quote like in x""4100"""`?

cobc/cobc.c
GitMensch: I _think_ those testcases are missing, right?

PR #88, opened 2023-02-19
[GCOS] Allow REPLACE after CONTROL DIVISION

lefessan:
`CONTROL DIVISION` in GCOS forbids `REPLACE` statements, but it is possible to have a `REPLACE` statement after the `CONTROL DIVISION` and before the `IDENTIFICATION DIVISION`.

PR #93, opened 2023-03-07
Add context to aliases in configuration

emilienlemaire:
While working on configurations with @nberth, we discoverd that rules like `reserved: ALIAS*=CONTEXTUAL-WORD` in configuration files, the context sensitivity of `CONTEXTUAL-WORD` was never reported to the aliases.
A small example with the following code:
```cobol
       IDENTIFICATION   DIVISION.
       PROGRAM-ID.      prog.
       DATA             DIVISION.
       WORKING-STORAGE  SECTION.
       01  BYTE-LENGTH  PIC 9.
       01  X            CONSTANT AS BYTE-LENGTH OF BYTE-LENGTH.
       PROCEDURE        DIVISION.
           MOVE X TO BYTE-LENGTH.
           DISPLAY BYTE-LENGTH NO ADVANCING
           END-DISPLAY.
           STOP RUN.
```
If we run the compiler with `reserved: XX*=BYTE-LENGTH` and replace all `BYTE-LENGTH` in the code with `XX` we get the following errors:
```sh
test.cob:5: error: syntax error, unexpected BYTE-LENGTH
test.cob:6: error: syntax error, unexpected BYTE-LENGTH
test.cob:6: error: constant item 'X' requires a VALUE clause
test.cob:8: error: syntax error, unexpected BYTE-LENGTH
test.cob:9: error: syntax error, unexpected BYTE-LENGTH, expecting (
```
when the file is expected to compile.
After the changes introduced in this PR we can compile with contextual aliases.

This PR also fixes a small error in `config/ibm.conf`

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,3 +1,8 @@
+2023-03-08  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
+
+	* reserved.c (get_user_specified_reserved_word): add check for
+	  context sensitivity in aliases
+
 
 2023-02-21  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/cobc/reserved.c
+++ b/cobc/reserved.c
@@ -4234,6 +4234,11 @@ get_user_specified_reserved_word (struct amendment_list user_reserved)
 		p = find_default_reserved_word (user_reserved.alias_for, 0);
 		if (p) {
 			cobc_reserved.token = p->token;
+            if (user_reserved.is_context_sensitive) {
+                cobc_reserved.context_sens =
+                    !!user_reserved.is_context_sensitive;
+                cobc_reserved.context_test = p->context_test;
+            }
 		} else {
 			/* FIXME: can we point to the fname originally defining the word? */
 			configuration_error (NULL, 0, 1,
--- a/tests/ChangeLog
+++ b/tests/ChangeLog
@@ -1,3 +1,7 @@
+2023-03-08  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
+
+	* testsuite.src/syn_misc.at: test context sensitive aliases
+
 
 2023-02-21  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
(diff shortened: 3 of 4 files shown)


config/ibm.words
```diff
@@ -357,7 +357,7 @@ reserved:	PLUS
 reserved:	POINTER
 #reserved:	POINTER-24       # potentially reserved
 #reserved:	POINTER-31       # potentially reserved
-reserved:	POINTER-32 X
+reserved:	POINTER-32
```
GitMensch, line 360: nice that you've spot the typo, the change was so small that I've directly checked that in (if I remember right the target branch will be updated tomorrow)

.gitignore
```diff
@@ -7,6 +7,7 @@
 
 .deps
 Makefile.in
+.cache
```
GitMensch, line 10: changes to this file are unrelated to any other PR and I think this overlaps with #76; please have a look there

GitMensch (changes requested):
The change itself looks good to me, please adjust the commit to remove the unrelated files (for example undo the commit, then revert/stash those) and add a test case (you already have presented one which is fine) to the testsuite (would go into syn_misc.at after the "aliases" test) and add an entry in cobc/Changelog.

For an example see https://github.com/OCamlPro/gnucobol/pull/88/files

nberth:
> ```cobol
>        01  X            CONSTANT AS BYTE-LENGTH OF BYTE-LENGTH.
> ```

@GitMensch Most documentations seem to make `BYTE-LENGTH` sensitive in "constant entry", which I *guess* in principle spans from the level (`01` above) to the terminating `.`. In practice, restricting "constant entry" to start right after `CONSTANT` makes sense, however I'm wondering whether the second `BYTE-LENGTH` above is actually legitimate.

GitMensch:
It is valid because the only thing that may be used there is a data-name, and therefore the data-names must be inspected.
Note that GnuCOBOL has (with the default std) uses context-sensitivity quite often and "with reduced scope", which enables it to compile a lot of different dialects by limiting their "special" reserved words to be tokenized as reserved word where it actually may be used as this - otherwise it is recognized as user-defined word (mostly data-name or label-name)

GitMensch:
@nberth : feel free to commit "upstream"

GitMensch (approved):

GitMensch:
Now upstream - thanks again.

PR #94, opened 2023-03-14
Variable is not referenced in report (SVN ticket : #863)

SBelondrade:
Hi Simon, I'm using my second account for this pull request.

Reference svn ticket : `#863`.

I replaced `printf` by `cb_warning_x` but I think I have some problems of understanding.

If I understand, I have to remove two checks and this condition in `codegen.c` :
```c
if((f->report_flag & COB_REPORT_LINE)
	&& f->children
	&& (f->children->report_flag & COB_REPORT_LINE)) {
		printf("Warning: Ignoring nested LINE %s %d\n",
			(f->report_flag & COB_REPORT_LINE_PLUS)?"PLUS":"",
			f->report_line);
		f->report_line = 0;
		f->report_flag &= ~COB_REPORT_LINE_PLUS;
		f->report_flag &= ~COB_REPORT_LINE;
	}
```

I added a new function and I called it `finalize_report` :

```c
static  void
check_report (struct  cb_report *r, cb_tree  ctl)
{
	if (!ctl) {
		return ;
	}
	
	cb_tree  nx = CB_CHAIN(ctl);
	
	if (nx) {
		check_report(r, nx);
	}
	int  bfound = 0;
	cb_tree  x = CB_VALUE (ctl);
	struct  cb_field *s = cb_code_field(x);
	if((s->report_flag & COB_REPORT_LINE)
	  && s->children
	  && (s->children->report_flag & COB_REPORT_LINE)) {
		cb_warning_x (COBC_WARN_FILLER,
		CB_TREE(s), _("Warning: Ignoring nested LINE %s  %d"),
		  (s->report_flag & COB_REPORT_LINE_PLUS)?"PLUS":"",
			s->report_line);
			s->report_line = 0;
			s->report_flag &= ~COB_REPORT_LINE_PLUS;
			s->report_flag &= ~COB_REPORT_LINE;
	}
	for(int  i = r->num_lines-1; i >= 0; i--) {
		if(r->line_ids[i]->report_control) {
			struct  cb_field *c = cb_code_field (r->line_ids[i]->report_control);
			if(c == s) {
				bfound = 1;
				break;
			}
		}
	}
	if (!bfound) {
		cb_warning_x (COBC_WARN_FILLER,
		CB_TREE(ctl), _("Control field %s is not referenced in report"),
			s->name);
		ctl = NULL;
	}
}
```

But I don't pass some unit test. Please tell me if I haven´t understood what I had to do ?


cobc/codegen.c
```diff
@@ -9681,6 +9681,25 @@ output_report_control (struct cb_report *p, int id, cb_tree ctl, cb_tree nx)
...
+			if(c == s) {
+				bfound = 1;
+				break;
+			}
+		}
+	}
+	if (!bfound) {
+		cb_warning_x (COBC_WARN_FILLER,
+			CB_TREE(ctl), _("Control field %s is not referenced in report"), s->name);
+		ctl = NULL;
```
lefessan, line 9698: `ctl = NULL;` is useless as `ctl` is a local variable.

cobc/codegen.c
```diff
@@ -9681,6 +9681,25 @@ output_report_control (struct cb_report *p, int id, cb_tree ctl, cb_tree nx)
...
+		p->controls = NULL;
+		return ;
```
lefessan, line 9700: This won't work if there is a list of more than one controls, and the first one is not used. If you only issue a warning and not an error, your code has to handle correctly such a case.

I would advise to clean up the controls at the beginning of the `output_report_definition` function, so that `p->controls` only contains controls that are used within the report.

lefessan (changes requested):

cobc/codegen.c
lefessan: Just to clarify: `output_control_report` is called on `nx` recursively just before your test, for other controls, so if you set `p->controls = NULL`, you discard all of them. The list of controls is used several times, so doing the test every time would be too complex, the only solution I see is to cleanup the list once at the beginning.

GitMensch (changes requested):
See my review comments for reasons that may be related to the test suite. Sorry for not paying that earlier (while I wrote those notes directly after you PR, I somehow missed to click on "submit review").

Furthermore please rebase.

cobc/codegen.c
```diff
@@ -9885,8 +9885,26 @@ output_report_control (struct cb_report *p, int id, cb_tree ctl, cb_tree nx)
...
+	for(i= p->num_lines-1; i >= 0; i--) {
+		if(p->line_ids[i]->report_control) {
+			struct cb_field *c = cb_code_field (p->line_ids[i]->report_control);
+			if(c == s) {
+				bfound = 1;
+				break;
+			}
+		}
+	}
+	if (!bfound) {
```
GitMensch, line 9901: This check, the warning and likely also the setting to null (or dropping the single unused control, as @lefessan pointed out) must be moved to typeck.c.

tests/testsuite.src/syn_reportwriter.at
```diff
@@ -689,3 +689,50 @@ prog.cob:37: warning: non-numeric PICTURE clause for SUM WSI-V
...
+               05   FILLER          PIC X(21).
+       88  E-TEST-2              VALUE " SM".
+           03   E-TEST-2-03.
+                   09   ERROR-2     PIC 9(3).
+])
+
+AT_CHECK([$COMPILE prog.cob], [0], [],
+[prog.cob: warning: control field ERROR-1 is not referenced in report
+prog.cob: warning: control field ERROR-2 is not referenced in report
+])
```
GitMensch, line 736: This ideally should be changed to

```
AT_CHECK([$COMPILE_ONLY prog.cob], [0], [],
[prog.cob: warning: control field ERROR-1 is not referenced in report
prog.cob: warning: control field ERROR-2 is not referenced in report
])
# additional check if we generate code that compiles, see bug #863
AT_CHECK([$COMPILE -w prog.cob], [0], [], [])
```

cobc/codegen.c
```diff
@@ -10229,7 +10246,8 @@ output_report_define_lines (int top, struct cb_field *f, struct cb_report *r)
 	if ((f->report_flag & COB_REPORT_LINE)
```
GitMensch, line 10246: (originating lines 10229-10235) This _should_ be moved to typeck.c, _at least_ the warning.

GitMensch (changes requested):
Rechecked: this issue still happens both in 3.2 release (and therefore current 3.x branch) and trunk. it would be nice if this PR can be finished so that we can include it upstream.

cobc/codegen.c
sbelondr: Ok. I moved this

TimVold:
This fix does not appear to have been merged to trunk, yet.
I have not "up-voted" the bug on SourceForge, yet, either.
However, I'm wondering if there is a work-around that may be applied to a user's COBOL code?

Thanks in advance for any insights and all your fabulous work on GnuCOBOL.

Cheers,
Tim

PR #95, opened 2023-03-22
New feature, EXTSM added

n-isaka:
Improvement of EXTFH and also add a similar function called EXTSM.
EXTSM is a programmable sorting process that exists in MicroFocus.

More details are discussed below.
https://sourceforge.net/p/gnucobol/discussion/cobol/thread/26d71dae9b/
https://github.com/opensourcecobol/gnucobol-osscons-patch

Please let me know if there are any problems with where to send pull requests, branches, etc.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,9 @@ NEWS - user visible changes				-*- outline -*-
    the error output for format errors (for example invalid indicator column)
    is now limitted to 5 per source file
 
+** support the COLLATING SEQUENCE clause on indexed files
+   (currently only with the BDB backend)
+
   more work in progress
 
 * Important Bugfixes
@@ -40,6 +43,9 @@ NEWS - user visible changes				-*- outline -*-
    a file for diagnostics; this flag can be activated if your editor and
    build system do not correctly work together to locate files from
    diagnostic output
+
+** New option -fdefault-file-colseq to specify the default
+   file collating sequence
 
 * More notable changes
 
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,17 @@
 
+2024-01-25  David Declerck <david.declerck@ocamlpro.com>
+
+	FR #459: support COLLATING SEQUENCE clause on SELECT / INDEXED files
+	* codegen.c (output_file_initialization): output the indexed
+	  file/keys collating sequence (were already present in the AST)
+	* tree.c (validate_indexed_key_field): process postponed
+	  key collating sequences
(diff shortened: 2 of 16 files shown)


tests/testsuite.src/run_misc.at
```diff
@@ -3617,6 +3617,111 @@ AT_CHECK([$COBCRUN_DIRECT ./prog2], [0], [], [])
...
+
+AT_CHECK([$COMPILE -std=mf prog.cob], [0], [], [])
+AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
+[AAA222
+BBB111
+CCC333
+], [])
+
+AT_CLEANUP
+
```
GitMensch, line 3724: This would be OK for an initial test. I do wonder if you have more EXTSM testcases, though.

GitMensch (changes requested):
The general approach is _very good_. I assume you've tested that already - because I don't have a test environment for that so far.
Confusing is only the change comment in the PR "Improvement of EXTFH" - as there isn't any change on it within the current commits; but actually I'd prefer to only keep this to EXTSM (where EXTFH changes aren't directly related).

There are some general requests concerning to possible refactoring to "combine" code parts from EXTFH and EXTSM, please see review comments.

From the docs it looks to me that the bit handling should be adjusted.

We'd need entries to libcob/Changelog and one entry for the NEWS.

What would be needed in any case allowing us to distribute those (=any) changes: a Copyright 
Assignment. See https://sourceforge.net/p/gnucobol/wiki/Copyright%20reassignment/ for some details.
Note: If all contributors that led to the code work for a single (or two) companies, then a copyright assignment by the company is often more reasonable than a copyright assignment of single contributors + the necessary copyright disclaimers from the employer(s).

We can either discuss the details here or via mail.

Reasonable _follow ups_ (=for another PR once this is included)
* addition of `MERGE` (could be integrated in this PR, too)
* addition of `CALLSORT` directive and `-fcallsort` option -> telling the compiler to build the FCD and call the new `EXTSM` (or other compatible sort functions) to do the sort/merge; the first is _very_ useful to test the EXTSM entry point (it also was quite useful to validate the EXTFH wrapper)

GitMensch:
@n-isaka Any progress on that? Note that there's now a link to this PR in [this StackOverflow question](https://stackoverflow.com/questions/75994932/error-call-to-undeclared-function-extsm-iso-c99-and-later-do-not-support-imp/75998219#75998219) from someone that needs `EXTSM`.

PR #96, opened 2023-04-20
Fix codegen bug with FD GLOBAL and nested programs

lefessan:
The program added in the testsuite generates a C error when compiled with earlier versions, for a missing label:

```
$ cobc -F test2.cob
test2.c: In function ‘NESTED__PROGRAM__1__2_0_’:
test2.c:799:5: error: label ‘l_1’ used but not defined
  799 |     goto l_1;
      |     ^~~~
```

The reason is that `has_global_file` is incorrectly unset in nested sub-programs.

This PR fixes this problem by only unsetting it depending on the level at which it was set.


tests/testsuite.src/syn_file.at
```diff
@@ -2282,3 +2282,74 @@ AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
...
+ DATA DIVISION.
+ WORKING-STORAGE SECTION.
+ PROCEDURE DIVISION.
+   GOBACK.
+ END PROGRAM NESTED-PROGRAM-2.
+ 
+ END PROGRAM NESTED-FILES.
+])
+AT_CHECK([$COBC -F -C prog.cob], [0], [], [])
+AT_CHECK([grep 'Paragraph Default Error Handler' prog.c], [0], [  /* Line: 0         : Paragraph Default Error Handler   : prog.cob */
```
GitMensch, line 2350: please a dd a line break before the last `[`

cobc/codegen.c
```diff
@@ -13302,11 +13302,20 @@ codegen (struct cb_program *prog, const char *translate_name)
 			break;
 		}
 		subsequent_call = 1;
-		if (current_program->flag_file_global
-		 && current_program->next_program->nested_level) {
-			has_global_file = 1;
+		/* set has_global_file if needed, only for sub-programs of this program */
+		if ( current_program->flag_file_global
+		     && current_program->next_program->nested_level >
```
GitMensch, line 13307: somehow the indent got mixed / lost; this would be the "GC default":

```c
		/* set has_global_file if needed, only for sub-programs of this program */
		if (current_program->flag_file_global
		 && current_program->next_program->nested_level >
		    current_program->nested_level) {
			if (!has_global_file) {
				has_global_file = 1 ;
				has_global_file_level = current_program->nested_level ;
			}
```

tests/testsuite.src/syn_file.at
```diff
@@ -2282,3 +2282,74 @@ AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
...
+  /* Line: 0         : Paragraph Default Error Handler   : prog.cob */
+  /* Line: 0         : Paragraph Default Error Handler   : prog.cob */
+], [])
+AT_CLEANUP
```
GitMensch, line 2354: Please also add a `$COMPILE` - Somehow I think it is likely best to move that run_file.at (would still have failed the `$COMPILE`) and verify with an error in the most nested entry that the default handler is triggered.
And if we have that test we wouldn't need the syn test any more.

tests/testsuite.src/syn_file.at
```diff
@@ -2282,3 +2282,74 @@ AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
...
 ])
 AT_CLEANUP
+
+# Verify that FD GLOBAL will create default handlers in sub-programs, but
+# only in its own sub-programs
+
+AT_SETUP([Scope of FD GLOBAL in nested programs])
+AT_KEYWORDS([FD GLOBAL nested])
+AT_DATA([prog.cob],[
+ IDENTIFICATION DIVISION.
```
GitMensch, line 2292: Could you try to stick with fixed-form if there's no explicit need for free-form reference format?
The main reason is that it allow to quickly (or at least much more quick) manually cross-test the source with other compilers.

tests/testsuite.src/syn_file.at
```diff
@@ -2282,3 +2282,74 @@ AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
 prog.cob:12: error: syntax error, unexpected INPUT, expecting Identifier
 ])
 AT_CLEANUP
+
+# Verify that FD GLOBAL will create default handlers in sub-programs, but
+# only in its own sub-programs
+
+AT_SETUP([Scope of FD GLOBAL in nested programs])
+AT_KEYWORDS([FD GLOBAL nested])
```
GitMensch, line 2290: all of those are already in the title (automatically in the keywords) - so I suggest to use `file` instead

tests/testsuite.src/syn_file.at
GitMensch: ...
and please "duplicate" it for a `DECLARATIVES` in the outer program, to verify not only the default handler but the file handler in any case can be executed.

GitMensch (changes requested):
Thanks for inspecting this - I think the codegen approach is reasonable and well done.
I've thought "the outer program should handle that" - but for the default handler this is not necessary.

If it already works with user-specified error-handling (a `DECLARATIVE` in the outer program) then we _may_ use the same approach for the default handler; if it doesn't then we should adjust the codegen for both cases (may need a special [=negative] entry point).

tests/testsuite.src/syn_file.at
lefessan: This code is automatically generated by `autofonce`, so I will wait for the tests to be well written before changing the formatting.

cobc/codegen.c
lefessan: should be ok now

tests/testsuite.src/syn_file.at
lefessan: I added a `DECLARATIVES`, but I am not sure of what the expected behavior should be.

Right now, it looks like the handlers are only generated where they are declared, and not in subprograms. Should they be duplicated in all sub-program inheriting from the files ?

Also, the bug only appears on `WRITE` instructions, and I don't really know how to generate an error easily on a `WRITE` that would trigger the handler execution at runtime.

tests/testsuite.src/syn_file.at
lefessan: ok, they are in fixed form now.

tests/testsuite.src/syn_file.at
lefessan: ok, done

GitMensch:
@lefessan Could you please add the necessary tests for the following (and rebase)?

> if it already works with user-specified error-handling (a DECLARATIVE in the outer program) then we may use the same approach for the default handler; if it doesn't then we should adjust the codegen for both cases (may need a special [=negative] entry point).

lefessan:
So, actually, the bug is not triggered if `USE GLOBAL` declaratives are provided in the outer program, because another flag is used (`prog->flag_gen_error` instead of `has_global_file`), and the outer declaratives are correctly called from the inner program.

lefessan:
Is it ready for SVN ?

tests/testsuite.src/run_file.at
```diff
@@ -13029,3 +13029,90 @@ Z9
...
+         NESTED-PROGRAM .
+         WRITE FILE-EXT-REC FROM "Hello".
+         GOBACK.
+       END PROGRAM NESTED-PROGRAM-1-1.
+
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. NESTED-PROGRAM-1-2.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 FILE-EXT-DAT PIC X(347).
```
GitMensch, line 13095: Seems unused, so maybe drop the complete `DATA DIVISION` in `NESTED-PROGRAM-1-2`, `NESTED-PROGRAM-1-1` and `NESTED-PROGRAM-2`?
Do you want to do anything with ``NESTED-PROGRAM-2`? If not then please drop it.

GitMensch (changes requested):
> So, actually, the bug is not triggered if `USE GLOBAL` declaratives are provided in the outer program, because another flag is used (`prog->flag_gen_error` instead of `has_global_file`), and the outer declaratives are correctly called from the inner program.

Can you please add a testcase for that?

And... shouldn't the change then also be to (additionally ?) adjust `prog->flag_gen_error` so that both programs use the same approach and both use the "outer" label, instead of generating an "inner" one (_guess_ this is done with the change as-is)?

tests/testsuite.src/run_file.at
lefessan: I think it's better to keep this second program, so that the test may detect more errors in the future.

GitMensch (changes requested):
likely only testcase adjustments needed

GitMensch (changes requested):
The change looks fine, but the result seems to be wrong?
See review of the testsuite entry.

tests/testsuite.src/run_file.at
```diff
@@ -13210,3 +13210,155 @@ Z9
...
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. NESTED-PROGRAM-1.
+       ENVIRONMENT DIVISION.
+       INPUT-OUTPUT SECTION.
+       FILE-CONTROL.
+         SELECT FILE-EXT ASSIGN "./TEST-FILE" FILE STATUS IS WSFS.
+       DATA DIVISION.
+       FILE SECTION.
+       FD FILE-EXT GLOBAL
+                 RECORD IS VARYING IN SIZE  CHARACTERS.
```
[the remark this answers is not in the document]
lefessan: ok, removed.

tests/testsuite.src/run_file.at
```diff
@@ -13210,3 +13210,155 @@ Z9
...
+          DISPLAY "ERROR ON FILE-EXT".
+       P02 SECTION.
+          USE AFTER ERROR PROCEDURE ON OUTPUT.
+       P0201.
+          DISPLAY "ERROR ON OUTPUT".
+       END DECLARATIVES.
+       MAIN SECTION.
+         OPEN OUTPUT FILE-EXT.
+         CALL    "NESTED-PROGRAM-1-1".
+      * This should trigger an error in NESTED-PROGRAM-1-1
```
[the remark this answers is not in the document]
lefessan: My understanding is that the DECLARATIVES is called, and then the code returns to the point where it was called. The program should only abort if an abort is asked in the DECLARATIVES, no ?

tests/testsuite.src/run_file.at
```diff
@@ -13210,3 +13210,155 @@ Z9
...
+         WRITE FILE-EXT-REC FROM "Hello".
+         GOBACK.
+       END PROGRAM NESTED-PROGRAM-1-1.
+
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. NESTED-PROGRAM-1-2.
+       PROCEDURE DIVISION.
+         NESTED-PROGRAM .
+         WRITE FILE-EXT-REC FROM "Hello".
+         GOBACK.
```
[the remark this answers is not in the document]
lefessan: I added DISPLAYs before and after each write to follow the execution.

lefessan:
@GitMensch It looks like the CI fails on Macosx on test 25, which is related to the diagnostics with caret. Did you test your changes on Macosx ?

GitMensch:
It should abort if no USE section for the file's EXCEPTION is defined.

GitMensch:
> @GitMensch It looks like the CI fails on Macosx on test 25, which is related to the diagnostics with caret. Did you test your changes on Macosx ?

Nope, I'm not sure how well the MacOSX CI works at all... Could you recheck this, please?

tests/testsuite.src/run_file.at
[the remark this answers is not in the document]
GitMensch: This comment says there should be an error, but that seems to not be correct?

GitMensch (approved):
Other than the misleading comment in prog.cob - fine for svn.

GitMensch:
ping (approved)

PR #97, opened 2023-06-09
CI: check for c89 declaration

GitMensch:
to prevent this to be caught late, as happened with #85


GitMensch:
Note, that works locally (the only reason that it currently doesn't fail is that the target branch is not up-to-date with the changes from #85).

PR #99, opened 2023-06-20
Add dialect option subscript-check to relax bound checks for ODO items (FR #437)

nberth:
Only ODO items for now, which actually makes this PR quite short.
I'm not sure what is to be relaxed about reference modifications.
If that stays like this, `-fcheck-record-bound` may not be the clearest of names; maybe `-frelax-odo-bound-checks` would suit better.

https://sourceforge.net/p/gnucobol/feature-requests/437/

### DIFF
--- a/cobc/typeck.c
+++ b/cobc/typeck.c
@@ -2521,7 +2521,9 @@ cb_build_identifier (cb_tree x, const int subchk)
 		   --> either cache field name here (dropping after each statement)
 		       or remove/skip later during codegen
 		   */
-		if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT) && f->odo_level != 0) {
+		if (cb_subscript_check == CB_SUB_CHECK_FULL
+		 && CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT)
+		 && f->odo_level != 0) {
 			for (p = f; p; p = p->children) {
 				if (CB_VALID_TREE (p->depending)
 				    /* Do not check length for implicit access to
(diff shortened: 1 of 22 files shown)


cobc/flag.def
```diff
@@ -170,6 +170,9 @@ CB_FLAG (cb_flag_section_exit_check, 1, "section-exit-check",
 CB_FLAG (cb_flag_implicit_goback_check, 1, "implicit-goback-check",
 	_("  -fimplicit-goback-check\tcheck that code execution does not end implicit at end of PROCEDURE DIVISION"))
 
+CB_FLAG (cb_flag_check_record_bound, 1, "check-record-bound",
+	_("  -fcheck-record-bound  check ODO item access against field founder instead of dynamic size"))
```
GitMensch, line 174: that description would be a different check (and _may_ be the right one for GCOS, as it is [per docs] the one that is used on IBM) - you are welcome to implement that (see adjusted specs in the upstream FR)

GitMensch (changes requested):
I've retested with MF and the ODO item (and the minimum item) are never checked there either, even if bounds are checked (which is the default) or all checks are enabled.

I therefore see that this _should_ be a dialect option.
Please adjust the PR according to the updated specs in the [upstream FR](https://sourceforge.net/p/gnucobol/feature-requests/437/).

cobc/typeck.c
```diff
@@ -2570,12 +2572,13 @@ cb_build_identifier (cb_tree x, const int subchk)
...
 				if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT)) {
-					if (p->depending && p->depending != cb_error_node) {
+					if (cb_subscript_check != CB_SUB_CHECK_MAX
+					 && p->depending && p->depending != cb_error_node) {
 						e1 = CB_BUILD_FUNCALL_4 ("cob_check_subscript",
-							 cb_build_cast_int (sub),
-							 cb_build_cast_int (p->depending),
-							 CB_BUILD_STRING0 (name),
-							 cb_int1);
+									 cb_build_cast_int (sub),
```
GitMensch, line 2578: please revert this whitespace change (enough to do during commit into svn)

GitMensch (approved):
there were some "beauty" issues which I've checked in quickly; you may want to review those, then: feel free to commit the branch's diff to svn

nberth:
Ok thanks for all the hints and specs.
Now upstream

PR #101, opened 2023-06-21
Actually obey `-fdefault-colseq`

nberth:
… which currently seems ignored for alphanumeric comparisons others than in `SORT`.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,13 @@
 
+2023-06-22  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* parser.y (setup_program, setup_default_collation)
+	  (program_init_without_program_id): properly record default collating
+	  sequence if required via -fdefault-colseq
+	* tree.c (cb_build_binary_op), typeck.c (cb_build_cond_default): inhibit
+	  -ffast-compare and -fconstant-folding for programs with a non-native
+	  collating sequence
+
 2023-06-20  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
 
 	FR #437: dialect option to limit bound checks for ODO to maximum
--- a/cobc/parser.y
+++ b/cobc/parser.y
@@ -272,7 +272,6 @@ static cb_tree			xml_encoding;
 static int			with_xml_dec;
 static int			with_attrs;
 
-static cb_tree			default_collation;
 static cb_tree			alphanumeric_collation;
 static cb_tree			national_collation;
 
@@ -887,6 +886,24 @@ check_relaxed_syntax (const cob_flags_t lev)
 }
 
 static void
+setup_default_collation (struct cb_program *program) {
+	switch (cb_default_colseq) {
+#ifdef COB_EBCDIC_MACHINE
+	case CB_COLSEQ_ASCII:
+#else
+	case CB_COLSEQ_EBCDIC:
+#endif
+		alphanumeric_collation = build_colseq (cb_default_colseq);
+		break;
+	default:
+		alphanumeric_collation = NULL;
+	}
+	national_collation = NULL; /* TODO: default national collation */
+	program->collating_sequence = alphanumeric_collation;
+	program->collating_sequence_n = national_collation;
+}
+
+static void
 program_init_without_program_id (void)
 {
 	cb_tree		l;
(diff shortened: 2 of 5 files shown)


nberth:
Here's a bit of context on this.

While investigating a weirdly functioning `SEARCH ALL`, I actually observed that the resulting search algorithm did not operate according to any user-given collating sequences if (and only if) that sequence was specified using `-fdefault-colseq`. Other means of specifying the collating sequence (e.g, via the `ENVIRONMENT DIVISION`) seemed to make the search work properly.

Investigating further, it seems constant-folding does not operate according to the proper collating sequence. And it may also be the case that fast-comparisons (`-ffast-compare`) are also impacted.

A seemingly "easy" way to mitigate the last points would be to check on the status of (`current_`)`prog->collating_sequence` when performing constant folding and/or optimizing comparisons.

GitMensch:
Ideally both points would do the check via libcob functions and passing the collation.

nberth:
> Ideally both points would do the check via libcob functions and passing the collation.

Yes that would be ideal indeed. However, I don't think I'll have the time to implement that new approach right now (but I'd happily give it a try). May I suggest pushing the current PR upstream first, as this is primarily a bugfix?

(In the meantime, I've also added a test for `SEARCH ALL`.)

cobc/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2023-06-22  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* parser.c (setup_program): properly record default collating sequence
```
GitMensch, line 4: you've adjusted parser.y ;-)

cobc/parser.y
```diff
@@ -1369,6 +1369,16 @@ setup_program (cb_tree id, cb_tree as_literal, const enum cob_module_type type,
 
 	/* Initalize default COLLATING SEQUENCE */
 	default_collation = build_colseq (cb_default_colseq);
+	switch (cb_default_colseq) {
```
GitMensch, line 1372: something is wrong here, as this is later adjusted in `object_computer_sequence`;
from a quick glance you want to adjust the default of `alphanumeric_collation` and `national_collation`

cobc/tree.c
```diff
@@ -6174,8 +6174,11 @@ cb_build_binary_op (cb_tree x, const enum cb_binary_op_op op, cb_tree y)
 		/*
 		 * If this is an operation between two literal strings
 		 * then resolve the value here at compile time -> "constant folding"
+		 *
+		 * TODO: build cob_fields and call cob_cmp from libcob.
 		 */
 		} else if (cb_constant_folding
+		 && current_program->collating_sequence == NULL
```
GitMensch, line 6181: it seems that the check for = and ~ below can still be done - and in the other cases the literal's class should likely be checked against national in which case `collating_sequence_n` must be compared

GitMensch (changes requested):
the general idea and the test cases are fine; I'm also OK with "relay to libcob" - but it seems that both the wrong fields are adjusted (don't change current_program directly but its source fields) and checked alphanumeric (collating sequence for national literals)

please rework this (and _possibly_ keep [NOT] EQUALS also if a collation is specified)

cobc/parser.y
GitMensch: I'd suggest to init both `alphanumeric_collation` and `national_collation` in `program_init_without_program_id` and `setup_program`, then later use it directly.

cobc/tree.c
GitMensch: The new version of this code is fine!

cobc/parser.y
```diff
@@ -886,6 +885,24 @@ check_relaxed_syntax (const cob_flags_t lev)
...
 
+static void
+setup_default_collation (struct cb_program *program) {
+	switch (cb_default_colseq) {
+#ifdef COB_EBCDIC_MACHINE
+	case CB_COLSEQ_ASCII:
+#else
+	case CB_COLSEQ_EBCDIC:
+#endif
+		alphanumeric_collation = national_collation =
```
GitMensch, line 896: Setting the national collation to the alphanumeric collation is likely wrong- but that was the case in some of the removed code paths too.
You may always set it to `NULL` for now and leave a TODO marker, or leave as is (hm a TODO/CHECKME marker would be good in any case)

GitMensch (approved):
LGTM, _consider_ the minor change to parser and feel free to commit.

PR #102, opened 2023-06-21
Fix buffer overflow causing CI on macos to fail

lefessan:


lefessan:
The bug was triggered by the longgy.cob test that was added for caret, containing a very long identitier.

GitMensch:
Thanks for checking.

Interestingly running the test with memcheck
`./pre-inst-env valgrind --vgdb-error=1 cobc/.libs/cobc  -fdiagnostics-plain-output -Wno-others -fdiagnostics-show-caret longgy.cob`
does not raise an error (it normally does in this case)

Looking at this function (actually its callers) I think it does too much and not enough.
Can you please adjust it as follows?

* signature changed to `static size_t get_word (const char *str, char *word_buff, size_t buff_len)`
   * using buff_len instead of len
   * returning the word length (or zero)
* callers:
   * only have enough space for the the max tested length of the word later on
   * adjust to new signature
   * check with length + memcmp instead of strcmp
   * test `cb_comment_paragraphs != CB_IGNORE` before calling `next_word_is_comment_paragraph_name`, not within
* do Changelog

This leads to a cleaner implementation, minimal speedup and the memory fix.

GitMensch (changes requested):
see comment - and again: thanks for checking!

cobc/pplex.l
```diff
@@ -1936,42 +1936,38 @@ is_cobol_word_char (const char c)
...
-	return ret;
+	return length;
 }
 
 static int
 is_condition_directive_clause (const char *buff)
 {
-	char word_buff[COB_MAX_WORDLEN + 1];
-	char *p;
+	char p[COB_MAX_WORDLEN + 1];
```
GitMensch, line 1963: while keeping `word_buff` (actually `buff` is enough) it only needs to be of size 14 (`END-EVALUATE` is the longest one) and down in the call `sizeof(buff)` can be used then

[I guess this good commit was mostly for checking the CI result, so the compare was not changed to `len == 3 && memcmp (buff, "IF", 3)`]

lefessan: New commit adds this.

GitMensch (approved):
Ideally you could finish the calling, too:

> test `cb_comment_paragraphs != CB_IGNORE` before calling `next_word_is_comment_paragraph_name`, not within

but in general, this is good to go

GitMensch:
I've just ran the result in Appveyor MacOS (took a while to fix the old brew installations).
And guess what: this new test is the only one that is failing... after the changes.

[build log](https://ci.appveyor.com/project/GitMensch/gnucobol-3-x-macos/build/job/wiwrhyj786htw9hl)
[testsuite.log](https://ci.appveyor.com/api/buildjobs/wiwrhyj786htw9hl/artifacts/testsuite.log)

but that's because of the sed invocation, which does not work:

```
25. used_binaries.at:923: testing cobc diagnostics show caret ...
./used_binaries.at:942: $COBC -Wall -fsyntax-only prog.cob
./used_binaries.at:947: $COBC -fdiagnostics-show-caret -fdiagnostics-show-line-numbers prog.cob
./used_binaries.at:967: $COBC -fdiagnostics-show-caret prog.cob
./used_binaries.at:988: sed -e 's/DIVISION\./DIVISION  \.\t  \t  /' prog.cob > progsp.cob
./used_binaries.at:990: $COBC -Wno-others -fdiagnostics-show-caret progsp.cob
--- -	2023-06-28 09:08:42.000000000 -0500
+++ /Users/appveyor/projects/gnucobol-3-x-macos/tests/testsuite.dir/at-groups/25/stderr	2023-06-28 09:08:42.000000000 -0500
@@ -2,6 +2,11 @@
           WORKING-STORAGE  SECTION.
           01 TEST-VAR PIC 9(2) VALUE 'A'.
  >        COPY 'CRUD.CPY'.
-          PROCEDURE        DIVISION  .
+          PROCEDURE        DIVISION  .t  t
               DISPLAY TEST-VAR NO ADVANCING
+progsp.cob:2: error: syntax error, unexpected Identifier, expecting FUNCTION-ID or PROGRAM-ID
+   
+ >        IDENTIFICATION   DIVISION  .t  t
+          PROGRAM-ID.      prog.
+          DATA             DIVISION  .t  t
 
25. used_binaries.at:923: 25. cobc diagnostics show caret (used_binaries.at:923): FAILED (used_binaries.at:990)
```

:-/

... in any case it should be `$SED`, not `sed`.

GitMensch:
The easiest option is likely to only run the test once and in the testsuite source already have those tab + spaces included, followed by `#`, which is then not `$SED` but `tr -d` removed...

PR #104, opened 2023-07-01
bin/cob-config: fix missing arguments and order of echos

lefessan:
* Some arguments are printed in the help, but not implemented
* If multiple arguments are given, echo values in the same order as arguments


GitMensch (approved):
Good to go.

Please finish the other PR as I should publish RC3 start of next week.

PR #107, opened 2023-07-09
fix code generation for INPUT/CLOSE with double arg

lefessan:
The bug was that DECLARATIVES for both arguments were called when the operation failed for only one of them.

The reason was that the parser would generate OPEN or CLOSE statements that were one inside the other one, instead of a list of independant statements.


GitMensch (changes requested):
needs some answers to the questions raised, a comment in some places (instead of "likely not needed") why the implicit statement is necessary and a minor re-indent

if provided then could be merged in tomorrow as it looks good otherwise

Can you please add `DELETE FILE` to the testcase?

cobc/parser.y
```diff
@@ -12900,15 +12900,17 @@ close_body:
 close_files:
   file_name _close_option
   {
-#if 0 /* CHECKME: likely not needed */
-	begin_implicit_statement ();
-#endif
-	cb_emit_close ($1, $2);
+	  struct cb_statement * saved_current_statement = current_statement ;
+	  begin_implicit_statement ();
```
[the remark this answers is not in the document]
lefessan: I removed the modification of `delete_file_list`. The reason is that the problem appears with `DECLARATIVES`, and it looks like `DELETE FILE` does not trigger them in case of error (in the C code generated, there is no test after `cob_delete_file()` to check for an exception condition). So, we could use the same approach to construct the tree, but it would not change the generated code at the end.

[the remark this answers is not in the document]
GitMensch: I've rechecked the standard. While it is not completely clear what to do for `DELETE RECORD`, it is clear for `DELETE FILE` - we _should_ go into `DECLARATIVES` for any exception.

Can you please fix this along in this PR?

cobc/parser.y
```diff
@@ -12900,15 +12900,21 @@ close_body:
...
+	   and restore it. */
+	struct cb_statement * saved_current_statement = current_statement ;
 	begin_implicit_statement ();
-#endif
 	cb_emit_close ($1, $2);
+	current_statement = saved_current_statement ;
   }
 | close_files file_name _close_option
   {
+	struct cb_statement * saved_current_statement = current_statement ;
```
GitMensch, line 12914: Where do we need the reset of the current_statement?

GitMensch: OK, I see it now - the issue is that `begin_implicit_statement()` both nests the new statement to the `current_statement` and then overwrites that. The approach you've used is reasonable, while not ideal.

Please adjust the signature of the implicit statement as follows:
`begin_implicit_statement (cb_tree real_statement, cb_tree pos)`

The callers then either pass the `saved_current_statement` into it (for lists like in `open_file_entry`) or restore the `current_statement` (like in `close_files`, where we don't have a list).
They also pass "whatever reference creates the implied statement" (in the case above: `file_name`), so the implicit statement of `file-2` and `file-3` in

```cobol
CLOSE file-1
     file-2
     file-3
```

ill point to the correct source file, line and column.

lefessan: I see what you mean for the location `cb_tree pos`, but not for `real_statement`. `begin_implicit_statement` needs to modify `current_statement` at the end, so that the next function will work on it (`cb_emit_open`, `cb_emit_close`, etc.), so it is not possible to restore `current_statement` inside `begin_implicit_statement` (unless we get rid of `current_statement` and pass the statement to all the functions `cb_emit*`, but that would be a big refactoring).

GitMensch: `begin_implicit_statement` will use the "original" statement only for attaching the implicit one:

```diff
-	current_statement->body = cb_list_add (current_statement->body,
+	real_statement->body = cb_list_add (real_statement->body,
 					    CB_TREE (new_statement));
```

This should work fine, no?

lefessan: The problem I see is not there, it's the fact that `begin_implicit_statement` needs to modify `current_statement` (because `cb_emit*` functions work on `current_statement`). For example, if I modify the code with the new prototype:

```
| close_files file_name _close_option
  {
	begin_implicit_statement (current_statement,$2);
	cb_emit_close ($2, $3);
  }
```
would not work, because for the next filename, `current_statement` is not the toplevel `CLOSE` anymore, but the one generated by `begin_implicit_statement`, that will be passed to the new call of `begin_implicit_statement`, so that the new statement will be nested inside the previous one.

GitMensch: I'm not sure if you like the result (I find it a bit more "clean").

cobc/parser.y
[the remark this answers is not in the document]
GitMensch: This one is the "open" part, as it is only "related", you may commit to svn as-is, but as it _should_ work it would be reasonable to at least do the same change to delete_file that you did to the others.

GitMensch (approved):
Please at least update delete_file as you did for the others (if you find the time to actually let it go to DECLARATIVE and adding it to the test that would be nice, but if you can't make that today then better commit without the DECLARATIVE part for DELETE FILE).

The CI shows one new warning (from your last commit) and one warn-as-error (that would not compile on some systems), please fix those when checking that in to SVN.

~~~
../../cobc/cobc.c: In function ‘cobc_plex_strsub’:
../../cobc/cobc.c:1346:17: warning: variable ‘n’ set but not used [-Wunused-but-set-variable]
 1346 |         int     n;
      |                 ^
../../cobc/parser.y: In function ‘yyparse’:
../../cobc/parser.y:15354:9: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
15354 |         struct cb_statement * top_statement = current_statement ;
      |         ^~~~~~
~~~

lefessan:
I have done the changes, except for enabling `DECLARATIVES` for `DELETE FILE`. I added the line (in `typeck.c`), and then commented it for later. The reason is that it breaks several tests in the testsuite (they expect `DELETE FILE` not to raise any error when the file is missing), and it's likely that it will break existing programs as well. One solution would be to add a config option for this specific feature, disabled by default.

PR #108, opened 2023-07-14
Add --fdump-tree=FILE option

lefessan:
Dump the AST/tree to a file, for debugging purpose.

A file dump_ast_gen.c is generated by an external tool directly from tree.h, it can then be edited manually if the types are modified, until the tool is run again later to clean it up.


GitMensch:
Would it be reasonable to use -fdump-something similar to GCC https://gcc.gnu.org/onlinedocs/gcc/Developer-Options.html ?
Note: I don't have any idea if there are options that output JSON.

Question: What is the OCAML ml and where can I read more about it?
What would this output be used for?

In any case please provide test cases with sample output (and to be sure that this doesn't break anything execute make checkall at least once passing this option in COBOL_FLAGS.

cobc/output_tree.c
```diff
@@ -0,0 +1,2122 @@
...
+{
+	output_constr_string ( "cob_statement", string_of_cob_statement (x));
+}
+
+
+static
+const char* string_of_cb_usage (enum cb_usage x)
+{
+	switch (x){
+	case CB_USAGE_BINARY: return "CB_USAGE_BINARY";		/* 0 */
```
GitMensch, line 602: Couldn't those and similar be "stringified" using our existing macros? This would still need a switch-case, but if the source is a macro then it can be just defined as additional macro instead of the full switch-case, otherwise a macro can still be used to only end in `CASE_STRING(some-enum)`.

Note: you don't need to include the numeric values (which possibly be changed later on in any case).

lefessan:
I changed the options to `-fdump-tree` and `-fdump-tree-flags`, it should be close enough `gcc` devs expectations.

I ran all the testsuite with both formats (using env variables to not modify scripts), and I could validate both OCaml and JSON files. 

I am not sure that adding a test in the testsuite would be a good idea: (1) the output depends on the internal types in `tree.h`, and (2) even for small examples, the output is pretty big, a few thousands lines...

The OCaml format is actually standard syntax for basic types in the OCaml language (https://ocaml.org). Since I mostly always program in OCaml (except for GnuCOBOL :-) ), it is easier for me to have this format than JSON, as I could easily write tools to simplify/filter information in the AST.

For examples, JSON format (with flags `+Ai`):
```
{
   "type_": "cb_program",
   "address_": " 0x559ec2ee75f8",
   "common": {
      "tag": { "type_": "cb_tag", "value_": "PROGRAM" },
      "source_file": "test.cob"
   },
   "program_name": "DOUBLE-OPEN",
   "program_id": "DOUBLE__OPEN",
   "orig_program_id": "DOUBLE-OPEN",
   "word_table": { "type_": "TODO", "value_": "struct cb_word **" },
   "local_include": { "type_": "TODO", "value_": "struct local_filename*" },
   "entry_list": [
      {
         "type_": "cb_list",
         "address_": " 0x559ec2f45138",
         "purpose": {
            "type_": "cb_label",
            "address_": " 0x559ec2f44f78",
            "common": {
               "tag": { "type_": "cb_tag", "value_": "LABEL" },
               "source_file": "test.cob",
               "source_line": 39
            },
            "name": "DOUBLE__OPEN",
            "orig_name": "DOUBLE-OPEN",
            "xref": { "type_": "TODO", "value_": "struct cb_xref" },
...
```

whereas OCaml format looks like:
```
RECORD [
   "type_", STRING "cb_program";
   "address_", POINTER 0x55acb46095f8L;
   "common", RECORD [
      "tag", CONSTR ("cb_tag", STRING "PROGRAM");
      "source_file", STRING "test.cob";
   ];
   "program_name", STRING "DOUBLE-OPEN";
   "program_id", STRING "DOUBLE__OPEN";
   "orig_program_id", STRING "DOUBLE-OPEN";
   "word_table", CONSTR ("TODO", STRING "struct cb_word **");
   "local_include", CONSTR ("TODO", STRING "struct local_filename*");
   "entry_list", LIST [
      RECORD [
         "type_", STRING "cb_list";
         "address_", POINTER 0x55acb4667138L;
         "purpose", RECORD [
            "type_", STRING "cb_label";
            "address_", POINTER 0x55acb4666f78L;
            "common", RECORD [
               "tag", CONSTR ("cb_tag", STRING "LABEL");
               "source_file", STRING "test.cob";
               "source_line", INT( 39);
            ];
            "name", STRING "DOUBLE__OPEN";
            "orig_name", STRING "DOUBLE-OPEN";
            "xref", CONSTR ("TODO", STRING "struct cb_xref");
...
```

lefessan:
Note that there are still some TODOs in the code. There are many fields in these data structures, and it's not clear what should be printed or not.

Also, it might be interesting at some point to have "levels" of importance for fields, so that, for example, only important fields would be printed at level 1, then less important fields at level 2, and so on... However, such levels should probably be decided after experimenting with using these outputs for real debugging.

GitMensch (changes requested):
I _guess_ the build_literal changes are part of a broken merge - you'd like to revert them then (at least with an extra cleanup commit)

To ensure that at least _something_ is tested you definitely want to add some testcases, possibly in used_binaries.at

cobc/ChangeLog.branch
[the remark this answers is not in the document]
lefessan: I just put the ChangeLog of the branch outside of the ChangeLog, because it creates a conflict everytime I rebased it. I will keep it that way until the work is finished.

cobc/dump_tree_gen.c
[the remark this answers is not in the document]
lefessan: As it is generated by a "non standard" tool, I am still wondering what to do. Adding a permanent dependency on the tool is probably a bad idea, I am more considering making this feature a configure-time option, disabled by default, and only available on-demand (with a specific workflow in the Github CI to check whether it is still working or not).

GitMensch:
That's strange, if you keep it inserted "down where it is" until there are possibly more changes, then there should be no conflict as no other commits as to the changelog down.

lefessan:
> That's strange, if you keep it inserted "down where it is" until there are possibly more changes, then there should be no conflict as no other commits as to the changelog down.

I see. For me, it has to be inserted at the point where it is merged, i.e. the ChangeLog before should correspond to what is in the code at the point where the work is done, and not when it was started, no ?

GitMensch:
That depends. If most of the old work is unchanged and "old", then I personally would have one "old" and one "new" entry.
If it is totally new like this I'd likely move the Changelog entry up - but only directly before the actual merge (so no conflict in between).

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+		if (!strcmp(filename, "stderr")){
+			close_fd = 0;
+			fd = stderr;
+		} else {
+			int len = strlen (filename);
+			struct stat st;
+			if (len > 0 && filename[len-1] == '/'){
+				filename = cobc_main_stradd_dup (filename, prog->program_id);
+			}
+			if (!stat (filename, &st)) unlink(filename);
```
GitMensch, line 551: Why the unlink? fopen as specified will overwrite this, no?
And if using `cobc prog1.cob prog2.cop -fdump-file=progs.dump` we'll have only one element included that way, no?

Just as a question: would the dump for multiple files in one be bad? If not allowing starting the filename with a "+" (adding that to the mode) would be good.

In any case: please add the check for the unix_lf (binary mode) here to not break the upcoming tests under Windows.
What about one source with multiple programs (contained and or not contained)?

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+			fd = fopen (filename, "w");
+		}
+	if (!fd){
+		cb_perror (0, "cobc: %s: %s", filename, cb_get_strerror ());
+	}
```
GitMensch, line 556: this should likely go after the `fopen` and if raised should `return` (currently would SIGSEGV)

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+void set_flags (const char* flags)
+{
+	int sign = 1;
+	int i;
+	int len = strlen(flags);
+
+	for(i=0; i< len; i++){
+		switch(flags[i]){
+		case '+': sign = 1; break;
+		case '-': sign = 0; break;
```
GitMensch, line 480: these flags duplicate the filename option and should therefore be dropped here and in the flag.def help

cobc/flag.def
```diff
@@ -282,3 +282,12 @@ CB_FLAG (cb_diagnostics_absolute_paths, 1, "diagnostics-absolute-paths",
...
 	_("  -fprof                enable profiling of the COBOL program"))
 
+CB_FLAG_NQ (1, "dump-tree", CB_FLAG_DUMP_TREE,
+	_("  -fdump-tree=<file>    dump parsed AST to <file>, format is JSON unless\n"
+	  "                        the file has a .ml extension for OCaml format."))
+
+CB_FLAG_NQ (1, "dump-tree-flags", CB_FLAG_DUMP_TREE_FLAGS,
+	_("  -fdump-tree-flags=<flags>    set flags used when dumping the parsed AST\n"
+	  "                        '+/-' to enable/disable, 'l' for locations,\n"
+	  "                        't' for types, 'p' for pointers, 'i' for indent."))
```
GitMensch, line 292: this misses some of the flags

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+
+	fprintf (fd, "%s", fmt.format_header);
+	cb_print_tree ( (cb_tree) prog);
+	fprintf (fd, "%s", fmt.format_trailer);
+
+	if (close_fd) {
+		fclose (fd);
+		if (cb_dump_tree_print_message){
+			fprintf (stderr, "File \"%s\" generated\n", filename);
+		}
```
GitMensch, line 566: should be output up-front depending on verbose mode (see cobc.c)

cobc/tree.h
GitMensch, line 1: all changes to this file but the definition of the new function can be checked in upstream at any time

GitMensch (changes requested):
a bunch of issues from the last iteration solved and cleaned up correctly - looking forward for the test

cobc/Makefile.am
```diff
@@ -67,3 +67,10 @@ endif
 
 .l.c:
 	LC_CTYPE=C $(LEX) -o$@ $<
+
+dump_tree.$(OBJEXT): $(top_srcdir)/cobc/dump_tree_gen.c
+
```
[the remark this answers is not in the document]
lefessan: removed

cobc/cobc.c
```diff
@@ -8303,6 +8320,11 @@ process_translate (struct filename *fn)
 		}
 	}
 
+	if (!dump_tree_to_file)
+		dump_tree_to_file = getenv("COB_DUMP_TREE");
```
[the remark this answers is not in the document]
lefessan: removed

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+{
+	char *env;
+	int close_fd = 1;
+
+	set_format_by_file_ext (filename);
+	env = getenv ("COB_DUMP_TREE_FLAGS");
+	if (env) set_flags (env);
+	if (flags) set_flags (flags);
+
+	if (!strcmp(filename, COB_DASH) || !strcmp(filename, "stdout")){
```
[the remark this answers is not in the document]
lefessan: I kept only the dash, no stdout and no stderr

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+		close_fd = 0;
+		fd = stdout;
+	} else
+		if (!strcmp(filename, "stderr")){
```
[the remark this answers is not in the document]
lefessan: removed

cobc/dump_tree.c
```diff
@@ -0,0 +1,568 @@
...
+			close_fd = 0;
+			fd = stderr;
+		} else {
+			int len = strlen (filename);
+			struct stat st;
+			if (len > 0 && filename[len-1] == '/'){
+				filename = cobc_main_stradd_dup (filename, prog->program_id);
```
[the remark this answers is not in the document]
lefessan: removed

cobc/dump_tree.c
lefessan: I added the return

cobc/dump_tree.c
lefessan: The goal is to be able to enable/disable flags in the same definition, for example `+z-i` (add zero-value fields and disable indent)

cobc/dump_tree.c
lefessan: removed

GitMensch:
Last set of errors from the CI:

~~~
In file included from ../../cobc/dump_ast.c:190:
../../cobc/dump_ast_gen.c: In function 'init_ptr_table':
../../cobc/dump_ast_gen.c:39:5: error: implicit declaration of function 'bzero' [-Wimplicit-function-declaration]
   39 |     bzero (ptr_table, ptr_table_size * sizeof(void*) );
mv -f .deps/replace.Tpo .deps/replace.Po
      |     ^~~~~
../../cobc/dump_ast_gen.c:39:5: warning: incompatible implicit declaration of built-in function 'bzero' [-Wbuiltin-declaration-mismatch]
../../cobc/dump_ast_gen.c: In function 'print_struct_cb_funcall':
../../cobc/dump_ast_gen.c:7826:40: warning: the comparison will always evaluate as 'true' for the address of 'argv' will never be NULL [-Waddress]
 7826 |     if( print_zero_fields || ptr->argv != NULL ){
      |                                        ^~
In file included from ../../cobc/dump_ast.c:37:
../../cobc/tree.h:1288:33: note: 'argv' declared here
 1288 |         cb_tree                 argv[CB_BUILD_FUNCALL_MAX];     /* Function arguments */
      |                                 ^~~~
../../cobc/dump_ast_gen.c: In function 'print_struct_cb_intrinsic':
../../cobc/dump_ast_gen.c:8420:58: warning: passing argument 2 of 'print_struct_cb_intrinsic_table_ptr' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
 8420 |        print_struct_cb_intrinsic_table_ptr (indent+2, ptr->intr_tab);
      |                                                       ~~~^~~~~~~~~~
../../cobc/dump_ast_gen.c:7995:89: note: expected 'struct cb_intrinsic_table *' but argument is of type 'const struct cb_intrinsic_table *'
 7995 | static void print_struct_cb_intrinsic_table_ptr (int indent, struct cb_intrinsic_table *ptr){
      |                                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
../../cobc/dump_ast.c: In function 'cb_dump_ast_to_file':
../../cobc/dump_ast.c:240:21: warning: unused variable 'len' [-Wunused-variable]
  240 |                 int len = strlen (filename);
      |                     ^~~
../../cobc/dump_ast_gen.c: At top level:
../../cobc/dump_ast_gen.c:2509:13: warning: 'print_struct_cob_prof_procedure_ptr' defined but not used [-Wunused-function]
 2509 | static void print_struct_cob_prof_procedure_ptr (int indent, struct cob_prof_procedure *ptr){
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../cobc/dump_ast_gen.c:2090:13: warning: 'print_struct_cb_call_xref_ptr' defined but not used [-Wunused-function]
 2090 | static void print_struct_cb_call_xref_ptr (int indent, struct cb_call_xref *ptr){
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../cobc/dump_ast_gen.c:2036:13: warning: 'print_struct_handler_struct_ptr' defined but not used [-Wunused-function]
 2036 | static void print_struct_handler_struct_ptr (int indent, struct handler_struct *ptr){
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../cobc/dump_ast_gen.c:1258:13: warning: 'print_struct_cb_xref_ptr' defined but not used [-Wunused-function]
 1258 | static void print_struct_cb_xref_ptr (int indent, struct cb_xref *ptr){
      |             ^~~~~~~~~~~~~~~~~~~~~~~~
../../cobc/dump_ast.c:162:13: warning: 'print_int_array' defined but not used [-Wunused-function]
  162 | static void print_int_array (int indent, int n, int *x)
      |             ^~~~~~~~~~~~~~~
make[3]: *** [Makefile:639: dump_ast.o] Error 1
~~~

The MSVC builds all fail because the new dump_ast files need to be added to build_windows.

cobc/flag.def
GitMensch: Do you consider the "no JSON/OCaml" temporary or is the idea to leave that as-is? Just asking....

lefessan: Initially, I thought that the AST of GnuCOBOL was "clean", in the sense that it was an exact representation of the COBOL code. However, it is not the case: the parser starts translating code to C (for example, replacing some nodes by direct calls to C functions), so the AST is actually half COBOL code, half translated code. So, the interest of sending this AST to another program (through OCaml or JSON) is less interesting. So, for now, I have given up on the idea of a correct JSON/OCaml encoding, focusing more on the ability to debug the compiler using a printed AST.

Yet, on the long term, I think it would be nice to move translations done in the parser to the codegen, and get a clean AST. At that point, it could be used by other programs and this output could be improved.

build_windows/vs2013/cobc.vcxproj.filters
```diff
@@ -92,6 +95,9 @@
     <None Include="..\..\cobc\warning.def">
       <Filter>Header Files</Filter>
     </None>
+    <None Include="..\..\cobc\dump_ast_gen.c">
+      <Filter>Header Files</Filter>
+    </None>
```
GitMensch, line 100: I'm confused - that is an auto-generated source (like parser.c) and therefore should be moved there and `ClCompile`d as well, no [that applies to all of those files under build_windows]?

cobc/dump_ast_gen.c
GitMensch, line 1: Why should we handle that as a header file? I _guess_ we call static functions in here directly from dump_ast.gen.c?
If this is the case and there are not too much, then dropping the `static` attribute via `sed` and adding those "entry prototypes" to tree.h would work better.

cobc/flag.def
GitMensch: I totally agree that this separation should be clearer; to do so we'd likely move most of the included C code to a new cb_tree like it currently is implemented, but into a new type of node that is only converted to C in codegen.
... but then - nearly all of the C parts should be FUNC_CALLs and most of those should be quite easily be separated and could be converted "on the fly", if wanted.

... but your human-readable AST dump may be ideal to show what is really in and consider adjustments that way.
I'm looking forward to some test cases showing what you consider to be bad in the AST. :-)

cobc/dump_tree.c
GitMensch: This one is open (why the unlink, what happens with multiple programs in the command line, addition of optional "+" and the unix-lf...)

cobc/dump_ast_gen.c
```diff
@@ -36,8 +36,8 @@ static void init_ptr_table ()
     ptr_table = (void**) calloc (ptr_table_size, sizeof (void*) );
     num_table = (int*)   calloc (ptr_table_size, sizeof (int) );
   } else {
-    bzero (ptr_table, ptr_table_size * sizeof(void*) );
-    bzero (num_table, ptr_table_size * sizeof(int) );
+    memset (ptr_table, 0, ptr_table_size * sizeof(void*) );
+    memset (num_table, 0, ptr_table_size * sizeof(int) );
```
GitMensch, line 40: Is this a manual change after generation? In this case you possibly want to define bzero as macro instead to do that change.

lefessan: The new file is generated by a new version of the tool (https://github.com/OCamlPro/ctypes-gen), but I guess most devs will edit it by hand, and the tool will only be used from time to time to get a clean version.

build_windows/vs2013/cobc.vcxproj.filters
lefessan: I don't really understand these files, I just grep for other similar files (warning.def and typeck.c) and modified the files accordinly...

cobc/flag.def
lefessan: Here is an example of the generated output for a small file:

[writefile.txt](https://github.com/user-attachments/files/17522224/writefile.txt)

I am not really sure it is interesting to create a test, as we don't really expect the format to be kept backward compatible. We may just want to make a test to verify that a file is generated, but not check the file format itself.

cobc/dump_ast_gen.c
lefessan: Maybe it should just be renamed as .def to be consistent with other included files that are not actually include files.

lefessan:
I close this PR for now, we have a working version in our private repository, that can be upstreamed if needed.

GitMensch:
If you have a working version that lives in a gnucobol tree: please reopen/push here or create a new item.

As you are aware, an upstream change can be "extra work" but has the benefit that you won't need to merge them in again into a newer version (and if that is of any use, also have SuperBOL expecting the feature).

lefessan:
The thing is, the code of this printer is completely dependent on the format of the tree, it has to be updated every time a new field is changed in the tree structure. Even in our internal codebase, it is usually out-of-date and updated only when we (I) need it... and most of the code is generated by a small tool written only for that ( https://github.com/OCamlPro/ctypes-gen ).

So, I don't feel very comfortable merging it in the official version...

GitMensch:
Ah, I see. Then it sounds reasonable to not include it.

PR #109, opened 2023-08-15
add dependencies options and -fcopybook-deps 

lefessan:
* Implements dependencies options similar to gcc
* `-fcopybook-deps` outputs only copybook names instead of file paths. `-fcopybook-deps` also forces `-E -foneline-deps -MT=copybooks`, disables errors on missing copybooks and remove output on stdout.


GitMensch:
As testcases are missing I cannot get a grasp what this is about....

We definitely should provide a way to "only" calculate the dependencies and in this case the paths are not available so we can only output the names, which _may_ be the idea behind that.

Not sure why we should have a "only on one line".

Adjustments to -MT and friends are useful and those were added back with an explicit note that there will be changes - but those should mimic what GCC does with them so please provide sample code and output of a C program with `#include <something.h>` and `#include <else/something.h>`  and `#include "other/something.h"`  along with some different -I and -M... options and add test cases that have the same scenario for cobc - making the later mimic the first is the way to go. Only when we have that we should inspect your scenario to check if that's what you need or what is missing.

And yes: if you could tackle the -M... options, that would be quite good (but those changes should mimic recent GCC, both from their options and from their results).

cobc/pplex.l
```diff
@@ -1629,10 +1633,21 @@ ppcopy (const char *name, const char *lib, struct cb_replace_list *replace_list)
 		}
 		/* otherwise fall-trough to error handling */
 	} else {
-		/* ensure to have errno from name as specified, not from another file */
-		(void)access (plexbuff1, R_OK);
-		/* pass file error as we have no more places to check */
-		cb_error ("%s: %s", plexbuff1, cb_get_strerror ());
+		/* do not output error if looking only for copybook deps */
+		if (!cb_flag_copybook_deps){
```
GitMensch, line 1637: How would one prevent the error of a not found include with GCC when generating the dependency file?

lefessan:
I added a test in `used_binaries.at`... and most arguments used by `gcc` that make sense for GnuCOBOL. I don't know which changes are being done in `gcc`, but I can import them accordingly.

For `-fcopybook-deps`, the idea is to output dependencies for other tools than `make`, in a very simple format and without file path resolution. For example, using this output, it is possible to create a tool that will use it to know the copybooks of a source file, and then search for them by scanning directories, and finally call `cobc` again with an automatically generated list of `-I <dir>`.

cobc/pplex.l
lefessan: I don't understand your question. You mean that we should add in the dependencies the dependencies that only appear in the generated C files ? FWIW, the `-fcopybook-deps` is a specific mode where we don't really care about missing COPY files, but other modes with `-M` will still complain about them missing.

GitMensch: The question was: is there an option to let GCC generate a dependency file without complaining? If yes we should provide it the same way.

lefessan: I think it's `-MG`:       -MG In conjunction with an option such as -M requesting dependency
           generation, -MG assumes missing header files are generated files
           and adds them to the dependency list without raising an error.  The
           dependency filename is taken directly from the "#include" directive
           without prepending any path.  -MG also suppresses preprocessed
           output, as a missing header file renders this useless.

My implementation of -MG still complains about the error, but adds the file to the list of dependencies. I can modify my PR to remove the error in this case, but it will not be possible to suppress the output as `cobc` will probably have already started generating it when the error occurs.

GitMensch: When this is in, then I'm quite sure we can suppress the error, too.

lefessan: Just pushed a version that does not complain with `-MG` about the error (see the last test in `used_binaries.at`)

lefessan:
@GitMensch I have a completely unrelated question ; I have a PR on the manual in https://github.com/ocamlpro/gnucobol-docs . Is there another maintainer for that part, or are you also in charge ?

tests/testsuite.src/used_binaries.at
```diff
@@ -1002,3 +1002,125 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+ sub/COPY3.CPY
+
+])
+
+AT_CHECK([$COMPILE_ONLY -M -fcopybook-deps prog.cob], [0],
+[prog.o: COPY1 COPY2 sub/COPY3
+
+])
+
+AT_CHECK([$COMPILE_ONLY -M -MT prog.so prog.cob], [0],
```
GitMensch, line 1063: please duplicate this one with multiple `-MT` options (like in the example above, which should be moved here and then doesn't need to use multiple `-MT` any more [it should primarily test copybooks with paths])

tests/testsuite.src/used_binaries.at
```diff
@@ -1002,3 +1002,125 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+ sub/COPY3.CPY
+
+])
+
+AT_CHECK([rm -f prog.d])
+
+AT_CHECK([$COMPILE -MD  prog.cob])
+
+AT_CHECK([cat prog.d], [0],
+[prog.o: \
```
GitMensch, line 1086: that test would fail when COB_OBJECT_EXT != o

tests/testsuite.src/used_binaries.at
```diff
@@ -1002,3 +1002,125 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+ prog.cob \
+ COPY1.CPY \
+ COPY2.CPY \
+ sub/COPY3.CPY
+
+])
+
+AT_CHECK([rm -f prog.d])
+
+AT_CHECK([$COMPILE -MD  prog.cob])
```
GitMensch, line 1083: we generate a file here, so it is reasonable to add `AT_CAPTURE` for that before;
please drop the double-space after `-MD` when adjusting the test file

tests/testsuite.src/used_binaries.at
```diff
@@ -1002,3 +1002,125 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+
+AT_CHECK([cat prog.d], [0],
+[prog.o: \
+ prog.cob \
+ COPY1.CPY \
+ COPY2.CPY \
+ sub/COPY3.CPY
+
+])
+
```
GitMensch, line 1093: `-MD` is for "creating it along" - so please check that the target executable is created here

tests/testsuite.src/used_binaries.at
```diff
@@ -1002,3 +1002,125 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+AT_CHECK([$COMPILE_ONLY -M -MT prog.so prog.cob], [0],
+[prog.so: \
+ prog.cob \
+ COPY1.CPY \
+ COPY2.CPY \
+ sub/COPY3.CPY
+
+])
+
+AT_CHECK([$COMPILE_ONLY -M -MQ '$(target)#toto' prog.cob], [0],
```
GitMensch, line 1072: `-M` is for "don't create output, but dependency list", so please use `$COMPILE` and check that the target executable is not generated.
Duplicate the test to verify, that also when adding `-g` no prog.c is generated.

tests/testsuite.src/used_binaries.at
```diff
@@ -271,7 +271,7 @@ AT_CHECK([$GREP 'PARAGRAPH_EX_l_7:'    prog.c], [0], ignore, [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [bluBb END], [])
 
 AT_CAPTURE_FILE([prog.d])
```
GitMensch, line 273: As the purpose of this test is (primarily) "check that cobc implies the right options depending on the output file specified" (and the output file _after your change **corectly**_ isn't even generated because of `-M`) and we now have a dependency generation test: please move this series of checks down to the new test.

tests/testsuite.src/used_binaries.at
```diff
@@ -1002,3 +1002,125 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+AT_CHECK([$COMPILE_ONLY -M -MF prog.dep prog.cob])
+
+AT_CHECK([cat prog.dep], [0],
+[prog.o: \
+ prog.cob \
+ COPY1.CPY \
+ COPY2.CPY \
+ sub/COPY3.CPY
+
+])
```
GitMensch, line 1056: Won't these tests currently fail on WIN32 because of the slash?

GitMensch (changes requested):
Should `-MM` and `-MMD` be added, ignoring the copybooks in `COB_COPY_DIR` (currently extfh and screenio)?

Shouldn't the dependency options be included in gnucobol.texi, possibly copying GCCs docs verbatim? This would also allow to note in NEWS that we adjusted these options (referencing the 3.2 note about the future change) while pointing to the manual for the details on the dependency generation.

In any case this should be split to two commits for SVN - one that adjusts the "standard" dependency options and a second one for the new flags.
I'd like to review them separately, can you please split the PR into two (obviously the new flags would not base on the original target but on the deps branch)?

tests/testsuite.src/used_binaries.at
lefessan: There are already several tests in the testsuite checking that multiple  `-MT` options correctly append all of them.

tests/testsuite.src/used_binaries.at
lefessan: Should I add the `AT_CAPTURE` before or after the `rm` ? There are multiple `prog.d` files generated in the sequence of tests, should I capture all of them ?

GitMensch: Before. the `AT_CAPTURE` stays active till the end of the test, so testsuite.log shows the entry that actually produced the fail - which is exactly what we want

tests/testsuite.src/used_binaries.at
GitMensch: this was about moving the multiple `-MT` from the test above here.

GitMensch:
> Should `-MM` and `-MMD` be added, ignoring the copybooks in `COB_COPY_DIR` (currently extfh and screenio)?
> 
> Shouldn't the dependency options be included in gnucobol.texi, possibly copying GCCs docs verbatim? This would also allow to note in NEWS that we adjusted these options (referencing the 3.2 note about the future change) while pointing to the manual for the details on the dependency generation.
> 
> In any case this should be split to two commits for SVN - one that adjusts the "standard" dependency options and a second one for the new flags. I'd like to review them separately, can you please split the PR into two (obviously the new flags would not base on the original target but on the deps branch)?

Friendly ping @lefessan

cobc/ChangeLog
```diff
@@ -1,4 +1,20 @@
...
+
+2023-08-16  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c, flags.def: add new flags to output dependencies:
+	  -foneline-deps outputs all dependencies on a single line instead of
+	  multiple lines; -fcopybook-deps outputs only copybook names instead
+	  of file paths. -fcopybook-deps also forces -E, -foneline-deps,
+	  -MT=copybooks, disables errors on missing copybooks and removes
+	  output on stdout.
+
```
GitMensch, line 17: if you rewrote the changes please updated the dates, if you made significant adjustments then add a single "new" entry with the new date on top, otherwise it is fine to leave dates as is, but in any case: the entries should be located in the file according to their dates.

GitMensch:
Thanks for the rebase. As guessed before, noted in the review and now seen in the CI: we have issues at least on Win32 as follows:

```
#                             -*- compilation -*-
28. used_binaries.at:1119: testing output dependencies ...
../../tests/used_binaries.at:1138: mkdir -p sub
../../tests/used_binaries.at:1144: $COMPILE_ONLY prog.cob
../../tests/used_binaries.at:1146: $COMPILE_ONLY -M prog.cob prog.cob
--- -	2024-03-12 14:25:51.094447400 +0000
+++ /d/a/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/28/stdout	2024-03-12 14:25:51.042599800 +0000
@@ -2,12 +2,12 @@
  prog.cob \
  COPY1.CPY \
  COPY2.CPY \
- sub/COPY3.CPY
+ sub\COPY3.CPY
 
 prog.o: \
  prog.cob \
  COPY1.CPY \
  COPY2.CPY \
- sub/COPY3.CPY
+ sub\COPY3.CPY
```

I _think_ all make implementations support / paths, in this case we _may_ use these here; otherwise the test may get a conditional to change the expected result for Win32 to backslash or we go with adjusting the output of (.)\\(.) to $1/$2.

lefessan:
So, in this version, I am using the `COB_IS_RUNNING_IN_TESTMODE` variable to replace backslashes by slashes in the compiler output, so that tests have the same results on Unix and Windows.

tests/testsuite.src/used_binaries.at
```diff
@@ -396,13 +372,28 @@ AT_DATA([prog.cob], [
...
+AT_CHECK([TMPDIR="" TMP="" TEMP="./prog.cob" $COMPILE prog.cob 2> compiler.output], [0], [],
+[],[
+    AT_CHECK([grep libcob: compiler.output], [0], [libcob: warning: Temporary directory TEMP is invalid, adjust TMPDIR!
+])
+  ],
+  [
+    AT_CHECK([cat compiler.output], [0], [libcob: warning: Temporary directory TEMP is invalid, adjust TMPDIR!
 ])
+  ])
+
```
GitMensch, line 396: that's an ... interesting approach, but please move that out to a separate PR

GitMensch (changes requested):
Nearly finished! Please see comments.

cobc/error.c
```diff
@@ -63,11 +63,42 @@ size_t				cb_msg_style;
...
 DECLNORET static void		cobc_too_many_errors (void) COB_A_NORETURN;
 
+static int is_test = -1;
+
+/* Returns a copy of the argument with backslashes replaced by slashes
+in filenames, only if COB_IS_RUNNING_IN_TESTMODE. This mode simplifies
+the test of outputs, as they will be similar on Unix and Windows.
+*/
+char *
+cobc_slashify (const char *src)
```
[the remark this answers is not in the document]
lefessan: Most strings are `const char*`, so it is not possible to mutate them...

tests/testsuite.src/used_binaries.at
```diff
@@ -1123,19 +1117,197 @@ AT_CHECK([$COBC -I . -m --include "file.h" -fstatic-call prog2.cob], [0], [], []
...
+AT_CHECK([$GREP 'prog.c prog.h '       prog.d], [0], [ignore])
+AT_CHECK([$GREP ' prog.i:'             prog.d], [0], [ignore])
+AT_CHECK([$GREP 'sub/copy/PROC.cpy'    prog.d], [0], [ignore], [], [
+  # Previous test "failed" --> no entry with slash available, check backslash for this and following tests
+  AT_CHECK([$GREP ' sub\\copy\\PROC.cpy \\'  prog.d], [0], [ignore])
+  AT_CHECK([$GREP ' sub\\PROCE.cpy'          prog.d], [0], [ignore])
+  AT_CHECK([$GREP 'sub\\copy\\PROC.cpy:'     prog.d], [0], [ignore])
+  AT_CHECK([$GREP 'sub\\PROCE.cpy:'          prog.d], [0], [ignore])
+], [
+  AT_CHECK([$GREP ' sub/copy/PROC.cpy \\'    prog.d], [0], [ignore])
```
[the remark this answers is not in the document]
lefessan: Indeed, it's probably not useful anymore, but without a Windows computer, it's painful to modify all these tests and then wait for Github CI for feedback...

tests/testsuite.src/used_binaries.at
```diff
@@ -1102,14 +1097,13 @@ AT_CHECK([$COBC -m -fstatic-call prog.cob], [0], [], [])
...
 AT_CHECK([$COBC -m --include "$PWD/file.h" -fstatic-call prog.cob], [1], [], [ignore])
+
 AT_DATA([prog2.cob], [
        IDENTIFICATION   DIVISION.
        PROGRAM-ID.      prog.
        DATA DIVISION.
-       WORKING-STORAGE SECTION.
-       01 long USAGE BINARY-C-LONG.
        PROCEDURE        DIVISION.
-           CALL "f" USING "Hello" BY VALUE long RETURNING NOTHING.
```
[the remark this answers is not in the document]
lefessan: So, under Windows, the function has to be available. `f` was not a real function, so it was failing with an "undefined function 'f'". I solved it by calling an existing function (`rename` is part of `stdio.h`, which is not included by default in `common.h`). My understanding was that the test was mostly testing the arity of the function, not the exact types, so I chose a function with two `char*` because I couldn't find an existing function with the same prototype as former `f`.

tests/testsuite.src/used_binaries.at
lefessan: The problem with moving that elsewhere is that the test will keep failing under Windows, which means that I might make changes in this PR and not notice that the failures under Windows have changed. I would prefer to have a working CI under Windows and find another fix in another PR later...

tests/testsuite.src/used_binaries.at
```diff
@@ -1072,12 +1063,16 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
-HOME/prog.cob:14: warning: ignoring redundant .
-])
+AT_CHECK([echo "$PWD/prog.cob:7: error: CRUD.CPY: No such file or directory" > expected.output])
+AT_CHECK([echo "$PWD/prog.cob:6: warning: numeric value is expected" >> expected.output])
+AT_CHECK([echo "$PWD/prog.cob:14: warning: ignoring redundant ." >> expected.output])
+AT_CAPTURE_FILE([expected.output])
+
+AT_CHECK([[cat compiler.output | tr '[:upper:]' '[:lower:]' | tr -d ':/\\' > compiler.output2]])
+AT_CHECK([[cat expected.output | tr '[:upper:]' '[:lower:]' | tr -d ':/\\' > expected.output2]])
+AT_CHECK([diff compiler.output2 expected.output2])
```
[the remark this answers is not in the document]
lefessan: I am not familar with HEREDOC notations. This version is not the most elegant one, but it does the job :-)

lefessan:
> > Should `-MM` and `-MMD` be added, ignoring the copybooks in `COB_COPY_DIR` (currently extfh and screenio)?

I don't think it is worth it. Though they are easy to implement (by modifying the `cb_copy_find_file` in `pplex.l` to set a flag when using the `COB_COPY_DIR` entry of `cb_include_list`), I don't see any reason why the external tool that would use the dependencies would need to remove system dependencies, or wouldn't be able to do it by itself. Adding low value features makes the code harder to maintain for little benefit.

lefessan:
This version has a `NEWS` section and a new section in `gnucobol.texi`  for dependency options.

tests/testsuite.src/used_binaries.at
lefessan: Should work now even with .obj, using $SED on compiler output...

lefessan:
@GitMensch I think I have replied and/or fixed all your points, is it for merge ?

tests/testsuite.src/used_binaries.at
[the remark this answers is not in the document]
GitMensch: This is all unrelated to the actual PR changes, please move this out to a separate PR for later review.

doc/gnucobol.texi
GitMensch, line 1: That's well done. Please add a ChangeLog entry for it.

cobc/error.c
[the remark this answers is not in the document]
GitMensch: I see. But `cob_slashify()` and its related testsuite changes are far from the original PR. Please move that out to a separate one, then revert the related parts in this one.

.github/workflows/windows.yml
```diff
@@ -79,15 +79,16 @@ jobs:
           path: _build/install.log
 
       - name: check
-        continue-on-error: true
```
GitMensch, line 82: What about moving this to two separate steps? One "check" with `continue-on-error` and another `test` that should never fail? After all these run as separate checks for win32 in any case.

GitMensch (changes requested):
Nah, "review pending" - I haven't clicked on send.

But this let me recheck another time now and after thinking "what do we want to achieve" I concluded that it is not the Win32 test working fine afterwards.

I therefore added some review comments asking for moving those changes out to separate PR(s) to be reviewed separately (and definitely later), allowing to merge this R hopefully next week.

Thank you and Sorry for the postponed review!

.github/workflows/windows.yml
GitMensch, line 1: the CI is separate with #170 and should be dropped here, then rebased after the CI PR is in the target branch

ddeclerck:
Took care of most changes ; as far as I can tell the only thing that remains to do is splitting the unrelated testsuite changes and `cobc_slashify` to another PR, right ?

ddeclerck:
@GitMensch I think I took all remarks into account. Is this ok for you ?

tests/testsuite.src/used_binaries.at
```diff
@@ -1172,3 +1148,192 @@ AT_CHECK([$COMPILE_MODULE -Wno-unfinished --copy "f.copy" -fstatic-call prog2.co
...
+        COPY COPY3 in "sub".
+        EX.
+            STOP RUN.
+])
+
+AT_CHECK([mkdir -p sub])
+AT_DATA([COPY1.CPY], [])
+AT_DATA([COPY2.CPY], [       DISPLAY "Hello".
+])
+AT_DATA([sub/COPY3.CPY], [])
```
GitMensch, line 1177: Please include in this test a copy that is included in another copy (=nested)= and include COPY2 two times in the program to cover these edge cases, possibly one time as `COPY COPY2.CPY`.

Also missing: compiling two programs at once. If I understand that correctly, this should result in two dependency files..

.github/workflows/windows-msvc.yml
GitMensch, line 1: please just commit these changes to the github workflow files directly into gcos4gnucobol-3.x, they seem to not be related to this

NEWS
```diff
@@ -57,6 +57,13 @@ NEWS - user visible changes				-*- outline -*-
 ** New option -fdefault-file-colseq to specify the default
    file collating sequence
 
+** New dependency options: output COPY dependencies of COBOL files with
+   -M to only print dependencies without compiling, -MD to print
+   dependencies which compiling, -MP to add phony -targets, MQ to
+   quote Makefile-specific characters and -MG to keep -missing
+   copybooks. More specific is -fcopybook-deps to print only dependency
```
GitMensch, line 64: Are there any "prints"? I've thought this is output commonly going to a dependency file`, with possibly the special case of output to std...out?
I think it is best to only note the options here or even say "related to -MT"( or similar) and then reference the new chapter in the manual to be more concise.

cobc/typeck.c
```diff
@@ -3588,6 +3588,7 @@ check_argument_conformance (struct cb_program *program, cb_tree argument_tripple
 	} else {
 		arg_field = NULL;
 	}
+	if (!CB_FIELD_P(param)) return;
```
GitMensch, line 3591: That's unrelated but important, I guess there's also  test that can (and should be) added for this.

cobc/pplex.l
```diff
@@ -159,6 +159,7 @@ static int	fill_continued_alnums = 1; /* whether continued alphanumeric
 					      spaces up to text column */
 static int	emit_area_a_tokens = 0;
 
+
```
GitMensch, line 162: that extra line seems to not have a purpose

cobc/cobc.c
```diff
@@ -9427,15 +9601,7 @@ main (int argc, char **argv)
...
-	if (cb_depend_file) {
-		struct cb_text_list	*l;
-		fprintf (cb_depend_file, "%s: \\\n", cb_depend_target);
-		for (l = cb_depend_list; l; l = l->next) {
-			fprintf (cb_depend_file, " %s%s\n", l->text, l->next ? " \\" : "\n");
-		}
-		for (l = cb_depend_list; l; l = l->next) {
-			fprintf (cb_depend_file, "%s:\n", l->text);
-		}
+	if (cb_depend_file && cb_depend_file != stdout){
```
GitMensch, line 9604: stdout cannot be null, so the second part of that check is enough

cobc/cobc.c
```diff
@@ -3001,6 +3019,80 @@ remove_trailing_slash (char *data)
 	}
 }
 
+static int
+string_is_dash (const char *s)
+{
+	return (strcmp (s, COB_DASH) == 0);
+}
```
GitMensch, line 3026: we should request inlining here, but ... I'm not sure if we maybe should revert these parts doing the inlinig ourselves (as it was done before)

cobc/help.c
```diff
@@ -135,8 +135,13 @@ cobc_print_usage_common_options (void)
...
 	puts (_("  --save-temps[=<dir>]  save intermediate files\n"
 	        "                        * default: current directory"));
+	puts (_("  -M                    output dependency list in Makefile format"));
 	puts (_("  -MT <target>          set/add target file used in dependency list"));
+	puts (_("  -MQ <target>          same as -MT but with Makefile-quoting of the target"));
 	puts (_("  -MF <file>            place dependency list into <file>"));
+	puts (_("  -MP                   create phony targets for all dependencies"));
+	puts (_("  -MG                   output missing dependencies without complaining"));
+	puts (_("  -MD                   output dependencies in .d files while compiling"));
 	puts (_("  -ext <extension>      add file extension for resolving COPY"));
```
GitMensch, line 145: -fcopybook-deps should be here for a grouped help, suppressed in flags.def, similar to others.

cobc/cobc.h
```diff
@@ -473,6 +473,9 @@ extern int			cb_saveargc;
 extern FILE			*cb_listing_file;
 extern FILE			*cb_src_list_file;
 extern FILE			*cb_depend_file;
+extern int                      cb_depend_output;
+extern int                      cb_depend_keep_missing;
+extern int                      cb_depend_keep_system;
```
GitMensch, line 478: This seems unused, possibly related to `-MMD`?

cobc/cobc.c
```diff
@@ -3001,6 +3019,80 @@ remove_trailing_slash (char *data)
...
+		cb_depend_target = cobc_strdup (s);
+	} else {
+		/* multiple invocations add to the list */
+		const size_t orig_len	= strlen (cb_depend_target);
+		const size_t new_len	= strlen (s);
+		const size_t buff_len	= orig_len + 1 + new_len + 1;
+		cb_depend_target = cobc_realloc (cb_depend_target, buff_len);
+		memset (cb_depend_target + orig_len, ' ', 1);
+		memcpy (cb_depend_target + orig_len + 1, s, new_len);
+		memset (cb_depend_target + orig_len + 1 + new_len, 0, 1);
```
GitMensch, line 3041: As those memset calls use only 1, maybe it is better to just assign that position? But I won't _request_ a change here.

cobc/cobc.c
```diff
@@ -3001,6 +3019,80 @@ remove_trailing_slash (char *data)
...
+
+static void
+add_depend_escape_target (const char *s)
+{
+	int i, len, nchars;
+	len = strlen (s);
+	nchars = 0;
+	for (i=0; i<len; i++){
+		char c = s[i];
+		if (c == '$' || c == '#' || c == '*') nchars++;
```
GitMensch, line 3053: Do we cover this in the testcases already?

cobc/cobc.c
```diff
@@ -3001,6 +3019,80 @@ remove_trailing_slash (char *data)
...
+static const char *
+file_replace_extension (const char *file, const char *ext)
+{
+	int len = strlen (file);
+	int extlen = strlen (ext);
+	int i;
+	for (i=len; i>0; i--){
+		char c = file[i];
+		if (c == '.'){
+			int newlen = i+extlen+1;
```
GitMensch, line 3085: the len's and c can be const

cobc/cobc.c
```diff
@@ -3001,6 +3019,80 @@ remove_trailing_slash (char *data)
...
+}
+
+static void
+add_depend_escape_target (const char *s)
+{
+	int i, len, nchars;
+	len = strlen (s);
+	nchars = 0;
+	for (i=0; i<len; i++){
+		char c = s[i];
```
GitMensch, line 3052: len and c can be const

tests/testsuite.src/used_binaries.at
```diff
@@ -1172,3 +1148,192 @@ AT_CHECK([$COMPILE_MODULE -Wno-unfinished --copy "f.copy" -fstatic-call prog2.co
...
+AT_CHECK([$SED "s/prog.$COB_OBJECT_EXT/prog.COB_OBJECT_EXT/g;s/sub\\\\/sub\//g" prog.dep], [0],
+[prog.COB_OBJECT_EXT: \
+ prog.cob \
+ COPY1.CPY \
+ COPY2.CPY \
+ sub/COPY3.CPY
+
+])
+
+AT_CHECK([$COMPILE_ONLY -M -fcopybook-deps prog.cob > compiler.output], [0])
```
GitMensch, line 1208: we should not need COMPILE_ONLY for -M, right?

GitMensch (changes requested):
Just a question: `-MMD` and `-isystem` are left for a later PR,. right?

This PR needs testcase adjustments, but otherwise loolks good.

.github/workflows/windows-msvc.yml
ddeclerck: They are not, but at some but I had a need for it to debug this PR.
Will merge them separately.

cobc/pplex.l
ddeclerck: Probably a leftover from the removed parts which have been moved to #180 ...

ddeclerck:
> Just a question: `-MMD` and `-isystem` are left for a later PR,. right?

Not sure ; quoting Fabrice's reply (https://github.com/OCamlPro/gnucobol/pull/109#issuecomment-2002510812):

> > > Should `-MM` and `-MMD` be added, ignoring the copybooks in `COB_COPY_DIR` (currently extfh and screenio)?
> 
> I don't think it is worth it. Though they are easy to implement (by modifying the `cb_copy_find_file` in `pplex.l` to set a flag when using the `COB_COPY_DIR` entry of `cb_include_list`), I don't see any reason why the external tool that would use the dependencies would need to remove system dependencies, or wouldn't be able to do it by itself. Adding low value features makes the code harder to maintain for little benefit.

GitMensch:
I agree in general (and COB_COPY_DIR would only be applied for the inbuilt one, because the env car is used by users for their copybooks as well).

If added, then this makes only sense with -isystem (= places for copybooks a user don't want to see), but we can leave that out until there is someone that would like to use it.
As noted in the review, I think they're one leftover var related to that.

tests/testsuite.src/used_binaries.at
ddeclerck: > Also missing: compiling two programs at once. If I understand that correctly, this should result in two dependency files..

This is actually already tested (using the same COBOL source twice):
```
AT_CHECK([$COMPILE -M prog.cob prog.cob > compiler.output], [0])
```
We do have an output for each .cob file.

NEWS
ddeclerck: I guess it's "print" in a wider definition;, whether to a file or to stdout (depends if using the `-MF` option).

cobc/typeck.c
ddeclerck: Not sure how to trigger this one.

cobc/cobc.c
ddeclerck: But `cb_depend_file` can be, and we do not want to "close" it if it is.

cobc/cobc.c
ddeclerck: I like the idea of a descriptive name such as `string_is_dash` for this (plus it's somewhat "easier" to write).
Would a macro be acceptable for this ? Otherwise I'll just reverse this change.

GitMensch:
Possibly passing a constant/literal? I'd need to check the context.

GitMensch:
Years ago I'd have said yes, but it is best to prevent function-like macros. I'm fine with an inline function.

Am 22. September 2024 12:33:39 MESZ schrieb OCP David Declerck ***@***.***>:

cobc/cobc.c
ddeclerck: Yup, at line 1227.

tests/testsuite.src/used_binaries.at
ddeclerck: >  and include COPY2 two times in the program to cover these edge cases

And indeed, we end up with duplicates in the dependency list. I have to fix that.

ddeclerck: Seems this is because `pp_text_list_add` does not check for duplicates. This function was present before this patch and seems to be only used to build the dependency list: is it okay to change it so that it check for duplicates *unconditionally* ?

GitMensch:
Yes.

Am 22. September 2024 13:33:17 MESZ schrieb OCP David Declerck ***@***.***>:

NEWS
GitMensch: As noted, please change it to only have an "overview" here and point to the manual for the details.

cobc/typeck.c
GitMensch: Possibly passing a constant/literal? I'd need to check the context.

cobc/pplex.l
```diff
@@ -2794,6 +2813,12 @@ pp_text_list_add (struct cb_text_list *list, const char *text,
 	struct cb_text_list	*p;
 	void			*tp;
 
+	for (p = list; p; p = p->next) {
+		if (!strncmp(text, p->text, size + 1)) {
```
GitMensch, line 2817: Why using not use strcmp here?

tests/testsuite.src/used_binaries.at
GitMensch: > > Also missing: compiling two programs at once. If I understand that correctly, this should result in two dependency files..
> 
> This is actually already tested (using the same COBOL source twice):
> 
> ```
> AT_CHECK([$COMPILE -M prog.cob prog.cob > compiler.output], [0])
> ```
> 
> We do have an output for each .cob file.

Hm, not sure if this would be the case for GCC? I've thought about two different programs as this is less questionable and more descriptive.

GitMensch:
From the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html) the following may be not "correct" yet:

> Passing `-M` to the driver implies `-E`, and suppresses warnings with an implicit `-w`.

Note that we can, of course, should be able to copy most of the GCC docs about those options to gnucobol.texi.


Would there be any reason to support [`DEPENDENCIES_OUTPUT` env var?](https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html#index-DEPENDENCIES_005fOUTPUT)

cobc/pplex.l
ddeclerck: Just a habit of using `strncmp`. Changed it.

tests/testsuite.src/used_binaries.at
ddeclerck: I changed it to have two different programs.

cobc/typeck.c
GitMensch: Let's move this and its Changelog entry to a separate PR and we can work out the right test (if @lefessan does not remember).

GitMensch (commented):
After moving that unrelated thing without a testcase out, this is nearly good to go now, it just needs ChangeLog entries for the changes done after Fabrice's work, then can be merged upstream.

GitMensch:
I'm quite sure that there were also recent adjustments from you :-)
But you can tackle this changelog change directly when committing upstream.

ddeclerck:
> I'm quite sure that there were also recent adjustments from you

Not really, just minor changes: allow duplicates in `pplex.l` (I added to the ChangeLog), moving the flag from `flag.def` to `cobc.c` (I adjusted the filenames in the ChangeLog entry), some whitespaces fixes (adding a space when I found `){` glued together),  and a two or three `const` added (to the new code, not to existing code).

Unless I'm missing something (I'm working on 10 things at the same time, I'm kinda lost)

GitMensch:
then let's only out the duplicate stuff in pplex.l to recent changelog entry from you before commit (and possibly leave flags.def in upstream as-is)

ddeclerck:
> then let's only out the duplicate stuff in pplex.l to recent changelog entry from you before commit (and possibly leave flags.def in upstream as-is)

Ah, I had added that to Fabrice's entry from 03/2024. It's only a 3-line patch, I'm okay with not taking the credit for that 😂

GitMensch:
I'd like to keep the Changelog including time entries relative clean, that's not (only) about attribution (also a the date is quite interesting) eeeeeeeeeI see from https://github.com/OCamlPro/gnucobol/compare/1a2949b4df2a55b7bca16f13a91c53a2e3807a9d..236c8b2c21dbaf470250d48bd2a014be4b10aaf1 that the adjustment is mostly testcases.

ddeclerck:
Didn't notice this was granted the "Ready for SVN" tag 😅. Rebasing before upstreaming.

GitMensch:
The most simple invocation needs to be adjusted, as this does not behave as GCC (and that's what for example IBM's cob2 follows)

~~~
$> touch  def.h
$> echo '#include "def.h"' > test.c
$> gcc -M test.c
 test.o: test.c /usr/include/stdc-predef.h def.h
$> ls -t | head
test.c def.h

$> touch def.cob
$> echo '       COPY "def.cob".' > test.cob
$> cobc -M test.cob
test.o: \
 test.cob \
 def.cob
$> ls -t | head
test.i test.cob def.cob
~~~

The format is minimal different but "logically (for make) identical", so I don't mind. The single part that needs an adjustment is that test.i should only be stored if `--save-temps` is used.

GitMensch:
tackled by #223 and now upstream

PR #112, opened 2023-10-02
New flag -fdiagnostics-absolute-path to display full paths within err…

lefessan:
…or locations

This flag is useful for editors that cannot correctly locate files with recursive build systems (typically, vscode running `make` to compile a gnucobol project)

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,11 +1,11 @@
 
 2023-10-02  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
-	* error.c (print_error_prefix), flag.def: add new flag
+	* error.c (print_error_prefix), flag.def: new flag
 	  -fdiagnostics-absolute-paths to print the full path of
-	  a file in case of error. This flag can be activated if
+	  a file for diagnostics; this flag can be activated if
 	  your editor and build system do not correctly work
-	  together to locate files after errors within your project.
+	  together to locate files from diagnostic output
 
 2023-07-26  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,12 @@
 
+2023-10-02  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* error.c (print_error_prefix), flag.def: add new flag
+	  -fdiagnostics-absolute-paths to print the full path of
+	  a file in case of error. This flag can be activated if
+	  your editor and build system do not correctly work
+	  together to locate files after errors within your project.
+
 2023-07-26  Simon Sobisch <simonsobisch@gnu.org>
 
 	* typeck.c (search_set_keys): improving SEARCH ALL syntax checks
--- a/cobc/error.c
+++ b/cobc/error.c
@@ -28,6 +28,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <stdarg.h>
+#include <sys/stat.h>
 
 #include "cobc.h"
 #include "tree.h"
@@ -57,6 +58,28 @@ static void
 print_error_prefix (const char *file, int line, const char *prefix)
 {
 	if (file) {
+		char *absfile = NULL ;
+		if (cb_diagnostics_absolute_paths
+		    && strcmp (file, COB_DASH) != 0
+		    && file[0] != '/'
+		    && file[0] != '\\'
+		    && ( file[0] != 0 && file[1] != ':' )
+			){
+			int filelen = strlen (file);
+			int dirlen = 256;
+			char *cwd ;
+			absfile = cobc_malloc( dirlen + 1 + filelen + 1 );
+			cwd = getcwd (absfile, dirlen);
+			if (cwd != NULL ){
+				struct stat st;
+				dirlen = strlen (cwd);
+				absfile[dirlen] = '/';
+				memcpy (absfile+dirlen+1, file, filelen+1);
+				if (!stat(absfile,&st)){
+					file = absfile;
+				}
+			}
+		}
 		if (line <= 0) {
 			fprintf (stderr, "%s: ", file);
 		} else if (cb_msg_style == CB_MSG_STYLE_MSC) {
@@ -64,6 +87,7 @@ print_error_prefix (const char *file, int line, const char *prefix)
 		} else {
 			fprintf (stderr, "%s:%d: ", file, line);
 		}
+		if (absfile) cobc_free (absfile);
 	}
 	if (prefix) {
 		fprintf (stderr, "%s", prefix);
--- a/cobc/error.c
+++ b/cobc/error.c
@@ -28,10 +28,19 @@
 #include <ctype.h>
 #include <errno.h>
 #include <stdarg.h>
+#ifdef HAVE_SYS_STAT_H
 #include <sys/stat.h>
+#endif
 
 #include "cobc.h"
 #include "tree.h"
+
+#ifdef	_WIN32
+#if !defined(__BORLANDC__) && !defined(__WATCOMC__) && !defined(__ORANGEC__)
+#include <direct.h> // _getcwd
+#define	getcwd		_getcwd
+#endif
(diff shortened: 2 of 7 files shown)


GitMensch:
How is the flag called by gcc and/or clang?

Note: a "plain" filename like "mycopy.cpy" should be resovlved in vscode by opening that file from the workspace; relative names will work (depending on the absolute/relative setting, the later taking the cwd/env you use for the task into account).
A real problem starts if you have the same name in multiple places and one of the paths specified with `-I` is used.
.. but then: wouldn't it be more reasonable to directly use the pathname as it was used on open (with no extra flag)?

This can also (with relative paths) be tested then.

lefessan:
`gcc` does not have this option, but `clang` has:
```
  -fdiagnostics-absolute-paths
                          Print absolute paths in diagnostics
```
Just a `s` to add...

For `vscode`, my experience is that, using the GnuCOBOL extension and problem matchers, and a project with a hierarchy of `Makefiles`, `vscode` was unable to correctly locate the files where the errors happened. It worked after adding this option.

GitMensch:
Please give a sample compile with / without this flag.

lefessan:
Without:
```
$ cobc -g --free -I copy/ -x -o 01_generate.exe 01_generate.cob
01_generate.cob:3: error: ENVIRONMENT DIVISION header missing
    1 | IDENTIFICATION DIVISION.
    2 | PROGRAM-ID. 01_generate.
    3 > x
    4 | ENVIRONMENT DIVISION.
    5 | 
```

With:
```
cobc -g --free -fdiagnostics-absolute-path -I copy/ -x -o 01_generate.exe 01_generate.cob
/home/lefessan/GIT/COBOL/cobol-project/src/01_generate.cob:3: error: ENVIRONMENT DIVISION header missing
    1 | IDENTIFICATION DIVISION.
    2 | PROGRAM-ID. 01_generate.
    3 > x
    4 | ENVIRONMENT DIVISION.
    5 | 
```

GitMensch:
I think I see where the issue is (executing a _recursive_ make, right?)

Please add a copybook that has a warning and is found in the copy folder.

lefessan:
```
cobc -g --free -fdiagnostics-absolute-path -I copy/ -x -o 01_generate.exe 01_generate.cob
/home/lefessan/GIT/COBOL/cobol-project/src/copy/integer-string.cpy:1: error: RECORD description missing or invalid
    1 > x
    2 | 01 :FILE:-REC.
    3 | 05 :FILE:-REC-INT PIC 9(5).
```

GitMensch:
What's the output with the unpatched version?

Is the problem a "make" that uses sub-folders?

lefessan:
The patch only modifies the behavior for relative paths, by adding the current working directory in front of it, and then checking whether such a file really exists. So, it will only modify the display in the cases where the display caret would work.

 cobc -g --free -I copy/ -x -o 01_generate.exe 01_generate.cob                                                                                                   [18:27]
```
copy/integer-string.cpy:1: error: RECORD description missing or invalid
    1 > x
    2 | 01 :FILE:-REC.
    3 | 05 :FILE:-REC-INT PIC 9(5).
```

Yes, my initial problem with vscode was using make in sub-folders, but this feature could be useful in the future for any editor having the same problem, even if it is fixed in vscode.

GitMensch:
The biggest issue I have is that we cannot add it reasonable to the automated test suite, can we?
If not then it may still be reasonable to add it to the _manual_ testsuite...

cobc/error.c
```diff
@@ -57,13 +58,36 @@ static void
...
+		    && strcmp (file, COB_DASH) != 0
+		    && file[0] != '/'
+		    && file[0] != '\\'
+		    && ( file[0] != 0 && file[1] != ':' )
+			){
+			int filelen = strlen (file);
+			int dirlen = 256;
+			char *cwd ;
+			absfile = cobc_malloc( dirlen + 1 + filelen + 1 );
+			cwd = getcwd (absfile, dirlen);
```
GitMensch, line 71: `getcwd()` needs a define handling for WIN32, feel free to include that from libcob/fileio.c

cobc/error.c
GitMensch, line 1: I just was reminded that <sys/stat.h> is not C89 and some environment actually don't have it.
We do have the necessary define, so you can already use `#ifdef HAVE_SYS_STAT_H` around the include and the reverse down in the usage.  [I do plan to fix that in other files for GC 3.3]
... but: wouldn't it be better to use `access()` there in any case?

tests/testsuite.src/used_binaries.at
```diff
@@ -1000,5 +1000,14 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
  > <EOF>
 ]])
 
+
+AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-absolute-paths -Wall prog.cob 2> compiler.output], [1])
```
GitMensch, line 1004: `AT_CHECK([$COMPILE -fdiagnostics-absolute-paths prog.cob 2> compiler.output], [1])`

should work fine, no?

tests/testsuite.src/used_binaries.at
```diff
@@ -1000,5 +1000,14 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+
+AT_CHECK([sed -e "s|$PWD|HOME|" compiler.output], [0],
+[HOME/prog.cob:7: error: CRUD.CPY: No such file or directory
```
GitMensch, line 1007: Instead of escaping the <> please use [[ ]] to check stdout and $SED instead of sed

GitMensch (changes requested):
please adjust the stat call, then this is ready for inclusion

cobc/error.c
lefessan: ok

cobc/error.c
lefessan: `HAVE_SYS_STAT_H` exists, but it is not used anywhere in GnuCOBOL...

For `access` vs `stat`, we are only interested in the existence of the file at that point, so `stat` would be better.

tests/testsuite.src/used_binaries.at
lefessan: ok

tests/testsuite.src/used_binaries.at
lefessan: There is no need to escape anymore with $COMPILE :-)

lefessan:
I am adding some documentation in gnucobol.texi, and I have a question:

Shouldn't `-fdiagnostics-plain-output` also imply `-fno-diagnostics-show-option` ?

GitMensch:
> Shouldn't `-fdiagnostics-plain-output` also imply `-fno-diagnostics-show-option` ?

Only if it is included with GCC, my understanding is that this isn't the case as the later only got introduced after the former (at least this is the way it was with GC). Please recheck to know.

cobc/error.c
GitMensch: > `HAVE_SYS_STAT_H` exists, but it is not used anywhere in GnuCOBOL...

That's what "I do plan to fix that in other files for GC 3.3" references.

Would it be better to not do any check if `stat` doesn't exist?
Note that `access (name, F_OK)` should possibly be fine, but I'll leave that to you.

GitMensch (approved):
Thanks for the documentation update!

cobc/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2023-10-02  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* error.c (print_error_prefix), flag.def: new flag
+	  -fdiagnostics-absolute-paths to print the full path of
+	  a file for diagnostics; this flag can be activated if
+	  your editor and build system do not correctly work
+	  together to locate files from diagnostic output
```
GitMensch, line 8: should we move this longer description to the NEWS and keep the reasoning out here (or have it in both)?

GitMensch: ping (NEWS entry would be necessary in any case)

PR #113, opened 2023-10-03
Fix decimal constant bug occuring when several programs in a COBOL file

ddeclerck:
This patch fixes a nasty bug dealing with decimal constants when there are several programs in a COBOL file.

Consider the following COBOL file:
```
       IDENTIFICATION DIVISION.
       PROGRAM-ID. prog.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01  X  PIC 9(2).
       PROCEDURE DIVISION.
           CALL "prog2"
           IF X + 42 = 0
               DISPLAY "OK".
           STOP RUN.
       END PROGRAM prog.

       PROGRAM-ID. prog2 INITIAL.
       PROCEDURE DIVISION.
           EXIT PROGRAM.
       END PROGRAM prog2.
```

Attempting to run this program will result in a segfault:
```
prog.cob:8: attempt to reference invalid memory address (signal)
```

We found out the bug was occurring because returning from a subprogram clears all decimal constants, i.e the generated code for `prog2` contains:
```
  P_clear_decimal:
  /* Clear Decimal Constant values */
  cob_decimal_clear (dc_1);
  dc_1 = NULL;
```

Hence when `prog` tries to use the constant 42 (through `dc_1`), we are greeted with a segfault since `dc_1` is now NULL.

The proposed fix is to move all the decimal constants to the "local" header files instead of the "global" header.
The testsuite has been run and does not complain.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -12602,7 +12602,8 @@ output_internal_function (struct cb_program *prog, cb_tree parameter_list)
 
 	seen = 0;
 	for (m = literal_cache; m; m = m->next) {
-		if (m->make_decimal) {
+		if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
+		 && m->make_decimal) {
 			if (!seen) {
 				seen = 1;
 				output_line ("/* Set Decimal Constant values */");
@@ -12809,7 +12810,8 @@ cancel_end:
 	output_line ("P_clear_decimal:");
 	seen = 0;
 	for (m = literal_cache; m; m = m->next) {
-		if (m->make_decimal) {
+		if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
+		 && m->make_decimal) {
 			if (!seen) {
 				seen = 1;
 				output_line ("/* Clear Decimal Constant values */");
@@ -13930,7 +13932,8 @@ codegen_internal (struct cb_program *prog, const int subsequent_call)
 		struct literal_list* m = literal_cache;
 		int comment_gen = 0;
 		for (; m; m = m->next) {
-			if (m->make_decimal) {
+			if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
+			 && m->make_decimal) {
 				if (!comment_gen) {
 					comment_gen = 1;
 					output_local ("\n/* Decimal constants */\n");
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -13926,6 +13926,29 @@ codegen_internal (struct cb_program *prog, const int subsequent_call)
 		/* Switch to main storage file */
 		output_target = cb_storage_file;
 	}
+
+	/* Decimal constants */
+	{
+		struct literal_list* m;
+		int comment_gen = 0;
+		for (m = literal_cache; m; m = m->next) {
+			if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
+			 && m->make_decimal) {
+				if (!comment_gen) {
+					comment_gen = 1;
+					output_local ("\n/* Decimal constants */\n");
+				}
+				output_local ("static\tcob_decimal\t%s%d;\n",
+						CB_PREFIX_DEC_FIELD, m->id);
+				output_local ("static\tcob_decimal\t*%s%d = NULL;\n",
+						CB_PREFIX_DEC_CONST, m->id);
+			}
+		}
+		if (comment_gen) {
+			output_local ("\n");
+		}
+	}
+
 }
 
 void
@@ -13962,28 +13985,6 @@ codegen_finalize (void)
 				cob_gen_optim (optidx);
 				output_storage ("\n");
 			}
-		}
-	}
-
-	/* Decimal constants */
-	{
-		struct literal_list* m;
-		int comment_gen = 0;
-		for (m = literal_cache; m; m = m->next) {
-			if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
-			 && m->make_decimal) {
-				if (!comment_gen) {
-					comment_gen = 1;
-					output_storage ("\n/* Decimal constants */\n");
-				}
-				output_storage ("static\tcob_decimal\t%s%d;\n",
-						CB_PREFIX_DEC_FIELD, m->id);
-				output_storage ("static\tcob_decimal\t*%s%d = NULL;\n",
-						CB_PREFIX_DEC_CONST, m->id);
-			}
-		}
-		if (comment_gen) {
-			output_storage ("\n");
 		}
 	}
 
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -12602,8 +12602,7 @@ output_internal_function (struct cb_program *prog, cb_tree parameter_list)
 
 	seen = 0;
 	for (m = literal_cache; m; m = m->next) {
-		if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
-		 && m->make_decimal) {
+		if (m->make_decimal) {
 			if (!seen) {
 				seen = 1;
 				output_line ("/* Set Decimal Constant values */");
@@ -12810,8 +12809,7 @@ cancel_end:
 	output_line ("P_clear_decimal:");
 	seen = 0;
 	for (m = literal_cache; m; m = m->next) {
-		if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
-		 && m->make_decimal) {
+		if (m->make_decimal) {
 			if (!seen) {
 				seen = 1;
 				output_line ("/* Clear Decimal Constant values */");
@@ -13929,11 +13927,10 @@ codegen_internal (struct cb_program *prog, const int subsequent_call)
 
 	/* Decimal constants */
 	{
-		struct literal_list* m;
+		struct literal_list* m = literal_cache;
 		int comment_gen = 0;
-		for (m = literal_cache; m; m = m->next) {
-			if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
-			 && m->make_decimal) {
+		for (; m; m = m->next) {
+			if (m->make_decimal) {
 				if (!comment_gen) {
 					comment_gen = 1;
 					output_local ("\n/* Decimal constants */\n");
--- a/tests/testsuite.src/run_misc.at
+++ b/tests/testsuite.src/run_misc.at
@@ -14570,3 +14570,32 @@ TST-DECIMAL IS < ZERO-DECIMAL
 
 
 AT_CLEANUP
+
+AT_SETUP([Decimal constants working after sub-program call])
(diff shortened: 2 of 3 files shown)


GitMensch:
That's a quite severe issue, please recheck if this happens with 3.1 and trunk, then create a bug issue with the details. Thank you.

I'll check the PR quite soon, likely tomorrow.

ddeclerck:
Confirmed on 3.1 and trunk as well.
Should I open the issue here or on SourceForge ?

GitMensch:
On SF please, that's the right place for any bug report or feature request :-)

Thanks!

ddeclerck:
Hi @GitMensch,
Did you have a bit of time to check this ?

lefessan:
The bug report is here: https://sourceforge.net/p/gnucobol/bugs/920/

ddeclerck:
You mean here : https://sourceforge.net/p/gnucobol/bugs/917/
(920 is a different thing)

GitMensch:
Without further checks I do wonder where this has gone bad, it seems this was fixed already?
https://sourceforge.net/p/gnucobol/bugs/431/

ddeclerck:
> Without further checks I do wonder where this has gone bad, it seems this was fixed already? https://sourceforge.net/p/gnucobol/bugs/431/

I think this is a different issue. The one you mention is about decimal constants not being initialized at all.
The issue we are facing right now is that when exiting a called module, every decimal constant are de-initialized, even those used by the calling module.

tests/testsuite.src/run_misc.at
```diff
@@ -14570,3 +14570,32 @@ TST-DECIMAL IS < ZERO-DECIMAL
 
 
 AT_CLEANUP
+
+AT_SETUP([Decimal constants working after sub-program call])
+AT_KEYWORDS([runmisc])
+
+# this used to cause a SIGSEGV
```
GitMensch, line 14577: please add ", see bug #917" there, which contains all discussion about that

tests/testsuite.src/run_misc.at
```diff
@@ -14570,3 +14570,32 @@ TST-DECIMAL IS < ZERO-DECIMAL
...
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01  X  PIC 9(2).
+       PROCEDURE DIVISION.
+           CALL "prog2"
+           IF X + 42 <> 0
+               DISPLAY "OK".
```
GitMensch, line 14588: as a suggestion: consider using "`DISPLAY "OK" WITH NO ADVANCING"`, as that allows to use a single line AT_CHECK below

cobc/codegen.c
```diff
@@ -13926,6 +13926,29 @@ codegen_internal (struct cb_program *prog, const int subsequent_call)
 		/* Switch to main storage file */
 		output_target = cb_storage_file;
 	}
+
+	/* Decimal constants */
+	{
+		struct literal_list* m;
+		int comment_gen = 0;
+		for (m = literal_cache; m; m = m->next) {
+			if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
```
GitMensch, line 13935: While unrelated to the original issue... I think we should drop the class check here, that's up to whatever defines `make_decimal`.
Suggestion: consider moving the initialization of `m` to its declaration.

tests/testsuite.src/run_misc.at
```diff
@@ -14570,3 +14570,32 @@ TST-DECIMAL IS < ZERO-DECIMAL
 
 
 AT_CLEANUP
+
+AT_SETUP([Decimal constants working after sub-program call])
```
GitMensch, line 14574: If I'm not wrong this is only happening on `INITIAL` sub-programs, as only those redo the init.
In this case please adjust to `decimal constants and INITIAL nested programs`, and add another empty line before AT_SETUP above

tests/testsuite.src/run_misc.at
```diff
@@ -14570,3 +14570,32 @@ TST-DECIMAL IS < ZERO-DECIMAL
...
+       PROCEDURE DIVISION.
+           CALL "prog2"
+           IF X + 42 <> 0
+               DISPLAY "OK".
+           STOP RUN.
+       END PROGRAM prog.
+
+       PROGRAM-ID. prog2 INITIAL.
+       PROCEDURE DIVISION.
+           EXIT PROGRAM.
```
GitMensch, line 14594: please use

```cobol
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01  X  PIC 99  VALUE 42.
       01  Y  PIC 9v9 VALUE 0.1.
       PROCEDURE DIVISION.
       MAIN.
      *    ensure that cobc cannot optimize the expression away
           IF FUNCTION CURRENT-DATE() = 0
              ADD 1 TO Y.
           IF (X + Y) / 42.1 = 1
               DISPLAY "OK".
           GOBACK.
```

here, so we use the constant, too

tests/testsuite.src/run_misc.at
```diff
@@ -14570,3 +14570,32 @@ TST-DECIMAL IS < ZERO-DECIMAL
...
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01  X  PIC 9(2).
+       PROCEDURE DIVISION.
+           CALL "prog2"
+           IF X + 42 <> 0
```
GitMensch, line 14587: to make that portable we should give it a `VALUE` and to ensure that also future versions of the compiler don't optimize the constant away too easily (trunk already uses integer arithmetic in most possible places) we need to make that more complex. I suggest the following (that's not ideal as it won't work with all `-std`, the only one that would work is replacing that function by `ACCEPT FROM DATE` to a 6 digit variable and use that)

```cobol
       01  X  PIC 9(2) VALUE 0.
       01  Y  PIC 9v9  VALUE 0.1.
       PROCEDURE DIVISION.
       MAIN.
      *    ensure that cobc cannot optimize the expression away
           IF FUNCTION CURRENT-DATE() = 0
              ADD 1 TO Y.
           CALL "prog2"
           IF X + Y + 42.1 <> 0
```

GitMensch (changes requested):
Only minor changes to do for now, mostly to the testsuite.

Please be so kind to adjust it, then upload the generated code of that program with and without the codegen change, to ease rechecking the result after switching from "storage" to local.

tests/testsuite.src/run_misc.at
GitMensch: Yes, I just want to ensure that we don't create another bug ;-)

tests/testsuite.src/run_misc.at
ddeclerck: Note that this bug does not only affect "nested" programs, but also consecutive programs in the same source file.

cobc/codegen.c
ddeclerck: Note there are two other loops whose prefix are similar in `cobc.c` at 12604 and 12812 ; should I remove the class check in those two too ?

tests/testsuite.src/run_misc.at
GitMensch: Good to know, then `decimal constants with multiple INITIAL programs in same source unit` would be better, but that's both too long and doesn't read that well...

Actually your test example doesn't use nested, but it seems to be good to also test that.
What do you think about copying the source to have that tested, either as separate test or in the same test (then possibly moving prog2 out to a copybook /  use conditional compilation)?

cobc/codegen.c
GitMensch: Thanks for being thorough! Everything that checks for `make_decimal` to be set should not test for the class of the literal any more.

ddeclerck:
> Only minor changes to do for now, mostly to the testsuite.
> 
> Please be so kind to adjust it, then upload the generated code of that program with and without the codegen change, to ease rechecking the result after switching from "storage" to local.

That has been done.

Here is the generated code in both case (relevant changes in `.h` files) :
[prog_nofix.zip](https://github.com/OCamlPro/gnucobol/files/12913739/prog_nofix.zip)
[prog_fix.zip](https://github.com/OCamlPro/gnucobol/files/12913738/prog_fix.zip)

GitMensch:
Thank you for your work. I've adjusted the [bug report](https://sourceforge.net/p/gnucobol/bugs/917/) with the findings - it isn't about `INITIAL` at all, but about `CANCEL`...

It seems reasonable to move the "INITIAL" from `AT_SETUP` to `AT_KEYWORDS` (also add `CALL` and `CANCEL` there), then drop `INITIAL` from prog.cpy and include that in the current places via `REPLACING` and finally add two other iterations of that copy using non-initial programs that are `CANCEL`ed after their `CALL`.

As far as I see your change will then still work fine - then please push that change to the testsuite here.


The second issue in both the old and new codegen: all constants from all runtime-elements (programs and user-defined functions) are initialized and clean in all runtime-elements, even if they aren't used at all (which is highlighted by your first test code where the constant was not used in prog2 and still got cleaned at its exit.
This is only _related_ to the issue tackled here and may be handled separately. I've [another bug report](https://sourceforge.net/p/gnucobol/bugs/923/) for this, which could be solved first (and drop the `make_decimal` check in that loop, too.


Another _related_ issue is, that we don't need multiple definitions in the first place. Adjusting that would also fix the bug at hand, but I currently see only two options and am not sure if those are good (both would leave the constant to be static and global):
1) move the init/cleanup to shared object loading/unloading (like `output_so_load_version_check`), but that would leave them active forever if `COB_PHYSICAL_CANCEL` is not used - and still needs a fallback if neither `HAVE_ATTRIBUTE_CONSTRUCTOR` nor `WIN32` is set, the longer I think about that, the less I like that option myself
2) if the compile-unit has more than one program (otherwise there is no problem in the first place), then adjust the codegen to check that variable for NULL before doing the init and add a reference counter for each cob_decimal that is incremented on each place where the init is done, decrement on cleanup and only do real cleanup + setting to NULL if the reference counter is zero.

We _could_ use the change as-is, which would fix the SIGSEGV with the downside of a bit increased cpu time and memory usage (common programs likely have 5 to 20 decimal constants).

What is your take on this?

GitMensch:
I'd declare this PR as "ready for svn" (even when svn committing is still postponed) as soon as the Changelog entry is added.

Can you work on #923? In this case I'd suggest to postpone the change on

		if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
		 && m->make_decimal) {

there.

ddeclerck:
I'll try. I'll add questions directly on the SVN issue if needed.

GitMensch (approved):

GitMensch:
@ddeclerck Please bring this and #115 to svn, as two commits in whatever order. Then reference the commits at the referenced issues on SF (not sure if you can close them there yourself, if yes, please do so).

PR #114, opened 2023-10-11
New option --copy COPYBOOK, to include a COPYBOOK

lefessan:
This option provides the same behavior as `--include FILE` for `gcc` and `clang`, i.e. including a file (or a group of files) before parsing the source. 
It can be used typically to include `REPLACE` statements.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -22,6 +22,16 @@ NEWS - user visible changes				-*- outline -*-
 
 * Changes to the COBOL compiler (cobc) options:
 
+** New option --copy COPYBOOK to load copybooks before parsing files. This
+   option can typically be used to perform replacements without modifying
+   the source code, or to add prototypes for external calls.
+
+** New option --include FILE.h to add a #include in the generated C file.
+   This option can typically be used to force the C compiler to check static
+   calls to externals. The files are put into quotes, unless they start by
+   '<'. Quoted files are expected to have absolute paths, as the C compiler
+   is called in a temp directory instead of the project directory.
+
 ** output of unlimited errors may be requested by -fmax-errors=0,
    to stop compiliation at first error use -Wfatal-errors
 ** default value for -fmax-errors was changed from 128 to 20
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,9 @@
 
+2023-10-11  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c, pplex.l: new option --copy COPYBOOK, to include a COPYBOOK
+	  before reading the source file
+
 2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
 	* cobc.c (cobc_clean_up): when save-temps specifies a directory,
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,13 @@
 
+2023-10-12  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c, codegen.c: new option --include FILE, to #include
+	  additional files in the C generated code. Such files can be
+	  used to statically check the number of arguments in static
+	  calls, for example. The files are put into quotes, unless
+	  they start by '<'. Since C files are compiled in a temp dir,
+	  quoted files should be absolute paths. Implementing FR #176
+
 2023-10-11  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
 	* cobc.c, pplex.l: new option --copy COPYBOOK, to include a COPYBOOK
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -107,6 +107,7 @@ enum compile_level {
 #define	CB_FLAG_GETOPT_DEFAULT_COLSEQ  15
 #define	CB_FLAG_GETOPT_MEMORY_CHECK    16
 #define	CB_FLAG_GETOPT_COPY_FILE       17
+#define	CB_FLAG_GETOPT_INCLUDE_FILE    18
 
 
 /* Info display limits */
@@ -234,6 +235,7 @@ const char		*demangle_name = NULL;
 const char		*cb_storage_file_name = NULL;
 const char		*cb_call_extfh = NULL;
 struct cb_text_list	*cb_copy_list = NULL;
+struct cb_text_list	*cb_include_file_list = NULL;
 struct cb_text_list	*cb_include_list = NULL;
 struct cb_text_list	*cb_depend_list = NULL;
 struct cb_text_list	*cb_intrinsic_list = NULL;
@@ -598,6 +600,7 @@ static const struct option long_options[] = {
 	{"std",			CB_RQ_ARG, NULL, '$'},
 	{"conf",		CB_RQ_ARG, NULL, '&'},
 	{"copy",                CB_RQ_ARG, NULL, CB_FLAG_GETOPT_COPY_FILE},
+	{"include",             CB_RQ_ARG, NULL, CB_FLAG_GETOPT_INCLUDE_FILE},
 	{"debug",		CB_NO_ARG, NULL, 'd'},
 	{"ext",			CB_RQ_ARG, NULL, 'e'},	/* note: kept *undocumented* until GC4, will be changed to '.' */
 	{"free",		CB_NO_ARG, NULL, 'F'},	/* note: not assigned directly as this is only valid for */
@@ -3911,6 +3914,17 @@ process_command_line (const int argc, char **argv)
 			}
 			CB_TEXT_LIST_ADD (cb_copy_list,
 					  cobc_strdup (cob_optarg));
+			break;
+
+		case CB_FLAG_GETOPT_INCLUDE_FILE: /* 18 */
+			/* -include=<file.h> : add #include "file.h" to 
+			   generated C file */
+			if (strlen (cob_optarg) > (COB_MINI_MAX)) {
+				cobc_err_exit (COBC_INV_PAR, "--include");
+			}
+			CB_TEXT_LIST_ADD (cb_include_file_list,
+					  cobc_strdup (cob_optarg));
+			cb_flag_c_decl_for_static_call = 0;
 			break;
 
 		case 'A':
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -105,7 +105,8 @@ enum compile_level {
 #define	CB_FLAG_GETOPT_NO_DUMP         13
 #define	CB_FLAG_GETOPT_EBCDIC_TABLE    14
 #define	CB_FLAG_GETOPT_DEFAULT_COLSEQ  15
-#define	CB_FLAG_MEMORY_CHECK           16
+#define	CB_FLAG_GETOPT_MEMORY_CHECK    16
+#define	CB_FLAG_GETOPT_COPY_FILE       17
 
 
 /* Info display limits */
@@ -171,8 +172,8 @@ enum compile_level {
 #define GC_C_VERSION	_("unknown")
 #endif
 
-#define	CB_TEXT_LIST_ADD(y,z)	y = cb_text_list_add (y, z)
-#define	CB_TEXT_LIST_CHK(y,z)	y = cb_text_list_chk (y, z)
+#define	CB_TEXT_LIST_ADD(list,z)	list = cb_text_list_add (list, z)
+#define	CB_TEXT_LIST_CHK(list,z)	list = cb_text_list_chk (list, z)
 
 #ifdef	_MSC_VER
 #define	CB_COPT_0	" /Od"
@@ -232,6 +233,7 @@ const char		*cb_cobc_build_stamp = NULL;
 const char		*demangle_name = NULL;
 const char		*cb_storage_file_name = NULL;
 const char		*cb_call_extfh = NULL;
+struct cb_text_list	*cb_copy_list = NULL;
 struct cb_text_list	*cb_include_list = NULL;
 struct cb_text_list	*cb_depend_list = NULL;
 struct cb_text_list	*cb_intrinsic_list = NULL;
@@ -595,6 +597,7 @@ static const struct option long_options[] = {
 	{"save-temps",		CB_OP_ARG, NULL, '_'},
 	{"std",			CB_RQ_ARG, NULL, '$'},
 	{"conf",		CB_RQ_ARG, NULL, '&'},
+	{"copy",                CB_RQ_ARG, NULL, CB_FLAG_GETOPT_COPY_FILE},
 	{"debug",		CB_NO_ARG, NULL, 'd'},
 	{"ext",			CB_RQ_ARG, NULL, 'e'},	/* note: kept *undocumented* until GC4, will be changed to '.' */
 	{"free",		CB_NO_ARG, NULL, 'F'},	/* note: not assigned directly as this is only valid for */
@@ -3282,12 +3285,12 @@ process_command_line (const int argc, char **argv)
 			cobc_wants_debug = 1;
 			break;
 
-		case 8:
+		case CB_FLAG_GETOPT_DUMP:  /* 8 */
 			/* -fdump=<scope> : Add sections for dump code generation */
 			cobc_def_dump_opts (cob_optarg, 1);
 			break;
 
-		case 13:
+		case CB_FLAG_GETOPT_NO_DUMP: /* 13 */
 			/* -fno-dump=<scope> : Suppress sections in dump code generation */
 			if (cob_optarg) {
 				cobc_def_dump_opts (cob_optarg, 0);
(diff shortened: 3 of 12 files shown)


tests/testsuite.src/syn_copy.at
```diff
@@ -1072,3 +1072,36 @@ AT_DATA([prog.cob], [
...
 
 AT_CLEANUP
+
+
+AT_SETUP([Option --copy COPYBOOK])
+AT_KEYWORDS([argument])
+
+AT_DATA([copybook.cpy], [
+       REPLACE ==BEGIN PROGRAM== BY ==IDENTIFICATION DIVISION.
+       PROGRAM-ID.==
```
GitMensch, line 1082: please split this to two files, so we can check the use with two `--copy` below

tests/testsuite.src/syn_copy.at
```diff
@@ -1072,3 +1072,36 @@ AT_DATA([prog.cob], [
...
+prog.cob:2: error: invalid system-name 'BEGIN'
+prog.cob:2: error: syntax error, unexpected PROGRAM, expecting CRT or Identifier
+prog.cob:2: error: invalid system-name 'prog'
+prog.cob:2: error: syntax error, unexpected ., expecting CRT or Identifier
+prog.cob:3: error: syntax error, unexpected PROCEDURE
+prog.cob:4: error: PROCEDURE DIVISION header missing
+])
+
+AT_CHECK([$COMPILE_ONLY --copy copybook prog.cob])
+
```
GitMensch, line 1106: Please add another file and `AT_CHECK`  below (to the same test case) where the command-line-included copybook has an error; just to ensure (and showcase) that we have an reasonable output then.

cobc/cobc.c
```diff
@@ -595,6 +597,7 @@ static const struct option long_options[] = {
 	{"save-temps",		CB_OP_ARG, NULL, '_'},
 	{"std",			CB_RQ_ARG, NULL, '$'},
 	{"conf",		CB_RQ_ARG, NULL, '&'},
+	{"copy",                CB_RQ_ARG, NULL, CB_FLAG_COPY_FILE},
```
GitMensch, line 600: needs to be added to cobc/help.c and doc/gnucobol.texi, too

GitMensch (changes requested):
I approve the idea and the general approach (that _could_ be an issue if we don't discuss it up-front).

This needs some more checks and documentation, then would be fine.

Can you please also add `--include` to actually include the specified header in every generated .c file after the default includes?
This would be the "command line" part of https://sourceforge.net/p/gnucobol/feature-requests/176/ and would work identical with all C compilers as backends - and is just _so_ related to this PR...

cobc/cobc.c
```diff
@@ -3887,6 +3890,14 @@ process_command_line (const int argc, char **argv)
 			}
 			break;
 
+		case CB_FLAG_COPY_FILE: /* 17 */
+			/* -copy=<file> : COPY file at beginning */
+			{
```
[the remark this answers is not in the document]
GitMensch: Please move that outside of the parsing, directly to this function.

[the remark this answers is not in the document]
lefessan: Ok, I will look at it, though I am not sure of the benefit, compared to the extra complexity it introduces.

For `-include`, I cannot check that the file exists, as `cobc` does not have the full path (additional `-I dir` could be introduced to the C compiler by `-A`).

Do you have a test example that I could add to the testsuite ? I don't really see the point of `-include file.h` as the generated C code is supposed to be only dependent on `libcob.h`.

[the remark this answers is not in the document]
GitMensch: Thank you. It is true that there's no point in checking the `--include` file at all (this is the same as `-A` or `COBC_CFLAGS` or similar), neither in cobc nor as a false example in the testsuite (we can test the positive one).

The use is primarily for `CALL`ing static into C functions with the correct prototype, especially on Win32 where things like STDCALL are important.
An example for the testsuite would be something like a `CALL STATIC mymemcpy USING BY REFERENCE fld1 fld2 BY VALUE LENGTH OF fld1`  with a `--include` of a header that only contains `#define mymemcpy memcpy`.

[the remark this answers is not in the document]
lefessan: I have added the corresponding commit, but I couldn't find a way to remove the check that the function always returns an integer.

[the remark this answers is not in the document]
GitMensch: Thank you for this very good commit!

I _think_ you've searched for `CALL something RETURNING NOTHING` (OC extension) or `CALL something RETURNING OMITTED`, didn't you?

In [588cbab](https://github.com/OCamlPro/gnucobol/pull/114/commits/588cbab6796adc48fc06df82ef6584ecc17b0623) it seems that you've manually did imply what `-fno-gen-c-decl-static-call` does. I think it is fine to imply that (it just has to be noted in help.c and _maybe_ in gnucobol.texi), please just don't check by the new list but instead set `cb_flag_c_decl_for_static_call = 0;`.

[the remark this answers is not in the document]
lefessan: I think I have implemented everything that was requested.

Any estimate on when we can restart landing commits into the SVN ?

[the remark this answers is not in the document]
GitMensch: > Any estimate on when we can restart landing commits into the SVN?

I try hard to not let this be later than Wednesday next week.

cobc/cobc.c
```diff
@@ -9252,6 +9277,19 @@ main (int argc, char **argv)
 	finish_setup_compiler_env ();
 	finish_setup_internal_env ();
 
+	{
+		struct cb_text_list *l;
+		for (l = cb_copy_list; l; l=l->next){
+			char *name = cobc_strdup(l->text);
```
GitMensch, line 9283: Do we need to duplicate the l->text? Why not using it directly?
If needed, then please use a local `char name[COB_MINI_BUFF]` here, this also removes the new failing under valgrind's memcheck when there is an error (because the err_exit leaves before we free the data).

cobc/cobc.c
```diff
@@ -4434,7 +4458,8 @@ process_filename (const char *filename)
 		fn->translate = cobc_main_strdup (output_name);
 	} else
 	if (save_all_src || save_temps || save_c_src
-	 || cb_compile_level == CB_LEVEL_TRANSLATE) {
+	    || cb_include_file_list
```
GitMensch, line 4461: Is there a reason that we need the change here? Please explain or revert that line.

GitMensch:
Hm, and another one - mostly a question @lefessan: how does this apply to `-fformat=auto`, or better to say: shouldn't the auto-format be "reset" after each of the included `--copy` elements?

cobc/cobc.c
lefessan: If we don't do that, `#include "file.h"` will be compiled in `/tmp/`, so it will lookup `file.h` in `/tmp/` instead of in the project.

GitMensch: Wouldn't it be more reasonable to auto-add `-I .` to the used compiler flags or doesn't this work?

lefessan:
> It would also provide a "similar" option to MicroFocus `-C DIRECTIVES "FILE"` (take the old file, copy as ".cpy" go in there and add ` $SET` in front of each line, then compile in cobc with `--copy your.dir.cpy`.

It should probably be part of a document of recipes to migrate from MicroFocus to GnuCOBOL. Unfortunately, the localisation of errors would make it difficult to debug in case of problem, it would be better to call a directive entry point in the parser for each directives in a list, juste before including the copies.

GitMensch:
> it would be better to call a directive entry point in the parser for each directives in a list, just before including the copies

That would be interesting, especially if it handles the different directive variants. Something to keep in mind for later.

lefessan:
> Hm, and another one - mostly a question @lefessan: how does this apply to `-fformat=auto`, or better to say: shouldn't the auto-format be "reset" after each of the included `--copy` elements?

Good point. I just checked: if the format is `auto`, the detection is performed on the main target, the format is then set, and all copybooks are expected to be of the same format. I think it's a reasonable behavior.

GitMensch:
I agree, that's a reasonable approach, possibly also to suggest users to use format-independent COBOL in those `--copy` included copybooks ;-)

cobc/cobc.c
lefessan: What's the problem with the current behavior ? Is it because it might pollute the current directory with binary artefacts ?

Asking the user to use `-I $PWD` would also have an impact on `#include <...>` macros, when the option is only adding a quote include, that is supposed to be the only ones looking for files inside the project.

cobc/cobc.c
lefessan: The reason is that `cb_copy_find_file()` expects a non-const first argument (it modifies `/` into '\\` under Windows for example), while `cb_text_list` contains a `const char*` string.

I used the local array to remove the warning from valgrind.

cobc/cobc.c
```diff
@@ -3887,6 +3893,24 @@ process_command_line (const int argc, char **argv)
 			}
 			break;
 
+		case CB_FLAG_GETOPT_COPY_FILE: /* 17 */
+			/* --copy=<file> : COPY file at beginning */
```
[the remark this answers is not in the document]
lefessan: done

cobc/cobc.c
GitMensch: Thank you for the explanation, adding a note that the buffer name point to is modified in its function doc is likely good (its enough to do before the commit to svn).

The change is good, was there also a reason (other than the late hour) to not use `strcpy`?

cobc/cobc.c
GitMensch: > Is it because it might pollute the current directory with binary artefacts ?

This is one of the reasons. Another one is that those artifacts may even overwrite actual source files (would be troublesome in any case, but we do have an issue about this in our trackers for a reason).

On the second thought I agree that **auto-adding** `-I .` was a bad idea, thank you for not jumping on that train.
It is up to the user to decide how he handles that, the one that will likely be _most_ used is on Win32 with `--include="<windows.h>"` or `--include="<pdcurses.h>"`  (= files that are in the system default include paths already, also on other systems).
Let's drop the _auto-adding_ of `--save-temps` and in the testsuite use `--include="$PWD/file.h"`.

lefessan: > Let's drop the _auto-adding_ of `--save-temps` and in the testsuite use `--include="$PWD/file.h"`.

But there is no auto-adding of `--save-temps`, this test only modifies the location of the generated C name, i.e. the current directory instead of the `/tmp` directory. That's what you mean ?

GitMensch: I see, it's only a partial version and then deletes the file after use.

But you're fine with not adding any path and just let the user specify `--include="$PWD/file.h"`, if this is what he wants, right? In this case we won't need that the C file to be generated locally, either (as noted, most often people actually want to use a system-installed header).

lefessan: What troubles me with the needed $PWD is that it’s not the same behavior as in gcc, most people will get an error the first time unless they check the doc first.

GitMensch: I've thought about that... and take full responsibility. Please check in with $PWD, then this is ready for svn.

GitMensch:
As these are new options: please add them to the NEWS. I suggest to do that here already to not forget it on the later commit.

lefessan:
I just added the NEWS section.

I am still not comfortable with using absolute paths with `--include`. Recently, we discussed about using Docker to provide simpler multi-platform installations for GnuCOBOL: in my experience, it works pretty well, but only if all your paths are relative (the idea is to mount the local directory somewhere and run the program inside the container in that directory, so that it can access all local files). So, the `--include "$PWD/FILE.h"` will fail in such a context, since the absolute path is not the one in the Docker image.

If we cannot make a relative path the default, I added a commit to add a `COBC_BUILD_INPLACE` env variable that makes the C file to be generated and build in the local directory. It's not my preferred solution, but it would allow the Docker version to work correctly by setting it inside and outside the container.

GitMensch:
Can you please explain where you see the problem with something like `--include "FILE.h" -I .`?

GitMensch:
What's the benefit of `COBC_BUILD_INPLACE` over `--save-temps`? I guess it would place the intermediate files locally, but then deletes them afterwards?

lefessan:
I didn't notice that `-g` could be used for that.

For me, the behavior of `--include FILE.h` should not depend on whether it is used with or without another unrelated option. Indeed, `--include FILE.h -g` will probably be the set of options used by most users, so it will work without `$PWD`, but then, they will not understand why removing `-g` suddenly make their build fail. 

For now, I only see scenarios where the option would only be used in development mode (so, with `-g`), but my experience is that one day or another, somebody will find a way to try to use this option in an unexpected way in release mode too (to replace some calls by macros ?), so we should let the door opened...

GitMensch:
> so we should let the door opened...

We do with `--save-temps` or `-I $TMPDIR`. You may consider checking gnucobol.tex if those three possibilities are mentioned there or should be mentioned.
I'm not sure we need that at all, because, as I've said, people most commonly use that option with including a file from the system include directories; for me a single half sentence about this "note that to find the file you may need to pass an appropriate `-I`" or alike would be enough.

GitMensch:
What's the benefit of `COBC_BUILD_INPLACE` over `--save-temps`?
I _guess_ the difference is it would place the intermediate files locally, but then deletes them afterwards, right?

lefessan:
Yes, it does not leave intermediate files in the local dir. (FWIW, my latency is high since last week as I am in a two-week vacation with my kids... and I didn't have as much time as I expected to hack on my spare time :-) )

GitMensch (changes requested):
Please adjust the Changelog to reference FR #176 (and close that as soon as the change is committed to svn).
Please drop `COBC_BUILD_INPLACE ` and replace as noted in the previous posts.

Afterwards this can go into svn (ideally after @ddeclerck's changes).

lefessan:
> Please drop COBC_BUILD_INPLACE and replace as noted in the previous posts.

I am still not fond of using `--save-temps` to get this behavior, as it will pollute the local directory with unwanted intermediate files. Is it possible to keep it ?

GitMensch:
Isn't it enough to have `-I .` if you want the local path?

In any case: #112 can be integrated upstream before.

NEWS
```diff
@@ -19,6 +19,16 @@ NEWS - user visible changes				-*- outline -*-
 
 * Changes to the COBOL compiler (cobc) options:
 
+** New option --copy COPYBOOK to load copybooks before parsing files. This
+   option can typically be used to perform replacements without modifying
+   the source code
```
GitMensch, line 24: or to include COBOL prototypes :-)

lefessan: I added a testcase for it, but indeed, it is not completely supported, but at least, it shows a working example despite the warnings.

cobc/cobc.c
```diff
@@ -3909,6 +3912,16 @@ process_command_line (const int argc, char **argv)
 					  cobc_strdup (cob_optarg));
 			break;
 
+		case CB_FLAG_GETOPT_INCLUDE_FILE: /* 18 */
+			/* -include=<file.h> : add #include "file.h" to 
+			   generated C file */
+			{
+				char *file = cobc_strdup (cob_optarg);
+				cb_flag_c_decl_for_static_call = 0;
+				CB_TEXT_LIST_ADD (cb_include_file_list, file);
```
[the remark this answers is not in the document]

NEWS
GitMensch: consider to add -Wno-unfinished in that testcase - and at some point we should fix at least the strange RETURNING.

tests/testsuite.src/used_binaries.at
```diff
@@ -1011,3 +1011,63 @@ AT_CHECK([$COBC -fdiagnostics-plain-output -fdiagnostics-show-caret -Wno-others
...
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 long USAGE BINARY-C-LONG.
+       PROCEDURE        DIVISION.
+           CALL "f" USING "Hello" BY VALUE long RETURNING NOTHING.
+])
+
+AT_CHECK([$COBC -m --include "$PWD/file.h" -fstatic-call prog2.cob])
+
+AT_CHECK([$COBC -I . -m --include "file.h" -fstatic-call prog2.cob])
```
GitMensch, line 1047: Mainly wondering: Is that better than using what we normally do: `, [0], [], [])`?
If yes we should someday apply this globally.

Hm, and for "being identical", _maybe_ add it back in the meanwhile?

cobc/ChangeLog
```diff
@@ -1,4 +1,18 @@
 
+2023-10-12  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c, codegen.c: new option --include FILE, to #include
```
GitMensch, line 4: please reference the FR (I thought there was one), then this Changelog is good

GitMensch (approved):
Looks ready to me. One more solved :-)

tests/testsuite.src/used_binaries.at
lefessan: Yes, it is supposed to be equivalent.

NEWS
lefessan: done

lefessan:
Merged.

GitMensch:
Thank you. I've just recognized that there's one missing part from FR # 176: Back in the discussions then the point was mot sources that call into C always need the same C header, and therefore a new directive was the goal.

Can you or @engboris have a look at adding the `>>IMP INCLUDE` directive and close the related FR by that?

GitMensch:
reopen for the related missing part of the FR

engboris:
> Thank you. I've just recognized that there's one missing part from FR # 176: Back in the discussions then the point was mot sources that call into C always need the same C header, and therefore a new directive was the goal.
> 
> Can you or @engboris have a look at adding the `>>IMP INCLUDE` directive and close the related FR by that?

I'm currently working on it. What should happen if both the compiler option and the directive are defined? Should one be ignored? Both be considered?

GitMensch:
ping @engboris

ddeclerck:
Shall we close this one again, since it is now superseded by #143 ?

GitMensch:
Yes, it is fine to close that as only the directive is missing and this is what the related PR is about (not sure when I find the time to inspect the other current PRs though...).

PR #115, opened 2023-10-17
Fix Bug 923: generated modules init/clear unused decimal constants

ddeclerck:
That should address [https://sourceforge.net/p/gnucobol/bugs/923/](https://sourceforge.net/p/gnucobol/bugs/923/).

Passes all of the testsuite.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -2,21 +2,21 @@
 2023-10-17  David Declerck <david.declerck@ocamlpro.com>
 
 	BUG #923: generated modules init/clear unused decimal constants
-	* tree.h (struct cb_program): new `decimal_constant` field to store
-	  decimal constants used by the current program
-	* codegen.c (literal_list): removal of the unused `c` field
+	* codegen.c (literal_list): removal of the unused x field,
+	  and type moved to tree.h
+	* tree.h (struct cb_program): new decimal_constant field
+	  to store decimal constants used by the current program,
 	* codegen.c (cb_cache_program_decimal_constant): new function
 	  that adds constants used by the current program to the new
-	  `decimal_constant` field in `struct cb_prog`
+	  decimal_constant field in struct cb_prog
 	* codegen.c (cb_lookup_literal): added the current program as
-	  argument to `cb_lookup_literal` to account for different
+	  argument to cb_lookup_literal to account for different
 	  usage contexts (parse/typecheck vs codegen)
 	* codegen.c (output_internal_function): iterate over
-	  `prog->decimal_constants` instead of `literal_cache` so as to only
+	  prog->decimal_constants instead of literal_cache so as to only
 	  output decimal constants actually used by the current program
-	* codegen.c (output_param): pass `current_prog` to `cb_lookup_literal`
-	* typeck.c (decimal_expand, cb_build_cond_fields): pass `current_program`
-	  to `cb_lookup_literal`
+	* typeck.c (decimal_expand, cb_build_cond_fields),
+	  codegen.c (output_param) : pass current_program to cb_lookup_literal
 
 2023-07-26  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,23 @@
 
+2023-10-17  David Declerck <david.declerck@ocamlpro.com>
+
+	BUG #923: generated modules init/clear unused decimal constants
+	* tree.h (struct cb_program): new `decimal_constant` field to store
+	  decimal constants used by the current program
+	* codegen.c (literal_list): removal of the unused `c` field
+	* codegen.c (cb_cache_program_decimal_constant): new function
+	  that adds constants used by the current program to the new
+	  `decimal_constant` field in `struct cb_prog`
+	* codegen.c (cb_lookup_literal): added the current program as
+	  argument to `cb_lookup_literal` to account for different
+	  usage contexts (parse/typecheck vs codegen)
+	* codegen.c (output_internal_function): iterate over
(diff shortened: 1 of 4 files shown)


cobc/ChangeLog
```diff
@@ -1,4 +1,23 @@
 
+2023-10-17  David Declerck <david.declerck@ocamlpro.com>
+
+	BUG #923: generated modules init/clear unused decimal constants
+	* tree.h (struct cb_program): new `decimal_constant` field to store
+	  decimal constants used by the current program
+	* codegen.c (literal_list): removal of the unused `c` field
```
GitMensch, line 7: that's likely "x", not "c"
rechecking the Changelog show that we don't use backticks in there, so please adjust the entries

cobc/ChangeLog
```diff
@@ -1,4 +1,23 @@
...
+	* codegen.c (cb_lookup_literal): added the current program as
+	  argument to `cb_lookup_literal` to account for different
+	  usage contexts (parse/typecheck vs codegen)
+	* codegen.c (output_internal_function): iterate over
+	  `prog->decimal_constants` instead of `literal_cache` so as to only
+	  output decimal constants actually used by the current program
+	* codegen.c (output_param): pass `current_prog` to `cb_lookup_literal`
+	* typeck.c (decimal_expand, cb_build_cond_fields): pass `current_program`
+	  to `cb_lookup_literal`
+
```
GitMensch, line 20: one entry is enough:

* codegen.c (output_param), typeck.c (decimal_expand, cb_build_cond_fields):
  pass `current_program` to `cb_lookup_literal`

cobc/codegen.c
```diff
@@ -2763,8 +2762,28 @@ output_source_cache (void)
 
 /* Literal */
 
+static void
+cb_cache_program_decimal_constant (struct cb_program *prog, struct literal_list *cached_literal)
+{
+	struct literal_list	*l;
+	if (prog != NULL && cached_literal->make_decimal) {
```
GitMensch, line 2769: Is `prog` ever `NULL` here?
Should we only call that function if `make_decimal` is true (=checked outside)?

In any case, please add a function doc comment for new functions (and wherever you work on a function without, feel invited to add one).

cobc/codegen.c
```diff
@@ -2763,8 +2762,28 @@ output_source_cache (void)
...
+		l->literal = cached_literal->literal;
+		l->make_decimal = cached_literal->make_decimal;
+		l->next = prog->decimal_constants;
+		prog->decimal_constants = l;
+	}
+}
+
 int
-cb_lookup_literal (cb_tree x, int make_decimal)
+cb_lookup_literal (struct cb_program *prog, cb_tree x, int make_decimal)
```
GitMensch, line 2786: this one for example, could use one saying that the literal id is resolved from the literal cache and the literal added, if missing and after the change also say that if the literal is used as decimal, it is attached to the passed program's decimal constants list

cobc/codegen.c
```diff
@@ -12601,9 +12621,8 @@ output_internal_function (struct cb_program *prog, cb_tree parameter_list)
 	}
 
 	seen = 0;
-	for (m = literal_cache; m; m = m->next) {
-		if (CB_TREE_CLASS (m->x) == CB_CLASS_NUMERIC
-		 && m->make_decimal) {
+	for (m = prog->decimal_constants; m; m = m->next) {
+		if (m->make_decimal) {
```
GitMensch, line 12625: all entries in that list have the decimal marker, so that additional check can be dropped, same below

cobc/codegen.c
```diff
@@ -127,7 +127,6 @@ struct attr_list {
 struct literal_list {
 	struct literal_list	*next;
 	struct cb_literal	*literal;
-	cb_tree			x;
 	int			id;
 	int			make_decimal;
 };
```
GitMensch, line 132: Shouldn't that struct be moved to tree.h?

GitMensch (changes requested):
Thank you for this well done PR.

There are only minor adjustments needed. Once there is a "go" for committing to svn ("soon"), I suggest to first commit this, then the other PR of yours.

cobc/codegen.c
ddeclerck: Well, if `prog` is NULL, it's actually a bug. Should I just let the compiler crash in this case ?

Performing the check on `make_decimal` inside the function looks safer to me (one does not have to think whether he should or shouldn't call the function) ; of course, as long as it is only ever used in `cb_lookup_literal`, this is not a big deal...

cobc/codegen.c
ddeclerck: Actually I just added a forward declaration `struct literal_list;` in `tree.h`; that is sufficient (I don't see any reason to expose the whole type in `tree.h`).

GitMensch: I confess that I don't know if that works like declarations of functions in ancient C compilers and also don't know how that looks like in GDB when inspecting the program. But we can test.

cobc/codegen.c
GitMensch: yes, please let the compiler SIGSEGV there, and if this happens in `make checkall` then there's something related to fix

as per your comment - as this is only a function with local state - let's do the check outside

ddeclerck: No SIGSEGV, we're all good ;)

cobc/codegen.c
GitMensch: You mean the program? Sure if that's possible without including tree.h in too many other sources - that would actually be preferable.

ddeclerck: I mean writing the full definition of `struct literal_list` in `tree.h` instead of just the forward declaration.

GitMensch (approved):

PR #117, opened 2023-10-17
[CANCELLED] Add a settings file to configure cobc at runtime

lefessan:
New --settings=FILE can be used to configure flags that are normally set at compile time. These settings can be used typically to create cross-compilers by retargetting an existing compiler.

Also add a setting to make the compiler relocatable, i.e. compute paths of include, share, etc. relative to cobc's location.

### DIFF
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -106,6 +106,7 @@ enum compile_level {
 #define	CB_FLAG_GETOPT_EBCDIC_TABLE    14
 #define	CB_FLAG_GETOPT_DEFAULT_COLSEQ  15
 #define	CB_FLAG_MEMORY_CHECK           16
+#define	CB_FLAG_GETOPT_SETTINGS        17
 
 
 /* Info display limits */
@@ -369,13 +370,7 @@ static struct cobc_mem_struct	*cobc_plexmem_base = NULL;
 
 static const char	*cobc_cc;		/* C compiler */
 static char		*cobc_cflags;		/* C compiler flags */
-#ifdef COB_DEBUG_FLAGS
 static const char		*cobc_debug_flags;		/* C debgging flags */
-#else
-#ifndef	_MSC_VER
-#error		missing definition of COB_DEBUG_FLAGS
-#endif
-#endif
 
 static char		*cobc_libs;		/* -l... */
 static char		*cobc_lib_paths;	/* -L... */
@@ -395,11 +390,7 @@ static size_t		cobc_libs_len;
 static size_t		cobc_lib_paths_len;
 static size_t		cobc_include_len;
 static size_t		cobc_ldflags_len;
-#ifdef COB_EXPORT_DYN
-static size_t		cobc_export_dyn_len;
-#else
-#define		cobc_export_dyn_len		0
-#endif
+static size_t		cobc_export_dyn_len = 0;
 #ifdef COB_SHARED_OPT
 static size_t		cobc_shared_opt_len;
 #else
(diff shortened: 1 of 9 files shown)


cobc/cobc.c
```diff
@@ -592,6 +583,7 @@ static const struct option long_options[] = {
 	{"O2",			CB_NO_ARG, NULL, '2'},
 	{"O3",			CB_NO_ARG, NULL, '3'},
 	{"Os",			CB_NO_ARG, NULL, 's'},
+	{"settings",		CB_OP_ARG, NULL, CB_FLAG_GETOPT_SETTINGS},
```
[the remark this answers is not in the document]
GitMensch: OK, after inspecting settings.c, this actually makes sense. From my gut it feels like a separate `--export-settings[=<file>]` would be more appropriate with <file> defaulting to `-` (stdout).

In any case - if `cobc --settings` is added then the same feature should work with `cobcrun --runtime-config` (either without a parameter or as separate export option) providing a way to "dump" the settings, allowing the to be adjusted and then used.
We _may_ want to export/dump the settings (in both the cobc and cobcrun case) match the "installed defaults" as a comment line (just uncomment, then adjust)

cobc/settings.c
```diff
@@ -0,0 +1,446 @@
...
+```
+
+You can then modify any option within the file.
+
+A file containing these settings can be specified in 3 different ways,
+the first working way in this order will be used:
+
+* Use of the argument `--settings=FILE`
+* Use of the environment variable `COBC_SETTINGS`
+* Use of `../etc/gnucobol/gnucobol.settings` from the directory
```
GitMensch, line 75: Shouldn't this be the `COB_CONFIG_DIR` (with its fallback) instead?

configure.ac
GitMensch, line 1: I'm not sure that we need a setting for that, why not always have that?

cobc/settings.c
```diff
@@ -0,0 +1,446 @@
...
+  containing the `cobc` executable
+
+If a file is specified, an error while loading the file is fatal.
+
+Once the settings have been loaded, a few variables can be modified if
```
GitMensch, line 80: I suggest to make that not depending on `COBC_IS_RELOCATABLE` (env var or its default) but instead just checking if the executable is in the place that we expect - if it isn't then o the relocation.
... but this part should possibly go elsewhere (see main review comment)

GitMensch (changes requested):
This is about https://sourceforge.net/p/gnucobol/feature-requests/450/ where additional important discussion happens. Please outline more about the need for this configuration there, keeping in mind that if you want to cross-compile for another environment you still have to have:

* GnuCOBOL built for that cross-target (to get "libcob"; normally this will also provide you with a "cobc" for that environment)
* a C cross-compiler (that you've likely used to generate that GnuCOBOL version in the first place, so all of its configuration is already inbuilt in the generated cobc executable)

The main culprit: when building GnuCOBOL itself with a cross-compiler you need a subsystem to also execute the resulting cobc binary (for example wine).
I do wonder if it isn't more reasonable to work on https://sourceforge.net/p/gnucobol/feature-requests/245/ which allows to not only cross-compile GnuCOBOL but also provide the option to generate a cross-compiling cobc.


A "side issue" here is the option to be relocatable, but that does only apply here to the compiler, not to the runtime.
I suggest to keep that out of this PR and instead improve the relocatable option that Ron already added in trunk (the "get name" function for example would be a reasonable addition).

Note: I do think that making GnuCOBOL well relocatable is important and this may also land in 3.x (likely to only happen if you provide a patch), but as we have that in trunk already it seems more reasonable to test and improve there (it was only tested with GNU/Linux x64), then either merge the revisions to 3.x or do a big backport instead.


If we still want to add a compiler setting, wouldn't it then be reasonable to use the existing configuration options in libcob?
To do so, the current entry functions in config.c would get an additional pointer to the configuration structure (we even can store it in a static var to ease access later on) and and we'd have a separate configuration table that only contains the cobc related settings, then pass the that to the config functions in libcob to set the values (single place of already well tested code, with guaranteed identical behavior in the future).


Nonetheless - first review while still in draft

I highly suggest to add ChangeLog entries at least before each push, this way you know more what you've changed and why, and it also helps for a possible review.

GitMensch:
Friendly ping for rebase and potentially handling some of the PR comments.

@lefessan What state would you see this PR in? Can you work on this (or let someone do that)? The general ideas are definitely good.

lefessan:
Wow, I had completely  forgotten this patch. I need to discuss this with the team, to check if it's worth taking somebody else's time for it, or if I need to take some time on my own week-end :-)

PR #126, opened 2023-11-29
Small fixes from profiling PR

lefessan:
Do not move object files if they were specified as an explicit target on the command line (typically `cobc -c --save-temps=DIR -o foo.o foo.cob` should keep `foo.o` in the current directory)

### DIFF
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -2180,7 +2180,8 @@ clean_up_intermediates (struct filename *fn, const int status)
 	if (fn->need_preprocess
 	 && (status
 		||  cb_compile_level > CB_LEVEL_PREPROCESS
-		|| (cb_compile_level == CB_LEVEL_PREPROCESS && save_temps))) {
+		|| (cb_compile_level == CB_LEVEL_PREPROCESS
+		    && save_temps && !save_temps_dir))) {
 		cobc_check_action (fn->preprocess);
 	}
 	/* CHECKME: we had reports of unexpected intermediate
@@ -2287,7 +2288,8 @@ cobc_clean_up (const int status)
 		if (fn->need_assemble
 		 && (status
 			||  cb_compile_level > CB_LEVEL_ASSEMBLE
-			|| (cb_compile_level == CB_LEVEL_ASSEMBLE && save_temps))) {
+			|| (cb_compile_level == CB_LEVEL_ASSEMBLE
+			    && save_temps && !save_temps_dir))) {
 			cobc_check_action (fn->object);
 		}
 		clean_up_intermediates (fn, status);
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -2,6 +2,7 @@
 2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
 	* common.c (cob_get_strerror), coblocal.h: export as utility function
+	* common.c (cob_expand_env_string): fix potention buffer overflow
 
 2023-07-28  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,8 +1,7 @@
 
 2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
-	* common.h/common.c: export "cob_get_strerror" as a public function
-	* coblocal.h: include "config.h"
+	* coblocal.h, common.c: export "cob_get_strerror" as a public function
 
 2023-07-28  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,4 +1,9 @@
 
+2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* common.h/common.c: export "cob_get_strerror" as a public function
+	* coblocal.h: include "config.h"
+
 2023-07-28  Simon Sobisch <simonsobisch@gnu.org>
 
 	* screenio.c, common.c: replace use of NCURSES_MOUSE_VERSION by
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,7 +1,7 @@
 
 2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
-	* coblocal.h, common.c: export "cob_get_strerror" as a public function
+	* common.c (cob_get_strerror), coblocal.h: export as utility function
 
 2023-07-28  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/libcob/coblocal.h
+++ b/libcob/coblocal.h
@@ -487,6 +487,7 @@ COB_HIDDEN int		cob_check_env_true	(char*);
 COB_HIDDEN int		cob_check_env_false	(char*);
 COB_HIDDEN const char	*cob_get_last_exception_name	(void);
 COB_HIDDEN void		cob_parameter_check	(const char *, const int);
+COB_HIDDEN char*        cob_get_strerror (void);
 
 enum cob_case_modifier {
 	CCM_NONE,
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,10 @@
 
+2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* cobc.c (cobc_clean_up): when save-temps specifies a directory,
+	  do not move object files and preprocess files when they were
+	  specified as an explicit target on the command line (-E, -c)
+
 2023-07-26  Simon Sobisch <simonsobisch@gnu.org>
 
 	* typeck.c (search_set_keys): improving SEARCH ALL syntax checks
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -907,7 +907,7 @@ cob_get_source_line ()
 }
 
 /* reentrant version of strerror */
-static char *
+char *
 cob_get_strerror (void)
 {
 	size_t size;
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -7785,9 +7785,10 @@ cob_expand_env_string (char *strval)
 				}
 			}
 			if (penv != NULL) {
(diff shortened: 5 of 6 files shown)


cobc/cobc.c
```diff
@@ -2287,7 +2287,11 @@ cobc_clean_up (const int status)
 		if (fn->need_assemble
 		 && (status
 			||  cb_compile_level > CB_LEVEL_ASSEMBLE
-			|| (cb_compile_level == CB_LEVEL_ASSEMBLE && save_temps))) {
+			|| (cb_compile_level == CB_LEVEL_ASSEMBLE
+			    && save_temps
+			    /* do not move object if a name was
+			     * specified on the command line */
+			    && !output_name))) {
```
GitMensch, line 2294: Does this needs to be applied to at least `-C` (translate only), too? What about `-S`?

Should it do a strcmp to output_name?

In any case: this change needs a testsuite entry.

cobc/parser.y
```diff
@@ -10969,6 +11009,8 @@ procedure_division:
   statements
   _dot_or_else_area_a
   _procedure_list
+  {
+  }
```
GitMensch, line 11013: What is the purpose of this change?

cobc/parser.y
```diff
@@ -407,6 +407,46 @@ emit_statement (cb_tree x)
...
+
+/* Insert a statement to replace the current statement. */
+static COB_INLINE COB_A_INLINE void
+replace_last_statement (cb_tree x)
+{
+	drop_last_statement ();
+	emit_statement (x);
+}
+#endif
+
```
GitMensch, line 449: we should add it if/when they are actually used, not in a previous commit

libcob/coblocal.h
```diff
@@ -555,19 +554,27 @@ DECLNORET COB_HIDDEN void	cob_hard_failure (void) COB_A_NORETURN;
 
 /* static inline of smaller helpers */
 
+/* The following functions are only defined on-demand using flags
+   COB_NEEDS_*. Otherwise, we get either defined-but-not-used
+   warnings, or not-always-inlined warnings. */
```
GitMensch, line 559: I haven't seen those warnings, do you have that locally / in CI?
Any "special flags" used? Which C compiler version?

libcob/common.h
```diff
@@ -1645,6 +1645,7 @@ COB_EXPIMP int		cob_last_exception_is	(const int);
 
 COB_EXPIMP int		cob_last_exit_code	(void);
 COB_EXPIMP const char*	cob_last_runtime_error	(void);
+COB_EXPIMP char*        cob_get_strerror (void);
```
GitMensch, line 1648: The (purpose of the) exteralization is missing. Note that this only can be reasonably called if the errors came from libcob, if not then we need to mirror this for example in cobc.

libcob/ChangeLog
```diff
@@ -1,4 +1,13 @@
 
+2023-11-29  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* common.h: export "cob_get_strerror" as a public function
+	* common.c: (cob_expand_env_string): use "getpid" instead
+	  of "cob_sys_getpid" to use the correct PID in case of "fork"
```
GitMensch, line 6: Do you have an example where this is probematic? It isn't if you use `CBL_GC_FORK`, of course. Note that (especially with 4.x) there are also other elements that need to be adjusted.

GitMensch (changes requested):
not sure about all of those, for the cobc change we should have a testcase after the commit

cobc/parser.y
lefessan: Is there some kind of place where we can keep such functions for later when we need them ? If I remove them now, they will have to be rewritten from scratch when needed...

GitMensch: Sure, you can add as much feature branches in git (locally or also in this mirror) as you want :-)

libcob/coblocal.h
lefessan: Hum... ok, this is getting weird:

lefessan: Hum... ok, this is getting weird:
```
#include "coblocal.h"
#include "config.h"
```
triggers warnings such as:
```
In file included from ../../libcob/profiling.c:29:
../../libcob/coblocal.h:577:1: warning: ‘always_inline’ function might not be inlinable [-Wattributes]
  577 | cob_max_int (const int x, const int y)
      | ^~~~~~~~~~~
../../libcob/coblocal.h:570:1: warning: ‘always_inline’ function might not be inlinable [-Wattributes]
  570 | cob_min_int (const int x, const int y)
      | ^~~~~~~~~~~
../../libcob/coblocal.h:577:1: warning: ‘cob_max_int’ defined but not used [-Wunused-function]
  577 | cob_max_int (const int x, const int y)
      | ^~~~~~~~~~~
../../libcob/coblocal.h:570:1: warning: ‘cob_min_int’ defined but not used [-Wunused-function]
  570 | cob_min_int (const int x, const int y)
      | ^~~~~~~~~~~
```

Whereas:
```
#include "config.h"
#include "coblocal.h"
```
does not.

There must be some flag in "config.h" that is needed. 

Maybe "coblocal.h" should have a `#include "config.h"` instead of these changes ?

libcob/ChangeLog
lefessan: No, except the use of `fork()` in external C code.

libcob/common.h
lefessan: This change is needed to get the correct error message, for example in the profiling PR when `fopen` fails.

cobc/parser.y
lefessan: Not really needed here, actually.

libcob/coblocal.h
GitMensch: No. config.h should just always be included first.

libcob/ChangeLog
GitMensch: I'm not sure if/how we should handle that. The scenario's I think of are:

* "external" -> COBOL (then back, then fork then to COBOL in both)
* COBOL -> "external with fork" -> to COBOL

In both cases the runtime needs to know about that, because it needs to reset some things. While this is only the cached PID in GC 3.1+3.2, we may need more entries. For example trunk has improved locking with BDB and this is done internally by a "locker id" - and this may not be used in more than a single thread (or even process). fbdb keeps its locker ids and resets them (later acquiring a new one) for a cbl_sys_fork'd process.

So... _can_ we handle that outside of `cob_sys_fork`? I currently see no option.
If there is no option, why should we support a specific case of `fork` that may won't work otherwise?

It _seems_ better to explicit document that after the COBOL runtime is initialized you may only run `cob_sys_fork()` but not `fork()` directly. We can of course come up with an API function which we call from within `cob_sys_fork()` for childs and that external programs may use when they do a `fork()` before getting to COBOL again.

Opinions?

cobc/parser.y
lefessan: Removed

cobc/parser.y
lefessan: removed

cobc/cobc.c
lefessan: * `-C` and -S` work as expected in the presence of `-save-temps=DEBUG`

libcob/coblocal.h
```diff
@@ -22,6 +22,8 @@
 #ifndef COB_LOCAL_H
 #define COB_LOCAL_H
 
+#include "config.h"
```
lefessan, line 25: @GitMensch Any reason why `config.h` is not included systematically in all other `.h` files ?

GitMensch:
Yes, because this generated file has no inclusion guard - it should be included only once - and should be included before any system header in any case. As tarstamp.h may be used in there (that was kind of common for environments without auto-configuration in the patchlevel, but we could drop support for that) it is included before config.h

GitMensch:
In the current code that would be

```c
#ifndef COBCRUN_NAME
#ifndef COB_TAR_DATE
#include "tarstamp.h"
#endif
#include "config.h"
#endif
```

in both cobc/cobc.h and libcob/coblocal.h

I'd prefer the more easy one to just have

```c
/* may be used in config.h, e.g. for PATCH_LEVEL */
#include "tarstamp.h"
/* (generated) configuration for this build */
#include "config.h"
```

in all source files as the first code line in the beginning.

... thinking of this again, tarstamp.h may only be used in config.h if it is manually created (like in build_windows/config.h.in), so we should just include it there and in the places using it (without checking: libcob/common.c, cobc/cobc.c, cobc/codegen.c); I'll do this change to the 3.x branch "soon".

Further: actually config.h should only be used if `HAVE_CONFIG_H` is defined "to be clean" but I think it isn't reasonable to expect anyone to pass that much defines on the command line, so plan to ignore this.

lefessan:
> I'd prefer the more easy one to just have in all source files as the first code line in the beginning.

I am not fond of such "informal convention", that may lead to bugs the next time somebody creates a new source file. Just during this PR, I had to debug weird behaviors twice, because of the position of `config.h` before or after other include files.

Another idea: create a header file `cob-config.h` that contains:
```
#ifndef COB_CONFIG_H
#ifndef COBCRUN_NAME
#ifndef COB_TAR_DATE
#include "tarstamp.h"
#endif
#include "config.h"
#endif
#endif
```

and have it included from all other include files, and replace `config.h` in source files.

GitMensch:
... but even then this cob-config.h must either be included everywhere or included in coblocal.h and cobc.h, no? In this case it would be better placed into those files directly.

lefessan:
I simplified this PR to only fix the `--save-temps=DIR` behavior, and export `cob_get_strerror` in `coblocal.h`.

For the `config.h`, I am not sure that adding the check in `cobc.h` would be ok, as this file is usually included only very late in files, while `config.h` must be the first included files (even before system includes).

I was also surprised that the configuration is not exported during installation, I would have expected `libcob/common.h` to include `config.h` and have `config.h` found there too (it might be useful to check the `config.h` after installation to see what flags where used).

Anyway, I propose to have this discussion in another PR, maybe for 4.x, when we start targetting it for next changes. By the way, is there a thread somewhere on how the switch between 3.x to 4.x is organized ?

GitMensch (approved):
that's fine now; and as my work on svn finally started (with a lot of delay) this should soon be able to actually be included upstream

PR #128, opened 2023-12-11
Fix testsuite 8 on Windows/MSYS2

ddeclerck:
EDIT: fixes for 808 and 809 already merged

This fixes tests number 8, 808 and 809 on Windows with MSYS2.

Test 8 fails randomly during the call to gcc, as it tries to access the temporary directory.
Since the objective of this test is to check the compilation behavior when tweaking TMP/TEMP/TMPDIR, I resorted to a workaround: I simply replaced $COMPILE by $COMPILE_ONLY.

As for tests 808 and 809, they fail because of differences in the way paths are shown on Windows VS Unix. Since the expected output uses relative Unix-style paths, I used sed to pre-process the output to remove any absolue Windows-style prefix and replace them with ./.

The only remaining failing test, 771, is a bit more tricky ; it will be handled separately.

### DIFF
--- a/tests/testsuite.src/used_binaries.at
+++ b/tests/testsuite.src/used_binaries.at
@@ -396,11 +396,16 @@ AT_DATA([prog.cob], [
            EXIT PROGRAM.
 ])
 
-AT_CHECK([TMPDIR="" TMP="notthere" TEMP="" $COMPILE prog.cob], [0], [],
+# Note: on Windows, GCC/MinGW randomly fails when it can't access the temporary directory,
+# so we only perform the COBOL-to-C translation. This is sufficient since the error we want
+# to check is a compile-time error. Note that we nevertheless perform a full compilation
+# afterwards, without tweaking the temporary, so that we can test the actual execution.
+AT_CHECK([TMPDIR="" TMP="notthere" TEMP="" $COMPILE_ONLY prog.cob], [0], [],
 [libcob: warning: Temporary directory TMP is invalid, adjust TMPDIR!
 ])
+AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [OK], [])
-AT_CHECK([TMPDIR="" TMP="" TEMP="./prog.cob" $COMPILE prog.cob], [0], [],
+AT_CHECK([TMPDIR="" TMP="" TEMP="./prog.cob" $COMPILE_ONLY prog.cob], [0], [],
 [libcob: warning: Temporary directory TEMP is invalid, adjust TMPDIR!
 ])
 # TMPDIR is only checked when actually needed which is currently only the case


tests/testsuite.src/run_misc.at
```diff
@@ -10921,7 +10919,8 @@ END OF DUMP - prog
 
 # AT_DATA workaround via sed:
 AT_CHECK([$SED -e 's/_$//' reference_fdls_tmpl > reference], [0], [], [])
-AT_CHECK([$SED -e 's/compiled ... .. .... ..:..:../compiled MMM DD YYYY HH:MM:SS/g' \
+AT_CHECK([$SED -e 's/compiled ... .. .... ..:..:../compiled MMM DD YYYY HH:MM:SS/g;
+                   s/Started by [[^[:cntrl:]]]*prog_fdls.exe/Started by .\/prog_fdls/g' \
```
GitMensch, line 10923: `[[^[:cntrl:]]]` is likely not portable, but at least here and below we can use `.*`, too.

tests/testsuite.src/used_binaries.at
```diff
@@ -387,11 +387,12 @@ AT_DATA([prog.cob], [
            EXIT PROGRAM.
 ])
 
-AT_CHECK([TMPDIR="" TMP="notthere" TEMP="" $COMPILE prog.cob], [0], [],
```
GitMensch, line 390: Please add a short comment above why we only use `COMPILE_ONLY` (maybe move that to the more complicated testcase to postpone, we actually do want to check that the C compile and link phase does work, here).

GitMensch (changes requested):
Thank you for tackling this.
Adjusting a "known format" is reasonable here (but at least above the first one: please add a note about the MSYS behavior of calling with full-path, which we need to remove here.

The replacement with `[[^[:cntrl:]]]` needs an adjustment, that won't work on AIX or Solaris outside of gsed.

tests/testsuite.src/run_misc.at
```diff
@@ -10820,30 +10813,35 @@ libcob: cpyabrt:4: warning: implicit CLOSE of FLATFILE ('RELFIX')
...
-	MAIN-1 at prog.cob:177
-	ENTRY sub1 at prog.cob:159
- Last statement of "sub2" was CALL
-	DO-CALL OF SubwaY at sub2.cob:48
-	ENTRY sub2 at sub2.cob:39
- Last statement of "prog" was CALL
-	CALL-IT-OMIT at prog.cob:118
-	MAIN-100 at prog.cob:85
-	ENTRY prog at prog.cob:77
- Started by ./prog
```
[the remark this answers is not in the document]
ddeclerck: Note sure what you mean here. We still compare the stack as before, except we do this through the intermediate file stderr.txt (check the redirection).

tests/testsuite.src/used_binaries.at
ddeclerck: > we actually do want to check that the C compile and link phase does work, here

That's why after this "compile-only" check, we also perform a regular compile (check line 393).

tests/testsuite.src/run_misc.at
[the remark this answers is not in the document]
GitMensch: When inspecting the diff at https://github.com/OCamlPro/gnucobol/pull/128/files I only see that we check the stacktrace for "started by" - all other elements seem to not be checked - but that was a wong impression, when checking the file it is clear that this is fine.

GitMensch:
> You suggested two solutions :
> 
>     * using `awk '{sub(/Started by [[:print:]]*prog.exe/, "Started by .\/prog")}1'`
>     * or just replacing `[[^[:cntrl:]]]` by `.*`
> 
> Which one do you like most ?

I don't mind. If we do use awk for that in the testsuite then this may be better (note: I haven't checked the portability of that and if it works!), otherwise I'd opt for the "more simple sed".

ddeclerck:
I went for the simpler `sed`, since that's what's already used in the testsuite.

GitMensch:
There are two generated Makefiles added in this PR, those should be dropped.

The changes in run_misc.at are ready for svn, not sure about the other as we explicit adjust variables to let the C compiler work fine, and this call is dropped by `COMPILE_ONLY`.

ddeclerck:
> There are two generated Makefiles added in this PR, those should be dropped.

Indeed, I guess I committed too quickly.

> The changes in run_misc.at are ready for svn, not sure about the other as we explicit adjust variables to let the C compiler work fine, and this call is dropped by `COMPILE_ONLY`.

Well, GCC seems to expect TMPDIR to point to some valid location. Under those circumstances, shouldnt an invalid TMPDIR be considered as an error instead of a mere warning ?

GitMensch:
At least with my installation setting `TMPDIR=/banana` doesn't raise any error with GCC.

ddeclerck:
Using GCC 13.2 under MSYS2/UCRT, seems it in fact only considers the `TMP` variable (while GCC docs only mentions `TMPDIR`).
`TMP=/banana gcc prog.c` systematically fails for me (works fine under Linux though).

GitMensch:
If `TMP` is unset, then it uses `TEMP`, if that's unset then it may use the userprofile...

I'm OK in general to adjust GnuCOBOL's behavior but a broken TMPDIR, but I'm not sure if aborting in common.c would be good...

Note that the _real_ issue is that the changed variable is _not_ correctly exported, this can break a bunch of things for other tests (in the future), too. Ideally we can tell the runtime environment (which seems to be some MSYS here) to correctly export this variable (and in the future: others).

ddeclerck:
> Note that the _real_ issue is that the changed variable is _not_ correctly exported, this can break a bunch of things for other tests (in the future), too. Ideally we can tell the runtime environment (which seems to be some MSYS here) to correctly export this variable (and in the future: others).

I'm not sure to understand. To you suggest this is a bug in the MSYS shell ?

GitMensch:
"Not a bug but a feature" - several environment variables are converted between Windows and Posix path, some are just not taken over (and it _seems_ that some variables will have the values which they had when the MSYS process was started).

Environment variables are also related C runtime used; when it "switches" you may see "old" values; though this is mostly an issue within a single process, calling a new one normally provides a "clear start".
If you want to see that then you may be able to check the chain of processes and the environment variables set, for example with process explorer (would need the compiler process to "wait" instead of end, but _maybe_ one could just create a gcc.cmd and have that in PATH before gcc, which then does a `set /p var=starting...` before the gcc (by full path) call.

Given that - I'm not sure how this should be tackled - as noted an error may be fine, but in this case cobc would have to pass the instruction "abort on error" (maybe a separate entry function?) when getting the path by libcob (or duplicate that code into cobc and adjust it there).

GitMensch:
If that's not too much to ask: yes, please do, this should get the false positives down to one.

Thank you!

ddeclerck:
Done !

GitMensch:
Can you please rebase and adjust the title?

ddeclerck:
That's been done.

tests/testsuite.src/used_binaries.at
```diff
@@ -396,11 +396,16 @@ AT_DATA([prog.cob], [
...
+# Note: on Windows, GCC/MinGW randomly fails when it can't access the temporary directory,
+# so we only perform the COBOL-to-C translation. This is sufficient since the error we want
+# to check is a compile-time error. Note that we nevertheless perform a full compilation
+# afterwards, without tweaking the temporary, so that we can test the actual execution.
+AT_CHECK([TMPDIR="" TMP="notthere" TEMP="" $COMPILE_ONLY prog.cob], [0], [],
 [libcob: warning: Temporary directory TMP is invalid, adjust TMPDIR!
 ])
+AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [OK], [])
```
GitMensch, line 407: The only point of testing the execution was to verify that this does work - without the recompile, because we adjust the broken environment before calling the C compiler.

The goal is good, but if we don't find a way to ensure that cobc's `setenv` reaches the program envrionment of the called C compiler, I'd rather let this fail "from time to time" instead of globally accepting a failure here.

I'd suggest to move this to a draft until someone finds the way to tweak this "correctly" (I _guess_ that means setting some environment variable; it would likely be reasonable to ask the MSYS2 folks (possibly by creating an issue in their tracker - if this happens under MSYS2).

Issue #125, opened 2023-10-28
Test 772 fails on Windows


lefessan:
From https://github.com/OCamlPro/gnucobol/actions/runs/6675975179?pr=109

```
772. run_misc.at:4507: testing direct CALL in from C w/wo error; no exit ...
../../tests/run_misc.at:4618: $COMPILE caller.c
../../tests/run_misc.at:4619: $COMPILE_MODULE callee.cob callee2.cob buggy.cob
../../tests/run_misc.at:4620: $COBCRUN_DIRECT ./caller callee 00
../../tests/run_misc.at:4623: $COBCRUN_DIRECT ./caller callee 42
../../tests/run_misc.at:4626: $COBCRUN_DIRECT ./caller callee2
--- -   2023-10-28 09:13:49.613224400 +0000
+++ /d/a/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/772/stderr      2023-10-28 09:13:49.548724300 +0000
@@ -1,2 +1 @@
-note: STOP RUN with return code 2
 
--- -   2023-10-28 09:13:49.690763800 +0000
+++ /d/a/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/772/stdout      2023-10-28 09:13:49.642475500 +0000
@@ -1 +1 @@
-STOP WITH 2
+
../../tests/run_misc.at:4626: exit code was 127, expected 0
772. run_misc.at:4507: 772. direct CALL in from C w/wo error; no exit (run_misc.at:4507): FAILED (run_misc.at:4626)
```

ddeclerck:
I stumbled on that one in #116 . Copy-pasting my analysis.

The failure is random, sometimes the test is OK, sometimes it fails.
The problem occurs specifically with STOP RUN (replacing by an EXIT program gives no error). Also, it does NOT occur if we define COB_WITHOUT_JMP. Investigating a bit, it seems this is because the COBOL module executing the STOP RUN statement is unloaded (with lt_dlclose) before cob_stop_run has finished its execution. This might be okay when calling exit, but when using longjmp, this probably messes up the stack frame.

GitMensch:
Instead of compiling with `COB_WITHOUT_JMP` it likely would also work to just disable the `dlclose()` by using `COB_PHYSICAL_CANCEL=never $COBCRUN_DIRECT ./caller callee2` (inline, only for that run), right?
[...]

PR #129, opened 2023-12-13
Fix cob_call_with_exception_check (test 771)

ddeclerck:


GitMensch (changes requested):
Is there a reason to skip the complete termination?
I _think_ we should do the screenio cleanup - but only if that was done after the call to `cob_call_with_exception_check()`.
We _very likely_ need to call the exit handlers, as those can be COBOL programs which need the state "at exit", not any previous state we longjmp to. But then those should not be executed after the longjmp.

We possibly should add some additional testcases that combine `cob_call_with_exception_check()` with the scenario "exception and exit handlers setup".

ddeclerck:
> Is there a reason to skip the complete termination?

Well, in #125, you suggest :
> I'd prefer the second - and document that modules will only be unloaded with this function if after the call a manual call to `cob_tidy()` is done.

Since `cob_tidy` calls `call_exit_handlers_and_terminate`, it seemed natural to just skip the whole termination.

Anyways, I pushed an improved version, where only the last bits of termination are delayed, ie only the following calls :
```
	cob_exit_call ();
	cob_exit_common ();
```

GitMensch (approved):
That's very clean. Please add a changelog entry, then this is good for svn (sight, I won't finish the parts to make svn ready for more commits until moid/end next week).

PR #132, opened 2024-01-25
Add support for collating sequences on indexed files

ddeclerck:
This PR adds support for COLLATING SEQUENCE on indexed files, as mentionned in [https://sourceforge.net/p/gnucobol/feature-requests/459/](https://sourceforge.net/p/gnucobol/feature-requests/459/).

Note this has. only been done for the BDB backend - that's what our client needs.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,9 @@ NEWS - user visible changes				-*- outline -*-
    the error output for format errors (for example invalid indicator column)
    is now limitted to 5 per source file
 
+** support the COLLATING SEQUENCE clause on indexed files
+   (currently only with the BDB backend)
+
   more work in progress
 
 * Important Bugfixes
@@ -40,6 +43,9 @@ NEWS - user visible changes				-*- outline -*-
    a file for diagnostics; this flag can be activated if your editor and
    build system do not correctly work together to locate files from
    diagnostic output
+
+** New option -fdefault-file-colseq to specify the default
+   file collating sequence
 
 * More notable changes
 
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -90,24 +90,25 @@ enum compile_level {
 	CB_LEVEL_EXECUTABLE	= 7
 };
 
-#define	CB_FLAG_GETOPT_STACK_SIZE       1
-#define	CB_FLAG_GETOPT_IF_CUTOFF        2
-#define	CB_FLAG_GETOPT_SIGN             3
-#define	CB_FLAG_GETOPT_FOLD_COPY        4
-#define	CB_FLAG_GETOPT_FOLD_CALL        5
-#define	CB_FLAG_GETOPT_TTITLE           6
-#define	CB_FLAG_GETOPT_MAX_ERRORS       7
-#define	CB_FLAG_GETOPT_DUMP             8
-#define	CB_FLAG_GETOPT_CALLFH           9
-#define	CB_FLAG_GETOPT_INTRINSICS      10
-#define	CB_FLAG_GETOPT_EC              11
-#define	CB_FLAG_GETOPT_NO_EC           12
-#define	CB_FLAG_GETOPT_NO_DUMP         13
-#define	CB_FLAG_GETOPT_EBCDIC_TABLE    14
-#define	CB_FLAG_GETOPT_DEFAULT_COLSEQ  15
-#define	CB_FLAG_GETOPT_MEMORY_CHECK    16
-#define	CB_FLAG_GETOPT_COPY_FILE       17
-#define	CB_FLAG_GETOPT_INCLUDE_FILE    18
+#define	CB_FLAG_GETOPT_STACK_SIZE            1
+#define	CB_FLAG_GETOPT_IF_CUTOFF             2
+#define	CB_FLAG_GETOPT_SIGN                  3
+#define	CB_FLAG_GETOPT_FOLD_COPY             4
+#define	CB_FLAG_GETOPT_FOLD_CALL             5
+#define	CB_FLAG_GETOPT_TTITLE                6
+#define	CB_FLAG_GETOPT_MAX_ERRORS            7
+#define	CB_FLAG_GETOPT_DUMP                  8
+#define	CB_FLAG_GETOPT_CALLFH                9
+#define	CB_FLAG_GETOPT_INTRINSICS           10
+#define	CB_FLAG_GETOPT_EC                   11
+#define	CB_FLAG_GETOPT_NO_EC                12
+#define	CB_FLAG_GETOPT_NO_DUMP              13
+#define	CB_FLAG_GETOPT_EBCDIC_TABLE         14
+#define	CB_FLAG_GETOPT_DEFAULT_COLSEQ       15
+#define	CB_FLAG_GETOPT_DEFAULT_FILE_COLSEQ  16
+#define	CB_FLAG_GETOPT_MEMORY_CHECK         17
+#define	CB_FLAG_GETOPT_COPY_FILE            18
+#define	CB_FLAG_GETOPT_INCLUDE_FILE         19
 
 
 /* Info display limits */
@@ -3819,6 +3820,13 @@ process_command_line (const int argc, char **argv)
 			}
 			break;
 
+		case CB_FLAG_GETOPT_DEFAULT_FILE_COLSEQ: /* 16 */
+			/* -fdefault-file-colseq=<ASCII/EBCDIC/NATIVE> */
+			if (cb_deciph_default_file_colseq_name (cob_optarg)) {
+				cobc_err_exit (COBC_INV_PAR, "-fdefault-file-colseq");
+			}
+			break;
+
 		case CB_FLAG_GETOPT_FOLD_COPY: /* 4 */
 			/* -ffold-copy=<UPPER/LOWER> : COPY fold case */
 			if (!cb_strcasecmp (cob_optarg, "UPPER")) {
@@ -3898,7 +3906,7 @@ process_command_line (const int argc, char **argv)
 			}
 			break;
 
-		case CB_FLAG_GETOPT_MEMORY_CHECK: /* 16 */
+		case CB_FLAG_GETOPT_MEMORY_CHECK: /* 17 */
 			/* -fmemory-check=<scope> :  */
 			if (!cob_optarg) {
 				cb_flag_memory_check = CB_MEMCHK_ALL;
@@ -3907,7 +3915,7 @@ process_command_line (const int argc, char **argv)
 			}
 			break;
 
-		case CB_FLAG_GETOPT_COPY_FILE: /* 17 */
+		case CB_FLAG_GETOPT_COPY_FILE: /* 18 */
 			/* --copy=<file> : COPY file at beginning */
 			if (strlen (cob_optarg) > (COB_MINI_MAX)) {
 				cobc_err_exit (COBC_INV_PAR, "--copy");
@@ -3916,7 +3924,7 @@ process_command_line (const int argc, char **argv)
 					  cobc_strdup (cob_optarg));
 			break;
 
-		case CB_FLAG_GETOPT_INCLUDE_FILE: /* 18 */
+		case CB_FLAG_GETOPT_INCLUDE_FILE: /* 19 */
 			/* -include=<file.h> : add #include "file.h" to 
 			   generated C file */
 			if (strlen (cob_optarg) > (COB_MINI_MAX)) {
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -9271,6 +9271,39 @@ output_key_components (struct cb_file* f, struct cb_key_component* key_component
 	}
 }
 
+
+static void
+output_indexed_file_key_colseq (const struct cb_file *f, const struct cb_alt_key *ak, int idx)
+{
+	const cb_tree	key = ak ? ak->key : f->key;
+	const cb_tree	key_col = ak ? ak->collating_sequence_key : f->collating_sequence_key;
+	const int	type = cb_tree_type (key, cb_code_field (key));
+	cb_tree		col = NULL;
+
+	/* We only apply a collating sequence if the key is alphanumeric / display */
+	if ((type & COB_TYPE_ALNUM) || (type == COB_TYPE_NUMERIC_DISPLAY)) {
+		col = key_col ? key_col : f->collating_sequence;
+#if 0	/* TODO: this should be done for national, when available */
+	} else if (type & COB_TYPE_NATIONAL) {
+		col = key_col_n ? key_col_n : f->collating_sequence_n;
+#endif
+	}
+
+	output_prefix ();
+	if (idx == 0) {
+		output ("%s%s->collating_sequence = ", CB_PREFIX_KEYS, f->cname);
+	} else {
+		output ("(%s%s + %d)->collating_sequence = ", CB_PREFIX_KEYS, f->cname, idx);
+	}
+	if (col != NULL && CB_REFERENCE_P (col)) {
+		output_param (cb_ref(col), -1);
+		output (";");
+	} else {
+		output ("NULL;");
+	}
+	output_newline ();
+}
+
 static void
 output_file_initialization (struct cb_file *f)
 {
@@ -9331,6 +9364,9 @@ output_file_initialization (struct cb_file *f)
 		} else {
 			output_line ("%s%s->offset = 0;", CB_PREFIX_KEYS, f->cname);
 		}
+		if (f->organization == COB_ORG_INDEXED) {
+			output_indexed_file_key_colseq (f, NULL, 0);
+		}
 		nkeys = 1;
 		for (l = f->alt_key_list; l; l = l->next) {
 			output_prefix ();
@@ -9352,6 +9388,9 @@ output_file_initialization (struct cb_file *f)
 				output_line ("(%s%s + %d)->offset = 0;", CB_PREFIX_KEYS,
 					f->cname, nkeys);
 				output_key_components (f, l->component_list, nkeys);
+			}
+			if (f->organization == COB_ORG_INDEXED) {
+				output_indexed_file_key_colseq (f, l, nkeys);
 			}
 			nkeys++;
 		}
--- a/cobc/parser.y
+++ b/cobc/parser.y
@@ -336,29 +336,6 @@ check_non_area_a (cb_tree stmt) {
 
 /* Collating sequences */
 
-/* Known collating sequences/alphabets */
-enum cb_colseq {
-	CB_COLSEQ_NATIVE,
-	CB_COLSEQ_ASCII,
-	CB_COLSEQ_EBCDIC,
-};
-enum cb_colseq cb_default_colseq = CB_COLSEQ_NATIVE;
-
-/* Decipher character conversion table names */
-int cb_deciph_default_colseq_name (const char * const name)
-{
-	if (!cb_strcasecmp (name, "ASCII")) {
-		cb_default_colseq = CB_COLSEQ_ASCII;
-	} else if (!cb_strcasecmp (name, "EBCDIC")) {
-		cb_default_colseq = CB_COLSEQ_EBCDIC;
-	} else if (!cb_strcasecmp (name, "NATIVE")) {
-		cb_default_colseq = CB_COLSEQ_NATIVE;
-	} else {
-		return 1;
-	}
-	return 0;
-}
-
 static cb_tree
 build_colseq_tree (const char *alphabet_name,
 		      int alphabet_type,
@@ -901,21 +878,32 @@ check_relaxed_syntax (const cob_flags_t lev)
 }
 
 static void
-setup_default_collation (struct cb_program *program) {
-	switch (cb_default_colseq) {
+prepare_default_collation (enum cb_colseq colseq) {
+	switch (colseq) {
 #ifdef COB_EBCDIC_MACHINE
 	case CB_COLSEQ_ASCII:
 #else
 	case CB_COLSEQ_EBCDIC:
 #endif
-		alphanumeric_collation = build_colseq (cb_default_colseq);
+		alphanumeric_collation = build_colseq (colseq);
 		break;
 	default:
 		alphanumeric_collation = NULL;
 	}
 	national_collation = NULL; /* TODO: default national collation */
+}
+
+static void
+setup_default_collation (struct cb_program *program) {
+	prepare_default_collation (cb_default_colseq);
 	program->collating_sequence = alphanumeric_collation;
 	program->collating_sequence_n = national_collation;
+}
+
+static void
+setup_default_file_collation (struct cb_file *file) {
+	prepare_default_collation (cb_default_file_colseq);
+	file->collating_sequence = alphanumeric_collation;
 }
 
 static void
@@ -5365,6 +5353,7 @@ file_control_entry:
 
 	}
 	key_type = NO_KEY;
+        setup_default_file_collation (current_file);
   }
   _select_clauses_or_error
   {
@@ -5752,7 +5741,7 @@ collating_sequence_clause:
 	check_repeated ("COLLATING", SYN_CLAUSE_3, &check_duplicate);
 	current_file->collating_sequence = alphanumeric_collation;
 	current_file->collating_sequence_n = national_collation;
-	CB_PENDING ("FILE COLLATING SEQUENCE");
+	CB_UNFINISHED ("FILE COLLATING SEQUENCE"); /* only implemented for BDB */
   }
 ;
 
@@ -5804,7 +5793,7 @@ collating_sequence_clause_key:
 	   and also attached to the correct key later, so just store in a list here: */
 	current_file->collating_sequence_keys =
 		cb_list_add(current_file->collating_sequence_keys, CB_BUILD_PAIR ($6, $4));
-	CB_PENDING ("KEY COLLATING SEQUENCE");
+	CB_UNFINISHED ("KEY COLLATING SEQUENCE"); /* only implemented for BDB */
   }
 ;
 
--- a/libcob/coblocal.h
+++ b/libcob/coblocal.h
@@ -570,6 +570,9 @@ cob_max_int (const int x, const int y)
 	return y;
 }
 
+COB_HIDDEN int		cob_cmps	(const unsigned char *, const unsigned char *,
+					 const size_t, const unsigned char *);
+
 #undef	COB_HIDDEN
 
 #endif	/* COB_LOCAL_H */
--- a/libcob/common.h
+++ b/libcob/common.h
@@ -1370,9 +1370,7 @@ typedef struct __cob_file_key {
 	unsigned int	offset;			/* Offset of field */
 	int		count_components;		/* 0..1::simple-key  2..n::split-key   */
 	cob_field	*component[COB_MAX_KEYCOMP];	/* key-components iff split-key   */
-#if 0	/* TODO (for file keys, not for SORT/MERGE) */
-	const unsigned char *collating_sequence;	/* COLLATING */
-#endif
+	const unsigned char *collating_sequence;	/* COLLATING (for file keys, not for SORT/MERGE) */
 } cob_file_key;
 
 
--- a/libcob/fileio.c
+++ b/libcob/fileio.c
@@ -689,6 +689,16 @@ static unsigned int	bdb_lock_id = 0;
 	key.data = fld->data;			\
 	key.size = (cob_dbtsize_t) fld->size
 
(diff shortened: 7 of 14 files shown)


ddeclerck:
Hmm, that broke the NIST testsuite.
Investigating further, it seems we should in fact NOT apply the program collating sequence to indexed files that do not have an explicit collating sequence (as opposed to SORT/MERGE files).

See 12.3.5.3 in ISO/IEC 1989:2014 :
```quote
11) The alphanumeric program collating sequence and national program collating sequence are used to determine
the truth value of any alphanumeric comparisons and national comparisons, respectively, that are:
a)Explicitly specified in relation conditions.
b)Explicitly specified in condition-name conditions.
c)Implicitly specified by the presence of a CONTROL clause in a report description entry.
When alphabet-name-1 or alphabet-name-2, or both, is associated with a locale, locale category LC_COLLATE
is used to carry out these comparisons.

12) The alphanumeric program collating sequence and national program collating sequence explicitly or implicitly
established by the OBJECT-COMPUTER paragraph are effective with the initial state of the runtime modules
to which they apply. If alphabet-name-1 or alphabet-name-2 references a locale, the associated collating
sequence is defined by category LC_COLLATE in the specific locale associated with that alphabet-name or, if
none, in the locale current at the time the collating sequence is used at runtime

13) The alphanumeric program collating sequence and national program collating sequence are applied to
alphanumeric and national sort or merge keys, respectively, unless the sort or merge collating sequence has
been modified by execution of a SET statement or a COLLATING SEQUENCE phrase is specified in the
respective SORT or MERGE statement.```

**Comment by GitMensch** [comment]
Found the answer in 12.4.4.3:

> If no COLLATING SEQUENCE clause is specified:
a) the  collating  sequence  for  alphanumeric  record  keys,  both  primary  and  alternate,  is  the  native alphanumeric collating sequence;
b) the collating sequence for national record keys, both primary and alternate, is the native national collating sequence.

So this is by default "native", not "program collating".
Please re-adjust, sorry if I was misleading here.

BUT: before adjusting the code, please add a _minimal_ version of the failing NIST test (ideally written from scratch) to our internal testsuite; this should fail before your change-adjustment and pass afterwards.

(and note that this is also one of the fixed file attributes, which in theory should be checked - we don't do up to 4.x, but don't check that there either...).

**Comment by ddeclerck** [comment]
I made the necessary changes. I also added a new `-fdefault-file-colseq` flags that works similarly to `-fdefault-colseq` but just for files.

As for the tests, I reworked them and included a dummy program collating sequence (taken from the failed NIST test), which was sufficient to cause the testsuite to fail priori to fixing the bug.

**Comment by GitMensch** [comment]
That sounds all good!

Please check the test coverage of your changes, the codecov bot says:

> Attention: 188 lines in your changes are missing coverage. Please review.

188 sounds a bit much.

File: cobc/parser.y [hunk 1]
```diff
@@ -5752,7 +5776,6 @@ collating_sequence_clause:
 	check_repeated ("COLLATING", SYN_CLAUSE_3, &check_duplicate);
 	current_file->collating_sequence = alphanumeric_collation;
 	current_file->collating_sequence_n = national_collation;
-	CB_PENDING ("FILE COLLATING SEQUENCE");
```
GitMensch, line 5755: This should be changed to `CB_UNFINISHED` - it only works for BDB (LMDB has this function internal only, D-ISAM and likely C-ISAM have no option for passing that from the outside (D-ISAM has `ISCUSTKEYS`, but that is only available with licensed versions built from source with custom compare function added), V*ISAM has the extension `issetcollate()` - but so far the passed collation is stored and then never used).

Please also add missing pieces for `collating_sequence_keys` - the `#ifdef 0` can be removed from `cob_file_key` so that you have the matching field (it is only allocated in libcob); the code in fileio.c to set the key into app_data would first check the key, then the file. The most work here is in the compiler and also those should be not that much..

ddeclerck:
> That sounds all good!
> 
> Please check the test coverage of your changes, the codecov bot says:
> 
> > Attention: 188 lines in your changes are missing coverage. Please review.
> 
> 188 sounds a bit much.

Fixed ! Just had to resynchronize with upstream.
The remaining uncovered lines are either benign or were not covered anyways.

libcob/coblocal.h
```diff
@@ -570,6 +570,9 @@ cob_max_int (const int x, const int y)
 	return y;
 }
 
+COB_HIDDEN int		common_cmps	(const unsigned char *, const unsigned char *,
```
GitMensch, line 573: so far all `COB_HIDDEN` have a `cob_` prefix. please refactor this to be identical

libcob/common.h
```diff
@@ -1395,7 +1395,7 @@ typedef struct __cob_file {
 	cob_file_key		*keys;			/* ISAM/RANDOM/SORT keys */
 	void			*file;			/* File specific pointer */
 	void			*linorkeyptr;		/* LINAGE or SPLIT KEY */
-	const unsigned char	*sort_collating;	/* SORT collating / CODE-SET (used for RE-/WRITE) */
+	const unsigned char	*sort_collating;	/* SORT collating */
```
GitMensch, line 1398: please adjust comment as this is now also the FILE collation for INDEXED files

libcob/common.h
```diff
@@ -1432,6 +1432,7 @@ typedef struct __cob_file {
 	char				*nxt_filename;	/* Next position in org_filename */
 	unsigned int		flag_is_concat:1;	/* SEQUENTIAL concatenated file names */
 
+	const unsigned char* code_set_write;	/* CODE-SET conversion for RE-/WRITE) */
```
GitMensch, line 1435: If you want to do this splitting - which seems reasonable - then you need to check the module version and "move" the field for GC 3.2 generated modules, as those will have used the field above. Also you'd have to move it at the end of the `cob_file`, as hinted in the doc-comment for `__cob_file`.

Note: If you _always_ place the key collation into _each_ `cob_key`, then you don't need to have the file-collation in `cob_file` at all (if set then those will never be empty).

libcob/fileio.c
```diff
@@ -5429,6 +5464,7 @@ indexed_read_next (cob_file *f, const int read_opts)
 		} else {
 			p->key.size = (cob_dbtsize_t) bdb_keylen(f,p->key_index);
 			p->key.data = p->last_readkey[p->key_index];
+			bdb_setkeycol (f, p->key_index);
```
GitMensch, line 5467: when placing this into each `cob_key` you can direclty do `p->key.app_data = f->keys[p->key_index]->collation` here

tests/testsuite.src/run_file.at
```diff
@@ -12429,6 +12429,241 @@ AT_CHECK([diff reference prog.out], [0], [], [])
 AT_CLEANUP
 
 
+AT_SETUP([INDEXED file manipulation under ASCII/EBCDIC collation])
+AT_KEYWORDS([runfile WRITE DELETE READ EBCDIC])
+
+AT_SKIP_IF([test "$COB_HAS_ISAM" = "no"])
```
GitMensch, line 12435: Change to `AT_SKIP_IF([test "$COB_HAS_ISAM" != "db"])` and add a comment that this is only included for the BDB backend so far.

GitMensch (changes requested):
There are some general changes (should not be too much work) and one change to `cob_key` that actually reduces the amount of changes (here the important part again):

> Note: If you _always_ place the key collation into _each_ `cob_key`, then you don't need to have the file-collation in `cob_file` at all (if set then those will never be empty).**

I'd like to also have support for key collation - as the runtime already handles that (just specifying always the same collation), which is possibly a bit more work, if you feel that's too much, then this part could be postponed.

libcob/fileio.c
```diff
@@ -4150,6 +4164,23 @@ indexed_file_delete (cob_file *f, const char *filename)
 #endif
 }
 
+static int
+bdb_bt_compare(DB *db, const DBT *k1, const DBT *k2)
+{
+	int res = 0;
+	const unsigned char *col = (const unsigned char *)k1->app_data;
```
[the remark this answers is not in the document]
ddeclerck: Done using a pair of macros, conditionally defined depending on BDB version.

libcob/common.h
ddeclerck: Actually canceled this change, since the file collation is now directly stored in the keys.

ddeclerck:
Should be good. Though I left the key collating sequence for later (although I enabled the field to store the file collating sequence).

NEWS
```diff
@@ -28,6 +28,8 @@ NEWS - user visible changes				-*- outline -*-
    build system do not correctly work together to locate files from
    diagnostic output
 
+** New option -fdefault-file-colseq to specify default file collating sequence
```
GitMensch, line 31: add  the file COLLATING SEQUENCE in general under "new features" (with the note "currently only with the BDB backend)

libcob/fileio.c
```diff
@@ -757,6 +767,22 @@ bdb_savekey (cob_file *f, unsigned char *keyarea, unsigned char *record, int idx
...
+static void
+bdb_setkeycol (cob_file *f, int idx)
+{
+	struct indexed_file	*p = f->file;
+	const cob_field	*fld = f->keys[idx].field;
+	/* We only apply a collating sequence if the key is alphanumeric */
+	if (COB_FIELD_IS_ANY_ALNUM(fld) || COB_FIELD_IS_NUMDISP(fld)) {
+		DBT_SET_APP_DATA(&p->key, (void *)f->keys[idx].collating_sequence);
+	/* TODO: this should be done for national, when available *//*
+	} else if (COB_FIELD_IS_NATIONAL(fld)) {
```
GitMensch, line 779: better use `#if 0 /* TODO ddd */` here, this leads to a better highlighting and less"that's a strange code here" analyzer warning

libcob/fileio.c
```diff
@@ -757,6 +767,22 @@ bdb_savekey (cob_file *f, unsigned char *keyarea, unsigned char *record, int idx
...
+		DBT_SET_APP_DATA(&p->key, (void *)f->sort_collating_n);*/
+	} else {
+		DBT_SET_APP_DATA(&p->key, NULL);
+	}
+}
+
```
GitMensch, line 785: The complete check part here should not be done each time, but in the codegen, no?

libcob/fileio.c
```diff
@@ -4150,6 +4179,23 @@ indexed_file_delete (cob_file *f, const char *filename)
 #endif
 }
 
+static int
+bdb_bt_compare(DB *db, const DBT *k1, const DBT *k2)
+{
+	int res = 0;
+	const unsigned char *col = (const unsigned char *)DBT_GET_APP_DATA(k1);
+	if (col == NULL) {
```
GitMensch, line 4187: Is there a reason to register `bdb_bt_compare` down there if `col == NULL`? I _think_ this check should be done outside.
Also: is it possible in our code that both DBT can have a different size? At least in the sort comparison they are always identical (and it would be a status 39 error if they wouldn't be here, no)?

GitMensch (changes requested):
looks like we're nearly done; I think this can be checked in next week (and _maybe_ there is a key collation available until then, too)

libcob/fileio.c
ddeclerck: I guess if no keys for the current file have a collation, then we don't need to register `bdb_bt_compare` indeed.

GitMensch: I've thought this was registered per key; in this case if the key has no collation, then we don't need to register, and can get rid of the `col == NULL` check, otherwise it needs to stay in and we only don't register if there's no collation at all.

ddeclerck: Oh, wait, you're right, the registration is indeed per key. So yeah, we can simplify the code further and register the comparison function only if the key has a collating sequence.

GitMensch: That's part of the beauty of code-reviews - the code gets improved even before it gets live :-)
Wait, who reviews my code...

GitMensch:
Note: I do have a bigger refactoring of the BDB code which I want to push start of next week, should I wait for this PR to be finished and merged before or would you possibly even prefer if I push the refactoring before?
(I just wanted to fix a bug, recognized that for doing so I should backport status 02 complex; then got from one thing to the next... and then did a refactoring, using the specific `f->keys[idx]` instead of using an index of a subfield and the same for the contained keydesc - the refactor got so big that I should push that separately in any case [the full code of course has a bug _somewhere_ and hides from me])

ddeclerck:
> Note: I do have a bigger refactoring of the BDB code which I want to push start of next week, should I wait for this PR to be finished and merged before or would you possibly even prefer if I push the refactoring before?

I plan to make the final changes to this PR on monday ; if that prevents you from merging, I don't mind rebasing my PR on your changes.

libcob/fileio.c
ddeclerck: > Also: is it possible in our code that both DBT can have a different size? At least in the sort comparison they are always identical (and it would be a status 39 error if they wouldn't be here, no)?

Yeah, they should always be of identical length in our context (and I made a quick check by adding an assert and running both NIST and the testsuite with no error). Anyway I prefered to handle the case anyways (the comparison function is rather "low-level" and does not really care about our use case).

ddeclerck:
Made the last changes, though there might be another one to do.
I'm concerned about the memcmp here (fileio.c: 5679) :
```
	/* Check record key */
	bdb_setkey (f, 0);
	if (!p->last_key) {
		p->last_key = cob_malloc ((size_t)p->maxkeylen);
	} else
	if (f->access_mode == COB_ACCESS_SEQUENTIAL
	 && memcmp (p->last_key, p->key.data, (size_t)p->key.size) > 0) {
		return COB_STATUS_21_KEY_INVALID;
	}
	memcpy (p->last_key, p->key.data, (size_t)p->key.size);
```
All other memcmp's on keys are fin because they just test (in)equality with 0, but this one tests for `> 0`. I guess I should replace this with a collation-aware compare, right ?

GitMensch:
I have a strange feeling now... Can you please verify that:

* the testsuite has at least one entry with variable length keys (which test case is that?) for each SEARCH and INDEXED
* if missing please add it
* check if we do store the variable length in the key fields

I agree that a memcpy should only be used for testing equality or difference, otherwise either a numeric compare (if key field is numeric - the byte order could be non-native!) or a collation aware check needs to be used

ddeclerck:
> That was my feeling. Please address the numeric key sorting, if possible.

I see two ways to fix that :
- either modifiny the `bdb_bt_compare` function - but that would fix only for BDB...
- or always store integer numeric keys in big-endian order (not sure how to deal with IEE754...), but then if I actually translate the keys so thay end up stored in the right order, I might as well do that for alphanumeric keys, making this PR les relevant...

In any case, I believe this require some thinking before actually doing that, and I'm not sure this belongs to this PR (as it did not introduce that problem).

> I _think_ this should work (and hopefully be tested) in SEARCH already.

Tested, seems to work fine for SEARCH.

> As both can use the same comparison logic (just using a different set of collations), I think it's useful to test that.

Could you elaborate ?

> Testing variable length keys belong to this PR if previously possibly completely handled by BDB and now in our own search function.

I'm not sure to understand ; could you rephrase that ?

GitMensch:
> * either modifying the `bdb_bt_compare` function - but that would fix only for BDB...

that would be enough for now and is likely "the most portable" solution. But that creates a HUGE problem: if we do that, then the files need to be rebuilt :-/  Note that using the key type and use the appropriate COBOL sort function (if the field structure within ´->app_data` [if not done: store a complete cb_field structure in there], then you only need to duplicate that cob_field to then set the two separate fields' data) should directly work for all types. This is effectively what is done for `SEARCH` (and the internal comparison during `SORT`).

Otherwise we could have small-enough numbers be converted to big-endian binary and bigger ones to be converted to USAGE DISPLAY SIGN SEPARATE LEADING in `bdb_savekey`, that would also work - but again leads to all files needing a rebuild. As this repacking/unpacking would be done on any key access, this is likely slower than the approach above.

As LMDB does not provide an external compare function, the second solution seems to be the only one that could be applied there (but note that LMDB exists only in trunk and even then is marked as experimental and possibly will be dropped).

Note: with ISAM there are key types that handle this (including IEE754), but we currently only use `CHARTYPE`, so the same problem seems to exists there. For this part I'll contact Ron.

> In any case, I believe this require some thinking before actually doing that, and I'm not sure this belongs to this PR (as it did not introduce that problem).

Hm. Please add a test with expected failure and a comment about the underlying problem.
We could otherwise leave that general problem out of this PR, but if a key is purely numeric, there should be no "additional" conversion applied which makes the current scenario worse.
That brings us to the good thing: if you implement key collation within the cobc site (you already have the codegen and use of that in libcob) then you can place a check there for a numeric type and just skip generating the collation for that.

>> As both can use the same comparison logic (just using a different set of collations), I think it's useful to test that.
> 
> Could you elaborate ?

`SEARCH`/`SORT` already compares "as cob_field" for any numeric type; if the same is applied to the INDEXED comparisons (you need a `cob_field` in ´app_data` for that), then everything "just works".
But as we noted: this may be kept to a separate PR, _especially_ as we have to consider what to do with the existing files (likely we'd have to insert a marker into the file itself, likely by an embedded table, where we record "that's type 2, use different search algorithm"). This is definitely useful, but definitely "quite big" (and yes, should be kept separate).

>> Testing variable length keys belong to this PR if previously possibly completely handled by BDB and now in our own search function.
> 
> I'm not sure to understand ; could you rephrase that ?

BDB compares keys. those are setup from `cob_field *component` - if this would match the COBOL field, then it may be variable-length; but as far as I now see those are always fixed-length; so the issue is again "likely important, but likely big and belongs to another PR".

So... do you add key collations in cobc now and skip to set the collation if the field is numeric (potentially also skip if it is NATIONAL)? That should provide enough for the original issue.

> Yeah, they should always be of identical length in our context (and I made a quick check by adding an assert and running both NIST and the testsuite with no error). Anyway I prefered to handle the case anyways (the comparison function is rather "low-level" and does not really care about our use case).

The new indexed_compare _is_ our specialized function to _our_ case - and we cannot even support two different sizes as we don't have it when the specialized function is used.
Therefore: please adjust `indexed_compare()` to take only a single size, which then also removes the need to use `cob_min_int()`.

For the runtime warning - I think that should be an error, but we don't expect it to be seen by a user so remove the translation part and add a coverage exclusion, like

```c
			/* LCOV_EXCL_START */
			if (k++ == MAX_MODULE_ITERS) {	/* prevent endless loop in case of broken list */
				/* not translated as highly unexpected */
				cob_runtime_warning ("max module iterations exceeded, possible broken chain");
				break;
			}
			/* LCOV_EXCL_STOP */
```

cobc/cobc.c
```diff
@@ -90,22 +90,23 @@ enum compile_level {
...
+#define	CB_FLAG_GETOPT_DUMP                  8
+#define	CB_FLAG_GETOPT_CALLFH                9
+#define	CB_FLAG_GETOPT_INTRINSICS           10
+#define	CB_FLAG_GETOPT_EC                   11
+#define	CB_FLAG_GETOPT_NO_EC                12
+#define	CB_FLAG_GETOPT_NO_DUMP              13
+#define	CB_FLAG_GETOPT_EBCDIC_TABLE         14
+#define	CB_FLAG_GETOPT_DEFAULT_COLSEQ       15
+#define	CB_FLAG_GETOPT_DEFAULT_FILE_COLSEQ  16
+#define	CB_FLAG_MEMORY_CHECK                17
```
GitMensch, line 109: Just wondering: do we want to make that an `enum` with `= 1` in the first entry, dropping all numbers in the comments of other places?
Ye, that is a refactoring that should be checked in separately up-front...

ddeclerck:
> > * either modifying the `bdb_bt_compare` function - but that would fix only for BDB...
> 
> that would be enough for now and is likely "the most portable" solution. But that creates a HUGE problem: if we do that, then the files need to be rebuilt :-/ Note that using the key type and use the appropriate COBOL sort function (if the field structure within ´->app_data`[if not done: store a complete cb_field structure in there], then you only need to duplicate that cob_field to then set the two separate fields' data) should directly work for all types. This is effectively what is done for`SEARCH`(and the internal comparison during`SORT`).
> 
> Otherwise we could have small-enough numbers be converted to big-endian binary and bigger ones to be converted to USAGE DISPLAY SIGN SEPARATE LEADING in `bdb_savekey`, that would also work - but again leads to all files needing a rebuild.

So, both solutions would imply a rebuilding of the files. What I like about the first one is that it uses backend's mechanisms intended for this purpose (i.e. custom key compare function) - but it has to be implemented specifically for each backend, and not all these backends expose such features. And what I like about the second one is that it is (I think ?) backend-agnostic.

> As this repacking/unpacking would be done on any key access, this is likely slower than the approach above.

I don't think we would ever have to translate the "encoded" keys backward, right ?

> As LMDB does not provide an external compare function, the second solution seems to be the only one that could be applied there

Isn't `mdb_set_compare` what we're looking for ?

> (but note that LMDB exists only in trunk and even then is marked as experimental and possibly will be dropped).

Even if slightly off-topic : I really like the idea of an LMDB backend. What would be the reasons for dropping it ? Not performing as expected ?

> Note: with ISAM there are key types that handle this (including IEE754), but we currently only use `CHARTYPE`, so the same problem seems to exists there. For this part I'll contact Ron.
> 
> > In any case, I believe this require some thinking before actually doing that, and I'm not sure this belongs to this PR (as it did not introduce that problem).
> 
> Hm. Please add a test with expected failure and a comment about the underlying problem.

I guess I'll do that.

> We could otherwise leave that general problem out of this PR, but if a key is purely numeric, there should be no "additional" conversion applied which makes the current scenario worse. That brings us to the good thing: if you implement key collation within the cobc site (you already have the codegen and use of that in libcob) then you can place a check there for a numeric type and just skip generating the collation for that.

Well, I think this is handled in `codegen.c: output_indexed_file_key_colseq` :
```
	int	type = cb_tree_type (key, cb_code_field (key));
	cb_tree	col = NULL;

	/* We only apply a collating sequence if the key is alphanumeric / display */
	if ((type & COB_TYPE_ALNUM) || (type == COB_TYPE_NUMERIC_DISPLAY)) {
		/* TODO: actually handle record keys (using file key for now) */
		col = f->collating_sequence;
	}
```
We agree that NUMERIC DISPLAY should be stored lexicographically, right ?

> > > As both can use the same comparison logic (just using a different set of collations), I think it's useful to test that.
> > 
> > Could you elaborate ?
> 
> `SEARCH`/`SORT` already compares "as cob_field" for any numeric type; if the same is applied to the INDEXED comparisons (you need a `cob_field` in ´app_data` for that), then everything "just works". But as we noted: this may be kept to a separate PR, _especially_ as we have to consider what to do with the existing files (likely we'd have to insert a marker into the file itself, likely by an embedded table, where we record "that's type 2, use different search algorithm"). This is definitely useful, but definitely "quite big" (and yes, should be kept separate).

Got it. As an alternative to an embedded "marker", how about an extra file for such metadata ?

> > > Testing variable length keys belong to this PR if previously possibly completely handled by BDB and now in our own search function.
> > 
> > I'm not sure to understand ; could you rephrase that ?
> 
> BDB compares keys. those are setup from `cob_field *component` - if this would match the COBOL field, then it may be variable-length; but as far as I now see those are always fixed-length; so the issue is again "likely important, but likely big and belongs to another PR".
> 
> So... do you add key collations in cobc now and skip to set the collation if the field is numeric (potentially also skip if it is NATIONAL)? That should provide enough for the original issue.

I think that's what my code does now ; except that it uses the file collation for every key - but I guess with all the things to do/fix we discussed here, I'd better implement complete support for (alphanumeric) key collations right now so we don't have to come back to it later.

> > Yeah, they should always be of identical length in our context (and I made a quick check by adding an assert and running both NIST and the testsuite with no error). Anyway I prefered to handle the case anyways (the comparison function is rather "low-level" and does not really care about our use case).
> 
> The new indexed_compare _is_ our specialized function to _our_ case - and we cannot even support two different sizes as we don't have it when the specialized function is used. Therefore: please adjust `indexed_compare()` to take only a single size, which then also removes the need to use `cob_min_int()`.

Alright, I'll change that.

> For the runtime warning - I think that should be an error, but we don't expect it to be seen by a user so remove the translation part and add a coverage exclusion, like
> 
> ```c
> 			/* LCOV_EXCL_START */
> 			if (k++ == MAX_MODULE_ITERS) {	/* prevent endless loop in case of broken list */
> 				/* not translated as highly unexpected */
> 				cob_runtime_warning ("max module iterations exceeded, possible broken chain");
> 				break;
> 			}
> 			/* LCOV_EXCL_STOP */
> ```

Will do.

cobc/cobc.c
ddeclerck: Yes indeed.

ddeclerck:
Rabses, added key collations and a test with expected failure regarding the numeric keys sorting problem.
@GitMensch would it be sufficient for now ?

GitMensch:
while reviewing: I've just checked in the refactoring

cobc/codegen.c
```diff
@@ -9271,6 +9271,40 @@ output_key_components (struct cb_file* f, struct cb_key_component* key_component
 	}
 }
 
+
+static void
+output_indexed_file_key_colseq (const struct cb_file *f, const struct cb_alt_key *ak, int idx)
+{
+	cb_tree	key = ak ? ak->key : f->key;
+	cb_tree	key_col = ak ? ak->collating_sequence_key : f->collating_sequence_key;
+	int	type = cb_tree_type (key, cb_code_field (key));
```
GitMensch, line 9280: please const that

cobc/codegen.c
```diff
@@ -9271,6 +9271,40 @@ output_key_components (struct cb_file* f, struct cb_key_component* key_component
...
+	cb_tree	col = NULL;
+
+	/* We only apply a collating sequence if the key is alphanumeric / display */
+	if ((type & COB_TYPE_ALNUM) || (type == COB_TYPE_NUMERIC_DISPLAY)) {
+		col = key_col ? key_col : f->collating_sequence;
+#if 0
+	/* TODO: this should be done for national, when available */
```
GitMensch, line 9287: combine that to one line please - so it is clear why that is deactivated

cobc/codegen.c
```diff
@@ -9271,6 +9271,40 @@ output_key_components (struct cb_file* f, struct cb_key_component* key_component
...
+#endif
+	}
+
+	output_prefix ();
+	if (idx == 0) {
+		output ("%s%s->collating_sequence = ", CB_PREFIX_KEYS, f->cname);
+	} else {
+		output ("(%s%s + %d)->collating_sequence = ", CB_PREFIX_KEYS, f->cname, idx);
+	}
+	if ((col != NULL) && CB_REFERENCE_P (col)) {
```
GitMensch, line 9299: Not necessarily a change, more a question - is there a reason to wrap the first part in parens?

cobc/parser.y
```diff
@@ -5752,7 +5741,7 @@ collating_sequence_clause:
...
+	CB_UNFINISHED ("FILE COLLATING SEQUENCE");
```
GitMensch, line 5744: please add an inline comment like `*/ only implemented for BDB /*`

libcob/fileio.c
```diff
@@ -4150,6 +4170,30 @@ indexed_file_delete (cob_file *f, const char *filename)
...
+}
+
+static int
+bdb_bt_compare(DB *db, const DBT *k1, const DBT *k2)
+{
+	const unsigned char *col = (unsigned char *)DBT_GET_APP_DATA(k1);
+	if (unlikely (col == NULL)) {
+		cob_runtime_error ("bdb_bt_compare was set but no collating sequence was stored in DBT");
+	}
+	if (unlikely (k1->size != k2->size)) {
```
GitMensch, line 4191: for ease of later migrating to trunk: please drop `unlikely`, as this was removed; instead add LCOV exclusion comments around that, to not decrease coverage because of those checks

libcob/fileio.c
```diff
@@ -757,6 +767,13 @@ bdb_savekey (cob_file *f, unsigned char *keyarea, unsigned char *record, int idx
 	return (int)f->keys[idx].field->size;
 }
 
+static void
```
GitMensch, line 770: should be forced-inlined

GitMensch (approved):
apart from the minor review comments and the rebase: that's beautiful - please: go, go, go

cobc/codegen.c
ddeclerck: Ah, that's just my over-parenthesing OCD kicking-in. I'll drop them.

ddeclerck:
All changes accounted for - will merge to SVN if it's okay (I see the tag is already there).

GitMensch:
I trust your changes will be fine - as the test is already set: go on with the commit.

Thanks!

GitMensch:
@ddeclerck Thank you for working on this.
In my review I've missed that the new functions were defined globally - that broke the !WITH_DB builds, just fixed that in svn.

PR #133, opened 2024-01-26
fix SEGFAULT in checking prototype arguments

lefessan:


tests/testsuite.src/syn_misc.at
GitMensch, line 1: This makes me wonder: what happens if you use a level 78 constant (should of course generate a nice error), can you add this to the test, "just because"`?

As the bug is not related to the codegen: please use the standard `$COMPILE_ONLY` here (and move the first line to the next one, dropping the line numbers at the left, which will increase all diagnostic number by 1).

GitMensch (changes requested):
thanks for the test and the fix, only a minor change to the testsuite, an added ChangeLog entry and that's good to go

tests/testsuite.src/syn_misc.at
lefessan: I added a test for a level-78 constant, it is replaced during the scanning phase, so there is an error, but it is not very understandable.

GitMensch:
I see, we expect WORD and have a LITERAL.

For level 78 (likely also for level 01 constants) we have several places that result in a strange error message (and like in that place a bit misleading position of the error message). A way to solve that would be passing an extra token type, but then we'd have to add something like the following - and use it in nearly all places where we have LITERAL and some, like that, where we currently use WORD.

```yacc
literal: LITERAL { $$ = $1; } | CONSTANT { $$ = CB_TREE (CB_CONST_USER ($1)->literal;) }
```

That's definitely too much for this PR.
As the parser error leads to a "totally different" view: please leave lvl78 in the test, but commented out with a note.

lefessan:
@engboris Maybe that could be a first job for you ? If I understand well, the idea is to remove 78-level constants from the lexer, and replace all the places where they can replace a LITERAL by a rule that also accepts a WORD and checks if it is a 78-level ?

GitMensch:
@lefessan that would be quite a big change and @engboris already took one of the other issues (towards stateless runtime), which is good to be finally handled.

lefessan:
I commented out the level 78 from the prototype call.

GitMensch (approved):

GitMensch:
Good as is now. For the additional part of the 78 handling - please create an issue in the wish-list tracker with as much technical information as reasonable for someon to work on this later.

GitMensch:
@lefessan @ddeclerck Who does the commit to SVN now?

ddeclerck:
Fabrice might be a bit busy at the moment, I'll take care of it (tomorrow I guess).

ddeclerck:
Merged in SVN at revision 5227.
(@GitMensch Can you reattribute it to Fabrice ?)

GitMensch:
metadata updated, no bug issue to be closed (nut that's quite an exotic point to reach, and described enough in the ChangeLog - not necessary to create a bug report)

On to the next PRs :-)

PR #135, opened 2024-02-19
Fix of screen section clauses

engboris:
Fix for [bug #947 on Sourceforge](https://sourceforge.net/p/gnucobol/bugs/947/)

The `VALUE ALL '-'` clause in `SCREEN SECTION` did not work (error message `error: syntax error, unexpected ALL`) although it should have.

Screen sections used clauses with `basic_literal` which does not allow the `ALL` keyword. I remplaced it by the `literal` construction (already defined) which extends `basic_literal` with the `ALL` keyword.

I added a test in `testsuite/syn_screen.at` (at the end) based on the code provided in the [bug report](https://sourceforge.net/p/gnucobol/bugs/947/).


GitMensch (changes requested):
Thank you for this first addition!

Your analysis and fix of the problem - from the parser side - is correct. See the first reviews  about "style" issues.

Please also add a runtime test (goes to run_manual_screen.at, tested by `make checkmanual`) and verify that this actually works as intended - there may be adjustments needed in libcob/screenio.c needed to handle this, too.

GitMensch (approved):
If the runtime test works for you, then this is good to go. Does it?

engboris:
I believe so. The test displays a sequence of "-" symbols and that's what I get.

engboris:
I believe it should have a "Ready for SVN" label?

ddeclerck:
Merged in SVN at revision 5226.
(@GitMensch Can you reattribute this to user "engboris" ?)

GitMensch:
Thanks for the check-in; metadata was updated in svn

Issue #103, opened 2023-06-27
CI: split "Windows workflow" into two: MSYS64 and VisualStudio


GitMensch:
The first will just be renamed, the second will get its dependencies with vcpkg (there's likely a pre-configured action for that, which caches its results), then compile (both win32 and x64) and run the testsuite with the current autofonce binaries.

lefessan:
To be honest, though we generate Windows binaries by cross-compilation, we haven't tried them yet... We will probably find bugs, typically related to how shells/which shells are started to run the testsuite. I haven't checked how the testsuite is currently ran under Windows on the current CI, which Unix env is being used for that and so on...

GitMensch:
To execute those with a cross-compile environment the normal handling with atlocal is used (the CI worker has to have Wine installed) - but the current GC environment is already MSYS64, as far as I remember. Testing those is therefore done by the normal `make check`.

To build and test VC generated binaries "manually" with the testsuite the following procedure is currently used:

* get all dependencies with vcpkg
* manually adjust the provided build_windows/config.h.in (set used libraries via sed to create config.h)
* create tarstamp.h with the provided PS script
* use msbuild to generate the binaries
* use the provided "makedist" to create the binary package including the "extras"
* [possibly upload the binaries as artifact]

This is the setup part (it seems reasonable to replace sed by powershell, but no one has done that so far):
```cmd
echo setup local configuration

sed "build_windows\config.h.in" -e's/\(#define CONFIGURED_ISAM *\) [A-Z]*/\1 BDB/g' -e's/\(#define CONFIGURED_CURSES *\) [A-Z]*/\1 PDCURSES/g' -e's/\(#define CONFIGURED_XML *\) [A-Z]*/\1 XML2/g' -e's/\(#define CONFIGURED_JSON *\) [A-Z]*/\1 CJSON_CJSON/g' >  "config.h"

powershell -ExecutionPolicy ByPass -File build_windows\maketarstamp.ps1 > tarstamp.h
```

then, for the actual test:
* execute the matching set_env.cmd (this will setup all the paths)
* from this open a msys or cygwin shell and run the testsuite from there, using the spacial atlocal_win (just copy)

Using autofonce could replace the last part, dropping the need for a POSIX shell environment (still either autofonce would have to either replace some of the tools or need sed/grep/diff/tr binaries in its PATH).

Issue #48, opened 2022-07-07
ci: consider additional windows-run with Visual Studio


GitMensch:
the GH Actions environment windows-2019 comes with Virtual Studio 2019 and vcpkg

A setup similar to this is actually running on Appveyor currently ([and btw compiled latest GCOS adjustments, too](https://ci.appveyor.com/project/GitMensch/gnucobol-3-x-vs)); it may be useful to test complilation (and maybe running) with at least one non-gcc build (macos is likely clang, even if named gcc, but even then much more "like gcc")

That's the CI definition that would work in appveyor, giving an overview about steps to possibly add some day:

```yml
version: 3.2-dev.{build}-vs

image:
- Visual Studio 2019

configuration:
- Release
- Debug

platform:
- x86
- x64

clone_folder: C:\projects\gnucobol_vs

environment:
  WINFLEXBISON_VERSION: 2.5.24
  WINFLEXBISON_ARCHIVE: win_flex_bison-%WINFLEXBISON_VERSION%.zip
  DEP_ARCHIVE: dependencies-%PLATFORM%.7z
  VCPKGS: zlib:%PLATFORM%-windows liblzma:%PLATFORM%-windows mpir:%PLATFORM%-windows pdcurses:%PLATFORM%-windows berkeleydb:%PLATFORM%-windows libxml2:%PLATFORM%-windows cjson:%PLATFORM%-windows json-c:%PLATFORM%-windows

install:

- cmd: >-
    set "OLDDIR=%CD%"
    echo Use vcpkg for pre-prerequisites &&   pushd c:\tools\vcpkg &&     git pull &&   .\bootstrap-vcpkg.bat &&   vcpkg integrate install &&   vcpkg install %VCPKGS% &&  vcpkg update &&  vcpkg upgrade %VCPKGS% --no-dry-run
    pushd %OLDDIR%

    echo Get flex/bison, include them in PATH
    if not exist "%WINFLEXBISON_ARCHIVE%" appveyor DownloadFile "https://github.com/lexxmark/winflexbison/releases/download/v%WINFLEXBISON_VERSION%/%WINFLEXBISON_ARCHIVE%"
    7z.exe x -y -owin_flex_bison\ "%WINFLEXBISON_ARCHIVE%" > nul
    set PATH=%CD%\win_flex_bison;%PATH%

cache:

- '%WINFLEXBISON_ARCHIVE%'

- c:\tools\vcpkg\installed\

- c:\tools\vcpkg\packages\

[...]

PR #140, opened 2024-03-19
Add/update Windows workflows

ddeclerck:
This PR adds an MSVC workflow to compile GnuCOBOL and run the testsuite.

Note 1: we manually generate the `tests/atconfig` and `tests/package.m4` files to avoid having to generate and call `./configure`

Note 2: about 80 tests fail (I get the same results locally on an actual Windows machine)


.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,125 @@
...
+        uses: actions/checkout@v3
+
+      - name: Restore VCPKG package cache
+        uses: actions/cache/restore@v3
+        id: restore-vcpkg-cache
+        with:
+          key: cache-vcpkg
+    
+      - name: Install VCPKG packages
+        if: steps.restore-vcpkg-cache.outputs.cache-hit != 'true'
```
GitMensch, line 44: You likely either want to run that in any case (should be quite fast after getting the cache) or want to do that once a day/week, so you do get updates also without manually dropping the cache.

.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,125 @@
...
+      - name: Save VCPKG package cache
+        uses: actions/cache/save@v3
+        id: save-vcpkg-cache
+        with:
+          key: cache-vcpkg
+          path: |
+            ${{ env.VCPKG_ROOT }}\installed
+            ${{ env.VCPKG_ROOT }}\packages
+
+      - name: Install WinFlexBison
```
GitMensch, line 63: That wont change and therefore would be reasonable to also cache (until url changes)

.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          cd $env:VCPKG_ROOT
+          git pull
+          .\bootstrap-vcpkg.bat -disableMetrics
+          vcpkg integrate install
+          vcpkg install (-split $env:VCPKGS)
+          vcpkg update
+          vcpkg upgrade (-split $env:VCPKGS) --no-dry-run
+
+      - name: Save VCPKG package cache
```
GitMensch, line 54: You likely want to force that also for failing builds.

GitMensch (commented):
The one missing and important thing is to make the testsuite log available as artifact (if this does not exist then the generated config.h will be useful, just from experience).

You likely also want to run the dist script from build_windows to additional provide the resulting binaries for easy test.

Currently all the run tests are missing. Guess: some of the dlls are not found, but the testsuite.log will show what the issue really is.

GitMensch:
Ah, now it is a draft ;-) Just thought about some issues I've seen when working on the appimage definition - while this isn't linked by commit message or manually, I _think_ this is thought to implement #48 #103.

ddeclerck:
> ... and of course, I am interested in this one:
> 
> > about 80 tests fail (I get the same results locally on an actual Windows machine)
> 
> Last time I've run the tests there were much less failures (and additional 1-2 when using the debug version as that raised some overflow/C RTS check).

I haven't investigated yet, but I seem to see a lot of linker errors: "LNK2005: DllMain already defined".
Have you met such errors before ?

ddeclerck:
Seems I can't go further until the [xz mess](https://www.phoronix.com/news/GitHub-Disables-XZ-Repo) is solved...
```
Downloading https://github.com/tukaani-project/xz/archive/v5.4.4.tar.gz
[DEBUG] Trying to hash C:\vcpkg\downloads\tukaani-project-xz-v5.4.4.tar.gz.7420.part
[DEBUG] C:\vcpkg\downloads\tukaani-project-xz-v5.4.4.tar.gz.7420.part has hash a9fd8148a6693d5b48848274fb48cca0b3f11ba4548399c7a232d26d9f07346a709eda8e010ca2d08e29bf085f208969fdeece19b37b8539fe99ea8d518c5788
error: Failed to download from mirror set
error: File does not have the expected hash:
url: https://github.com/tukaani-project/xz/archive/v5.4.4.tar.gz
```

GitMensch:
It seems the downloads on GitHub and vcpkg are working again (see https://github.com/microsoft/vcpkg/issues/37893).

Can you go on with this PR, please? Making it visible which tests do fail is useful and allows us to work on fixing them, especially after comparing them with the failed MinGW tests.

GitMensch:
> we manually generate the `tests/atconfig` and `tests/package.m4` files to avoid having to generate and call `./configure`

Note in general: there is a reason for the atlocal_win. If not done already I suggest checking the README under build_windows.


Comments: 

* test 849 is a hairy beast: we write to internal storage to test if GnuCOBOL can recognize that - but as the memory layout outside of lvl 01 is not guaranteed (at not until this extra feature is available under GC4, and even then we want to check the then "old 'random' memory layout") it is hard to get it right: write too much / bad place and there's an abort before GnuCOBOL generated modules can even recognize; write too less and it won't be a problem on environment X [MacOS for example was different, iff I remember correctly, also enabled C runtime checks for hardening may as well break the test "before"]...
  --> I'd suggest to inspect this last and be prepared for ignoring this (to get it right you'd need a bunch of memory dump output or better a source-level debugger and inspecting memory there) - and if you get it right, then the test mail fail under a different environment
* tests 803-805: that's related to PDCurses - you either need to disable this (see atlocal_win) or _potentially_ use a different version/port of PDCurses
* tests 26+27 -see discussion in #109 (where I've request to move this out to a separate PRs); but test 26 should be fixable by minor adjustment in the testsuite's `$SED` use
* test 10, maybe the filename is not `prog.s` - check with a listing directory what is happening; I'm quite sure that this worked before with MSVC (and we explicit dropped the "now compile the prog.s and run it", because MSVC generated assembly does not necessary compile, it is "informational only" (whatever this means in this context)
* 18, 118+119 - these look like necessary adjustments in the cl.exe/(link.exe  command line cobc uses

ddeclerck:
> > we manually generate the `tests/atconfig` and `tests/package.m4` files to avoid having to generate and call `./configure`
> 
> Note in general: there is a reason for the atlocal_win. If not done already I suggest checking the README under build_windows.

I do use `atlocal_win` (almost unmodified, just using the Release configuration - should I use Debug instead ?)

I think I checked the README file when I wrote this workflow (it was over a month ago). Took some liberties with the instructions...

GitMensch:
Release is fine, but as Debug includes runtime checks it is often more useful.
... of course: best would be running _both_ Release and Debug after each other or, as in this case: just as a parallel workflow that uses the same definition, just two configurations.

ddeclerck:
> * test 10, maybe the filename is not `prog.s` - check with a listing directory what is happening; I'm quite sure that this worked before with MSVC (and we explicit dropped the "now compile the prog.s and run it", because MSVC generated assembly does not necessary compile, it is "informational only" (whatever this means in this context)

Indeed, the file is named `prog.asm`. According to the [documentation](https://learn.microsoft.com/en-us/cpp/build/reference/fa-fa-listing-file?view=msvc-170), when the file name given to the `/Fa` option has no extension, it will use `.asm` by default. In `cobc.c`, function `process_compile`, we have :
```
	if (output_name) {
		name = output_name;
	} else {
		name = file_basename (fn->source, NULL);
#ifndef	_MSC_VER
		strcat (name, ".s");
#endif
	}
```
I'd say we should just add the `.s` extension regardless of the compiler used. Was there any reason not to do that ? (maybe different behavior with older `cl.exe` versions ?)

ddeclerck:
> * tests 26+27 -see discussion in [add dependencies options and -fcopybook-deps  #109](https://github.com/OCamlPro/gnucobol/pull/109) (where I've request to move this out to a separate PRs); but test 26 should be fixable by minor adjustment in the testsuite's `$SED` use

27 seems to be a different issue. We're referring to a function that is only declared but not defined. That seems to be okay when building a *library* with gcc, but I don't think MSVC allows that (I belive it needs an import library for that).

GitMensch:
Rechecked:
* all VC versions that GC3 supports (2008+) have the same rule
* earlier versions of Visual Studio and msbuiild (I _guess_ before 2013/2015) only recognize  ".asm" as assembly source (newer also can cope with .s) and then will execute masm on the file, creating an object file
* the masm binary is either ml or ml64

I therefore think that:
* For _MSVC we should explicit strcat `".asm"` - so it is clearly "on purpose", ideally with a comment `/* earlier versions of msbuild don't recognize .s */`.
* The test (10 is about the output files) can then test for prog.s || prog.asm (maybe more if there's a default)
* For MSVC we should explicit call ml / ml64 - can you please add a define in build_windows/config.h.iin ` #ädefine AS "ml.exe"` and use it for the MSVC compile of assembly?
* We _likely_ can then also use another `$COMPILE prog.cob -S -o prog.s` followed by `$COMPILE prog.s`.

GitMensch:
Your analysis for the failing test 27 is correct - could you try to create an import library (for any compiler) up-front?

ddeclerck:
Worked on bit on this. Had to fix a few things.

Now only those failed tests remain : 10 (save-temps in sub-directory), 26 (cobc diagnostics show caret) and 850 (runtime check: write to internal storage)
Note sure about 8 (is it still random, or solved for good ?).

libcob/profiling.c
```diff
@@ -336,7 +336,7 @@ print_monotonic_time (FILE *file, cob_ns_time t) {
 	if (seconds > 1000) {
 		fprintf (file, "%d s", seconds);
 	} else {
-		fprintf (file, "%d.%03Ld s", seconds, milliseconds);
+		fprintf (file, "%d.%03d s", seconds, (unsigned int)milliseconds);
```
ddeclerck, line 339: As far as I can tell, "%Ld" does not exist. This caused crashes under MSVC.

tests/testsuite.src/used_binaries.at
```diff
@@ -1104,7 +1104,7 @@ AT_CHECK([$COMPILE_MODULE prog.cob], [0], [], [])
 
 # mismatch in function signature - we ignore the error output,
 # as it depends on the C compiler in use
-AT_CHECK([$COMPILE_MODULE --include $(_return_path "$(pwd)/filec.h") -fstatic-call prog.cob], [1], [], [ignore])
+AT_CHECK([$COMPILE_MODULE --include $(_return_path "$(pwd)/filec.h") -fstatic-call prog.cob], [1], [ignore], [ignore])
```
ddeclerck, line 1107: Under MSVC, the error seems to be displayed on stdout instead of stderr...

libcob/profiling.c
GitMensch: I'll fix this one with the next commit

tests/testsuite.src/used_binaries.at
GitMensch: Did you verify that with plain use of MSVC? I _guess_ this may be rooted in our process_filtered function which is only active under MSVC to filter the dumb output of the fiile that is processed.

GitMensch:
The fail wit 8 _may_ be only applicable to GCC, not sure. What do you think of using make checkall here and also `|| make check TESTSUITEFLAGS=--recheck --verbose`, to have the few failing test details also in the console log?

Note: the -llibsomething seems not portable... Any idea how to work around that?

tests/testsuite.src/used_binaries.at
ddeclerck: Same by calling `cl` directly:
```
C:\Users\David>cl /nologo test.c 1>out.txt 2>err.txt

C:\Users\David>type out.txt
test.c
test.c(11): error C2198: 'f' : pas assez d'arguments pour un appel

C:\Users\David>type err.txt

C:\Users\David>
```

ddeclerck:
> The fail wit 8 _may_ be only applicable to GCC, not sure.

Testing locally, it seems MSVC is affected by this too (though it seems to work in the CI  for some reason).

> What do you think of using make checkall here and also `|| make check TESTSUITEFLAGS=--recheck --verbose`, to have the few failing test details also in the console log?

Well, I was willing to avoid the "make" path, to speed up the CI (less MSYS2 packages to install, no need to run configure...). Plus, I've just tested on my Windows machine, and I can't seem to get `make check` to work with the GnuCOBOL built for MSVC. Depending on the path I take, either it insists on (re)building an MSYS2 GnuCOBOL and using that one, or it accepts using the MSVC GnuCOBOL but passes additional GCC-only flags, resulting in failure...

> Note: the -llibsomething seems not portable... Any idea how to work around that?

Not really. To my understanding this is just because MSVC does not add a "lib" prefix to libraries. Maybe this could be tweaked directly in GnuCOBOL, when passing a library, check whether the file exists with or without the "lib" prefix and adjust accordingly (works as long as both do not exist at the same time).

tests/testsuite.src/used_binaries.at
GitMensch: Thanks for checking - I've adjusted the testsuite.

GitMensch:
> Well, I was willing to avoid the "make" path, to speed up the CI (less MSYS2 packages to install, no need to run configure...). Plus, I've just tested on my Windows machine, and I can't seem to get `make check` to work with the GnuCOBOL built for MSVC. Depending on the path I take, either it insists on (re)building an MSYS2 GnuCOBOL and using that one, or it accepts using the MSVC GnuCOBOL but passes additional GCC-only flags, resulting in failure...

This should only be the case when you don't use the atlocal_win part.

For the -llib - my recent commit should work around that (by compiling to libfilec and filec).
Checking the library name at compile time would mean to know the complete lookup path of the linker, that's unlikely to be added completely in a portable way.

I  haven't checked the caret test yet and we possibly need the ml.exe/ml64.exe part done.

GitMensch:
> Well, I was willing to avoid the "make" path, to speed up the CI (less MSYS2 packages to install, no need to run configure...). Plus, I've just tested on my Windows machine, and I can't seem to get `make check` to work with the GnuCOBOL built for MSVC. Depending on the path I take, either it insists on (re)building an MSYS2 GnuCOBOL and using that one, or it accepts using the MSVC GnuCOBOL but passes additional GCC-only flags, resulting in failure...

Checking again what you did, that's fine.
My suggested adjustment is:

```diff
-bash -c "./testsuite"
+bash -c "./testsuite || ./testsuite --recheck --verbose"
```

GitMensch:
Tests 6 + 10 seems to be the same and likely an error in how cobc executes cl:

~~
executing:	cl.exe /c /I "D:\a\gnucobol\gnucobol" /I
		"C:\vcpkg\installed\x64-windows\include" /nologo /MD /c
		/Fa"prog.asm" /Fo"prog.asm"
		"C:\Users\RUNNER~1\AppData\Local\Temp\cob1364_0.c"
return status:	1
stdout:
cobc (GnuCOBOL) 3.3-dev.20240516
Built     May 16 2024 12:40:00	Packaged  May 16 2024 12:38:35 UTC
C version (Microsoft) 1939
  C:\Users\RUNNER~1\AppData\Local\Temp\cob1364_0.c : fatal error C1083: Cannot open compiler generated file: 'prog.asm': Permission denied
~~~

Can you have a look at that command line and a fix?

For now the other two tests fail reproducible, I suggest for the MSVC workflow you edit `AT_XFAIL` in there before generating the testsuite - then this could be pulled in and monitor for new failures.

ddeclerck:
```
executing: cl.exe /c /I "D:\a\gnucobol\gnucobol" /I "C:\vcpkg\installed\x64-windows\include" /nologo /MD /c /Fa"prog.asm" /Fo"prog.asm" "C:\Users\RUNNER~1\AppData\Local\Temp\cob1364_0.c"
```

Looks like the reason is because both `/Fa` and `/Fo` point to the same file (I was able to reproduce the error just calling cl.exe).

ddeclerck:
In `cobc.c:8216` we have :
```
	sprintf (cobc_buffer, cb_source_debugging ?
		"%s /c %s %s /Od /MDd /Zi /FR /c /Fa\"%s\" /Fo\"%s\" \"%s\"" :
		"%s /c %s %s     /MD          /c /Fa\"%s\" /Fo\"%s\" \"%s\"",
			cobc_cc, cobc_cflags, cobc_include, name,
			name, fn->translate);
```
Changing this way solves the problem :
```
	sprintf (cobc_buffer, cb_source_debugging ?
		"%s /c %s %s /Od /MDd /Zi /FR /c /Fa\"%s\" /Fo\"%s\" \"%s\"" :
		"%s /c %s %s     /MD          /c /Fa\"%s\" /Fo\"%s\" \"%s\"",
			cobc_cc, cobc_cflags, cobc_include, name,
			fn->object, fn->translate);
```
Does that sound reasonable ?

GitMensch:
Yes, and if that works for you locally and in the CI build - just commit that upstream.

ddeclerck:
As for test 26, making this change in the `_return_path` function in `atlocal_win` seems to work locally under both MSYS2 and Cygwin (waiting for CI to complete) :
```
	if test "$OSTYPE" = "cygwin" -o "$OSTYPE" = "msys"; then
		cygpath -pm "$1"
	else
```
(ie adding `msys` and using `-pm` instead of `-pw`)

GitMensch:
~~~
simon@mingw-msys ~
$ echo $OSTYPE
msys

simon@mingw-msys ~
$ cygpath
bash.exe": cygpath: command not found

simon@mingw-msys ~
$ env | grep -i msys
PS1=\[\033]0;$MSYSTEM:\w\007
MSYSTEM=MINGW32

simon@msys2-mingw32 ~
MSYS2_NOSTART=yes
WD=D:\dev\msys64\usr\bin\
MSYSTEM=MINGW32
MSYSTEM_CARCH=i686
MSYSTEM_CHOST=i686-w64-mingw32
PS1=\[\e]0;\w\a\]\n\[\e[32m\]\u@\h \[\e[35m\]$MSYSTEM\[\e[0m\] \[\e[33m\]\w\[\e[0m\]\n\$
MSYSTEM_PREFIX=/mingw32
~~~

We therefore cannot check only for "msys", but a separate test should work, `-o x$MSYSTEM_CHOST != x`  seems to be ok.
This could be set manually, of course, but at least I've only seen that for MSYSTEM_PREFIX so far.

ddeclerck:
After checking, it messes up the CI anyways, so...

cobc/cobc.c
```diff
@@ -8217,7 +8217,7 @@ process_compile (struct filename *fn)
 		"%s /c %s %s /Od /MDd /Zi /FR /c /Fa\"%s\" /Fo\"%s\" \"%s\"" :
 		"%s /c %s %s     /MD          /c /Fa\"%s\" /Fo\"%s\" \"%s\"",
 			cobc_cc, cobc_cflags, cobc_include, name,
-			name, fn->translate);
```
GitMensch, line 8220: check if that's reasonable my manually doing:

```cmd
cobc -v -S hello.cob
cobc -v hello.asm
cobcrun hello
```

ddeclerck: Well, since MSVC's assembly listing is "informational", we're not expecting this to work, do we ?
Even calling ml64.exe directly on the .asm file, I get some errors.

ddeclerck:
Seems we're all good now.

GitMensch:
We'd expect a hand written valid one to be assembled correctly. As noted earlier, we possibly need to define and use a macro with ml/ml64 to reach that.

Am 17. Mai 2024 16:31:45 MESZ schrieb OCP David Declerck ***@***.***>:

ddeclerck:
> ```
> simon@mingw-msys ~
> $ echo $OSTYPE
> msys
> 
> simon@mingw-msys ~
> $ cygpath
> bash.exe": cygpath: command not found
> 
> simon@mingw-msys ~
> $ env | grep -i msys
> PS1=\[\033]0;$MSYSTEM:\w\007
> MSYSTEM=MINGW32
> ```

I'm curious about how you obtain these environments. Whether I start the msys2 shell from the exe wrappers (msys2.exe, ucrt64.exe...) or the `msys2_shell.cmd` script, I have all of the `MSYSTEM`, `MSYSTEM_CARCH`, `MSYSTEM_CHOST` and `MSYSTEM_PREFIX` variables defined, and `cygpath` available in PATH.

GitMensch:
That's the original MSYS (the one that has WinXP and possible older support), not MSYS2. As their site is still down and the Internet archive use is cumbersome, I'd grab them from Arnold's site - this is what he also uses to create the official (small) MinGW package.

GitMensch:
@ddeclerck What's the state with this PR (for both the 3.x and the gc4 branch)?

ddeclerck:
> @ddeclerck What's the state with this PR (for both the 3.x and the gc4 branch)?

Well, I consider this to be "good enough" for its purpose, i.e. ensuring we don't break Windows MSVC builds.
That's also what I'm using for the GC4 CI.

Note the changes to `cobc.c` and `atlocal_win`, which should probably be commited upstream independently for this PR.

GitMensch:
feel free to check in those two changes upstream (in this case also an entry in tests/Changelog, please)

GitMensch:
It seems there is an issue with caching WinFlexBison which you may want to inspect
in any case I suggest to run the `build_windows/makedist.cmd` and upload the result as artifact

as there are more C RTS checks in the debug build you may consider building both Release and Debug (if I remember right the debug one actually shows an not-yet-resolved issue in the MSVC builds)

Would it be reasonable to run NIST as well?

GitMensch:
In any case there should be a build matrix to build and test both win32 and x64 platforms (and possibly both Release and Debug).

Oh and just out of interest - would it be complicated to also test with VC2013 (as oldest-supported version in 3.x - not relevant for trunk)?

In this case vcpkg cannot be used but something like the following would work instead

```cmd
wget https://versaweb.dl.sourceforge.net/project/gnucobol/dependencies/windows/old/win_prerequistes_vc11.7z
7z.exe x win_prerequistes_vc11.7z
move gmp.h mpir.h
cd Win32  &&  copy "Release\libvbisam.dll" "Debug\"
```

GitMensch:
Looks nice, so if this runs then it should be merged (and then the same commit also be applied to gc4).

ddeclerck:
> Looks nice, so if this runs then it should be merged (and then the same commit also be applied to gc4).

Well, there are indeed failures under debug configurations, as well as two tests that I had to exclude as they would hang (only in debug):
- CURRENCY SIGN WITH PICTURE SYMBOL (run_misc.at)
- FUNCTION RANDOM (run_functions.at)

(aren't you on vacation ? :D)

GitMensch:
ping @ddeclerck for checking the issues above (even if they are 3.x - but they are likely identical in 4.x)

engboris:
> > ping @ddeclerck for checking the issues above (even if they are 3.x - but they are likely identical in 4.x)
> 
> I'm overwhelmed, can't do that for now. @engboris, could you give me a hand ?

I'd like to help (and I'm always eager to learn new things) but I'll probably need a significant onboarding as I have currently no knowledge of Github CI / Windows MSVC.

ddeclerck:
> > > ping @ddeclerck for checking the issues above (even if they are 3.x - but they are likely identical in 4.x)
> > 
> > 
> > I'm overwhelmed, can't do that for now. @engboris, could you give me a hand ?
> 
> I'd like to help (and I'm always eager to learn new things) but I'll probably need a significant onboarding as I have currently no knowledge of Github CI / Windows MSVC.

I'll show you (hopefully by the end of the week ?).

GitMensch:
Just checking about the state @ddeclerck @engboris. I think there was more discussion about the CI somewhere (including which tests and artifacts would be reasonable), but at least https://github.com/OCamlPro/gnucobol/pull/147#issuecomment-2226872297 hints at running `make mingwdist` and providing the two resulting artifacts (for both MSYS1 and MSYS2) and similar there is build_windows\makedist.cmd` to create artifacts for the the MSVC builds.

ddeclerck:
Forgot to mention it, but the MYSY1 and MSYS2 CIs (on the GC4 branch) have already been updated according to [https://github.com/OCamlPro/gnucobol/pull/147#issuecomment-2226872297](https://github.com/OCamlPro/gnucobol/pull/147#issuecomment-2226872297).

As for MSVC, it's on the toto-list, but as a matter of fact, all the team is overwhelmed, so it will take some time.

GitMensch:
Ah, I did indeed missed 9bd5c8cfaf34d648d93d5397fc56d1a5dfb356a8, maybe "just" bring the GC4 and GC3 CI definitions up to the same state for now? Should nearly be a copy/paste, no?

ddeclerck:
> Ah, I did indeed missed [9bd5c8c](https://github.com/OCamlPro/gnucobol/commit/9bd5c8cfaf34d648d93d5397fc56d1a5dfb356a8), maybe "just" bring the GC4 and GC3 CI definitions up to the same state for now? Should nearly be a copy/paste, no?

Yes, that should be doable.

GitMensch:
hm,, the last push changed the name to the wrong file again ? [I've previously edited it it to the correct one: run_fundamental.at)

ddeclerck:
> hm,, the last push changed the name to the wrong file again ? [I've previously edited it it to the correct one: run_fundamental.at)

I'm not sure what you're talking about; is this related to this PR ?

GitMensch:
yes, I've pushed because I've seen the test still hang, and then found that the sed used operates on the wrong file - as it currently does again.

ddeclerck:
> yes, I've pushed because I've seen the test still hang, and then found that the sed used operates on the wrong file - as it currently does again.

Ah, you pushed to my own branch, that's why. I assume people never do that, and I happily force-push to my branch to keep a single commit.

GitMensch:
No problem, but as noted you need:

```diff
- sed -i '/AT_SETUP(\[MOVE to edited item (4)\])/a AT_SKIP_IF(\[true\])' testsuite.src/run_functions.at
+ sed -i '/AT_SETUP(\[MOVE to edited item (4)\])/a AT_SKIP_IF(\[true\])' testsuite.src/run_fundamental.at
```

GitMensch:
> Of course, on the long term, those will be investigated locally.

Sounds good. That may have been faster (but that's hard to know up front) - the CI now hangs at test 577 / 1304.

GitMensch:
https://learn.microsoft.com/en-us/windows/win32/debug/error-mode says:

> Best practice is that all applications call the process-wide [SetErrorMode](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-seterrormode) function with a parameter of SEM_FAILCRITICALERRORS at startup. This is to prevent error mode dialogs from hanging the application.

So, in both CI cases and when running as a service, we'd like to crash the application instead of opening a popup.

During debugging on a GnuCOBOL developer environment the messages as is, including the "retry to debug", is useful.

What are your thoughts about that? Adding a call on MSVC to the startup code in libcob (also executed from cobc) to disable that popup in all cases but a special environment variable is set?

ddeclerck:
> https://learn.microsoft.com/en-us/windows/win32/debug/error-mode says:
> What are your thoughts about that? Adding a call on MSVC to the startup code in libcob (also executed from cobc) to disable that popup in all cases but a special environment variable is set?

That sounds reasonable. We could use the `CI` environment variable, which *should* be defined by most CI frameworks (at least it is on GitHub, and I believe also on Appveyor ?)

ddeclerck:
> As you have the setup: can you check those occurrences and use bitmask for explicit loss of data?

I really have to go back to the GC3/4 merge.

I spent yesterday's afternoon giving @engboris all the info about this MSVC CI and prepared for him a script that reproduces the CI steps locally, so he should be able to take over this matter.

I'd go as far as suggesting to merge this PR in its current state (as it is already very helpful for its purpose), and open a new one dedicated to the fixes.

GitMensch:
I'd say %CI% is a possible option if we leave it otherwise on; I'm looking forward for @engboris to implement this and start checking the data-loss, bitmasking where expected, which should go into two separate PRs.

Thanks for improving the CI so far. The only thing that seems to be missing and would be nice if you could add it is the binary artifacts for MSVC, created after build by executing `build_windows\makedist.cmd` and upload that, _maybe_ you can inspect that before getting back to the merge, or Boris can handle this as a first PR.

ddeclerck:
> The only thing that seems to be missing and would be nice if you could add it is the binary artifacts for MSVC, created after build by executing `build_windows\makedist.cmd` and upload that, _maybe_ you can inspect that before getting back to the merge, or Boris can handle this as a first PR.

Oh I couldn't get this to work ; here's what I get (using the commented bits at the end of `windows-msvc.yml`):
```
Setup using created GnuCOBOL distribution -x64-...
Setup Visual Studio (x64/amd64)...

cl.exe already in PATH
no further initialization is done for the C compiler

cl.exe now in PATH:
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.40.33807\bin\Hostx64\x64\cl.exe
Compilateur d'optimisation Microsoft (R) C/C++ version 19.40.33812 pour x64
Copyright (C) Microsoft Corporation. Tous droits réservés.

cl : Ligne de commande error D8003 : nom de fichier de la source absent

Setting environment for GnuCOBOL.
Building extras with

cobc had unexpected return value -1073741515, running verbose again...
C:\Users\David\Programmes\msys64\home\David\gc4-msys1\gc3-msvc\build_windows\distnew\bin_x64\cobc.exe
Abort!
```

PR #141, opened 2024-03-20
[CANCELLED] Support for JOR files

lefessan:
JOR files are journals of COBOL executions. This commit implements
minimalist support for JOR files with the following features:
* extensible: the format can be easily extended in the future
* compact: the format is binary and does not take too much space on disk

Currently, the JOR file contains the arguments, the execution start
and stop times and per-file statistics of opens, reads, writes, starts
and closes operations.

### DIFF
--- a/libcob/common.h
+++ b/libcob/common.h
@@ -1380,9 +1380,24 @@ typedef struct __cob_file_key {
 #define	COB_FILE_VERSION	1
 
 
-#define COB_JOR_FILE_OPERATIONS 10
+enum cob_jor_file_operation {
+/* order must match field_file_op_name[] in jor.c */
+	COB_JOR_WRITE_TRY = 0,
+	COB_JOR_WRITE_OK,
+	COB_JOR_READ_TRY,
+	COB_JOR_READ_OK,
+	COB_JOR_START_TRY,
+	COB_JOR_START_OK,
+	COB_JOR_OPEN_TRY,
+	COB_JOR_OPEN_OK,
+	COB_JOR_CLOSE_TRY,
+	COB_JOR_CLOSE_OK,
+	COB_JOR_OPERATIONS_MAX
+};
+
 typedef struct __cob_file_jor {
-	cob_u32_t   *ops [COB_JOR_FILE_OPERATIONS] ;
+	int id ;
+	cob_u32_t   *ops [COB_JOR_OPERATIONS_MAX] ;
 } cob_file_jor;
 
 /* File structure */
@@ -2877,19 +2892,26 @@ COB_EXPIMP cob_field* cob_intr_hex_to_char (cob_field*);
 /* Functions in jor.c */
 /************************/
 
-enum cob_jor_file_operation {
-/* order must match field_file_op_name[] in jor.c */
-	COB_JOR_WRITE_TRY = 0,
-	COB_JOR_WRITE_OK,
-	COB_JOR_READ_TRY,
-	COB_JOR_READ_OK,
-	COB_JOR_START_TRY,
-	COB_JOR_START_OK,
-	COB_JOR_OPEN_TRY,
-	COB_JOR_OPEN_OK,
-	COB_JOR_CLOSE_TRY,
-	COB_JOR_CLOSE_OK,
-	COB_JOR_AFTER_LAST_OPERATION
+#define JOR_MAGIC "GNUCOJOR"
+#define JOR_MAGIC_LEN 8
+#define JOR_HEADER_SIZE 16
+
+#define OPCODE_NEW_NAME 0
+#define OPCODE_LOCAL_FIELD 1
+
+enum jor_operation_type {
+	TYPE_RECORD =     0,
+	TYPE_UINT8,
+	TYPE_INT8,
+	TYPE_UINT16,
+	TYPE_INT16,
+	TYPE_UINT32,
+	TYPE_INT32,
+	TYPE_UINT64,
+	TYPE_INT64,
+	TYPE_FLOAT,
+	TYPE_STRING8,
+	TYPE_STRING16 /* = 11 */
 };
 
 COB_EXPIMP void cob_jor_exit (int, const char*, ...);
--- a/libcob/common.h
+++ b/libcob/common.h
@@ -1724,7 +1724,7 @@ COB_EXPIMP void	cob_set_locale			(cob_field *, const int);
 COB_EXPIMP int 	cob_setenv		(const char *, const char *, int);
 COB_EXPIMP int 	cob_unsetenv		(const char *);
 COB_EXPIMP char	*cob_getenv_direct		(const char *);
-COB_EXPIMP char *cob_expand_env_string	(char *);
+COB_EXPIMP char *cob_expand_env_string	(const char *);
 COB_EXPIMP char	*cob_getenv			(const char *);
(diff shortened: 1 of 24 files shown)


libcob/common.h
```diff
@@ -2864,6 +2871,37 @@ COB_EXPIMP cob_field *cob_intr_bit_to_char		(cob_field *);
...
+	COB_JOR_READ_TRY,
+	COB_JOR_READ_OK,
+	COB_JOR_START_TRY,
+	COB_JOR_START_OK,
+	COB_JOR_OPEN_TRY,
+	COB_JOR_OPEN_OK,
+	COB_JOR_CLOSE_TRY,
+	COB_JOR_CLOSE_OK,
+	COB_JOR_AFTER_LAST_OPERATION
+};
```
GitMensch, line 2893: put that above, add a trailing "_MAX" entry and use that max entry as index instead of `COB_JOR_FILE_OPERATIONS `

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
+/*
+   Copyright (C) 2023 Free Software Foundation, Inc.
+   Written by Emilien Lemaire and Fabrice Le Fessant.
```
GitMensch, line 3: So you did start that last year already, together with Emilien?

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+#ifdef	_WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#undef MOUSE_MOVED
+#include <process.h>
+#include <io.h>
+#include <fcntl.h>	/* for _O_BINARY only */
+#endif
+
+#include <sys/time.h>
```
GitMensch, line 51: as clangd and some compilers will tell you - these are not all used in the source, please cleanup

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+static int field_status ;
+static int field_reason ;
+static int field_file ;
+static int field_name ;
+static int field_truncated ;
+static int field_start ;
+static int field_exit ;
+static int field_duration ;
+static int field_time ;
+static int field_args ;
```
GitMensch, line 77: likely `unsigned int`, if you don't need a negative marker (and if that is used, it is often useful to double-check if that's a  good implementation)

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+#define TYPE_INT8       2
+#define TYPE_UINT16     3
+#define TYPE_INT16      4
+#define TYPE_UINT32     5
+#define TYPE_INT32      6
+#define TYPE_UINT64     7
+#define TYPE_INT64      8
+#define TYPE_FLOAT      9
+#define TYPE_STRING8   10
+#define TYPE_STRING16  11
```
GitMensch, line 227: enum?

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+		jor_position += len;
+	}
+	RECORD_END ();
+}
+
+static
+char* jor_allocate(int *size)
+{
+	return cob_malloc(*size);
+}
```
GitMensch, line 281: likely replace in the caller

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+		fprintf (stderr, "COB_JOR_AFTER_LAST_OPERATION = %d\n",
+			 COB_JOR_AFTER_LAST_OPERATION);
+		exit (2);
+	}
+	if ( field_file_op_name[COB_JOR_FILE_OPERATIONS-1] == NULL ){
+		fprintf (stderr,
+			 "field_file_op_name[%d] not initialized\n",
+			 COB_JOR_FILE_OPERATIONS-1);
+		exit (2);
+	}
```
GitMensch, line 328: use the common abort and message functions here or drop that part

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+
+	if (!cobsetptr->cob_jor_enable &&
+	    /* testsuite clears COB_JOR_ENABLE... */
+	    !getenv ("COB_JOR_ENABLED")) return ;
```
GitMensch, line 332: that will be already be handled in the settings pointer, so this part can be removed

libcob/fileio.c
GitMensch, line 1: `cob_jor_file_operation` is called often, so I think we _may_ check `!cobsetptr->cob_jor_enable` in this file instead (saving the function call and its resulting stack/register changes on in and out), possibly wrap that in a macro so you keep the single lines here. But I've done no perf tracing on this...

libcob/jor.c
```diff
@@ -0,0 +1,464 @@
...
+	jor_funcs.save = f->save;
+	jor_funcs.allocate = f->allocate;
+}
+
+void cob_init_jor (cob_global *lptr, cob_settings *sptr,
+		   int cob_argc, char** cob_argv)
+{
+	int i;
+
+	if (!!cobsetptr) return ;
```
GitMensch, line 308: that would be an error, I therefore would suggest to drop that

libcob/jor.c
GitMensch, line 1: the outline at the start is well done; please add doxygen comments to each of the functions as well

GitMensch (changes requested):
obviously this has one commit too much in currently; for the first draft see some notes in the review; the biggest question is (likely answered in NEWS and gnucobol.texi): why should a user want that, what is its practical purpose?

Should there be binaries under /bin that take an existing jor file and convert it to xml/json/text?

GitMensch:
What do you think of moving the stats to be only/mostly set on `save_status` as per Ron's implementation https://github.com/OCamlPro/gnucobol4/commit/4a61f7cab2bb4a3e8f3a5971cb3cd67fc8290775 https://github.com/OCamlPro/gnucobol4/commit/24a3751e47fdec140be50e445a4fb80c3223385d? 

What about merging the fileio stats (only active for programs that are compiled with tracing) to GC 3.3 (that would also include https://github.com/OCamlPro/gnucobol4/commit/980d0b41829de8952157c8264ec61d497e504497)?

lefessan:
The idea behind this PR is to be able to get a minimal trace of the execution, in production mode. This is not possible with standard tracing options that are too expensive. You can imagine a batch failing during the night, people might want to understand what happened exactly during the run, and maybe to preceding steps, without running them again in trace mode. Also, you get information like the time spent, the IO stats, and probably more in the future.

There are maybe other solutions, but I think it's worth investigating this mode.

And yes, there needs to be some program that will translate these binary files into json or something like that...

By the way, I am not really targeting 3.3, ideally, it's something that could land in 4.0 once 3.3 has been released.

GitMensch:
> By the way, I am not really targeting 3.3, ideally, it's something that could land in 4.0 once 3.3 has been released.

Ah, then please edit the PR to have that based on master then.
Still, it would be useful to check the review comments, including the question to do it at the same place: save_status and in 4.x consider to provide the option to disable/enable the file recording by file (using the IO_ options, see referenced commits).

> This is not possible with standard tracing options that are too expensive.

We still can consider to use these as basis -> check for a new trace value "binary" and in this case do the tracing you currently do.

In any case I suggest to write a plan (which includes your binary format), then send that to the developer mailing list for additional input - _especially_ when you target trunk.

lefessan:
I am not planning to continue this PR, but instead, maybe try using Open Telemetry

PR #143, opened 2024-04-25
Add >>IMP INCLUDE directive

engboris:
Closes #114  and completes FR176 on SourceForge.

A new directive `>> IMP INCLUDE FILE_1 ... FILE_n` is introduced, allowing to include multiple C headers. Only files with name finishing by `.h` are allowed.

I'm not sure about that but I marked the feature with `/* GnuCOBOL 3.3 extension*/`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,13 +6,9 @@ NEWS - user visible changes				-*- outline -*-
 
 * New GnuCOBOL features
 
-** the leading space for all internal directives is removed in the lexer.
-   Source previously processed may need to be ajusted to be processed by
-   GnuCOBOL 3.3.
-
-** cobc now checks for binary files and early exit parsing those;
-   the error output for format errors (for example invalid indicator column)
-   is now limitted to 5 per source file
+** cobc now checks for binary and multi-byte encoded files and early exit
+   parsing those; the error output for format errors (for example invalid
+   indicator column) is now limitted to 5 per source file
 
 ** support the COLLATING SEQUENCE clause on indexed files
    (currently only with the BDB backend)
--- a/NEWS
+++ b/NEWS
@@ -6,9 +6,13 @@ NEWS - user visible changes				-*- outline -*-
 
 * New GnuCOBOL features
 
-** cobc now checks for binary and multi-byte encoded files and early exit
-   parsing those; the error output for format errors (for example invalid
-   indicator column) is now limitted to 5 per source file
+** the leading space for all internal directives is removed in the lexer.
+   Source previously processed may need to be ajusted to be processed by
+   GnuCOBOL 3.3.
+
+** cobc now checks for binary files and early exit parsing those;
+   the error output for format errors (for example invalid indicator column)
+   is now limitted to 5 per source file
 
 ** support the COLLATING SEQUENCE clause on indexed files
    (currently only with the BDB backend)
--- a/NEWS
+++ b/NEWS
@@ -54,6 +54,9 @@ NEWS - user visible changes				-*- outline -*-
    calls to externals. The files are put into quotes, unless they start by
    '<'. Quoted files are expected to have absolute paths, as the C compiler
    is called in a temp directory instead of the project directory.
+   The directive >>IMP INCLUDE "file.h", >>IMP INCLUDE "<file.h>",
+   >>IMP INCLUDE <file.h> or >>IMP INCLUDE file.h can be used as an alternative
+   to this compiler option.
 
 ** output of unlimited errors may be requested by -fmax-errors=0,
    to stop compiliation at first error use -Wfatal-errors
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -149,9 +149,12 @@
 
 2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
 
-	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
-	  new >>IMP INCLUDE directive to include one or multiple header files in
-	  the generated C code (same behavior as the --include compiler option)
+	FR #176: "Implementation of GC directive to include .h (c/c++) files"
+	* pplex.l, ppparse.y, cobc.h, codegen.c: new >>IMP INCLUDE directive to
+	  include one or multiple header files in the generated C code (same behavior
+	  as the --include but with one directive per file)
+	* scanner.l: the leading space for all internal directives is removed in the
+	  lexer. Source previously preprocessed may need to be adjusted
 
 2024-04-24  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -163,6 +163,12 @@
 	  replace stream) and we use a circular buffer for the temporary
 	  queue of tokens instead of a list.
 
+2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
+
+	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
+	  new >>IMP INCLUDE directive to include one or multiple header files in
+	  the generated C code (same behavior as the --include compiler option)
+
 2024-03-17  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
             Emilien Lemaire <emilien.lemaire@ocamlpro.com>
 
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -150,9 +150,10 @@
 2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
 
 	FR #176: "Implementation of GC directive to include .h (c/c++) files"
-	* pplex.l, ppparse.y, cobc.h, codegen.c: new >>IMP INCLUDE directive to
-	  include one or multiple header files in the generated C code (same behavior
-	  as the --include but with one directive per file)
+	* pplex.l, ppparse.y, cobc.h, codegen.c (output_gnucobol_defines):
+	  new >>IMP INCLUDE directive to include one or multiple header files in the
+	  generated C code (same behavior as the --include but with one directive per
+	  file)
 	* scanner.l: the leading space for all internal directives is removed in the
 	  lexer. Source previously preprocessed may need to be adjusted
 
@@ -165,12 +166,6 @@
 	  ones passed internally from the copy-replacing stream to the
 	  replace stream) and we use a circular buffer for the temporary
 	  queue of tokens instead of a list.
-
-2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
-
-	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
-	  new >>IMP INCLUDE directive to include one or multiple header files in
-	  the generated C code (same behavior as the --include compiler option)
 
 2024-03-17  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
             Emilien Lemaire <emilien.lemaire@ocamlpro.com>
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -149,9 +149,9 @@
 
 2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
 
-	* cobc.h, pplex.l, ppparse.y: new >>IMP INCLUDE directive to include
-	  multiple header files in the C generated code. Has the same behavior as the
-	  --include compiler option.
+	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
+	  new >>IMP INCLUDE directive to include one or multiple header files in
+	  the generated C code (same behavior as the --include compiler option)
 
 2024-04-24  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
@@ -162,12 +162,6 @@
 	  ones passed internally from the copy-replacing stream to the
 	  replace stream) and we use a circular buffer for the temporary
 	  queue of tokens instead of a list.
-
-2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
-
-	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
-	  new >>IMP INCLUDE directive to include one or multiple header files in
-	  the generated C code (same behavior as the --include compiler option)
 
 2024-03-17  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
             Emilien Lemaire <emilien.lemaire@ocamlpro.com>
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -147,6 +147,12 @@
 	  compiler aborts on broken expressions, bugs #933, #938 and #966
 	* typeck.c: minor refactoring within functions
 
+2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
+
+	* cobc.h, pplex.l, ppparse.y: new >>IMP INCLUDE directive to include
+	  multiple header files in the C generated code. Has the same behavior as the
+	  --include compiler option.
+
 2024-04-24  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
 	* replace.c: optimize speed and memory usage. For speed, we add two
@@ -156,6 +162,12 @@
 	  ones passed internally from the copy-replacing stream to the
 	  replace stream) and we use a circular buffer for the temporary
 	  queue of tokens instead of a list.
+
+2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
+
+	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
+	  new >>IMP INCLUDE directive to include one or multiple header files in
+	  the generated C code (same behavior as the --include compiler option)
 
 2024-03-17  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
             Emilien Lemaire <emilien.lemaire@ocamlpro.com>
--- a/cobc/cobc.h
+++ b/cobc/cobc.h
@@ -481,7 +481,7 @@ extern struct cb_text_list	*cb_copy_list;
 extern struct cb_text_list	*cb_include_file_list; /* global */
 extern struct cb_text_list	*cb_include_file_list_directive; /* temporary */
 extern struct cb_text_list	*cb_include_list;
-extern struct cb_text_list *cb_include_file_list_directive;
+extern struct cb_text_list	*cb_include_file_list_directive;
 extern struct cb_text_list	*cb_intrinsic_list;
 extern struct cb_text_list	*cb_extension_list;
 extern struct cb_text_list	*cb_static_call_list;
--- a/cobc/cobc.h
+++ b/cobc/cobc.h
@@ -478,9 +478,9 @@ extern int			cb_depend_keep_missing;
 extern int			cb_flag_copybook_deps;
 extern struct cb_text_list	*cb_depend_list;
 extern struct cb_text_list	*cb_copy_list;
-extern struct cb_text_list	*cb_include_file_list;
+extern struct cb_text_list	*cb_include_file_list; /* global */
+extern struct cb_text_list	*cb_include_file_list_directive; /* temporary */
 extern struct cb_text_list	*cb_include_list;
-extern struct cb_text_list *cb_include_file_list_directive;
 extern struct cb_text_list	*cb_intrinsic_list;
 extern struct cb_text_list	*cb_extension_list;
 extern struct cb_text_list	*cb_static_call_list;
--- a/cobc/cobc.h
+++ b/cobc/cobc.h
@@ -481,6 +481,7 @@ extern struct cb_text_list	*cb_copy_list;
 extern struct cb_text_list	*cb_include_file_list; /* global */
 extern struct cb_text_list	*cb_include_file_list_directive; /* temporary */
 extern struct cb_text_list	*cb_include_list;
+extern struct cb_text_list *cb_include_file_list_directive;
 extern struct cb_text_list	*cb_intrinsic_list;
 extern struct cb_text_list	*cb_extension_list;
 extern struct cb_text_list	*cb_static_call_list;
--- a/cobc/cobc.h
+++ b/cobc/cobc.h
@@ -481,7 +481,6 @@ extern struct cb_text_list	*cb_copy_list;
 extern struct cb_text_list	*cb_include_file_list; /* global */
 extern struct cb_text_list	*cb_include_file_list_directive; /* temporary */
 extern struct cb_text_list	*cb_include_list;
-extern struct cb_text_list	*cb_include_file_list_directive;
 extern struct cb_text_list	*cb_intrinsic_list;
 extern struct cb_text_list	*cb_extension_list;
 extern struct cb_text_list	*cb_static_call_list;
--- a/cobc/cobc.h
+++ b/cobc/cobc.h
@@ -276,7 +276,7 @@ enum cb_sub_check {
 struct cb_text_list {
 	struct cb_text_list	*next;			/* next pointer */
 	struct cb_text_list	*last;
-	const char		*text;
+	char			*text;
 };
 
 /* Structure for extended filenames */
@@ -480,6 +480,7 @@ extern struct cb_text_list	*cb_depend_list;
 extern struct cb_text_list	*cb_copy_list;
 extern struct cb_text_list	*cb_include_file_list;
 extern struct cb_text_list	*cb_include_list;
+extern struct cb_text_list *cb_include_file_list_directive;
 extern struct cb_text_list	*cb_intrinsic_list;
 extern struct cb_text_list	*cb_extension_list;
 extern struct cb_text_list	*cb_static_call_list;
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -1832,24 +1832,24 @@ output_gnucobol_defines (const char *formatted_date)
 		current_compile_tm.tm_sec;
 	output_line ("#define  COB_MODULE_TIME\t\t%d", i);
 
-	{
-		struct cb_text_list *l;
-		for (l = cb_include_file_list; l; l = l->next) {
-			if (l->text[0] == '<') {
-				output_line ("#include %s", l->text);
-			} else {
-				output_line ("#include \"%s\"", l->text);
-			}
-		}
-
-		for (l = cb_include_file_list_directive; l; l = l->next) {
-			if (l->text[0] == '<') {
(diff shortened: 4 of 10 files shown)


cobc/pplex.l
```diff
@@ -360,6 +361,12 @@ MAYBE_AREA_A	[ ]?#?
 	return CALL_DIRECTIVE;
 }
 
+^{MAYBE_AREA_A}[ ]*">>"[ ]?"IMP"	{
+	/* GnuCOBOL 3.3 extension */
```
GitMensch, line 365: ` >> IMP` is standard - the `INCLUDE` part is the extension

cobc/ppparse.y
```diff
@@ -1368,6 +1373,30 @@ leap_second_directive:
 | OFF
 ;
 
+imp_directive:
+  _include imp_include_sources
+  {
+	cb_include_file_list = ppp_list_append(cb_include_file_list, $2);
```
GitMensch, line 1379: You won't set the option here but in scanner.l

the rule of thumb is: both of those should resolve in the same output

* `cobc test.cob -o test.c` 
* `cobc test.cob -o test.i && cobc test.i -o test.c`

but as-is the directive is lost.

Also; it is likely reasonable to have `cb_include_file_list_directive` as you can then free it and set it to NULL between the sources:

* `cobc --include "<windows.h>" prog1.cob prog2.cob --save-temps` should have the include in both prog1.c + prog2.c
* `cobc prog1.cob prog2.cob --save-temps`  with `>> IMP INCLUDE "<windows.h>"` in the first program should not have it included in prog2,c

cobc/ppparse.y
```diff
@@ -1895,5 +1924,6 @@ _printing:	| PRINTING ;
 _on:		| ON ;
 _than:		| THAN ;
 _to:		| TO ;
+_include:	| INCLUDE ;
```
GitMensch, line 1927: please have this as mandatory, not as optional

cobc/pplex.l
```diff
@@ -1867,6 +1883,14 @@ int cobc_deciph_source_format (const char *sfname) {
 	return 1;
 }
 
+int cobc_deciph_header_filename (const char *filename) {
+	size_t n = strlen(filename);
+	if (filename[n-2] == '.' && filename[n-1] == 'h') {
```
GitMensch, line 1888: I think this breaks support for `>> IMP INCLUDE "<system.h>`. In any case it should be handled identical as `--include` does.

GitMensch (changes requested):
Thanks for this first take - please copy the tests for `--include` and adjust those for the directive. In gnbucobol.texi and NEWS add a remark in the `--include` description, that the directive can be used as an alternative.

cobc/ppparse.y
engboris: > Also; it is likely reasonable to have cb_include_file_list_directive as you can then free it and set it to NULL between the sources

I believe I can use `cb_include_file_list`? Since it has been introduced for that purpose in Fabrice's previous PR. Do we really need two lists?

GitMensch:
We don't need two lists, but we need to reset the list for each source file specified on the command line to the state from cobc command line. 

Therefore either backup the whole list and restore it (similar to cb_source_format) or just use two lists and reset the second before each new source file. I'm ok with both ways but think the two-list approach is easier to handle.

cobc/pplex.l
```diff
@@ -1867,6 +1882,19 @@ int cobc_deciph_source_format (const char *sfname) {
...
+	if (filename[0] == '<' && filename[n-1] == '>' &&
+      filename[n-3] == '.' && filename[n-2] == 'h') {
+			return 0;
+	} else if (filename[n-2] == '.' && filename[n-1] == 'h') {
+		return 0;
+	}
+	else {
+		return 1;
+	}
+}
```
GitMensch, line 1896: That is validation, not deciphering; also: is there a reason to limit to .h files? I'd just pass as-is - we do the same for `--include`, no?`
If the validation should stay in then move to the single file that actually uses this, rename it and check n before breaking with `>>IMP INCLUDE "no"` (access out of bounds)

cobc/scanner.l
```diff
@@ -323,6 +324,14 @@ AREA_A		"#AREA_A"\n
 	cobc_areacheck = 0;
 }
 
+<*>^[ ]?"#INCLUDE".*\n {
+	yytext[strlen(yytext)-1] = '\0';
```
GitMensch, line 328: instead of dropping te last position, just use `.*/\n` instead to not include that in yytext

cobc/scanner.l
```diff
@@ -2583,6 +2603,22 @@ clear_constants (void)
 	top_78_ptr = NULL;
 }
 
+static struct cb_text_list *
+scan_list_add (struct cb_text_list *list, const char *text)
+{
+	struct cb_text_list	*p;
+
+	p = cobc_plex_malloc (sizeof (struct cb_text_list));
+	p->text = cobc_plex_strdup (text);
```
GitMensch, line 2612: we're in the scanner, so this should be `cobc_strdup` and friends

cobc/codegen.c
```diff
@@ -1830,14 +1830,28 @@ output_gnucobol_defines (const char *formatted_date)
...
 		}
+		while (ld) {
+			if (ld->text[0] == '<') {
+				output_line ("#include %s", ld->text);
+			} else {
+				output_line ("#include \"%s\"", ld->text);
+			}
+			next = ld->next;
+			if (next != NULL)
+				free (ld);
```
GitMensch, line 1851: when using `cobc_strdup` and friends, use `cobc_free` here - but shouldn't that be done on both the text (after output) and the list item? Why not doing that for the last?

cobc/scanner.l
```diff
@@ -1297,6 +1306,17 @@ H#[0-9A-Za-z]+ {
...
+<PICTURE_STATE>{
+  "IS" {
+	/* Ignore */
+  }
+  [^ \n;]+ {
+	BEGIN INITIAL;
+	scan_picture (yytext);
+	RETURN_TOK (PICTURE);
+  }
+}
```
GitMensch, line 1318: How is that related?

tests/testsuite.src/syn_misc.at
```diff
@@ -8385,3 +8385,105 @@ prog.cob:18: error: ANY LENGTH items may only be BY REFERENCE formal parameters
...
+[prog.cob:2: error: PROGRAM-ID header missing
+prog.cob:2: error: ENVIRONMENT DIVISION header missing
+prog.cob:2: error: CONFIGURATION SECTION header missing
+prog.cob:2: error: SPECIAL-NAMES header missing
+prog.cob:2: error: invalid system-name 'file2'
+prog.cob:2: error: invalid symbol '#' - skipping word
+prog.cob:2: error: invalid system-name 'h'
+prog.cob:4: error: syntax error, unexpected IDENTIFICATION, expecting CRT or Identifier
+prog.cob:4: error: syntax error, unexpected DIVISION
+])
```
GitMensch, line 8487: where's the error here?

GitMensch (changes requested):

tests/testsuite.src/syn_misc.at
engboris: Since this directive is part of the syntax of programs (unlike the compiler option), there is apparently no explicit restriction. Hence, I thought of allowing only quoted file names.

Or should I be more permissive and also allow unquoted file names? Should I also assume that GnuCOBOL's philosophy is to be as permissive as possible?

cobc/codegen.c
engboris: Tell me if I'm wrong but the text is of type `const char*` and thus, it shouldn't be freed since it is constant (?).

tests/testsuite.src/syn_misc.at
GitMensch: The directive should take what the command line option takes. I think that if you put garbage in there on the command line, you will get garbage during compile, so the same can apply here. In any case there should be no COBOL parser error like the one that is currently expected here

cobc/codegen.c
GitMensch: If the data comes from `cobc_strdup`, then it is allocated (permanently, not cached), not a constant.
`const char *` may be a wrong type.

cobc/codegen.c
```diff
@@ -1830,14 +1830,31 @@ output_gnucobol_defines (const char *formatted_date)
...
+				output_line ("#include \"%s\"", ld->text);
+			}
+			next = ld->next;
+			if (ld != NULL) {
+				cobc_free (ld);
+				cobc_free (ld->text);
+				ld = NULL;
+			}
+			ld = next;
+		}
```
GitMensch, line 1856: the code is nearly good (we cannot free ld first, then access its element to free that), but please change it to have no `ld`:

```c
		struct cb_text_list *l;
		for (l = cb_include_file_list; l; l = l->next) {
		  ...
		}
		l = cb_include_file_list_directive;
		while (l) {
		  struct cb_text_list last = l;		
		  ...
		  l = l->next:
		  cobc_free (last->text);
		  cobc_free (last);
		}
```

cobc/cobc.c
```diff
@@ -237,6 +237,7 @@ const char		*cb_storage_file_name = NULL;
 const char		*cb_call_extfh = NULL;
 struct cb_text_list	*cb_copy_list = NULL;
 struct cb_text_list	*cb_include_file_list = NULL;
+struct cb_text_list	*cb_include_file_list_directive = NULL;
```
GitMensch, line 240: this should be moved to scanner.l

cobc/ChangeLog
```diff
@@ -9,6 +9,12 @@
 	  replace stream) and we use a circular buffer for the temporary
 	  queue of tokens instead of a list.
 
+2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
+
+	* cobc.h, pplex.l, ppparse.y: new >>IMP INCLUDE directive to include
```
GitMensch, line 14: please add the other files here as well

tests/testsuite.src/syn_misc.at
```diff
@@ -8385,3 +8385,52 @@ prog.cob:18: error: ANY LENGTH items may only be BY REFERENCE formal parameters
 
 AT_CLEANUP
 
+AT_SETUP([IMP INCLUDE directive])
```
GitMensch, line 8388: please use a double empty line before `AT_SETUP` to structure the file a bit better (also for the other file)

tests/testsuite.src/used_binaries.at
```diff
@@ -1139,3 +1139,34 @@ AT_CHECK([$COMPILE_MODULE -Wno-unfinished --copy "f.copy" -fstatic-call prog2.co
...
+])
+AT_DATA([prog.cob], [
+       >> IMP INCLUDE "file.h"
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       PROCEDURE        DIVISION.
+           CALL "f" USING "Hello".
+])
+
+AT_CHECK([$COBC -m -I . -fstatic-call prog.cob], [1], [], [ignore])
```
GitMensch, line 1157: That duplicates the check for the signature mismatch, as we have that with `--include` already, we can drop it here (note: I hope to check in a change to the `--include` test which _may_ improve the win32 failure this evening).

GitMensch (changes requested):
still some requested changes - but we're on the finish line with this PR

tests/testsuite.src/used_binaries.at
engboris: > As the --include test also has --copy covered - please add the >> IMP INCLUDE to the same test, adding IMP INCLUDE directive as keywords

I'm not sure to understand this part. Can you give more details, please?

cobc/codegen.c
```diff
@@ -1830,16 +1830,28 @@ output_gnucobol_defines (const char *formatted_date)
...
+		while (l) {
+		  struct cb_text_list *last = l;
+			if (l->text[0] == '<') {
+				output_line ("#include %s", l->text);
+			} else {
+				output_line ("#include \"%s\"", l->text);
+			}
+		  l = l->next;
+		  cobc_free (last->text);
+		  cobc_free (last);
```
GitMensch, line 1852: minor formatting: please indent with tab instead of space

cobc/scanner.l
```diff
@@ -2772,3 +2798,5 @@ cb_find_defined_program_by_id (const char *orig_id)
 
 	return NULL;
 }
+
+
```
GitMensch, line 2802: please drop the additional newlines

tests/testsuite.src/used_binaries.at
GitMensch: please add this test into the previous one (a separate AT_DATA with AT_TEST call + the two additional entries in AT_KEYWORDS)

cobc/codegen.c
```diff
@@ -1830,16 +1830,28 @@ output_gnucobol_defines (const char *formatted_date)
...
+			if (l->text[0] == '<') {
 				output_line ("#include %s", l->text);
 			} else {
 				output_line ("#include \"%s\"", l->text);
 			}
 		}
-	}
 
+		l = cb_include_file_list_directive;
+		while (l) {
```
GitMensch, line 1843: What is the reason for using a "full for loop" above (that's fine) and here go with a different one?`If there isn't any please use `for` in both cases.

GitMensch (changes requested):
changes to the testcases needed, overal fine

tests/testsuite.src/used_binaries.at
```diff
@@ -1171,4 +1171,20 @@ AT_CHECK([$COMPILE_MODULE -Wno-unfinished --copy "f.copy" -fstatic-call prog2.co
...
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 long USAGE BINARY-C-LONG.
+       PROCEDURE        DIVISION.
+           CALL "f" USING "Hello" BY VALUE long RETURNING NOTHING.
+])
+
+AT_CHECK([$COBC -m -I . -fstatic-call prog2.cob], [0], [], [])
```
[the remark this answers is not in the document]
engboris: Could you be more specific? I'm not sure to see what you have in mind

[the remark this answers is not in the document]
engboris: Are you saying I should comment the test? And have
```
# static build with correct function signature
#AT_CHECK([$COBC -m -I . -fstatic-call prog2.cob], [0], [], [])
```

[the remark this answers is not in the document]
engboris: friendly ping

engboris:
@GitMensch Could you check this? It was almost finished and it was a rather minor change but I wasn't sure about the tests. Thank you!

cobc/ppparse.y
```diff
@@ -1368,6 +1373,29 @@ leap_second_directive:
...
+  /* GnuCOBOL 3.3 extension */
+  INCLUDE imp_include_sources
+  {
+	struct cb_text_list *p = $2;
+	while (p != NULL) {
+		fprintf (ppout, "#INCLUDE %s\n", p->text);
+		p = p->next;
+	}
+  }
+;
```
GitMensch, line 1386: please add error handling here,, something like a variant `errror` where you reset the error; this should help that the later parser does not output "missing program-id" (see your test case)

GitMensch (changes requested):
code changes requested for @engboris - I'll adjust the used_binaries.at, so please don't force-push

GitMensch:
I've rebased this PR and adjusted the "real" test for IMP INCLUDE. There are still open changes (see review notes), but I'm sure @engboris will have a look at those over the next days (or hours)?

cobc/ppparse.y
```diff
@@ -1368,6 +1373,33 @@ leap_second_directive:
 | OFF
 ;
 
+imp_directive:
+  /* GnuCOBOL 3.3 extension */
+  INCLUDE
+  {
+	ppparse_error (_("Missing argument for IMP INCLUDE directive"));
```
engboris, line 1380: I'm not sure if this was what you expected. I only get this error when arguments are missing and not others. You talked about "resetting the error".

cobc/codegen.c
```diff
@@ -1830,16 +1830,29 @@ output_gnucobol_defines (const char *formatted_date)
...
 			}
 		}
-	}
 
+		for (l = cb_include_file_list_directive; l; l = l->next) {
+			if (last != NULL) {
+				cobc_free (last->text);
+				cobc_free (last);
+			}
+			last = l;
```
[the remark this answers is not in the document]
engboris: If I understand correctly, I should free those variables outside/independently of codegen but still after the running of codegen? Maybe after `ylex_clear_all ()` in `cobc.c` and add a `parse_clear_all ()` function or something like that?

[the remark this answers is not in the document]
GitMensch: The thought was to drop all the explicit `cobc_free` calls and move this to "parser memory" using `cobc_parse_malloc` instead for the allocation; then, potentially at ylex_clear_all or clear_local_codegen_vars unset the var...

Can you please check if this works?

Note: you currently allocate `cb_include_file_list_directive` with `cobc_parse_strdup` - so it is parser allocated already and will automatically be freed - you may only need to drop setting the "last" attribute (or create a separate structure) and the free calls here.

cobc/ppparse.y
GitMensch: I've meant to let the parser do the error output, but catch that (so it doesn't fall through and consumes more):
instead of INCLUDE and nothing (=manually catched) you do `INCLUDE error` and do an yyerrok call (see existing examples).
As the directive is `>> IMP` and we only support one of those per line, it may be more reasonable to do that kind of error handling there -> just directly a single `error`.

You should be able to tinker with that, seeing what the difference in the testsuite it makes.

cobc/scanner.l
```diff
@@ -323,6 +326,13 @@ AREA_A		"#AREA_A"\n
 	cobc_areacheck = 0;
 }
 
+<*>^[ ]?"#INCLUDE".*/\n {
```
GitMensch, line 329: I think we shouldn't allow an optional leading space here (and also drop it for everything that is "new" like the areacheck) but already scan the trailing one.

tests/testsuite.src/syn_misc.at
```diff
@@ -8405,3 +8405,56 @@ prog.cob:18: error: ANY LENGTH items may only be BY REFERENCE formal parameters
...
+AT_SETUP([IMP INCLUDE directive multiple files])
+AT_KEYWORDS([IMP INCLUDE])
+
+AT_DATA([prog.cob], [
+       >>IMP INCLUDE "file1.h" file2.h <file3.h> "<file4.h>"
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. PROG.
+       PROCEDURE DIVISION.
+           GOBACK.
+])
```
GitMensch, line 8456: You possibly want to do --save-temps and `$GREP` for "file" afterwards to verify here that it was included correctly.

GitMensch (changes requested):
overall a good state, we should be able to add the missing pieces and be able to upstream that, soon

cobc/scanner.l
engboris: "new" relatively to what?

GitMensch: Every internal directive newer than GC 3.1.2 - because afterwards we only generated _without_ leading space, so it is reasonable not to check those entries for an optional space.

... we could also drop it from any internal directive and make an entry in the NEWS that source previously processed may need to be adjusted to be processed by GnuCOBOL 3.3.

cobc/codegen.c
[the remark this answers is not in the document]
engboris: > you may only need to drop setting the "last" attribute (or create a separate structure) and the free calls here.

Looks like it works with that with no leak from valgrind. I didn't unset any variable (tell me if it is necessary).

engboris:
@GitMensch friendly ping

NEWS
```diff
@@ -6,6 +6,10 @@ NEWS - user visible changes				-*- outline -*-
 
 * New GnuCOBOL features
 
+** the leading space for all internal directives is removed in the lexer.
+   Source previously processed may need to be ajusted to be processed by
```
GitMensch, line 10: typo: "adjusted"; the whole block needs to be moved down to "Changes that potentially effect recompilation of existing programs:" and can skip the "by gc 3.3" part; as this is only about **pre**processed files  (please adjust the text) that whole block is only important for the NEWS file if any GC version created leading spaces before the internal directives - is this the case? (To be sure it should be enough to test 1 file with some directives in GC 2.2, 3.1 and 3.2, I _think_ there weren't such cases)

cobc/scanner.l
GitMensch, line 1: while this possibly is not relevant for NEWS - please add an entry for this in cobc/Changelog

cobc/cobc.h
```diff
@@ -478,8 +478,10 @@ extern int			cb_depend_keep_missing;
 extern int			cb_flag_copybook_deps;
 extern struct cb_text_list	*cb_depend_list;
 extern struct cb_text_list	*cb_copy_list;
-extern struct cb_text_list	*cb_include_file_list;
+extern struct cb_text_list	*cb_include_file_list; /* global */
+extern struct cb_text_list	*cb_include_file_list_directive; /* temporary */
 extern struct cb_text_list	*cb_include_list;
+extern struct cb_text_list	*cb_include_file_list_directive;
```
GitMensch, line 484: that's a duplicate line, no?

tests/testsuite.src/syn_misc.at
```diff
@@ -8414,3 +8414,60 @@ prog.cob:18: error: ANY LENGTH items may only be BY REFERENCE formal parameters
...
+
```
GitMensch, line 8466: If you add prog2.cob here without any 2>>IMP INCLUDE" the prog2.c has those includes as well (or cobc aborts), no?

I _think_ `cb_include_file_list_directive` should be reset to `NULL` before the next non-nested program is generated, similar to how the source format is reset.

cobc/ChangeLog
```diff
@@ -125,6 +125,12 @@
 	  compiler aborts on broken expressions, bugs #933, #938 and #966
 	* typeck.c: minor refactoring within functions
 
+2024-04-25  Boris Eng <boris.eng@ocamlpro.com>
+
+	* pplex.l, ppparse.y, scanner.l, cobc.h, codegen.c:
+	  new >>IMP INCLUDE directive to include one or multiple header files in
```
GitMensch, line 131: please mention FR #176 here

cobc/ChangeLog
```diff
@@ -125,6 +125,12 @@
...
+	  the generated C code (same behavior as the --include compiler option)
```
GitMensch, line 132: "as the --include option, but per file instead of for all files)"  [that's too long, you likely find a better wording]

GitMensch (changes requested):
_nearly_ done, I'm positive that we reach the commit-state tomorrow :-)

Thanks for working on this again!

NEWS
engboris: I compiled some tests from the testsuite with the `-E` option. The directives appear but no leading space for 2.2, 3.1 and 3.2. Should I remove the whole block from the NEWS file?

GitMensch: In this case I'd suggest to do that - we still have Changelogs for these kind of things :-)

tests/testsuite.src/syn_misc.at
engboris: I'm not sure to understand. Are you thinking of something like this:

```cobol
       >>IMP INCLUDE "file1.h" file2.h <file3.h> "<file4.h>"
       IDENTIFICATION DIVISION.
       PROGRAM-ID. prog1.
       PROCEDURE DIVISION.
           STOP RUN.
       END PROGRAM prog1.

       IDENTIFICATION DIVISION.
       PROGRAM-ID. prog2.
       PROCEDURE DIVISION.
           STOP RUN.
       END PROGRAM prog2.
```
?

In this case, the includes only occur once.

GitMensch: No, something like `prog1.cob prog2.cob`, which will generate two independent C files.

engboris:
Should be good now. You were right, cobc aborted. I added new checks:

```
AT_SETUP([IMP INCLUDE directive multiple files])
AT_KEYWORDS([IMP INCLUDE])

AT_DATA([prog.cob], [
       >>IMP INCLUDE "file1.h" file2.h <file3.h> "<file4.h>"
       IDENTIFICATION DIVISION.
       PROGRAM-ID. PROG.
       PROCEDURE DIVISION.
           GOBACK.
])

AT_DATA([prog2.cob], [
       IDENTIFICATION DIVISION.
       PROGRAM-ID. prog.
       PROCEDURE DIVISION.
           GOBACK.
])

AT_CHECK([$COMPILE --save-temps -E -o prog.i prog.cob], [0], [], [])
AT_CHECK([$COMPILE -C prog.cob prog2.cob], [0], [], [])
AT_CHECK([$GREP "#INCLUDE file1.h"   prog.i], [0], ignore, [])
AT_CHECK([$GREP "#INCLUDE file2.h"   prog.i], [0], ignore, [])
AT_CHECK([$GREP "#INCLUDE <file3.h>" prog.i], [0], ignore, [])
AT_CHECK([$GREP "#INCLUDE <file4.h>" prog.i], [0], ignore, [])
AT_CHECK([$GREP "#INCLUDE file1.h"   prog.c], [1], ignore, [])
AT_CHECK([$GREP "#INCLUDE file2.h"   prog.c], [1], ignore, [])
AT_CHECK([$GREP "#INCLUDE <file3.h>" prog.c], [1], ignore, [])
AT_CHECK([$GREP "#INCLUDE <file4.h>" prog.c], [1], ignore, [])
```

engboris:
I did some mistakes while rebasing and fixing conflicts. I'm trying to fix it now.

engboris:
@GitMensch Could you do a last quick check? I rebased and fixed conflicts but I may have introduced unwanted code although it is unlikely (all the tests are successful). I'm sorry for the inconvenience!

GitMensch:
That's fine. [I've just checked the last "Compare", then the additional fix]

ddeclerck:
What is the status of this PR ? Is it ready to be merged ?

engboris:
> What is the status of this PR ? Is it ready to be merged ?

Yes, it should be ready (already marked with "Ready for SVN")

ddeclerck:
> > What is the status of this PR ? Is it ready to be merged ?
> 
> Yes, it should be ready (already marked with "Ready for SVN")

Ah, yeah, I overlooked the "Ready for SVN" tag.
So I guess I can merge this to SVN right away, right @GitMensch ?

GitMensch:
yes, just waits for the commit since two months

ddeclerck:
Merged in SVN @ 5448.

PR #144, opened 2024-05-13
Fix MSVC builds - missing profiling.c

ddeclerck:
Follow-up of https://github.com/OCamlPro/gnucobol/pull/110#issuecomment-2091486360.

Add missing profiling.c file in MSVC projects.
Also fix usage of preprocessor directives inside macro in `flag.def` (MSVC does not seem to like that).

### DIFF
--- a/build_windows/ChangeLog.txt
+++ b/build_windows/ChangeLog.txt
@@ -1,4 +1,8 @@
 
+2024-05-13  David Declerck <david.declerck@ocamlpro.com>
+
+	* general for libcob: add missing profiling.c
+
 2023-07-07  Simon Sobisch <simonsobisch@gnu.org>
 
 	* general for cobc: include new replace.c
--- a/cobc/flag.def
+++ b/cobc/flag.def
@@ -119,11 +119,16 @@ CB_FLAG (cb_flag_winmain, COB_SHOW, "winmain",
 #undef COB_SHOW
 
 #ifdef HAVE_ATTRIBUTE_CONSTRUCTOR
-CB_FLAG_ON (cb_flag_use_constructor, 1, "use-constructor",
+#define CB_FLAG_M CB_FLAG_ON
+#define COB_SHOW 1
 #else
-CB_FLAG (cb_flag_use_constructor, 0, "use-constructor",
+#define CB_FLAG_M CB_FLAG
+#define COB_SHOW 0
 #endif
+CB_FLAG_M (cb_flag_use_constructor, COB_SHOW, "use-constructor",
 	_("  -fuse-constructor     generate internal one-time code via constructor"))
+#undef CB_FLAG_M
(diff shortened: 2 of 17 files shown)


build_windows/ChangeLog.txt
```diff
@@ -1,4 +1,8 @@
 
+2024-05-13  David Declerck <david.declerck@ocamlpro.com>
+
+	* vs20xx/libcob.vcxproj[.filters]: add missing profiling.c
```
GitMensch, line 4: we don't use wildcards, as this is quite global, the "global" entry as below would be fine

cobc/flag.def
```diff
@@ -120,10 +120,11 @@ CB_FLAG (cb_flag_winmain, COB_SHOW, "winmain",
 
 #ifdef HAVE_ATTRIBUTE_CONSTRUCTOR
 CB_FLAG_ON (cb_flag_use_constructor, 1, "use-constructor",
+	_("  -fuse-constructor     generate internal one-time code via constructor"))
 #else
 CB_FLAG (cb_flag_use_constructor, 0, "use-constructor",
-#endif
 	_("  -fuse-constructor     generate internal one-time code via constructor"))
+#endif
```
GitMensch, line 127: thank you for inspecting this - please use the `COB_SHOW` variant as in the item above instead

GitMensch (changes requested):
some minor changes necessary; please do the adjustment also to the ocide folder

after these changes are done, feel free to commit upstream directly

cobc/flag.def
ddeclerck: Question though : the two branches do not use the same macro (`CB_FLAG_ON` vs `CB_FLAG`) ; is this intentional ?

GitMensch: yes, we enable it by default and show it in the help if it is supported, otherwise disable by default and hide from the help

ddeclerck: So if the two branches must use different macros, I can't really proceed as you did with `COB_SHOW`, right ?

GitMensch:
Yes. You can also use a different approach, but ideally we have the help text only once.

cobc/flag.def
ddeclerck: Should be okay now.

GitMensch (approved):
please send upstream - I'll also run the appveyor builds then

GitMensch:
Would it be reasonable to get back to #140 after checking in Win32+MacOS changes?

ddeclerck:
Merged upstream.

> Would it be reasonable to get back to #140 after checking in Win32+MacOS changes?

I would like to, but I have things to do for our main client first (in particular : PR #136 , collation support on indexed files in GCSORT, and the GC3 => GC4 forward porting).

PR #148, opened 2024-05-30
Fix warnings in GC4

ddeclerck:
@GitMensch Here are my changes to fix the warnings on the GC4 branch. If that's okay I'll commit that to SVN (with an appropriate ChangeLog entry).

### DIFF
--- a/bin/cobfile.c
+++ b/bin/cobfile.c
@@ -351,7 +351,7 @@ dropField (cob_field *f)
 
 /* Does the 'word' match within the string */
 static int
-matchWord (const char *word, unsigned char *defs, char *str, int *pos )
+matchWord (const char *word, char *defs, char *str, int *pos )
 {
 	int	j, k, doeq, doend;
 	int	ln = strlen (word);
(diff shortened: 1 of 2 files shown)


GitMensch:
For the two new files that looks good. For call.c - why don't we have that issue in 3.x?

ddeclerck:
> For the two new files that looks good. For call.c - why don't we have that issue in 3.x?

Actually the problem also occurs in GC3, if you explicitly disable NLS (`configure --disable-nls`). In GC4 the strings where these error occurs are not localized. Makes sense, the truncation analysis can not deduce the size of strings returned by `gettext`.

Maybe you just want to turn this analysis off ?

GitMensch:
OK, if that warning is visible also in 3.x, then commit the changes to the other two files upstream and we can consider this PR finished.

ddeclerck:
> OK, if that warning is visible also in 3.x, then commit the changes to the other two files upstream

Not sure what you want me to do with the third one (`libcob/call.c`) ; should I commit it as well ?

GitMensch:
no, let's inspect that later and have a separate fix (so we can easily merge it)

ddeclerck:
Alright.

Interestingly, the `unsigned` qualifiers were added in a previous commit to solve warnings:
[https://sourceforge.net/p/gnucobol/code/4993/](https://sourceforge.net/p/gnucobol/code/4993/)
My PR actually undoes these changes. Pretty weird.

GitMensch:
oops..

GitMensch:
The fields that are passed to `isspace()` and friends need to stay unsigned.

ddeclerck:
> The fields that are passed to `isspace()` and friends need to stay unsigned.

Indeed, I overlooked this. I'll just add appropriate casts when needed (as suggested by `man`).

PR #149, opened 2024-06-11
Initial PR for jni integration

xevor11:
An initial PR to get feedback on style, standard, and structuring functions. Also checking if I am in the correct direction for the JNI integration steps.


GitMensch:
Before more checking: please add the file references to libcob/Makefile.am and build_windows/* so it is already compiled and linked in the CI.

Am 12. Juni 2024 00:39:52 MESZ schrieb Vedant Tewari ***@***.***>:

libcob/java.c
```diff
@@ -0,0 +1,183 @@
...
+        (*jvm)->DestroyJavaVM(jvm);
+        return;
+    }
+    
+    jmethodID mid = get_from_cache(cls, methodName, methodSig);
+    if (mid == NULL) {
+        char* signature = generate_method_signature(paramTypes, paramCount, returnType);
+        mid = (*env)->GetStaticMethodID(env, cls, methodName, signature);
+        if (mid == NULL) {
+            printf("Method not found\n");
```
GitMensch, line 119: printf is ok for intermediate versions (but even then you possibly will go for the runtime_warning), for later you possibly want a COBOL exception to be raised

libcob/java.c
```diff
@@ -0,0 +1,183 @@
...
+            free(signature);
+            (*jvm)->DestroyJavaVM(jvm);
+            return;
+        }
+        add_to_cache(cls, mid);
+        free(signature);
+    }
+    cob_call_static_method(env, jvm, cls, mid);
```
GitMensch, line 127: better separation needed: function name "lookup" should not include "call"

libcob/java.c
```diff
@@ -0,0 +1,183 @@
...
+
+    const char *nativeResult = (*env)->GetStringUTFChars(env, result, 0);
+
+    (*env)->ReleaseStringUTFChars(env, input, nativeInput);
+    (*env)->ReleaseStringUTFChars(env, result, nativeResult);
+
+    free(methodSig);
+}
+
+// extern "C"
```
GitMensch, line 163: style issue: for commenting out blocks use `#if 0 /* comment why not active */`

GitMensch (changes requested):
You likely want at least one "internal entry point" (libcob/coblocal.h) to show the function(s) that will be called from libcob/call.c (and possibly a cleanup function to be called from common.h, even if that is only a stub).

Before more checking: please add the file references to libcob/Makefile.am and build_windows/* so it is already compiled and linked in the CI.

GitMensch:
Most important: Please rebase and retarget, you want to both branch from and target this PR to the [java-interop](https://github.com/OCamlPro/gnucobol/tree/java-interop) branch, not [gcos4gnucobol-3.x](https://github.com/OCamlPro/gnucobol/tree/gcos4gnucobol-3.x)

libcob/jni.h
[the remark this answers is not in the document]
xevor11: So in general the jni functions that are being used to initialize the VM, identify the method and classes, I'm curious how they are used internally by the configure script?

xevor11:
I see, I will make another PR with the target to the correct branch

libcob/jni.h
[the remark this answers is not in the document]
GitMensch: the configure would just use the header check (like it is done with curses) to recognize if the headers are available, similar with the needed library

PR #150, opened 2024-06-12
requested changes to java.c, configure.ac, and internal state to cobl…

xevor11:
Rebased and Made requested changes in PR #149

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -468,6 +468,23 @@ AC_DEFUN([AC_PROG_CXX], [])
 AC_PROG_LN_S
 AC_PROG_INSTALL
 
+PRINT_VAR(JAVA_HOME)
+PRINT_VAR(JAVA)
+PRINT_VAR(JAVAC)
+PRINT_VAR(JAVAH)
+PRINT_VAR(JAR)
+PRINT_VAR(JNI_INCLUDES)
+
+AC_PATH_PROG(javahome,java, "/usr/bin")
+javahome=`dirname ${javahome}`
+javahome=`dirname ${javahome}`
+echo "JAVAHOME=${javahome}"
+for each in `ls ${javahome}/include` ;do
+  if [ test -d ${javahome}/include/${each} ] ;then
+     javaos=${each}
+  fi;
+done;
+
 AC_PROG_MAKE_SET
 AC_LIB_RPATH
 
@@ -480,6 +497,10 @@ dnl AC_CHECK_HEADERS([stdint.h whcar.h malloc.h])
 # mandatory:
 AC_CHECK_HEADERS([sys/types.h signal.h stddef.h], [],
 	[AC_MSG_ERROR([mandatory header could not be found or included])])
+AC_CHECK_HEADERS([jni.h], [],
+  [AC_MSG_ERROR([jni.h is required but could not be found])])
+AC_CHECK_HEADERS([jni_md.h], [],
+  [AC_MSG_ERROR([jni_md.h is required but could not be found])])
 # optional:
 AC_CHECK_HEADERS([sys/time.h locale.h fcntl.h dlfcn.h sys/wait.h sys/sysmacros.h])
 
--- a/libcob/Makefile.am
+++ b/libcob/Makefile.am
@@ -19,6 +19,12 @@
 # You should have received a copy of the GNU General Public License
 # along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
 
+lib_LTLIBRARIES = libcob.la
+libcob_la_SOURCES = common.c move.c numeric.c strings.c \
+	fileio.c call.c intrinsic.c termio.c screenio.c reportio.c cobgetopt.c \
+	java.c \
+	mlio.c coblocal.h cconv.c system.def profiling.c
+
 if LOCAL_CJSON
 nodist_libcob_la_SOURCES = cJSON.c
 DISTCLEANFILES = cJSON.c cJSON.h
--- a/libcob/coblocal.h
+++ b/libcob/coblocal.h
@@ -1,6 +1,6 @@
 /*
-   Copyright (C) 2007-2012, 2014-2022 Free Software Foundation, Inc.
-   Written by Roger While, Simon Sobisch, Ron Norman
+   Copyright (C) 2007-2012, 2014-2024 Free Software Foundation, Inc.
+   Written by Roger While, Simon Sobisch, Ron Norman, Vedant Tewari
 
    This file is part of GnuCOBOL.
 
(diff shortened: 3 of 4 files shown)


configure.ac
```diff
@@ -468,6 +468,23 @@ AC_DEFUN([AC_PROG_CXX], [])
...
+PRINT_VAR(JAVA_HOME)
+PRINT_VAR(JAVA)
+PRINT_VAR(JAVAC)
+PRINT_VAR(JAVAH)
+PRINT_VAR(JAR)
+PRINT_VAR(JNI_INCLUDES)
+
+AC_PATH_PROG(javahome,java, "/usr/bin")
+javahome=`dirname ${javahome}`
+javahome=`dirname ${javahome}`
```
GitMensch, line 480: Why twice?
Do we  need this for anything later on?

configure.ac
```diff
@@ -480,6 +497,10 @@ dnl AC_CHECK_HEADERS([stdint.h whcar.h malloc.h])
 # mandatory:
 AC_CHECK_HEADERS([sys/types.h signal.h stddef.h], [],
 	[AC_MSG_ERROR([mandatory header could not be found or included])])
+AC_CHECK_HEADERS([jni.h], [],
+  [AC_MSG_ERROR([jni.h is required but could not be found])])
+AC_CHECK_HEADERS([jni_md.h], [],
+  [AC_MSG_ERROR([jni_md.h is required but could not be found])])
```
GitMensch, line 503: We can leave this as-is for now, later on it should only be done if `--with-java` is specified.
You may opt to do this part after your mid-term evaluation, whenever you do check how `--with-xml2` is handled.

libcob/Makefile.am
```diff
@@ -19,6 +19,12 @@
 # You should have received a copy of the GNU General Public License
 # along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
 
+lib_LTLIBRARIES = libcob.la
+libcob_la_SOURCES = common.c move.c numeric.c strings.c \
+	fileio.c call.c intrinsic.c termio.c screenio.c reportio.c cobgetopt.c \
+	java.c \
+	mlio.c coblocal.h cconv.c system.def profiling.c
```
GitMensch, line 26: you likely just wanted to add java.c to the list below; later on you _may_ include it conditional on a variable (similar but not identical to libodbci).

In any case that's an edjustment, so please add `,2024` to the copyright year of this file

libcob/coblocal.h
```diff
@@ -1,6 +1,6 @@
 /*
-   Copyright (C) 2007-2012, 2014-2022 Free Software Foundation, Inc.
-   Written by Roger While, Simon Sobisch, Ron Norman
+   Copyright (C) 2007-2012, 2014-2024 Free Software Foundation, Inc.
```
GitMensch, line 2: we don't just include year numbers, so until @ddeclerck has reached a state where he considers to have many important commit merges done (or you miss a specific 3.x commit that he already merged) and we rebase the [java-interop](https://github.com/OCamlPro/gnucobol/tree/java-interop) branch which likely will add 2023 commits, the change should say ,2024`.

libcob/coblocal.h
```diff
@@ -449,6 +449,29 @@ COB_HIDDEN const char	*cob_get_last_exception_name	(void);
...
+	CCM_NONE,
+	CCM_LOWER,
+	CCM_UPPER,
+	CCM_LOWER_LOCALE,
+	CCM_UPPER_LOCALE
+};
+COB_HIDDEN unsigned char	cob_toupper (const unsigned char);
+COB_HIDDEN unsigned char	cob_tolower (const unsigned char);
+COB_HIDDEN int		cob_field_to_string	(const cob_field *, void *,
+						 const size_t, const enum cob_case_modifier target_case);
```
GitMensch, line 474: something has gone wrong here - see the same content directly above

GitMensch (changes requested):
please add an entry to DEPENDENCIES about the new optional dependency, and an entry to ChangeLog and libcob/ChangeLog; this will also make it easier to understand what you added why

Please just update your source branch, currently jni-correct and those commits will land in here.

configure.ac
xevor11: That was by accident, however I was slightly confused about the formatting

libcob/coblocal.h
xevor11: I do not see any duplicates in the file?

xevor11:
Added the requested changes

libcob/coblocal.h
GitMensch: cob_field_to_string is duplicate, cob_tolower/cob_toupper/cob_get_strerror are shown to be added while not related to this PR

GitMensch:
Do you have a first working call to Java (which likely includes and adjustment to call.c as well), maybe `System.getProperty("java.version")`?

libcob/coblocal.h
xevor11: I might have changed the line number causing get strerror and the enum to appear in the PR

xevor11:
More changes for the following:

Integrated JNI with java.c by adding types and data structures in java.c for bidirectional communication

Working towards:

Do you have a first working call to Java (which likely includes and adjustment to call.c as well), maybe System.getProperty("java.version") which is then DISPLAYed by COBOL?

cobc/codegen.c
```diff
@@ -146,6 +146,8 @@ static struct literal_list	*literal_cache = NULL;
 static struct field_list	*field_cache = NULL;
 static struct field_list	*local_field_cache = NULL;
 static struct call_list		*call_cache = NULL;
+extern JavaVM *jvm;
+extern JNIEnv *env;
```
nberth, line 150: I'm not sure these are needed in the compiler (`cobc`) itself.

cobc/codegen.c
```diff
@@ -387,6 +389,21 @@ lookup_source (const char *p)
 	return source_id++;
 }
 
+static void lookup_java_call(const char *p) {
+    struct call_list *clp;
+
+    for (clp = call_cache; clp; clp = clp->next) {
```
nberth, line 395: I think it's another list you need to manipulate in this function. Something like `call_java_cache`.

configure.ac
```diff
@@ -480,9 +496,34 @@ dnl AC_CHECK_HEADERS([stdint.h whcar.h malloc.h])
...
 # optional:
 AC_CHECK_HEADERS([sys/time.h locale.h fcntl.h dlfcn.h sys/wait.h sys/sysmacros.h])
-
+AC_ARG_VAR([JAVA_HOME], [Java Runtime Environment (JRE) location])
+AC_ARG_ENABLE([java],
+              [AC_HELP_STRING([--disable-java],
+                              [disable Java Interoperatibility])])
```
nberth, line 508:
```suggestion
                              [disable Java Interoperability])])
```

xevor11:
@GitMensch Here are the final changes, we fixed issues with autoconf, config and when `make` it runs without errors

nberth:
> @GitMensch Here are the final changes, we fixed issues with autoconf, config and when `make` it runs without errors. We need to write some tests, but after that we can merge

I guess you want to check #157 instead. This PR (#150) is outdated and can probably be closed.

PR #153, opened 2024-06-20
XML PARSE Review

ammaralmorsi:


ddeclerck:
Hi !
Quick note: we're trying to move development from the GC3 branch to GC4 ; maybe it would be better for you to target GC4 (master in this repo) ?

GitMensch:
Thanks for having a quick look. But be aware that the two GSOC projects with the small project size go to 3.x first, only java-interop is trunk only.

GitMensch:
as this project was taken out of GSOC, we had to do that "one day" upstream, which is now dowe with a good progess, so closing here

Issue #154, opened 2024-07-04
Error "`bdb_bt_compare` was given keys of different length" is actually triggered


ddeclerck:
Yeah, that's in some code I wrote 😅

This error should not happen, yet it does happen in one of our customer use case.
I was provided the program that reads the "offending" indexed file, but I'm still expecting the program that creates it so I can analyze further.

Side note : the error does not trigger termination ; should we add a call to `cob_hard_failure` after `cob_runtime_error` ?

ddeclerck:
Interestingly, the indexed BDB file has been produced by an external tool. Analyzing with `db_dump` reveals some internal pages containing short data (length 3/4), possibly interpreted as keys by GnuCOBOL (not sure what those are - maybe partial/intermediate keys ? - I don't know about BDB's internals). Feeding this BDB file through GCSORT using a COPY operation generates a new BDB file that does not contain these short datas, but preserving the actual file contents.

Anyways, should GnuCOBOL accept these files ? If so, the fix is rather easy: we should order keys even if they do not have the same length. I tried it and the given program bbehaves as expected.

ddeclerck:
(probably should have notified you about this one, @GitMensch )

GitMensch:
I've subscribed to all PRs and issue here, so get the _starting_ post by mail :-)

ddeclerck:
> I've subscribed to all PRs and issue here, so get the _starting_ post by mail :-)

Ah, alright, so eyes everywhere 👀
:D

GitMensch:
I see, we speak about https://github.com/OCamlPro/gnucobol/blob/39344cf66085620e1dfdb16d54c7ba081efd1da8/libcob/fileio.c#L914-L936

> Side note : the error does not trigger termination ; should we add a call to `cob_hard_failure` after `cob_runtime_error` ?

Definitely not within the function (a runtime error may also consist of multiple outputs) but it is commonly used the way that after the first we do call the second function.

> Anyways, should GnuCOBOL accept these files ?

Depends. in general the `ORGANIZATION INDEXED` format has an internal structure, so if it is effort or possibly costly, then we may should either not do that or not by default.
As-is we already have the size check there and would within this case do what `cob_cmp_alnum()` does (which we could  extract from there into two functions and the "inner one" (that gets 2 times data pointer 2 sizes and one optional col) be made available via coblocal and use in fileio.c.

GitMensch:
[...]

PR #156, opened 2024-07-09
Handle BDB keys of different length

engboris:
Fix for issue #154 

# Problem

The function `libcob/fileio.c:bdb_bt_compare` fails on keys of different length passed in arguments. This is necessary since valid BDB files containing additional data interpreted as keys of different length may be used with GnuCOBOL  (example given from one of our customers). Such files may be produced by external tools but still meant to be readable.

# Solution

Use the already implemented `libcob/common.c:cob_cmp_alnum` function (now made external) to compare keys as alphanumeric fields of possibly different length (padded with spaces).

This PR introduces a new preparser flag `USE_BDB_KEYDIFF` with may be passed in the following way when compiling GnuCOBOL sources:
```
CPPFLAGS=-DUSE_BDB_KEYDIFF ./configure
make
```

This flag allows such BDB keys of different length as an option since it has an overhead cost.

#  Notes

I have also been suggested to

> also include the variant that operates on not 100% confirming BDB files (always using a sort function) (from @GitMensch, cf. issue #154)

Which is something I do not currently understand.

### DIFF
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -20,6 +20,12 @@
 	  (indirect_move): fixed bug when calculating scale of 
 	  the intermediate cob field
 
+2024-07-09  Boris Eng <boris.eng@ocamlpro.com>
+
+	* fileio.c (bdb_bt_compare, indexed_open): handle BDB keys of different
+	  length with a flag USE_BDB_KEYDIFF (passed with preparser flag CPPFLAGS)
+	* common.c (cob_cmp_strings), coblocal.h (cob_cmp_strings):
+	  extracted from (cob_cmp_alnum)
 
 2024-07-02  Chuck Haatvedt <chuck.haatvedt+cobol@gmail.com>
 
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -1975,16 +1975,14 @@ cob_cmp_all (cob_field *f1, cob_field *f2)
 	return ret;
 }
 
-/* compare content of field 'f1' to content of 'f2', space padded,
-   using the optional collating sequence of the program */
-static int
-cob_cmp_alnum (cob_field *f1, cob_field *f2)
+/* compare string 'data1' to string 'data2', of size 'size1' and 'size2'
+   respectively, space padded, using a given collating sequence */
+int
+cob_cmp_strings (
+	unsigned char* data1, unsigned char* data2,
+	size_t size1, size_t size2,
+	const unsigned char *col)
 {
-	const unsigned char	*col = COB_MODULE_PTR->collating_sequence;
-	const unsigned char	*data1 = COB_FIELD_DATA (f1);
-	const unsigned char	*data2 = COB_FIELD_DATA (f2);
-	const size_t		size1 = COB_FIELD_SIZE (f1);
-	const size_t		size2 = COB_FIELD_SIZE (f2);	
 	const size_t	min = (size1 < size2) ? size1 : size2;
 	int		ret;
 
@@ -2023,6 +2021,20 @@ cob_cmp_alnum (cob_field *f1, cob_field *f2)
 	}
 
 	return 0;
+}
+
+/* compare content of field 'f1' to content of 'f2', space padded,
+   using the optional collating sequence of the program */
+static int
+cob_cmp_alnum (cob_field *f1, cob_field *f2)
+{
+	return cob_cmp_strings (
+		(unsigned char *)COB_FIELD_DATA (f1),
+		(unsigned char *)COB_FIELD_DATA (f2),
+		COB_FIELD_SIZE (f1),
+		COB_FIELD_SIZE (f2),
+		COB_MODULE_PTR->collating_sequence
+	);
 }
 
 /* comparision of all key fields for SORT (without explicit collation)
--- a/libcob/fileio.c
+++ b/libcob/fileio.c
@@ -921,19 +921,24 @@ bdb_bt_compare (DB *db, const DBT *k1, const DBT *k2
 {
 	const unsigned char *col = (unsigned char *)DBT_GET_APP_DATA (k1);
 	COB_UNUSED (db);
-
+#ifdef USE_BDB_KEYDIFF /* flag passed with CPPFLAGS */
+	return cob_cmp_strings (k1->data, k2->data, (size_t)k1->size, (size_t)k2->size, col);
+#else
 	/* LCOV_EXCL_START */
 	if (col == NULL) {
 		cob_runtime_error ("bdb_bt_compare was set but no collating sequence was stored in DBT");
+		cob_hard_failure ();
 	}
 	if (k1->size != k2->size) {
 		cob_runtime_error ("bdb_bt_compare was given keys of different length");
+		cob_hard_failure ();
 	}
 	/* LCOV_EXCL_STOP */
 #if DB_VERSION_MAJOR >= 6
 	locp = NULL;	/* docs: must be set to NULL or corruption can occur ... */
 #endif
 	return indexed_key_compare (k1->data, k2->data, k2->size, col);
+#endif /* USE_BDB_KEYDIFF */
 }
 
 #endif	/* WITH_DB */
@@ -4672,9 +4677,14 @@ dobuild:
 				if (f->keys[i].tf_duplicates) {
 					p->db[i]->set_flags (p->db[i], DB_DUP);
 				}
+				/* TODO: add national compare function later */
+#ifdef USE_BDB_KEYDIFF
+				p->db[i]->set_bt_compare(p->db[i], bdb_bt_compare);
+#else
 				if (f->keys[i].collating_sequence) {
 					p->db[i]->set_bt_compare(p->db[i], bdb_bt_compare);
 				}
+#endif
 			}
 		} else {
 			handle_created = 0;
--- a/libcob/coblocal.h
+++ b/libcob/coblocal.h
@@ -494,6 +494,9 @@ COB_HIDDEN const char	*cob_get_last_exception_name	(void);
 COB_HIDDEN void		cob_parameter_check	(const char *, const int);
 COB_HIDDEN char*        cob_get_strerror (void);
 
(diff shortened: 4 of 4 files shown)


libcob/fileio.c
```diff
@@ -919,20 +919,34 @@ bdb_bt_compare (DB *db, const DBT *k1, const DBT *k2
 )
 {
 	const unsigned char *col = (unsigned char *)DBT_GET_APP_DATA (k1);
+#ifdef USE_BDB_KEYDIFF /* flag passed with CPPFLAGS */
+	COB_MODULE_PTR->collating_sequence = col;
```
ddeclerck, line 923: This changes the current module's collating sequence; we don't want that.

libcob/common.h
```diff
@@ -1932,6 +1932,7 @@ COB_EXPIMP int	cob_bcd_cmp		(cob_field *, cob_field *);
 COB_EXPIMP int	cob_numeric_display_cmp		(cob_field *, cob_field *);
 COB_EXPIMP int	cob_numeric_display_cmp_zero	(cob_field *);
 COB_EXPIMP int	cob_bcd_cmp_zero	(cob_field *);
+COB_EXPIMP int	cob_cmp_alnum	(cob_field *, cob_field *);
```
GitMensch, line 1935: no need to export that outside of libcob - coblocal.h is the better place

libcob/fileio.c
GitMensch: agreed, we also don't want cob_fields to be used here - so extract the logic that only operates on data + size + col from cob_cmp_alnum and locally export it, then calling this here

libcob/fileio.c
```diff
@@ -919,20 +919,34 @@ bdb_bt_compare (DB *db, const DBT *k1, const DBT *k2
...
+	unsigned char *data1 = k1->data;
+	unsigned char *data2 = k2->data;
+	size_t size1 = (size_t)k1->size;
+	size_t size2 = (size_t)k2->size;	
+	const cob_field_attr alphanum = {0x21, 0, 0, 0x0000, NULL};
+	cob_field f1 = {size1, data1, &alphanum};
+	cob_field f2 = {size2, data2, &alphanum};
+	return cob_cmp_alnum (&f1, &f2);
+#else
 	COB_UNUSED (db);
```
GitMensch, line 933: you still want this unused line, so have that up front outside the `#if`

GitMensch (changes requested):
the main open point is the previously unclear registration:

when `USE_BDB_KEYDIFF` is used then `bdb_bt_compare` needs to be always registered (it currently is only registered if there is a col in use)

For your example: parameters should be passed as arguments to configure, instead of being set inline/exported (if you are interested why: the autotools manual has an entry on that, but the biggest part is that passed arguments are checked to be identical between runs).

GitMensch (changes requested):
Please adjust as noted in the fix-commit (I've screwed up and commented there instead of on the PR).

In any case you want to further add to the ChangeLog.

One thing I do wonder: should `cob_cmp_strings()` have a parameter `int national` and do a different compare for the spaces in this case? `cob_cmp_alnum()` would pass zero, `bdb_bt_compare()` would pass depending on the national flag in k1.

In any case this can be done in a later refactoring on as well.

engboris:
It might be wiser for us to leave that for a later refactoring as there is a need from one of our customer and I'm not very familiar with character sets and national yet (I currently don't know what this `int national` argument here would refer to). What do you think?

GitMensch:
I agree, let's finish this PR just leaving a note at the place where we register the compare function, possibly something like `/* TODO: add national compare function later */`.

libcob/common.c
```diff
@@ -2025,6 +2021,20 @@ cob_cmp_alnum (cob_field *f1, cob_field *f2)
...
+static int
+cob_cmp_alnum (cob_field *f1, cob_field *f2)
+{
+	const unsigned char	*col = COB_MODULE_PTR->collating_sequence;
+	const unsigned char	*data1 = COB_FIELD_DATA (f1);
+	const unsigned char	*data2 = COB_FIELD_DATA (f2);
+	const size_t		size1 = COB_FIELD_SIZE (f1);
+	const size_t		size2 = COB_FIELD_SIZE (f2);	
+	return cob_cmp_strings ((unsigned char *)data1, (unsigned char *)data2,
+		size1, size2, col);
```
GitMensch, line 2035: Is there a reason to not pass the arguments directly without intermediate variables?

libcob/fileio.c
GitMensch: That's still open... just drop that line

libcob/fileio.c
```diff
@@ -920,19 +920,29 @@ bdb_bt_compare (DB *db, const DBT *k1, const DBT *k2
...
 	const unsigned char *col = (unsigned char *)DBT_GET_APP_DATA (k1);
 	COB_UNUSED (db);
-
+#ifdef USE_BDB_KEYDIFF /* flag passed with CPPFLAGS */
+	COB_MODULE_PTR->collating_sequence = col;
+	unsigned char *data1 = k1->data;
+	unsigned char *data2 = k2->data;
+	size_t size1 = (size_t)k1->size;
+	size_t size2 = (size_t)k2->size;
+	return cob_cmp_strings (data1, data2, size1, size2, col);
```
GitMensch, line 929: Is there a reason for intermediate variables here?

libcob/fileio.c
```diff
@@ -920,19 +920,29 @@ bdb_bt_compare (DB *db, const DBT *k1, const DBT *k2
...
 	if (k1->size != k2->size) {
 		cob_runtime_error ("bdb_bt_compare was given keys of different length");
+		cob_hard_failure ();
 	}
 	/* LCOV_EXCL_STOP */
 #if DB_VERSION_MAJOR >= 6
 	locp = NULL;	/* docs: must be set to NULL or corruption can occur ... */
 #endif
 	return indexed_key_compare (k1->data, k2->data, k2->size, col);
+#endif
```
GitMensch, line 945: That's a longer ifdef with nesting - I suggest (not require) to add `/* USE_BDB_KEYDIFF */` after that.

libcob/ChangeLog
```diff
@@ -1,4 +1,11 @@
 
+2024-07-09  Boris Eng <boris.eng@ocamlpro.com>
+
+	* fileio.c (bdb_bt_compare): handle BDB keys of different length with a flag
```
GitMensch, line 4: add `indexed_open` here

GitMensch (changes requested):
some finalization, then ready to check-in upstream

libcob/common.c
GitMensch: @engboris you've marked that as solved but the question is unanswered?

libcob/common.c
```diff
@@ -1975,16 +1975,12 @@ cob_cmp_all (cob_field *f1, cob_field *f2)
 	return ret;
 }
 
-/* compare content of field 'f1' to content of 'f2', space padded,
-   using the optional collating sequence of the program */
```
GitMensch, line 1979: let's include the old doc comment, adjusted to "strings, space padded, using an optional collating sequence" here so this "local exported" function has a doc comment

GitMensch (changes requested):
Apart from the unanswered answer (and the doc comment) that's finished; would you like to commit that upstream then?
[note: upstream 3.x is currently failing CI, you may want to wait until @chuck-haatvedt did his final fix-commit tomorrow

libcob/fileio.c
engboris: There is no reason. It was temporary and they were meant to be removed.

engboris:
It should be good now!

libcob/fileio.c
GitMensch: you've removed those here, but left them in line 2031++ https://github.com/OCamlPro/gnucobol/pull/156/files#diff-16de7d2ecd4918cd696dff002aebea147395b0d09288c2e8e0310155dc943da1R2031-R2037

GitMensch (approved):

chuck-haatvedt:
My final fix-commit is done now.
      This fixes the issue with a build without curses. It also fixes
      warnings in libcob related to C90 ==>
      declaration-after-statement.
      
      sorry about the delay...
    
    
      
      
      
      
      
      
		@page { margin: 0.79in }
		p { margin-bottom: 0.08in }
	
      Chuck
              Haatvedt
      ***@***.***
    
    On 7/11/2024 3:13 AM, Simon Sobisch
      wrote:
    
    
      
      @GitMensch requested changes on this pull request.
      Apart from the unanswered answer (and the doc
        comment) that's finished; would you like to commit that upstream
        then?
        [note: upstream 3.x is currently failing CI, you may want to
        wait until @chuck-haatvedt did his final
        fix-commit tomorrow
      
      In libcob/common.c:
      > -/* compare content of field 'f1' to content of 'f2', space padded,
-   using the optional collating sequence of the program */

      let's include the old doc comment, adjusted to
        "strings, space padded, using an optional collating sequence"
        here so this "local exported" function has a doc comment
      
      In libcob/common.c:
      > +	return cob_cmp_strings ((unsigned char *)data1, (unsigned char *)data2,
+		size1, size2, col);

      @engboris
        you've marked that as solved but the question is unanswered?
      —
        Reply to this email directly, view it on GitHub, or unsubscribe.
        You are receiving this because you were mentioned.Message
          ID: ***@***.***>
      [
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/OCamlPro/gnucobol/pull/156#pullrequestreview-2171214176",
"url": "https://github.com/OCamlPro/gnucobol/pull/156#pullrequestreview-2171214176",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

ddeclerck:
Merged in SVN @ 5296.

PR #157, opened 2024-07-09
Rebased and cleaned up all the changes

xevor11:
Cleaned up java.c, added the requested changes from [PR#150](https://github.com/OCamlPro/gnucobol/pull/150), ready for feedback and comments


configure.ac
```diff
@@ -483,6 +483,30 @@ AC_CHECK_HEADERS([sys/types.h signal.h stddef.h], [],
...
+  dnl     i?86)   JVM_ARCH=i386 ;;
+  dnl     *)      JVM_ARCH=$target_cpu ;;
+  dnl esac
+  dnl AC_SUBST([JVM_ARCH])
+  dnl AS_IF([test -z "$JAVA_HOME"],
+  dnl   [LDFLAGS="-L$JAVA_HOME/lib/$JVM_ARCH/client -L$JAVA_HOME/lib/$JVM_ARCH/server $LDFLAGS"]
+  AC_CHECK_LIB([jvm], [JNI_CreateJavaVM], [LIBS=$LIBS],
+                      [AC_MSG_WARN([no libjvm found; please set JAVA_HOME appropriately])])
+])
+AC_MSG_ERROR([stop here])
```
nberth, line 509: That would make the configure script stop here. I think that's not intended.

configure.ac
```diff
@@ -483,6 +483,30 @@ AC_CHECK_HEADERS([sys/types.h signal.h stddef.h], [],
 # optional:
 AC_CHECK_HEADERS([sys/time.h locale.h fcntl.h dlfcn.h sys/wait.h sys/sysmacros.h])
 
+# Check for JNI
+AC_ARG_ENABLE([java],
+              [AC_HELP_STRING([--disable-java],
+                              [disable Java Interoperability])])
+AS_IF([test X$enable_java != Xno], [
+   AX_JNI_INCLUDE_DIR
```
nberth, line 491: Right after that you need to traverse `JNI_INCLUDE_DIRS` and maybe append them to `CPPFLAGS`.

libcob/common.h
```diff
@@ -1294,6 +1294,18 @@ struct cob_call_struct {
 	cob_call_union		cob_cstr_cancel;	/* Cancel entry */
 };
 
+#ifdef HAVE_JNI
+typedef struct __cob_java_static_method {
+	jclass cls;
+	jmethodID mid;
+} cob_java_handle;
```
nberth, line 1301: Here the definition requires something like
```c
#ifdef HAVE_JNI
#include "jni.h"
#endif 
```
Otherwise, that definition could actually become abstract (as below) if we only ever manipulate pointer to its value outside of `java.c`, and in that case the inclusion of `jni.h` is not needed (which may be desirable).
```c
typedef struct __cob_java_handle cob_java_handle;
```

m4/ax_jni_include_dir.m4
```diff
@@ -0,0 +1,151 @@
...
+#   the variable JNI_INCLUDE_DIRS.
+#
+#   Example usage follows:
+#
+#     AX_JNI_INCLUDE_DIR
+#
+#     for JNI_INCLUDE_DIR in $JNI_INCLUDE_DIRS
+#     do
+#             CPPFLAGS="$CPPFLAGS -I$JNI_INCLUDE_DIR"
+#     done
```
nberth, line 26: Here is the code needed in `configure.ac`.

m4/ax_prog_java.m4
```diff
@@ -0,0 +1,115 @@
...
+#   *Warning*: its success or failure can depend on a proper setting of the
+#   CLASSPATH env. variable.
+#
+#   TODO: allow to exclude virtual machines (rationale: most Java programs
+#   cannot run with some VM like kaffe).
+#
+#   Note: This is part of the set of autoconf M4 macros for Java programs.
+#   It is VERY IMPORTANT that you download the whole set, some macros depend
+#   on other. Unfortunately, the autoconf archive does not support the
+#   concept of set of macros, so I had to break it for submission.
```
nberth, line 52: I'm not sure that particular file is needed. But if it is, I think its companion files (mentioned in that paragraph) are missing.

cobc/codegen.c
```diff
@@ -121,6 +121,12 @@ struct call_list {
 	const char		*call_name;
 };
 
+struct java_call_list {
+    struct java_call_list *next;
+    const char *call_name;
+    cob_java_handle *handle;
```
nberth, line 127: Why do you need the handle here? This should be in the generated code, not in the compiler itself.

m4/ax_jni_include_dir.m4
xevor11: Would be inside an AS_IF?
AS_IF([test -n "$JNI_INCLUDE_DIRS"], [
	for JNI_INCLUDE_DIR in $JNI_INCLUDE_DIRS
	do
		CPPFLAGS="$CPPFLAGS -I$JNI_INCLUDE_DIR"
	done
  ])

cobc/codegen.c
xevor11: It is due to this, I am assuming this function must return a void:

`static cob_java_handle* 
lookup_java_call(const char *p, struct cb_call s)`

However:

`cob_java_handle* static_java_method = lookup_java_call(s + 6, p->argv[0]);`

we had this inside the string comparison, so I wanted to confirm this

m4/ax_prog_java.m4
xevor11: Fixed

m4/ax_jni_include_dir.m4
nberth: No I don't think such a guard is needed. Right after `AX_JNI_INCLUDE_DIR` should be enough.

libcob/common.h
xevor11: I was going to suggest for now we should include jni.h inside the header guard, as we fix the config and check for other issues. However having an abstract definition definitely is a better option, I did check it `cob_java_handle` is only used in java.c

configure.ac
xevor11: Fixed

cobc/codegen.c
nberth: No the call to `cob_resolve_handle` should just be printed via calls to the `output_*` functions in `codegen.c`. Similarly to `cob_resolve_cobol` which is output a few lines below.

m4/ax_prog_java.m4
nberth: You can probably do `git commit -p --amend` to amend the current commit and then `git push origin +jni-config` to update the PR.

cobc/codegen.c
```diff
@@ -7532,230 +7553,256 @@ output_call (struct cb_call *p)
...
+				char *first_dot = strchr(s + 6, '.');
+				if (first_dot != NULL) {
+					*first_dot = '\0';
+					char *method_name = first_dot + 1;
+					char *class_name = s;
+					output("cob_resolve_java (\"%s\", \"%s\", \"()V\");", class_name, method_name);
+				}
+				output ("cob_resolve_java ();");
+				output_newline ();
+				output_block_close ();
```
nberth, line 7578: Right here you need to generate a call to `cob_call_java(…)` with the appropriate handle name `call_java_%s`. Arguments will come in a subsequent PR.

cobc/codegen.c
```diff
@@ -7532,230 +7553,256 @@ output_call (struct cb_call *p)
...
+					output_string ((const unsigned char *)s,
+							(int)strlen (s), 0);
+					output (", %d, %d);", cb_fold_call, !p->stmt1);
+				}
+				output_newline ();
+				output_block_close ();
 			}
-			output_newline ();
-			output_block_close ();
 		} else {
```
nberth, line 7603: Note: please restore the indentation of everything below this line.

cobc/ppparse.output
nberth, line 1: This file should not be committed.

libcob/java.c
```diff
@@ -0,0 +1,97 @@
...
+static int			/* non-zero means there's an error */
+cob_java_initialize() {
+    /* JDK/JRE 6 VM initialization arguments */
+    JavaVMInitArgs args;
+    JavaVMOption* options = (JavaVMOption*)cob_malloc(sizeof(JavaVMOption) * 1);
+    args.version = JNI_VERSION_1_6;
+    /* inline */
+    args.nOptions = 1;
+    options[0].optionString = cob_malloc(strlen("-Djava.class.path=/usr/lib/java") + 1);
+    strcpy(options[0].optionString, "-Djava.class.path=/usr/lib/java");
```
nberth, line 40: This is a bit weird. In any case, I guess this should be given via an environment variable (`CLASSPATH`)? Maybe setting this `optionString` field is not needed at all.

po/en@boldquot.insert-header
nberth, line 1: Everything under `po` should not be committed either.

configure.ac
```diff
@@ -483,6 +485,41 @@ AC_CHECK_HEADERS([sys/types.h signal.h stddef.h], [],
...
+   for JNI_INCLUDE_DIR in $JNI_INCLUDE_DIRS
+	do
+		CPPFLAGS="$CPPFLAGS -I$JNI_INCLUDE_DIR"
+	done
+	AC_DEFINE([WITH_JNI], [1])
```
nberth, line 498: We may also need to check there was no failure in `AX_JNI_INCLUDE_DIR` and then skip in such a case.

configure.ac
```diff
@@ -483,6 +485,45 @@ AC_CHECK_HEADERS([sys/types.h signal.h stddef.h], [],
 # optional:
 AC_CHECK_HEADERS([sys/time.h locale.h fcntl.h dlfcn.h sys/wait.h sys/sysmacros.h])
 
+# Check for JNI
+AC_ARG_ENABLE([java],
+              [AC_HELP_STRING([--disable-java],
+                              [disable Java Interoperability])])
+AS_IF([test X$enable_java != Xno], [
```
GitMensch, line 492: `ARC_ARG_ENABLE` is more about optional features (like the debug log) and less about external packages where `AC_ARG_WITH` is preferable (and would also match what configure has for example for libxml2).

See https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/External-Software.html and adjust (you may postpone that part until you have actual running tests, though).

In any case please move the JNI block directly before the libxml2 part.

libcob/common.h
```diff
@@ -1294,6 +1294,16 @@ struct cob_call_struct {
 	cob_call_union		cob_cstr_cancel;	/* Cancel entry */
 };
 
+// #ifdef HAVE_JNI
+// #include "jni.h"
```
GitMensch, line 1298: these commented line and the one above should be dropped

GitMensch (changes requested):
the real next change that should be done is an actual test for calling a static java method that has no parameters, to verify that java.c indeed works

GitMensch:
I hope my commit to java.c did not break anything... (I just felt that the explanations would take much longer and those are more related to libcob in general than the changes of this project itself).
... if not: thank you for cleaning it up.

And yes, once there is a running simple test, we should update the branch from gcos4 which will then also include the CI. for further commits.

cobc/codegen.c
```diff
@@ -7532,6 +7579,10 @@ output_call (struct cb_call *p)
 		if (name_is_literal_or_prototype) {
 			s = get_program_id_str (p->name);
 			name_str = cb_encode_program_id (s, 1, cb_fold_call);
+			if(strncasecmp("Java.", s, 5) == 0) {
+				output_java_call(s);
```
nberth, line 7583: Note: you could directly give `s + 5` to `output_java_call`.

cobc/codegen.c
```diff
@@ -7042,6 +7066,29 @@ output_field_constant (cb_tree x, int n, const char *flagname)
 	output_newline ();
 }
 
+static void output_java_call(const char *p)
```
nberth, line 7069: To be clear, I suggest to rename `p` into `class_n_method_name`.

cobc/codegen.c
```diff
@@ -7532,6 +7579,10 @@ output_call (struct cb_call *p)
 		if (name_is_literal_or_prototype) {
 			s = get_program_id_str (p->name);
 			name_str = cb_encode_program_id (s, 1, cb_fold_call);
```
nberth, line 7581: This line may go right before `lookup_call` below.

cobc/codegen.c
```diff
@@ -7042,6 +7066,33 @@ output_field_constant (cb_tree x, int n, const char *flagname)
...
+    output_prefix();
+    output("call_java_%s = ", p);
+    last_dot = strrchr(p, '.');
+    if (last_dot != NULL) {
+        *last_dot = '\0';
+        method_name = last_dot + 1;
+        class_name = p;
+        output("cob_resolve_java(\"%s\", \"%s\", \"()V\");", class_name, method_name);
+		*last_dot = '.';
+    } else {
```
nberth: missing: `output_newline ();`

cobc/codegen.c
```diff
@@ -7042,6 +7066,33 @@ output_field_constant (cb_tree x, int n, const char *flagname)
...
+		cobc_err_msg (_("malformed '%s' call to a Java method"), p);
+	}
+    output("cob_call_java(call_java_%s);\n", p);
```
nberth, line 7099: I think: extraneous `\n`.

cobc/codegen.c
```diff
@@ -387,6 +394,22 @@ lookup_source (const char *p)
...
+
+    for (clp = call_java_cache; clp; clp = clp->next) {
+        if (strcmp (p, clp->call_name) == 0) {
+            return;
+        }
+    }
+    clp = cobc_parse_malloc (sizeof(struct java_call_list));
+    clp->call_name = p;
+    clp->next = call_java_cache;
+    call_java_cache = clp;
```
GitMensch: please indent on the left with tab

cobc/codegen.c
```diff
@@ -408,17 +431,18 @@ lookup_func_call (const char *p)
...
-	func_call_cache = clp;
+    for (clp = func_call_cache; clp; clp = clp->next) {
+        if (strcmp(p, clp->call_name) == 0) {
+            return;
+        }
+    }
+    clp = cobc_parse_malloc(sizeof(struct call_list));
+    clp->call_name = p;
+    clp->next = func_call_cache;
+    func_call_cache = clp;
```
GitMensch, line 444: indentation got wrong, please clean up, which will also lead to a smaller diff

cobc/codegen.c
```diff
@@ -7531,6 +7582,10 @@ output_call (struct cb_call *p)
 		/* Dynamic link */
 		if (name_is_literal_or_prototype) {
 			s = get_program_id_str (p->name);
+			if(strncasecmp("Java.", s, 5) == 0) {
+				output_java_call(s + 5);
+				return;
+			} else {
```
GitMensch, line 7596: drop that else and the closing brace as you have a `return` already and don't need the additional scope (and create smaller diffs that way)

GitMensch (changes requested):
Requesting mostly formatting changes (and the variable name pointed out by Nicolas) for now - we really would need a testcase that showcase a call to a static, preferably parameterless, jre internal function works fine.

The one "real" change is to move the check for the java call into the parser - see review comments for the details.

And even if it doesn't work it should still be added to a new testcase at the end of run_extensions.at

xevor11:
Addressing this specific system call:

`System.getProperty("java.version")`

I wrote a simple method call:

```
if ((cob_global_exception & 0x0b00) == 0x0b00) cob_global_exception = 0;
    if (call_java_class_method_call.funcvoid == NULL || cob_glob_ptr->cob_physical_cancel)
  {
    call_java_class_method_call.funcvoid = cob_resolve_cobol ("java_class_method_call", 0, 0);
  }
```

however internally it must be create the VM and retrieve the methods from the JNI Function Table?

GitMensch:
> however internally it must be create the VM and retrieve the methods from the JNI Function Table?

Yes the first fist to be done in java.c once and the other is done by the `cob_resolve_java` (which will also set the exception and possible internal registers if resolving it does not work).

cobc/codegen.c
```diff
@@ -7531,6 +7582,10 @@ output_call (struct cb_call *p)
 		/* Dynamic link */
 		if (name_is_literal_or_prototype) {
 			s = get_program_id_str (p->name);
+			if(strncasecmp("Java.", s, 5) == 0) {
```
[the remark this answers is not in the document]
nberth: @xevor11 To help with that, I've just rebased `java-interop` on top of `master` (which is `trunk` on the upstream). Now you can rebase your `jni-config` branch on top of that so you'll have access to the piece of code that's linked above.

After that you can define `CB_CONV_JAVA` in `tree.h`, and then do the check for the `"Java."` prefix in `parser.y` (I *think* right before the call to `cb_check_conformance`, but I'm not sure what that latter function does exactly).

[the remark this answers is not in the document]
GitMensch: yes right before that function, which checks that something is according to a known prototype - in the case of Java that should check that only "java interop compatible" arguments are passed.

xevor11:
Additional changes got added due to new state of jni-config

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2024-07-12  Vedant Tewari <vedanttewari541@gmail.com>
```
GitMensch, line 1: formatting: please have ChangeLog entries start with an empty line, this helps to reduce merge problems

xevor11:
Yes we have a test that verifies the version, I was wondering for `git diff` I can compare my local against upstream/java-interop?

cobc/codegen.c
```diff
@@ -7099,29 +7099,29 @@ output_field_constant (cb_tree x, int n, const char *flagname)
...
-		transformed_p[i] = (p[i] == '.') ? '_' : p[i];
+	for (size_t i = 0; i < strlen(class_and_method_name) + 1; i++) {
+		transformed_p[i] = (class_and_method_name[i] == '.') ? '_' : class_and_method_name[i];
 	}
 
-	last_dot = strrchr(p, '.');
+	last_dot = strrchr(class_and_method_name, '.');
 	if (last_dot == NULL) {
-		cobc_err_msg (_("malformed call '%s' to a Java method"), p);
+		cobc_err_msg (_("malformed call '%s' to a Java method"), class_and_method_name);
```
nberth, line 7118: For later: I think this check should actually end up (also) earlier during parsing.

libcob/Makefile.am
```diff
@@ -126,7 +136,7 @@ lib_lm =
 endif
 
 lib_LTLIBRARIES = libcob.la $(lib_ci) $(lib_di) $(lib_vb) $(lib_vc) \
-	$(lib_od) $(lib_oc) $(lib_db) $(lib_lm)
+	$(lib_od) $(lib_oc) $(lib_db) $(lib_lm) $(lib_jni)
```
nberth, line 139: @GitMensch With these changes, it is not clear what the actual requirements for loading that additional `libcobjni` library at runtime are. Should it be via some additional code inserted by `codegen.c` for modules that have `CALL "Java.…" statements?

@xevor11 for the moment, I guess you could just add
```
if COB_HAVE_JNI
libcob_la_SOURCES += java.c
libcob_la_LIBADD += $(JVM_LIBS)
libcob_la_CFLAGS += $(JVM_CFLAGS)
endif
```
(or something equivalent — I'm not sure what are the limitations of what can be done in these automake files.)

GitMensch: for now this can be done as noted here, later on there should be a separate libcob-java.so which is delay-loaded on the first call to a java function (just as it is with libcob-bdb.so)

nberth: With the current design, most, if not all, function calls that are generated for each `CALL "Java.…"`, should end up being implemented in that `libcobjni.so` library. I guess there could be an intermediate `cob_load_java` in `libcob.so` itself, or something similar, to be called upon module initialization; or else a `cob_resolve_java_call` that is itself implemented in `call.c`, loading `libcobjni.so` if it is not loaded already, and then calling `cob_recolve_java` from `java.c`…

GitMensch:
The later would match the design we have in fileio best and would also be easily prepared now by just doing a direct call for the time being - the delay load addition is then only minimal.

For now I'd primarily like to see the working java calls in the testsuite and running in CI - simple static one, followed by more complex examples similar to the one from IBM.

libcob/Makefile.am
nberth: @xevor11 As it's more about the build and autotest system I've push a new commit on top of `jni-config` with a draft approach for doing the aforementioned dynamic loading of `libcobjni`. At least that appears to be working on my setup.

cobc/cobc.c
```diff
@@ -2620,6 +2620,12 @@ cobc_print_info (void)
 
 	cobc_var_print (_("JSON library"),		_(WITH_JSON), 0);
 
+#ifdef WITH_JNI
+	cobc_var_print (_("Java interoperability"),	_("enabled"), 0);
+#else
+	cobc_var_print (_("Java interoperability"),	_("disabled"), 0);
+#endif
```
GitMensch, line 2627: @nberth please add also to libcob/common.c - similar to fileio, where it  will output the state (=can it be loaded) and possibly provide "own" version output

libcob/call.c
GitMensch, line 1: @nbrth: this delay-load part is missing in the changedoc

nberth: Sure! There are still many missing ChangeLog entries; I am just rushing to try and find out where any remaining "hard-to-do" parts about the dynamic loading are. Based on our tests with @xevor11, things appear to depend heavily on the Java version. Like for now the "Hello World!" test works on my setup but exists silently on his machine (or else exhibits a segfault within `JNI_CreateJavaVM` when in a `gdb` session…).

libcob/java.h
```diff
@@ -0,0 +1,50 @@
...
+
+   You should have received a copy of the GNU Lesser General Public License
+   along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef COB_JAVA_H
+#define COB_JAVA_H
+
+#include "config.h"
+#include "common.h"
```
GitMensch, line 25: inclusion guard COB_JAVA_H should not be necessary, given its inclusion in only C files, right?
Please include config.h. + common.h in the including sources before the `#include jjava.h`, not in this header

nberth: Hum… for the moment it is included  from `common.h` as well.

GitMensch:
I suggest to give it a go with gitpod and see hows the result there - as you can use the debugger if it doesn't work.

libcob/java.h
nberth: In the end I simply removed `java.h`.

GitMensch:
As seen in the Test for the MacOS CI:
* configure.ac: has to be adjusted COB_HAVE_JNI->COB_HAS_JNI  (already fixed)
* configure.ac: `--with-java` should abort if no applicable jni.h can be found
* either a part of the configure macros does not work (for the MacOS CI just issue a `find -name` to look for installations of jni.h) or needs an extra install (I'd prefer to get the current path working first, before supporting the Homebrew installation next)
* it will be interesting to see Win32 builds - I guess they need an OpenJDK installation outside of pacman/vcpkg

GitMensch:
As noted in the other PR where we had the last commit before: please adjust libcob/Changelog for that and revert the definition of `lt_dladdsearchdir` and the changes to the defines above; the use of JAVA_HOME should definitely be documented in gnucobol.texi.
Please adjust the new CI files as well (no install.log needed, adding the `|| make check TESTSUITEFLAGS="--recheck --verbose"` for getting details on the failing testcase runs, replacing the commented NIST tests by a single note that those are not useful to be executed with this build as the java parts are not used there).

Afterwards (and after integrating the Windows CI that is work in progress) - what would be next for this PR? Checking that the arguments are one of the cobol java types and handle those?

libcob/java.c
```diff
@@ -0,0 +1,128 @@
...
+	for (char *p = jni_class_name; *p; ++p) {
+		if (*p == '.') {
+			*p = '_';
+		}
+	}
+
+	cls = (*env)->FindClass(env, jni_class_name);
+	free(jni_class_name);
+	if (!cls) {
+		return NULL;
```
nberth, line 82: Here you might want to call something like `cob_runtime_error`.

nberth: Same for the case of `!mid` below.

GitMensch:
It will be quite beneficial if we use the x86 MSYS2 version for testing an old JDK (version 8).

Can you please try to put a cached download of https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u422-b05/OpenJDK8U-jdk_x86-32_windows_hotspot_8u422b05.zip in use - only for mingw32?

nberth:
> It will be quite beneficial if we use the x86 MSYS2 version for testing an old JDK (version 8).
>
> Can you please try to put a cached download of https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u422-b05/OpenJDK8U-jdk_x86-32_windows_hotspot_8u422b05.zip in use - only for mingw32?

Yes that's a good idea.

Note Java 8 from Oracle seems to be the default on the Windows CIs (https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#java) — although I might not be aware of all intricacies behind versioning of Java VMs/JDKs.

GitMensch:
That's an interesting doc and we may should reset the JAVA_HOME to the 21 one for all other Windows environments then.

nberth:
> That's an interesting doc and we may should reset the JAVA_HOME to the 21 one for all other Windows environments then.

*cf* #173

cobc/parser.y
```diff
@@ -12253,6 +12253,20 @@ call_body:
 
 	/* Check parameter conformance, if we can work out what is being called. */
 	if (CB_LITERAL_P ($3)) {
+		/* Check for "Java." prefix and set call convention */
+		char* s = (char *)CB_LITERAL ($3)->data;
+		if (strncasecmp("Java.", s, 5) == 0) {
+			call_conv = CB_CONV_JAVA;
+		}
+		/* Check for malformed Java method names */
```
nberth, line 12261: For clarity, I think these two new checks should only be performed if `strncasecmp("Java.…) == 0` holds (so: in the block above.

cobc/parser.y
```diff
@@ -12253,6 +12253,20 @@ call_body:
...
+		char* class_and_method_name = s + 5;
```
nberth, line 12262: You can probably avoid this intermediate declaration (which is not at the beginning of a block), and just use `s + 5` in `strchr` instead.

tests/testsuite.src/run_java.at
```diff
@@ -0,0 +1,67 @@
...
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       PROCEDURE        DIVISION.
+           CALL "Java.Test.missingMethod"
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE prog.cob], [0], [], [])
+# TODO: should fail!
+AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [])
```
nberth, line 66: I guess this should now fail and show the runtime warning.

xevor11:
There is another PR #174 which includes the non-void method handling as well

cobc/parser.y
```diff
@@ -12257,15 +12257,14 @@ call_body:
...
-		char* class_and_method_name = s + 5;
-		if (strchr(class_and_method_name, '.') == NULL) {
-			cb_error_x ($3, _("malformed Java method name '%s', expected format 'Java.ClassName.methodName'"), s);
-		}
-		/* Check for unsupported Java method calls with parameters or return values */
-		if($7 != NULL || $8 != NULL) {
-			CB_PENDING("Java method calls with parameters or return values");
+			/* Check for malformed Java method names */
+			if (strchr(s + 5, '.') == NULL) {
+				cb_error_x ($3, _("malformed Java method name '%s', expected format 'Java.ClassName.methodName'"), s);
```
nberth, line 12262: @xevor11 Actually this should also be tested. You can probably copy/paste and adapt the test on missing method in `run_java.at`, and just omit the `.`.

tests/testsuite.src/run_java.at
```diff
@@ -65,3 +65,19 @@ AT_CHECK([$COMPILE prog.cob], [0], [], [])
...
+AT_SETUP([CALL Java with malformed method name])
+AT_KEYWORDS([extensions jni malformed])
+
+AT_SKIP_IF([test "$COB_HAS_JNI" = "no"])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       PROCEDURE        DIVISION.
+           CALL "JavaInvalidName"
```
nberth, line 78: @xevor11 I think that should be `"Java.InvalidName"` instead (otherwise it's just a regular call).

tests/testsuite.src/run_java.at
nberth: https://github.com/OCamlPro/gnucobol/actions/runs/10789037700/job/29921110531?pr=157#step:11:2677
```
libcob: prog.cob:5: error: Java class 'Test' not found
```

This message is what is missing from this test. To deal with this it should be edited and look like:
```
AT_CHECK([$COBCRUN_DIRECT ./prog], [1], [],
[libcob: prog.cob:5: error: Java class 'Test' not found
])
```

tests/testsuite.src/run_java.at
```diff
@@ -0,0 +1,112 @@
...
+      System.out.println ("Hello world!");
+   }
+}
+])
+AT_DATA([prog.cob], [
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       PROCEDURE        DIVISION.
+           CALL "Java.Test.printHelloWorld"
+           STOP RUN.
```
GitMensch, line 40: add `NOT ON EXCEPTION DISPLAY "Java call worked"` to this call

GitMensch (changes requested):
The overall state is quite good!

Biggest issues:

* from glancing at the code I think `CALL...[NOT] ON EXCEPTION` is broken (adjust the tests as requested and we'll see)
* the build system need some changes
* @nberth has a bit to do on the delay-load part (incl. ChangeLog entries)
* if possible I'd like to see the parameter check already, but we can leave this for later as well

GitMensch:
I've got word from the original author of the java related macros used and that's (part of) his answer:

>  I'm afraid I cannot help and that it would be better to make these macros as deprecated/abandoned.

I'm likely to adjust the part configure for that (but that likely takes some days until I get to this), @nberth can you please check the review related to the delay-load part and @xevor11 could you please check on the rest?

tests/testsuite.src/run_java.at
```diff
@@ -0,0 +1,134 @@
...
+## GNU General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
+
+### GnuCOBOL Test Suite
+
+### Java interoperability tests
+
+AT_SETUP([CALL Java static void (void)])
```
nberth, line 23: Missing: test case with a class name inside a package (*e.g* `testpackage.Test`).

m4/ax_prog_javac.m4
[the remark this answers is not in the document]
nberth: Actually it's used by one of the other `.m4`, so I'm reintroducing that one (at least for now, so we can rely on the CI).

cobc/codegen.c
```diff
@@ -7068,6 +7090,45 @@ output_field_constant (cb_tree x, int n, const char *flagname)
...
+    	COBC_ABORT ();
+	}
+	char* full_name = (char *)CB_LITERAL(p->name)->data; /* Assume java.prefix (enforced in `parser.y`, rule `call_body`)*/
+	char* class_and_method_name = full_name + 5;
+	char *last_dot;
+	char *method_name;
+	const char *class_name;
+	char* mangled;
+
+	mangled = strdup(class_and_method_name);
```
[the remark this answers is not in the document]
nberth: This actually ends up in a list of "Java" calls that's printed later, so it should not be freed. Maybe in that case there's a proper function for string duplication to be used instead.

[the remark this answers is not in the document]
GitMensch: `lookup_java_call` does the duplication for later; `mangled` is otherwise not used after the function ends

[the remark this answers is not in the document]
nberth: Ah you mean `lookup_java_call` *should* deal with the duplication — it does not at the moment. That makes totally sense now that I think of it as there's no point in duplicating if the call is already present. I'll correct that.

cobc/codegen.c
```diff
@@ -7068,6 +7090,74 @@ output_field_constant (cb_tree x, int n, const char *flagname)
...
+}
+
+static void
+output_java_call (struct cb_call *p)
+{
+	char		*full_name, *class_and_method_name, *last_dot, *c;
+	const char	*class_name, *method_name;
+	char		mangled[COB_NORMAL_BUFF];
+
+	/* TODO: move that back into cb_check_conformance (*) */
```
nberth, line 7122: @xevor11 I think you have started moving this check and the one below back into `cb_check_conformance`; so I'll stop pushing changes to this branch to let you advance with that.

PR #160, opened 2024-08-02
Fix some tests under MSVC debug

ddeclerck:
This attemps to fix 3 tests that fail under MSVC Debug while OK in Release :
- MF FIGURATIVE to NUMERIC (run_misc)
- Default file external name (run_file)
- EXTFH: SEQUENTIAL files (run_file)

Note1: only "System routine CBL_GC_HOSTED" (run_extensions) remains failing under MSVC Debug while OK in Release, but this one seems a bit more tricky (as the test itself says: "test_errno.c must be compiled with same C runtime as libcob to match...")

Note2: all the tests mentionned above are not related to the runtime checker popups

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -166,9 +166,6 @@ jobs:
         run: |
           cd tests
 
-          sed -i '/AT_SETUP(\[MF FIGURATIVE to NUMERIC\])/a AT_SKIP_IF(\[true\])' testsuite.src/run_misc.at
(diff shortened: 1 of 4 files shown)


GitMensch (approved):
As these are old bugs in the testsuite, consider adding a short entry to tests/ChangeLog.

Good to commit upstream - thanks for your investigation!

GitMensch:
Can you commit that soon or is this unlikely to happen in the next two weeks (in which case I'd try to do that)?

ddeclerck:
> Can you commit that soon or is this unlikely to happen in the next two weeks (in which case I'd try to do that)?

I have a but of time to do that this evening ;)

PR #162, opened 2024-08-03
Support literals encoding conversions according to the literal type.

AhmedMaher309:
Added the support for encoding conversion for National literals to be in UTF-16 and for alphanumeric literals to be in ISO-8859-15 and for the UTF-8 literals to be in UTF-8


cobc/scanner.l
```diff
@@ -1513,214 +1512,201 @@ read_literal (const char mark, const enum cb_literal_type type)
...
+	if (unlikely (required_size > plex_size)) {
+	        plex_size = required_size;
+	        cobc_free (plex_buff);
+	        plex_buff = cobc_malloc (plex_size);
 	}
 	memcpy (plex_buff, text, curr_len);
-	if (type[0] == 'X' || type [0] == 'U') {
-		/* characters, two half-bytes (hex) = 1 byte, add 1 halfbyte to round up
-		   to full bytes in case of odd digits */
+	if(type[0] == 'X' || type[0] == 'U'){
```
GitMensch, line 1629: use a `switch` statement here

cobc/scanner.l
```diff
@@ -1513,214 +1512,201 @@ read_literal (const char mark, const enum cb_literal_type type)
...
+		/* characters, two half-bytes (hex) = 1 byte, add 1 halfbyte to round up 
+		to full bytes in case of odd digits */
 		result_len = (curr_len + 1) / 2;
 	} else if (type[0] == 'B') {
-		result_len = curr_len * 4; /* boolean characters B -> 1110 */
+		result_len = curr_len * 4; // boolean characters B -> 1110
```
GitMensch, line 1634: revert the wrong comment type

cobc/tree.c
```diff
@@ -2769,7 +2769,7 @@ cb_build_numsize_literal (const void *data, const size_t size, const int sign)
 }
 
 cb_tree
-cb_build_alphanumeric_literal (const void *data, const size_t size)
```
GitMensch, line 2772: place the new `cb_build_literal_by_category` here instead of those two functions and use it accordingly

cobc/tree.c
```diff
@@ -2782,18 +2782,175 @@ cb_build_alphanumeric_literal (const void *data, const size_t size)
...
+			}
+	    }
+
+        outsize -= outbytesleft;
+    }
+
+    l = CB_TREE (build_literal (CB_CATEGORY_NATIONAL, outdata, outsize));
+    free(outdata);
+#else
+    l = CB_TREE (build_literal (CB_CATEGORY_NATIONAL, data, size));
```
GitMensch, line 2892: "poor man's conversion" here?

cobc/tree.c
```diff
@@ -2782,18 +2782,175 @@ cb_build_alphanumeric_literal (const void *data, const size_t size)
...
-	l = CB_TREE (build_literal (CB_CATEGORY_NATIONAL, data, size));
+	l = CB_TREE (build_literal (CB_CATEGORY_ALPHANUMERIC, data, size));
 
 	l->source_file = cb_source_file;
 	l->source_line = cb_source_line;
 
 	return l;
 }
 
+cb_tree
```
GitMensch, line 2797: For those new/adjusted functions below: possibly call the `cb_build_literal_by_category` as well instead of creating the tree the same way?

GitMensch (changes requested):
while the overall approach is fine (misses documentation and tests for utf8, but both should come after the new command line argument) there are some things to adjust as detailed below; please try to get to those until Tuesday and push when finished

cobc/typeck.c
```diff
@@ -3879,23 +3879,26 @@ validate_alphabet (cb_tree alphabet)
...
+		int			dupvals[maxchar];
 
-		for (n = 0; n < 256; n++) {
+		for (n = 0; n < maxchar; n++) {
 			values[n] = -1;
 			charvals[n] = -1;
 			dupvals[n] = -1;
 			ap->values[n] = -1;
 			ap->alphachr[n] = -1;
 		}
```
GitMensch, line 3894: As mentioned in the call this needs to be fixed to be portable (the variable length arrays [VLAs] are not C90 standard code).
I'll try implementing the support for national alphabets "upstream", fixing this while improving both the amount of storage and performance. Therefore I mark all of the related review-comments as "resolved" (= other than understanding them in principle there is nothing to do for you there), allowing you to concentrate on the other "non-resolved" ones and trying to add the command line options.

I'll drop a note as soon this is is available (expected until tomorrow), allowing you to revert the parts locally and rebase (we can do that together) - after pushing that there should also be more CI running again.

cobc/tree.c
AhmedMaher309: Why?

GitMensch: the "poor man's" conversion that was originally in is still useful for the `!HAVE_ICONV` case ( = the `#else` if iconv is not available) - other than iconv it has to _assume_ that the source was encoded in an iso-8859 variant

cobc/cobc.c
```diff
@@ -3855,6 +3862,32 @@ process_command_line (const int argc, char **argv)
...
 			break;
 
+		case CB_FLAG_GETOPT_SOURCE_ENCODE:
+			/* -fsource-encode=encoding*/
+			const char* valid_encodings[] = {
+				"UTF-8",
+				"ASCII",
+				"ISO-8859-1",
+				"ISO-8859-15"
+			};
```
GitMensch, line 3872: won't iconv return an error if it is executed with a "bad" encoding identifier? Then we need that only for the !`HAVE_ICONV` case an d can otherwise just pass it to iconv as-is.

cobc/cobc.c
```diff
@@ -9270,6 +9301,14 @@ main (int argc, char **argv)
 	/* Process command line arguments */
 	iargs = process_command_line (argc, argv);
 
+/* initialize the iconv struct after reading the command line*/
+#ifdef HAVE_ICONV_H
+	cb_iconv.alphanumeric = iconv_open("ISO-8859-15", cb_iconv.source);
+	cb_iconv.national = iconv_open("UTF-16LE", cb_iconv.source);
+	cb_iconv.utf8 = iconv_open("UTF-8", cb_iconv.source);
+#endif
+
```
GitMensch, line 9310: I think this should go after "other one-time initialization" down a bunch of lines below, but even better - be only initialized (at least national + utf8) if one national/utf8 literal / data is seen - as na huge amount of programs won't ever use that.

cobc/flag.def
```diff
@@ -68,6 +68,11 @@ CB_FLAG_RQ (cb_max_errors, 1, "max-errors", 20, CB_FLAG_GETOPT_MAX_ERRORS,
 	  "                        * default: 20\n"
 	  "                        * if <number> is 0, there's no limit"))
 
+CB_FLAG_RQ (cb_source_encode, 1, "source-encode", 0, CB_FLAG_GETOPT_SOURCE_ENCODE, 
+	_("  -fsource-encode=[encoding]\tdefine the source file encoding\n"
+	  "                        * default: UTF-8"))
```
GitMensch, line 73: I think it is reasonable to use the existing code as basis for the default - and that would be iso-8859-15; to override it people could use a BOM in the source file (pplex.l)

GitMensch (changes requested):
nice next commits; testcase for that feature and Changelog for the last commits seems too be missin

cobc/cobc.c
AhmedMaher309: iconv will return an error, but what if the input source encoding is correct but not supported like UTF-32 or UTF-16?

GitMensch:
In this case (after my push) opening the file will create an error if that was really the encoding, and iconv will return an error on first use if the given encoding does not match the input.

cobc/ChangeLog
```diff
@@ -25,6 +50,18 @@
...
 
+2024-07-09  Ahmed Maher <ahmed.maher.ahmed.309@gmail.com>
+            James K. Lowden  <jklowden@symas.com>
+
+	* typeck.c: refactor validate_move function
+
+2024-07-06 Ahmed Maher <ahmed.maher.ahmed.309@gmail.com>
+
+	* tree.c: added encoding conversion support to cb_build_national_literal and
+	cb_build_alphanumeric_literal
```
GitMensch, line 61: indent by two spaces

GitMensch:
For the failing CI build - you may want to use { } around that :-)
For the UTF8 test - let's inspect that later this week.

GitMensch:
I know that's a very late thought (post GSOC project timeline)...

But we definitely should add the option to set `cb_iconv.alphanumeric` via command line - otherwise we force it to `ISO-8859-15` where previously people did use CP850 on Windows to draw TUI symbols which won't work any more and return iconv errors as these are not part of iso-8859-1 at all.
... and of course this may also have an effect at runtime... so we should at least add this to the cob_module structure as `char *alphanumeric_encoding` and set this if it isn't `ISO-8859-15`. A later addition in libcob would then need to handle a very specific case: encodings that have digits at the non-native place - 0x30-x39 vs: 0xF0-xF9, for ease this likely should be a separate flag and the conversion of digits won't need iconv but a simple lookup table.

GitMensch:
You haven't seem to pushed your last changes, do you?

cobc/cobc.h
```diff
@@ -214,6 +221,7 @@ enum cb_std_def {
 	CB_STD_BS2000,
 	CB_STD_ACU,
 	CB_STD_RM,
+	CB_STD_GCOS,
```
GitMensch, line 224: drop this line...

tests/testsuite.src/configuration.at
```diff
@@ -331,7 +331,7 @@ AT_CHECK([$COMPILE_ONLY \
 -fstandard-define=99 prog.cob], [1], [],
 [configuration error:
 -fstandard-define=99: invalid value '99' for configuration tag 'standard-define';
-	maximum value: 9
+	maximum value: 10
```
GitMensch, line 334: ... and that line

PR #163, opened 2024-08-06
Object-Oriented COBOL - DRAFT VERSION

RemiBHardy:
This is a preliminary work for adding the capability to support COBOL objects on GnuCOBOL. FORMAL REVIEW IS NOT NEEDED AS IT WON'T BE MERGED FOR NOW

you can find here : 

*  In libcob , Object.c & Object.h : extend the library to provide functions necessary to build and use objects
* In cobc : parser.y reserved.c are modified to extend the syntax ; tree.h and tree.c are modified to build  the corresponding AST


cobc/parser.tab.c
nberth, line 1: @RemiBHardy This file, along with `parser.tab.h`, should not be indexed by git.

cobc/parser.y
```diff
@@ -23,7 +23,7 @@
 
 %defines
 %verbose
-%error-verbose
+%define parse.error verbose
```
GitMensch, line 26: This is a fix of a Bison warning, but the fix is only possible with "more recent" versions of Bison (a GC4 thing).

cobc/parser.y
```diff
@@ -12912,6 +13143,19 @@ _end_call:
...
+
+
+
+
+
+
+
+
+
+
```
GitMensch, line 13158: magic happens here 😁

cobc/parser.y
```diff
@@ -11610,7 +11842,6 @@ statement:
 | initiate_statement
 | inquire_statement
 | inspect_statement
-/* | TODO: invoke_statement */
```
GitMensch, line 11613: yes, that's still not there, so a TODO

GitMensch (commented):
as noted this is obviously a draft so a formal review will be for possibly much later; I just want to point out that:

* It is good to early write ChangeLog entries (also helps "visitors" to get a grasp about what was changed why, including when coming back to this PR
* you likely have a testcase: so just add it to something like syn_misc.at (or move out directly to a new syn_oo.at) [if this is from some doc, then recreate it a bit different to not raise questions about copyright], this also allows you to run it during tests and see what doesn't work yet  and ensures that if you make changes the other tests still work

Actually I would suggest to split this into two PRs: 1 with the syntax and the AST, ideally with some checks and including CB_PENDING; another one based on this with the codegen and implementation (which likely will change parts in the AST as well).
 
The benefit: the first one can be "early" included and leads to nice handling in cobc and overall checks instead of strange parser errors.

cobc/parser.tab.c
GitMensch: actually, those two files shouldn't even be generated (but parser.c, parser.h used, which are in gitignore)

cobc/parser.y
GitMensch, line 1: strangely the file attributes are changed - we don't need to execute that

RemiBHardy:
I closed this PR to create two separate PRs as suggested

PR #164, opened 2024-08-09
[CANCELLED] OO COBOL - AST

RemiBHardy:
This is a preliminary work for adding the capability to support COBOL objects on GnuCOBOL. FORMAL REVIEW IS NOT NEEDED AS IT WON'T BE MERGED FOR NOW

you can find here :

    In cobc : parser.y reserved.c are modified to extend the syntax

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,9 @@
 
+2024-08-08  Remi Bertrand-Hardy <remi.bhardy@gmail.com>
+
+	* parser.y : added syntax rules for object-oriented COBOL
+	* reserved.c : added token for object-oriented COBOL
+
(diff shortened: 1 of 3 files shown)


GitMensch:
Please add tests verifying that OO syntax (standart / ibm / mf?) can be parsed and add a NEWS entry.

cobc/ChangeLog
```diff
@@ -21,6 +26,8 @@
 	* reserved.c: make MENU context-sensitive
 	* reserved.c, parser.y: added MODAL + MODELESS to acu extension windows
 
+
+
```
GitMensch, line 30: that seems useful to revert :-)

GitMensch (changes requested):
Please add tests verifying that OO syntax (standart / ibm / mf?) can be parsed and add a NEWS entry.

PR #168, opened 2024-08-15
Actually configure `--with-java` on MacOS, and add `ubuntu-20.04` CI

nberth:
Also upgrade versions of github actions


GitMensch:
Is the action update part something that should be done in the gcos4 and gcos3 branches as well?

nberth:
> Is the action update part something that should be done in the gcos4 and gcos3 branches as well?

Yes while doing these I observed we used some actions that trigger deprecation warnings. Upgrade does not appear to be urgent (but should be done before November).

PR #170, opened 2024-08-20
CI adjustments for `gcos4gnucobol-3.x`

nberth:
- Upgrade versions of github actions (some are being deprecated);
- Perform tests before install when doing the opposite is not very well justified.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -100,6 +100,7 @@ jobs:
           path: _build/tests/cobol85/newcob.val
           key: newcob-val
           save-always: true
+          enableCrossOsArchive: true
 
       - name: NIST85 Test Suite
         run: |
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -94,24 +94,14 @@ jobs:
           name: testsuite-${{ matrix.os }}.log
           path: _build/tests/testsuite.log
 
-      - name: Cache newcob.val.tar.gz
+      - name: Cache newcob.val
         uses: actions/cache@v4
-        id: newcob
         with:
-          path: tests/cobol85/newcob.val.tar.gz.cached
-          key: newcob
+          path: _build/tests/cobol85/newcob.val
+          key: newcob-val
           save-always: true
-
-      - name: Download newcob.val.tar.gz
-        if: steps.newcob.outputs.cache-hit != 'true'
-        run: |
-          make -C _build/tests/cobol85 newcob.val.tar.gz
-          ln   -f _build/tests/cobol85/newcob.val.tar.gz        \
-                         tests/cobol85/newcob.val.tar.gz.cached
 
       - name: NIST85 Test Suite
         run: |
-          ln   -f        tests/cobol85/newcob.val.tar.gz.cached \
-                         tests/cobol85/newcob.val.tar.gz
           make -C _build/tests/cobol85 EXEC85 test              \
-                         --jobs=$(($(nproc)+1))
+                         --jobs=$((${NPROC}+1))
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -94,22 +94,24 @@ jobs:
           name: testsuite-${{ matrix.os }}.log
           path: _build/tests/testsuite.log
 
-      - name: Cache newcob.val.Z
-        uses: actions/cache@v3
+      - name: Cache newcob.val.tar.gz
+        uses: actions/cache@v4
         id: newcob
         with:
-          path: _build/tests/cobol85/newcob.val.Z.cached
+          path: tests/cobol85/newcob.val.tar.gz.cached
           key: newcob
+          save-always: true
 
-      - name: Download newcob.val.Z
+      - name: Download newcob.val.tar.gz
         if: steps.newcob.outputs.cache-hit != 'true'
         run: |
-          cd _build/tests/cobol85
-          make newcob.val.Z
-          ln -f newcob.val.Z newcob.val.Z.cached
+          make -C _build/tests/cobol85 newcob.val.tar.gz
+          ln   -f _build/tests/cobol85/newcob.val.tar.gz        \
+                         tests/cobol85/newcob.val.tar.gz.cached
 
       - name: NIST85 Test Suite
         run: |
-          cd _build/tests/cobol85
-          ln -f newcob.val.Z.cached newcob.val.Z
-          make EXEC85 && make --jobs=$(($(nproc)+1)) test
+          ln   -f        tests/cobol85/newcob.val.tar.gz.cached \
+                         tests/cobol85/newcob.val.tar.gz
+          make -C _build/tests/cobol85 EXEC85 test              \
+                         --jobs=$(($(nproc)+1))
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -24,7 +24,7 @@ jobs:
 
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install packages
         run: |
@@ -55,44 +55,44 @@ jobs:
       - name: configure
         run: |
           cd _build
-          ../configure --enable-cobc-internal-checks --enable-hardening --prefix /opt/cobol/gnucobol-gcos --exec-prefix /opt/cobol/gnucobol-gcos
+          ../configure --enable-cobc-internal-checks \
+                       --enable-hardening \
+                       --prefix /opt/cobol/gnucobol-gcos \
+                       --exec-prefix /opt/cobol/gnucobol-gcos
 
       - name: Upload config.log
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: config.log
+          name: config-${{ matrix.os }}.log
           path: _build/config.log
         if: failure()
 
       - name: make
         run: |
-          cd _build
-          make --jobs=$((${NPROC}+1))
+          make -C _build --jobs=$((${NPROC}+1))
 
 # make install must be done before make check, otherwise execution of
 # generated COBOL files fail for a missing /usr/local/lib/libcob.dylib
       - name: make install
         run: |
-          cd _build
-          sudo make install
-          find /opt/cobol > install.log
+          sudo make -C _build install
+          find /opt/cobol > _build/install.log
 
       - name: Upload install.log
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: install.log
+          name: install-${{ matrix.os }}.log
           path: _build/install.log
 
       - name: check
         run: |
-          cd _build
-          make check TESTSUITEFLAGS="--jobs=$((${NPROC}+1))"
+          make -C _build check TESTSUITEFLAGS="--jobs=$((${NPROC}+1))"
 
       - name: Upload testsuite.log
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: testsuite.log
+          name: testsuite-${{ matrix.os }}.log
           path: _build/tests/testsuite.log
 
       - name: Cache newcob.val.Z
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -56,7 +56,6 @@ jobs:
           ../configure --enable-cobc-internal-checks \
                        --enable-hardening \
                        --prefix /opt/cobol/gnucobol-gcos \
-                       --exec-prefix /opt/cobol/gnucobol-gcos
 
       - name: Upload config.log
         uses: actions/upload-artifact@v4
@@ -69,8 +68,9 @@ jobs:
         run: |
           make -C _build --jobs=$((${NPROC}+1))
 
-# make install must be done before make check, otherwise execution of
-# generated COBOL files fail for a missing /usr/local/lib/libcob.dylib
+      # make install must be done before make check, otherwise
+      # execution of generated COBOL files fail for a missing
+      # /usr/local/lib/libcob.dylib
       - name: make install
         run: |
           sudo make -C _build install
@@ -84,7 +84,8 @@ jobs:
 
       - name: check
         run: |
-          make -C _build check TESTSUITEFLAGS="--jobs=$((${NPROC}+1))"
+          make -C _build check \
+                         TESTSUITEFLAGS="--jobs=$((${NPROC}+1))"
 
       - name: Upload testsuite.log
         uses: actions/upload-artifact@v4
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -4,7 +4,6 @@ on:
   pull_request:
     branches: [ gcos4gnucobol-3.x ]
   push:
-    branches: [ gcos4gnucobol-3.x ]
   # manual run in actions tab - for all branches
   workflow_dispatch:
 
@@ -14,7 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
+          # - macos-latest-large  # macos 14, amd64
+          - macos-latest        # macos 14, arm64
 
     runs-on: ${{ matrix.os }}
 
@@ -41,9 +41,7 @@ jobs:
 
       - name: bootstrap
         run: |
-          ./autogen.sh
-          autoconf
-          autoreconf --install --force
+          ./build_aux/bootstrap install
 
       - name: Build environment setup
         run: |
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -13,12 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os:
-          - ubuntu-latest
-        include:
-          - os: ubuntu-latest
-            skip_test: true
-
+        os: [ ubuntu-latest ]
     runs-on: ${{ matrix.os }}
 
     steps:
@@ -37,7 +32,7 @@ jobs:
           git config --global user.name github-actions
           git config --global user.email github-actions-bot@users.noreply.github.com
 
-      - name: bootstrap
+      - name: Bootstrap
         run: |
           ./build_aux/bootstrap
 
@@ -62,7 +57,7 @@ jobs:
           echo "TERM=$TERM" >> $GITHUB_ENV
           echo "INSTALL_PATH=$(pwd)/_install" >> $GITHUB_ENV
 
-      - name: configure
+      - name: Configure
         run: |
           cd _build
           ../configure --enable-cobc-internal-checks \
@@ -78,12 +73,12 @@ jobs:
           name: config-${{ matrix.os }}.log
           path: _build/config.log
 
-      - name: make
+      - name: Build
         run: |
           make -C _build --jobs=$(($(nproc)+1))
 
       # note: distcheck also creates the dist tarball
-      - name: distcheck
+      - name: Build distribution archive & run tests
         run: |
           make -C _build distcheck                                \
                          TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"  \
@@ -107,27 +102,30 @@ jobs:
           if-no-files-found: error
           retention-days: 0
 
-      - name: Cache newcob.val.Z
-        uses: actions/cache@v3
+      - name: Cache newcob.val.tar.gz
+        uses: actions/cache@v4
         id: newcob
         with:
-          path: _build/tests/cobol85/newcob.val.Z.cached
+          path: tests/cobol85/newcob.val.tar.gz.cached
           key: newcob
+          save-always: true
 
-      - name: Download newcob.val.Z
+      - name: Download newcob.val.tar.gz
         if: steps.newcob.outputs.cache-hit != 'true'
         run: |
-          cd _build/tests/cobol85
-          make newcob.val.Z
-          ln -f newcob.val.Z newcob.val.Z.cached
+          make -C _build/tests/cobol85 newcob.val.tar.gz
+          ln   -f _build/tests/cobol85/newcob.val.tar.gz        \
+                         tests/cobol85/newcob.val.tar.gz.cached
 
       - name: NIST85 Test Suite
         run: |
-          cd _build/tests/cobol85
-          ln -f newcob.val.Z.cached newcob.val.Z
-          make EXEC85 && make --jobs=$(($(nproc)+1)) test
+          ln   -f        tests/cobol85/newcob.val.tar.gz.cached           \
+                         tests/cobol85/newcob.val.tar.gz
+          make -C _build/tests/cobol85 EXEC85 test \
+                                       --jobs=$(($(nproc)+1))
 
-      - uses: actions/upload-artifact@v4
+      - name: Upload NIST85 Test Suite results
+        uses: actions/upload-artifact@v4
         with:
           name: NIST85 results on ${{ matrix.os }}
           path: |
@@ -140,7 +138,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout code
+        uses: actions/checkout@v4
 
       # note: less dependencies as we don't generate a dist tarball, one additional for lcov
       - name: Install dependencies
@@ -148,7 +147,7 @@ jobs:
           sudo apt-get install automake libtool libdb5.3-dev libxml2-dev \
                                libcjson-dev bison flex help2man gettext lcov
 
-      - name: bootstrap
+      - name: Bootstrap
         run: |
           ./build_aux/bootstrap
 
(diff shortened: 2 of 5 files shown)


nberth:
@ddeclerck Some Windose workflows appear to hang during tests (but not always). Do you know if that's an "expected" state of affairs?

ddeclerck:
> @ddeclerck @GitMensch Some Windows workflows appear to hang during tests (but not always). Do you know if that's an "expected" state of affairs?

Yeah, this is a known problem with MSVC Debug targets. Under a proper machine with a GUI you'd get a warning in a popup window, but in a CI, without a GUI, it just hangs.

It was working before though. Now it gets stuck on `132: PROGRAM COLLATING SEQUENCE`. Seems this is caused by a recent commit on SVN: [5310](https://sourceforge.net/p/gnucobol/code/5310/).

.github/workflows/windows-msys1.yml
```diff
@@ -234,6 +223,17 @@ jobs:
           name: testsuite-${{ matrix.target }}.log
           path: ${{ env.GITHUB_WORKSPACE }}/_build/tests/testsuite.log
 
+      - name: Install GnuCOBOL
```
ddeclerck, line 226: Any reason for swapping these steps ?

nberth: > Any reason for swapping these steps ?

https://github.com/OCamlPro/gnucobol/pull/169#discussion_r1720904756

> Note that in debug mode, it has to be installed before running the testsuite, else you end up with tons of failures (don't ask me why...).

Ah…

nberth: Just out of curiosity, is there an equivalent to `ulimit` on these systems?

ddeclerck: No idea...

.github/workflows/ubuntu.yml
```diff
@@ -127,30 +129,30 @@ jobs:
...
+          name: NIST85 results on ${{ matrix.os }}
           path: |
             _build/tests/cobol85/**/*.log
             _build/tests/cobol85/**/*.out
 
       - name: install
         run: |
-          cd _build
-          make install
+          make -C _build install
```
GitMensch, line 141: This is only reasonable (because it is already done internally in `make distcheck?`) if you want to create a binary package and make it available as artifact, but then `make distbin` may be preferable; if you don't want to provide that artifact (now) then drop that last step in this CI.

.github/workflows/ubuntu.yml
```diff
@@ -166,33 +168,33 @@ jobs:
       - name: configure
         run: |
           cd _build
-          ../configure --enable-code-coverage CPPFLAGS="-Werror=declaration-after-statement"
+          ../configure --enable-code-coverage \
```
GitMensch, line 171: it may be reasonable to also add `-pedantic` to CPPFLAGS and specify `CC="gcc -std=c89"` here

.github/workflows/windows-msys2.yml
```diff
@@ -75,41 +75,39 @@ jobs:
...
+          find /opt/cobol > _build/install.log
 
-      - name: Upload testsuite-${{ matrix.target }}.log
+      - name: Upload install-${{ matrix.target }}.log
         uses: actions/upload-artifact@v4
         with:
-          name: testsuite-${{ matrix.target }}.log
-          path: ${{ env.GITHUB_WORKSPACE }}/_build/tests/testsuite.log
+          name: install-${{ matrix.target }}.log
+          path: ${{ env.GITHUB_WORKSPACE }}/_build/install.log
```
GitMensch, line 105: Is there any reason to do the install and provide the install log here? There is no artifact using that and we do have the distmgw for a "ready-to-use" package.

.github/workflows/windows-msys2.yml
GitMensch, line 1: Just recognized the first time: the setup part specifies `msystem: ucrt64` maybe it would be useful to use a build matrix adding mingw64 (most common used) and mingw32?


In any case I suggest to:

```diff
-          ./autogen.sh
-          autoconf
-          autoreconf --install --force
+          ./autogen.sh install
```

GitMensch (approved):
For the changes done: LGTM; we may want to add some more pieces though - see review comments.

.github/workflows/windows-msys2.yml
ddeclerck: AFAIR under some circumstances I have not investigated, `./autogen.sh install` is not always sufficient. But let's just try it.

.github/workflows/windows-msys2.yml
ddeclerck: That's something Fabrice did a long time ago, we just kept that until now...

.github/workflows/windows-msvc.yml
```diff
@@ -205,6 +205,7 @@ jobs:
           sed -i '/AT_SETUP(\[BCD ADD and SUBTRACT, DEFAULT ROUNDING MODE\])/a AT_SKIP_IF(\[true\])' testsuite.src/data_packed.at
           sed -i '/AT_SETUP(\[BCD ADD and SUBTRACT, all ROUNDED MODEs\])/a AT_SKIP_IF(\[true\])' testsuite.src/data_packed.at
           sed -i '/AT_SETUP(\[CURRENCY SIGN WITH PICTURE SYMBOL\])/a AT_SKIP_IF(\[true\])' testsuite.src/run_misc.at
+          sed -i '/AT_SETUP(\[PROGRAM COLLATING SEQUENCE\])/a AT_SKIP_IF(\[true\])' testsuite.src/syn_definition.at
```
ddeclerck, line 208: @GitMensch This new exception is needed because of commit [5310](https://sourceforge.net/p/gnucobol/code/5310/). To avoid these "surprises", would it be reasonable to check SVN-targetted modifications with this GitHub CI before committing to SVN ?

GitMensch: To avoid these surprises we should do the right thing - drop all those skips and instead tell Windows to not do the popup but error until told otherwise via a special setting.

We discussed that before and have checked the msdn for the necessary function call - so far no one has implemented that.

.github/workflows/windows-msys2.yml
GitMensch: as the CI is now stable "in general" it seems like a good time to adjust then :-)

GitMensch:
Oh, and one general thing - to be easier to review failures, please use, after make check a `|| make check TESTSUITEFLAGS="--recheck --verbose"` in all environments, this will show the output nicely on the CI output with most details.

.github/workflows/windows-msys2.yml
nberth: `./build_aux/bootstrap install` seems to do it.

GitMensch: ah https://github.com/OCamlPro/gnucobol/commit/ed33f5f4e509513e02d8cde3848e9e3d9e788f55 is missing in gc4

GitMensch:
ci changes from #172 should go in here as wellö

ddeclerck:
> ci changes from #172 should go in here as well

For @nberth: it just amounts to removing the whole "Adjust testsuite for Debug target" step in MSVC workflows, and adding `continue-on-error: true` to the `name: Run testsuite` step (or maybe something like `continue-on-error: ${{ matrix.target == 'Debug' }}` - until we resolve all the problems).

.github/workflows/ubuntu.yml
```diff
@@ -4,7 +4,6 @@ on:
   pull_request:
     branches: [ gcos4gnucobol-3.x ]
   push:
-    branches: [ gcos4gnucobol-3.x ]
```
ddeclerck, line 7: Don't we want to keep that ? (when merging upstream changes from `gnucobol-3.x`, we don't go through a PR)

nberth: Maybe we should PR the merges. Here it's only removing a restriction, the CI will run on *any* push, not just a push on that branch.

GitMensch:
> https://github.com/OCamlPro/gnucobol/actions/runs/10522831045/job/29156289216?pr=170#step:8:3908
> 
> `-pedantic -std=c89` was a bit bold; won't be fixed in this PR though.

Yes, and it wouldn't be reasonable to look for fixing all of those either.
The following should:

```
 In file included from ../../cobc/scanner.l:100:
../../cobc/cobc.h:108:23: warning: comma at end of enumerator list [-Wpedantic]
  108 |         CB_FORMAT_AUTO,         /* Auto-detect format */
      |                       ^
../../cobc/cobc.h:272:28: warning: comma at end of enumerator list [-Wpedantic]
  272 |         CB_SUB_CHECK_RECORD,    /* PENDING */
      |                            ^
In file included from ../../cobc/scanner.l:100:
../../cobc/config.def:91:21: warning: ISO C forbids forward references to ‘enum’ types [-Wpedantic]
   91 | CB_CONFIG_ANY (enum cb_assign_type, cb_assign_type_default, "assign-clause",
      |                     ^~~~~~~~~~~~~~
../../cobc/cobc.h:551:8: note: in definition of macro ‘CB_CONFIG_ANY’
  551 | extern type                     var;
      |        ^~~~
In file included from ../../cobc/scanner.l:101:
../../cobc/tree.h:1263:30: warning: comma at end of enumerator list [-Wpedantic]
 1263 |         BOP_SHIFT_RC    = 'd',  /* ( x >> y circular-shift ) */
      |                              ^
../../cobc/tree.h:1645:36: warning: comma at end of enumerator list [-Wpedantic]
 1645 |         CB_REPLACE_TRAILING     = 2,
      |                                    ^
../../cobc/tree.h:2802:25: warning: comma at end of enumerator list [-Wpedantic]
 2802 |         CB_COLSEQ_EBCDIC,
      |                         ^
../../cobc/scanner.l:149:22: warning: comma at end of enumerator list [-Wpedantic]
  149 |         CB_LITERAL_NC,
      |                      ^
../../cobc/scanner.l: In function ‘scan_ebcdic_char’:
../../cobc/scanner.l:1370:14: error: C++ style comments are not allowed in ISO C90
```

and I'm looking to fix that upstream now.

GitMensch:
Just to share some finding which was new to me: bison generates code like the following:

```c
  case 3291: /* _dot_or_else_area_a: "Identifier (Area A)"  */
#line 20470 "../../cobc/parser.y"
  {
	/* No need to raise the error for *_IN_AREA_A tokens */
	(void) cb_verify (cb_missing_period, _("optional period"));
	cobc_repeat_last_token = 1;
  }
#line 34469 "parser.c"
    break;
```

which is the reason that `gcc -std=c89 -pedantic` raises the following error:

```
../../cobc/parser.y:20475:7: warning: line number out of range
20475 | ;
      |       ^    
```

the "computed" line number 20475 is the line from the generated code above

```c
#line 34469 "parser.c"
```

and the message is because [C89 only allows SHORT_MAX there](https://gcc.gnu.org/legacy-ml/gcc-help/2007-03/msg00119.html):

> You will notice that you only get the warning with `-pedantic`. gcc is warning because C89 has a limit of 32767.
>
>> According to C99,
>> 
>> `# line digit-sequence new-line`
>> 
>> causes the implementation to behave as if the following sequence of source lines begins with a source line that has a line number as specified by the digit sequence (interpreted as a decimal integer). The digit sequence shall not specify zero, nor a number greater than 2147483647.
>
>[...] C99 raised the limit.

The only way to get around that would be to move ~2000 lines out of parser.y (reducing this huge file is a goal in general).
The "original" source parser.y is the biggest we have, currently ~ 20600 LOC, and then of course the generated C file is even bigger.

The easiest way to get the 2000 lines out is to move nearly everything in `%{  }%` to a separate "header" file, like parserc.h.
After doing that we may move longer code blocks to more static functions in there as well.
... or, more useful. move the functions into a separate C file, having only the defines and the function declarations in this header which then also speeds up building (often we only adjust the content within those functions, which now triggers a regeneration of parser.c (.h is already restored if not different by the build system), which triggers a recompile of that beast. Moving the functions out will tame the beast a bit and for the case of only changing the functions we'd drop that completely.

Note that GnuCOBOL generated _functions_ can easily get > 100.000 LOC, so this may raise additional compatibility issues with generated sources on ancient environments.

nberth:
> Note that GnuCOBOL generated _functions_ can easily get > 100.000 LOC, so this may raise additional compatibility issues with generated sources on ancient environments.

Slipping in a `s/\n\([^#]\)/\1/g` (sort of) could help with LOC :laughing:

GitMensch:
After several adjustments there are now only 3 error in the testsuite with a compile of `gcc -std=c89 -pedantic`:

~~~
544. run_fundamental.at:3527: testing Entry point visibility (1) ...
../../tests/run_fundamental.at:3552: $COMPILE prog.cob
../../tests/run_fundamental.at:3553: $COMPILE_MODULE module.cob
--- /dev/null   2024-08-23 10:21:01.764081026 +0000
+++ /workspace/gnucobol/build/tests/testsuite.dir/at-groups/544/stderr  2024-08-23 16:08:20.093913497 +0000
@@ -0,0 +1,4 @@
+/tmp/cob289710_0.c: In function 'module_module_init':
+/tmp/cob289710_0.c:235:34: warning: ISO C forbids conversion of object pointer to function pointer type [-Wpedantic]
+  235 |   module->module_entry.funcptr = (void *(*)())module;
+      |                                  ^
544. run_fundamental.at:3527:  FAILED (run_fundamental.at:3553)

783. run_misc.at:4772: testing CALL C with callback, PROCEDURE DIVISION EXTERN ...
../../tests/run_misc.at:4826: $COMPILE -Wno-unfinished -o prog prog.cob cprog.c
--- /dev/null   2024-08-23 10:21:01.764081026 +0000
+++ /workspace/gnucobol/build/tests/testsuite.dir/at-groups/783/stderr  2024-08-23 16:08:20.365913248 +0000
@@ -0,0 +1,11 @@
+cprog.c: In function 'cprog':
+cprog.c:5:8: warning: ISO C does not support omitting parameter names in function definitions before C2X [-Wpedantic]
+    5 | cprog (void *)
+      |        ^~~~~~
+cprog.c:12:33: error: expected ')' before 'cb'
+   12 |    (int (*)(char *, int, char *)cb)(p1, p2, p3);
+      |    ~                            ^~
+      |                                 )
+cprog.c:12:4: warning: ISO C forbids conversion of object pointer to function pointer type [-Wpedantic]
+   12 |    (int (*)(char *, int, char *)cb)(p1, p2, p3);
+      |    ^
../../tests/run_misc.at:4826: exit code was 1, expected 0
783. run_misc.at:4772:  FAILED (run_misc.at:4826)

784. run_misc.at:4841: testing CALL C with callback, ENTRY-CONVENTION EXTERN ...
../../tests/run_misc.at:4902: $COMPILE -Wno-unfinished -o prog prog.cob cprog.c
--- /dev/null   2024-08-23 10:21:01.764081026 +0000
+++ /workspace/gnucobol/build/tests/testsuite.dir/at-groups/784/stderr  2024-08-23 16:08:20.465913157 +0000
@@ -0,0 +1,4 @@
+cprog.c: In function 'cprog':
+cprog.c:13:5: warning: ISO C forbids conversion of object pointer to function pointer type [-Wpedantic]
+   13 |    ((int (*)(char *, int, char *))cb)(p1, p2, p3);
+      |     ^
784. run_misc.at:4841:  FAILED (run_misc.at:4902)
~~~

and these are not depending on `-std=c89` at all (we get requests to fix those from package managers as well, I think that was in libcob, though).

libcob has only a `long long` warning on c89 (we commonly use something else in that case);
cobc has the amount of line in parser.c + some strange warnings on stdio.h not included in generated flex sources (which is included in its generation itself).

This project did help me to understand why we had - but only in one place - a "split memcpy" generation to 509 bytes - that is because since C89 there is only the guarantee for strings supported with this size.
I've adjusted this code and applied something similar also to other places, but coverage shows that not all of those were even tested in `make checkall`. configure now checks for a supported length of 2048, and if this isn't given (which is likely embedded where GMP is hard to get or very ancient or ... pedantic) stays below 509 chars.

and just FYI: json.h uses variadic macros, so for those environments above cjson.h is a necessary falback.

I did also add a testcase for one of the previous untested cases in that area and should commit that until Sunday (Changelogs missing).

The split of parser.y is planned since at least 2020 - https://sourceforge.net/p/gnucobol/feature-requests/371 and a user report of

> #20099-D: Exceeding compiler resource limits, routine: yyparse; some optimizations skipped. Use +Onolimit if override desired"

on HP-UX.

.github/workflows/windows-msvc.yml
```diff
@@ -41,8 +41,8 @@ jobs:
         target:
           - Debug
           - Release
-
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
```
ddeclerck, line 45: Might be a bit short when dependencies are not cached (I remember runs between 20-30 min)

nberth: All cases I've observed did not seem to exceed 20 to 25 minutes. I'll expand to 45 minutes for a better margin.

.github/workflows/macos.yml
```diff
@@ -55,44 +53,44 @@ jobs:
       - name: configure
         run: |
           cd _build
-          ../configure --enable-cobc-internal-checks --enable-hardening --prefix /opt/cobol/gnucobol-gcos --exec-prefix /opt/cobol/gnucobol-gcos
+          ../configure --enable-cobc-internal-checks \
+                       --enable-hardening \
+                       --prefix /opt/cobol/gnucobol-gcos \
+                       --exec-prefix /opt/cobol/gnucobol-gcos
```
GitMensch, line 59: that's a duplicate and should be dropped as it is by default identical to `prefix`

.github/workflows/ubuntu.yml
```diff
@@ -166,33 +162,34 @@ jobs:
       - name: configure
         run: |
           cd _build
-          ../configure --enable-code-coverage CPPFLAGS="-Werror=declaration-after-statement"
+          ../configure --enable-code-coverage \
+                       CPPFLAGS="-Werror=declaration-after-statement -pedantic" \
```
GitMensch, line 166: I suggest that we drop `-pedantic` in the CI for now (while I still want to fix a bunch of those soon, see my extra branch if you are interested; note: if this is used then GnuCOBOL limits the string size in several places, which _could_ also make a performance difference).

.github/workflows/ubuntu.yml
```diff
@@ -166,33 +162,34 @@ jobs:
...
-          cd _build
-          make --jobs=$(($(nproc)+1))
+          make -C _build --jobs=$(($(nproc)+1))
 
       - name: coverage
         run: |
-          cd _build
-          make check-code-coverage TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"
+          make -C _build check-code-coverage \
+                         TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"
```
GitMensch, line 182: Can we have both the "normal" and the extended coverage?
Also: should we use a two step approach for generated the "normal one"?

As an explanation for both - here's what we use in Appveyor:

```sh
# note: 'check-code-coverage' ignores errors, so doing it as two-step manually
# make check-code-coverage TESTSUITEFLAGS="--jobs=$(($(nproc)+1))" CODE_COVERAGE_DIRECTORY=$(realpath .)
make check TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"
make code-coverage-capture CODE_COVERAGE_DIRECTORY=$(realpath .)
appveyor AddMessage "internal testsuite and coverage check are finished" -Details "$(grep -A3 "Lines:\|Functions:\|Branches:" GnuCOBOL-$VERSION-coverage/index.html | sed -e "s/.*>\(.*\)<.*/\1/g" | tr '\n' ' ')"
zip -r GnuCOBOL-$VERSION-coverage.zip GnuCOBOL-$VERSION-coverage
appveyor PushArtifact GnuCOBOL-$VERSION-coverage.zip
make -k --jobs=$(($(nproc)+1)) -C tests/cobol85 test
# TODO: add NIST test summary in details
appveyor AddMessage "NIST testsuite is finished"
make code-coverage-capture CODE_COVERAGE_OUTPUT_DIRECTORY=extended-coverage CODE_COVERAGE_OUTPUT_FILE=extended-coverage.info CODE_COVERAGE_DIRECTORY=$(realpath .)
appveyor AddMessage "extended coverage check is finished" -Details "$(grep -A3 "Lines:\|Functions:\|Branches:" extended-coverage/index.html | sed -e "s/.*>\(.*\)<.*/\1/g" | tr '\n' ' ')"
zip -r GnuCOBOL-$VERSION-coverage-extended.zip extended-coverage
appveyor PushArtifact GnuCOBOL-$VERSION-coverage-extended.zip
```

.github/workflows/macos.yml
```diff
@@ -55,44 +53,44 @@ jobs:
...
         if: failure()
 
       - name: make
         run: |
-          cd _build
-          make --jobs=$((${NPROC}+1))
+          make -C _build --jobs=$((${NPROC}+1))
 
 # make install must be done before make check, otherwise execution of
 # generated COBOL files fail for a missing /usr/local/lib/libcob.dylib
```
GitMensch, line 73: Indent here?

GitMensch (changes requested):
Thanks for those changes - they are really useful.

There are still some remarks in this review; one thing that just came to mind that the only x86 test (32bit) we do are the runs on Windows... but we don't _need_ to adjust the ubuntu one for that now (it may be easy to do with the build matrix - or not, haven't checked).

.github/workflows/ubuntu.yml
nberth: I've also temporarily switched to `-std=99` as otherwise C++ comments raise errors  (even without `-pedantic`). I don't think there's an easy way to change that into a warning (*cf* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80529).

nberth:
@GitMensch Note: it appears that https://www.itl.nist.gov/div897/ctg/suites/newcob.val.Z is now a dead link. I've changed the workflows for GC3 to use the `tar.gz` that's on sourceforge, but I think there are other places where a similar change is requires (for GC4 probably, and maybe on appveyor?).

GitMensch:
Please revert that part as we explicitly don't support C++ comments and a falling CI ID therefore the current thing if used.

GitMensch:
The CI should have no download, only the cache of newcob.val - "make test" will do the download  and unpacking as needed, using multiple URLs if necessary.

Am 27. August 2024 12:09:17 MESZ schrieb Nicolas Berthier ***@***.***>:

nberth:
> The CI should have no download, only the cache of newcob.val - "make test" will do the download and unpacking as needed, using multiple URLs if necessary.

The issue is that the URL is not *dead* per se: the download operation (via `wget -t1 -T5) does not fail and brings and HTML page instead (which then ends up in the workflow cache). This is why I think it is better to use the sourceforge address, at least while this link is not fixed.

.github/workflows/ubuntu.yml
nberth: Ok I've reverted back to `-std=c89`.

GitMensch:
> > The CI should have no download, only the cache of newcob.val - "make test" will do the download and unpacking as needed, using multiple URLs if necessary.
> 
> The issue is that the URL is not _dead_ per se: the download operation (via `wget -t1 -T5) does not fail and brings and HTML page instead (which then ends up in the workflow cache). This is why I think it is better to use the sourceforge address, at least while this link is not fixed.

That's really bad. Please switch URLs to SF fir in the Makefile (upstream).

.github/workflows/ubuntu.yml
```diff
@@ -107,27 +102,30 @@ jobs:
           if-no-files-found: error
           retention-days: 0
 
-      - name: Cache newcob.val.Z
-        uses: actions/cache@v3
+      - name: Cache newcob.val.tar.gz
+        uses: actions/cache@v4
```
GitMensch, line 106: directly cache newcob.val, not the intermediate format

nberth:
> > > The CI should have no download, only the cache of newcob.val - "make test" will do the download and unpacking as needed, using multiple URLs if necessary.
> > 
> > 
> > The issue is that the URL is not _dead_ per se: the download operation (via `wget -t1 -T5) does not fail and brings and HTML page instead (which then ends up in the workflow cache). This is why I think it is better to use the sourceforge address, at least while this link is not fixed.
> 
> That's really bad. Please switch URLs to SF fir in the Makefile (upstream).

Ok I just did. The parts on `newcob.val.Z` are just commented, in case we find a fixed URL for that file (if that matters at all).

.github/workflows/windows-msvc.yml
```diff
@@ -160,54 +159,6 @@ jobs:
           cd tests
           sed -i '/AT_SETUP(\[runtime check: write to internal storage (1)\])/a AT_SKIP_IF(\[true\])' testsuite.src/run_misc.at
 
-      - name: Adjust testsuite for Debug target
```
ddeclerck, line 163: I tackled most of these issues, only two remain and are explicitly skipped in the workflow file (beware when rebasing)

.github/workflows/windows-msvc.yml
```diff
@@ -230,6 +181,7 @@ jobs:
           autom4te --lang=autotest -I ./testsuite.src ./testsuite.at -o ./testsuite
 
       - name: Run testsuite
+        continue-on-error: ${{ matrix.target == 'Debug' }}
```
ddeclerck, line 184: This should no longer be necessary (after rebase)

.github/workflows/windows-msvc.yml
nberth: Do you mean I should introduce the changes to this file from https://github.com/ddeclerck/gnucobol/commit/8ae65ac260aef95e850972b72eef1bdef1ab0ed3?

ddeclerck: That's the idea. Though if you just rebase on `gcos4gnucobol-3.x` you'll get those changes.

nberth: I still see all the old fixes on that branch. Maybe there's something missing…

https://github.com/OCamlPro/gnucobol/blob/gcos4gnucobol-3.x/.github/workflows/windows-msvc.yml

ddeclerck: Ah, yeah, forgot to update the MSVC workflow after the SVN merge. Should be fine now.

nberth:
Ouch https://github.com/OCamlPro/gnucobol/actions/runs/10595691899/job/29362006666?pr=170#step:14:22
@ddeclerck @GitMensch  would you have any idea how to solve that kind of issue? (it's MSYS2)

ddeclerck:
> Ouch https://github.com/OCamlPro/gnucobol/actions/runs/10595691899/job/29362006666?pr=170#step:14:22 @ddeclerck @GitMensch would you have any idea how to solve that kind of issue? (it's MSYS2)

Seems to be an open issue on V4 of upload-artifact. Cf [https://github.com/actions/upload-artifact/issues/485](https://github.com/actions/upload-artifact/issues/485)
I'd be tempted to downgrade to V3 until this is fixed...

GitMensch:
This is fine so I pull that in. Note that the new failure is because of C++ comments which will be fixed with my changes that are current "under work" (which will also fix most of C89 pedantic).

GitMensch:
@nberth Any chance to rebase and work on #109?

PR #172, opened 2024-08-21
Disable Windows error popups under testing environments

ddeclerck:
This silences the Windows error popups under testing environments (turns them into actual errors on stderr).

This required the use of `_CrtSetReportMode`, as the use of `SetErrorMode` did not help.

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -160,54 +160,6 @@ jobs:
           cd tests
           sed -i '/AT_SETUP(\[runtime check: write to internal storage (1)\])/a AT_SKIP_IF(\[true\])' testsuite.src/run_misc.at
 
-      - name: Adjust testsuite for Debug target
-        if: ${{ matrix.target == 'Debug' }}
-        shell: C:\shells\msys2bash.cmd {0}
-        run: |
-          cd tests
-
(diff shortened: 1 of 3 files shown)


GitMensch (changes requested):
That sounds very good in general.

Don't check for CI to not enable but check for something else, maybe
`if (!IsDebuggerPresent || !getenv ("DEBUG_POPUPS_WANTED"))`

The idea seems good but I can't see in the CI run if there are any testresults (and there is no testsuite.log artifact, so it _seems_ something didn't worked out).-

And please add the changelog entry.

ddeclerck:
Should be good this time.

Just wondering about the condition: should we keep the popups only when there is a debugger present AND  DEBUG_POPUPS_WANTED is set, or when either there is a debugger present OR DEBUG_POPUPS_WANTED is set ?

GitMensch:
"or", so that's fine for commit upstream (libcob) and the CI definition(s) here

GitMensch (approved):

ddeclerck:
> "or", so that's fine for commit

Is it ? I think I have to replace `||` by `&&` in the condition (the test is backwards ; when it evaluates to true, the popups are disabled).

GitMensch:
I'm sure you'll do the right thing for the spec "either debugger is running or that environment variable is set" 💪 .

ddeclerck:
Merged on SVN @ 5315.

GitMensch:
after a rebase this leaves only the CI adjustment - which should go to https://github.com/OCamlPro/gnucobol/pull/172 - so closing this

PR #176, opened 2024-08-30
Handle nested elements correctly when with attributes is specified

ammaralmorsi:
fixes [bug#961](https://sourceforge.net/p/gnucobol/bugs/961/)


tests/ChangeLog
GitMensch, line 1: We commonly add only entries here when moving things around or do general changes.

tests/testsuite.src/run_ml.at
```diff
@@ -117,6 +124,14 @@ AT_DATA([prog.cob], [
                DISPLAY 'Test 10 failed: ' FUNCTION TRIM (out)
            END-IF
            .
+
```
GitMensch, line 127: Please copy this testcase adding a new one XML GENERATE WITH ATTRIBUTES, this helps to easier spot later where issues are and to stay focused when debugging.
Also you can add a comment referencing the original bug report :-)

GitMensch (changes requested):
Testsuite adjustments needed, otherwise looks fine.

GitMensch:
rebased and squashed commits for upstreaming

GitMensch:
which was done with [[r5421]](https://sourceforge.net/p/gnucobol/code/5421/), so closing here

PR #177, opened 2024-09-12
Support collating sequence for indexed file keys of alphanumeric class

nberth:
- Warn when a collating sequence is ignored for keys of indexed files
- Use file collating sequence for any key of alphanumeric class

Before this change, only keys that were alphanumeric elementary items or numeric display were supported.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -9341,11 +9341,11 @@ output_indexed_file_key_colseq (const struct cb_file *f, const struct cb_alt_key
 {
 	const cb_tree	key = ak ? ak->key : f->key;
 	const cb_tree	key_col = ak ? ak->collating_sequence_key : f->collating_sequence_key;
-	const int	type = cb_tree_type (key, cb_code_field (key));
 	cb_tree		col = NULL;
 
-	/* We only apply a collating sequence if the key is alphanumeric / display */
-	if ((type & COB_TYPE_ALNUM) || (type == COB_TYPE_NUMERIC_DISPLAY)) {
+	/* We only apply a collating sequence if the key is of class alphanumeric;
+	   Warned in `validate_indexed_key_field`. */
+	if (CB_TREE_CLASS (key) == CB_CLASS_ALPHANUMERIC) {
 		col = key_col ? key_col : f->collating_sequence;
 #if 0	/* TODO: this should be done for national, when available */
 	} else if (type & COB_TYPE_NATIONAL) {
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -9347,10 +9347,8 @@ output_indexed_file_key_colseq (const struct cb_file *f, const struct cb_alt_key
 	   Warned in `validate_indexed_key_field`. */
 	if (CB_TREE_CLASS (key) == CB_CLASS_ALPHANUMERIC) {
 		col = key_col ? key_col : f->collating_sequence;
-#if 0	/* TODO: this should be done for national, when available */
-	} else if (type & COB_TYPE_NATIONAL) {
-		col = key_col_n ? key_col_n : f->collating_sequence_n;
-#endif
+	} else if (CB_TREE_CLASS (key) == CB_CLASS_NATIONAL) {
+		col = f->collating_sequence_n;
 	}
 
 	output_prefix ();
--- a/cobc/tree.c
+++ b/cobc/tree.c
@@ -4781,6 +4781,21 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
 			}
 		}
 	}
+
+	/* check collating sequence is not ignored */
+	if (CB_TREE_CLASS (k) != CB_CLASS_ALPHANUMERIC) {
+		cb_tree colseq = (cbak == NULL)
+			? f->collating_sequence_key
+			: cbak->collating_sequence_key;
+		if (colseq == NULL) {
+			colseq = f->collating_sequence;
+		}
+		if (colseq != NULL) {
+			cb_warning_x (COBC_WARN_FILLER, CB_TREE(colseq),
+				      _("COLLATING SEQUENCE is ignored for non-alphanumeric"
+					" key '%s'"), k->name);
+		}
+	}
 }
 
 void
--- a/cobc/tree.c
+++ b/cobc/tree.c
@@ -4783,7 +4783,8 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
 	}
 
 	/* check collating sequence is not ignored */
-	if (CB_TREE_CLASS (k) != CB_CLASS_ALPHANUMERIC) {
+	if (get_warn_opt_value (cb_warn_filler) != COBC_WARN_DISABLED
+	 && CB_TREE_CLASS (k) != CB_CLASS_ALPHANUMERIC) {
 		const char *source = "KEY";
 		cb_tree colseq = (cbak == NULL)
 			? f->collating_sequence_key
--- a/cobc/tree.c
+++ b/cobc/tree.c
@@ -4784,16 +4784,21 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
 
 	/* check collating sequence is not ignored */
 	if (CB_TREE_CLASS (k) != CB_CLASS_ALPHANUMERIC) {
+		const char *source = "KEY";
 		cb_tree colseq = (cbak == NULL)
 			? f->collating_sequence_key
 			: cbak->collating_sequence_key;
+		cb_tree pos = colseq;
 		if (colseq == NULL) {
+			source = "FILE";
 			colseq = f->collating_sequence;
+			pos = key_ref;
 		}
 		if (colseq != NULL) {
-			cb_warning_x (COBC_WARN_FILLER, CB_TREE(colseq),
-				      _("COLLATING SEQUENCE is ignored for non-alphanumeric"
-					" key '%s'"), k->name);
+			cb_warning_x (COBC_WARN_FILLER, CB_TREE (pos),
+				      _("%s COLLATING SEQUENCE '%s' is ignored "
+					"for non-alphanumeric key '%s'"),
+				      source, CB_NAME (colseq), k->name);
 		}
 	}
 }
--- a/tests/testsuite.src/run_file.at
+++ b/tests/testsuite.src/run_file.at
@@ -12685,15 +12685,15 @@ CCC 888 +0000000043 1
 DONE
 ])
 
+# Note: `-Wno-others` is for ignored COLLATIONS for non-alphanumeric keys
+
 # Testing ASCII file collating sequence using clause
 AT_DATA([prog1.cob], [
        COPY "prog.cpy" REPLACING
          ==FILE-COLSEQ== BY ==COLLATING SEQUENCE IS ASCII==
          ==KEY-COLSEQ== BY ====.
 ])
-AT_CHECK([$COMPILE -Wno-unfinished prog1.cob], [0], [],
-[prog.cpy:31: warning: FILE COLLATING SEQUENCE 'ASCII' is ignored for non-alphanumeric key 'MY-AKEY2'
-])
+AT_CHECK([$COMPILE -Wno-unfinished -Wno-others prog1.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog1 1>prog1.out], [0], [], [])
 AT_CHECK([diff reference_ascii prog1.out], [0], [], [])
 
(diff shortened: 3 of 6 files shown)


cobc/tree.c
```diff
@@ -4721,6 +4721,15 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
 	}
 	k = CB_FIELD_PTR (key_ref);
 
+	/* check collating sequence is not ignored */
+	if (f->collating_sequence != NULL &&
+	    cbak == NULL &&	/* only if no alternate key is given (bit of hack for now) */
+	    CB_TREE_CLASS (k) != CB_CLASS_ALPHANUMERIC) {
```
GitMensch, line 4727: style: please check around line 4890 to see our special style :-)

cobc/tree.c
```diff
@@ -4721,6 +4721,15 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
 	}
 	k = CB_FIELD_PTR (key_ref);
 
+	/* check collating sequence is not ignored */
+	if (f->collating_sequence != NULL &&
+	    cbak == NULL &&	/* only if no alternate key is given (bit of hack for now) */
```
GitMensch, line 4726: Why not checking the alternative keys in a loop as well? Aren't the collations used for those?
What about split keys?

tests/testsuite.src/syn_file.at
```diff
@@ -680,14 +688,15 @@ AT_DATA([prog.cob], [
 ])
 
 # FIXME: "is not defined" should be changed in "is not defined in file ..."
-AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+AT_CHECK([$COMPILE_ONLY -Wno-unfinished prog.cob], [1], [],
```
GitMensch, line 691: I suggest using a numeric binary instead of the bit item - that should raise the same message and is fully implemented, no?

tests/testsuite.src/run_file.at
```diff
@@ -12769,6 +12769,97 @@ AT_CHECK([diff reference_ebcdic_ascii prog8.out], [0], [], [])
 AT_CLEANUP
 
 
+
+# This is, so far, only supported by the BDB backend
+AT_SETUP([INDEXED file with collation on group key])
+AT_KEYWORDS([runfile WRITE READ EBCDIC])
+
+AT_SKIP_IF([test "$COB_HAS_ISAM" != "db"])
```
GitMensch, line 12777: I think this should be changed to an expected fail if not BDB

tests/testsuite.src/run_file.at
```diff
@@ -12769,6 +12769,97 @@ AT_CHECK([diff reference_ebcdic_ascii prog8.out], [0], [], [])
...
+223
+335
+447
+])
+
+AT_CHECK([$COMPILE -Wno-unfinished -fdefault-file-colseq=ASCII prog.cob -o ascii], [0], [], [])
+AT_CHECK([$COBCRUN_DIRECT ./ascii 1>ascii.out], [0], [], [])
+AT_CHECK([diff reference_ascii ascii.out], [0], [], [])
+
+AT_CHECK([$COMPILE -Wno-unfinished -fdefault-file-colseq=EBCDIC prog.cob -o ebcdic], [0], [], [])
```
GitMensch, line 12856: What is unfinished here? That should be changed to unsupported (and if not already done, if the collations in general are not-applied (because other backend as BDB) then there should be an unsupported warning.

GitMensch (changes requested):
minor changes requested, especially to the test cases and likely to a missing warning

GitMensch:
ping @nberth to commit/comment on the review notes above

tests/testsuite.src/syn_file.at
nberth: Actually this is for `prog.cob:30: warning: handling of FILE COLLATING SEQUENCE is unfinished; implementation is likely to be changed`. Not sure what's best here: silence the warning? or include it in the test output?

tests/testsuite.src/run_file.at
nberth: Yes you're right, that option can be dropped here.

tests/testsuite.src/syn_file.at
GitMensch: Make a conditional to expect for BDB, not in other cases (like the 32bit test, depending on "failure" or "pass" run the actual test)?

tests/testsuite.src/run_file.at
nberth: Actually found out `autotest` has some nice constructs to mimic that.

cobc/tree.c
nberth: Finding a way to avoid emitting "too much" warnings for the file-level collating sequence seems a little tedious here.
Split keys appear to always be handled as alphanumerics (as per the code).

ddeclerck:
Looks like this could be merged ; is anything missing here @GitMensch ?

tests/testsuite.src/syn_file.at
```diff
@@ -681,13 +724,23 @@ AT_DATA([prog.cob], [
 
 # FIXME: "is not defined" should be changed in "is not defined in file ..."
 AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
-[prog.cob:10: error: 'TEST-P2' is not defined
+[prog.cob:33: warning: handling of FILE COLLATING SEQUENCE is unfinished; implementation is likely to be changed
+prog.cob:33: warning: handling of KEY COLLATING SEQUENCE is unfinished; implementation is likely to be changed
+prog.cob:39: warning: handling of FILE COLLATING SEQUENCE is unfinished; implementation is likely to be changed
+prog.cob:39: warning: handling of KEY COLLATING SEQUENCE is unfinished; implementation is likely to be changed
+prog.cob:46: warning: handling of FILE COLLATING SEQUENCE is unfinished; implementation is likely to be changed
+prog.cob:46: warning: handling of KEY COLLATING SEQUENCE is unfinished; implementation is likely to be changed
```
GitMensch, line 732: Why do we have two diagnostics on the same line?

GitMensch (changes requested):
Wondering: don't we need a `--without-db` and `--with-vbisam` setup in the workflow to verify the tests work as expected in those environments?

tests/testsuite.src/run_file.at
```diff
@@ -12762,13 +12772,112 @@ AT_DATA([prog8.cob], [
...
+])
+
+# Note: ignore any unfinished warning as the test is about the runtime behavior:
+AT_CHECK([$COMPILE -Wno-unfinished -fdefault-file-colseq=ASCII prog.cob -o ascii], [0])
+AT_CHECK([$COMPILE -Wno-unfinished -fdefault-file-colseq=EBCDIC prog.cob -o ebcdic], [0])
+
+# This is, so far, only supported by the BDB backend
+AT_CHECK([test "$COB_HAS_ISAM" = "db"], [0], [], [],
+# Previous test "failed" --> other ISAM, skip the test
+[AT_CHECK([true])],
```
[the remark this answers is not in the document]
nberth: I get an "UNEXPECTED PASS" when if I add an `AT_XFAIL_IF([true])` here and when I configure `--with-db`.
I'm guessing `AT_XFAIL_IF` may not be a regular autotest macro that just does in-place shell-script expansion.

We could just turn the non-db case into a somewhat "artificial" expected failure with:
```
AT_XFAIL_IF([test "$COB_HAS_ISAM" != "db"])
AT_CHECK([test "$COB_HAS_ISAM" = "db"], [0], [], [],
# Previous test "failed" --> other ISAM, fail the test
[AT_CHECK([false])],
```
… but that may be a bit strange to do it like this.

Side note: the version of VBISAM I found [here](https://sourceforge.net/projects/vbisam/files/vbisam2/) for testing this required to `#define vbisam_off_t off_t` in `vbisam.h` to compile properly.

ddeclerck:
> Wondering: don't we need a `--without-db` and `--with-vbisam` setup in the workflow to verify the tests work as expected in those environments?

Last time I checked VBISAM with 3.x (was a while ago), there were testsuite failures.

GitMensch:
> Last time I checked VBISAM with 3.x (was a while ago), there were testsuite failures.

That would be unrelated to a `--without-db --without-curses --without-xml2 --without-json` build :-)

But even with a set of failures (even the NEWS file says that depending on the version of vbisam used there will be some unexpected passes and failures and will do so until we have a V-ISAM version released) this gets "only" down to working it out - we tinker the MSYS2 and MSVC builds for some testsuite expectations in GitHub CI as well - and this does help to catch any _new_ issues that arise (and may also show some positive side effects we otherwise don't catch).
Note: I'm not planning to adjust *any* test expectations specific _for_ VBISAM in GnuCOBOL 3.3 but the "no isam" and "not bdb" ones should definitely work.

GitMensch:
I'm taking on the "minimal build" and send a PR for that (this should pass here in any case) and would let the vbisam part optional to you.

ddeclerck:
> > Last time I checked VBISAM with 3.x (was a while ago), there were testsuite failures.
> 
> That would be unrelated to a `--without-db --without-curses --without-xml2 --without-json` build :-)
> 
Sure, but you mentionned `--with-vbisam` :-)

cobc/tree.c
```diff
@@ -4781,6 +4781,21 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
...
+		cb_tree colseq = (cbak == NULL)
+			? f->collating_sequence_key
+			: cbak->collating_sequence_key;
+		if (colseq == NULL) {
+			colseq = f->collating_sequence;
+		}
+		if (colseq != NULL) {
+			cb_warning_x (COBC_WARN_FILLER, CB_TREE(colseq),
+				      _("COLLATING SEQUENCE is ignored for non-alphanumeric"
+					" key '%s'"), k->name);
```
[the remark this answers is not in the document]
nberth: Done and improved with a `KEY` or `FILE` prefix to better show what kind of key is ignored.

ddeclerck:
I think Nicolas took care of all the requested changes ?
If so, is this OK to be merged @GitMensch ? (it's another burning issue for our customer)

tests/testsuite.src/run_file.at
```diff
@@ -12691,7 +12691,9 @@ AT_DATA([prog1.cob], [
          ==FILE-COLSEQ== BY ==COLLATING SEQUENCE IS ASCII==
          ==KEY-COLSEQ== BY ====.
 ])
-AT_CHECK([$COMPILE -Wno-unfinished prog1.cob], [0], [], [])
+AT_CHECK([$COMPILE -Wno-unfinished prog1.cob], [0], [],
```
GitMensch, line 12694: please add `-Wno-others` here and below (or the appropriate new `-Wno-ignored-something`, if you for some reason want to add that), the warning is checked in syn_file.at

tests/testsuite.src/run_file.at
```diff
@@ -12762,13 +12772,112 @@ AT_DATA([prog8.cob], [
...
+AT_CHECK([test "$COB_HAS_ISAM" = "db"], [0], [], [],
+# Previous test "failed" --> other ISAM, fail the test
+[AT_CHECK([false])],
+
+# Previous test "passed" --> using BDB, perform the actual test
+[AT_CHECK([
+  echo "ASCII:"  && $COBCRUN_DIRECT ./ascii && \
+  echo "EBCDIC:" && $COBCRUN_DIRECT ./ebcdic
+], [0], [expout])		# <- compare stdout with exiting `expout`
+])
```
GitMensch, line 12876: `AT_XFAIL_IF` is enough - please run the common test without another conditional.

That use of the testsuite is indeed interesting and saves us from running multiple calls and then an extra diff.
We may should use that in other places (but still, please do it without the conditional.

GitMensch (changes requested):
Please adjust the testcases, then it is good to go (with the option to add a new warning group and/or national collation "up to libcob" now instead of later)

cobc/tree.c
```diff
@@ -4781,6 +4781,26 @@ validate_indexed_key_field (struct cb_file *f, struct cb_field *records,
...
+			? f->collating_sequence_key
+			: cbak->collating_sequence_key;
+		cb_tree pos = colseq;
+		if (colseq == NULL) {
+			source = "FILE";
+			colseq = f->collating_sequence;
+			pos = key_ref;
+		}
+		if (colseq != NULL) {
+			cb_warning_x (COBC_WARN_FILLER, CB_TREE (pos),
```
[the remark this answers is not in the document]
nberth: May `ignored-collations` be too precise for this?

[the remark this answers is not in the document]
GitMensch: That would work but I'd prefer a warning option that we can use for any ignored elements (so yes, too specific ;-)

cobc/codegen.c
```diff
@@ -9341,11 +9341,11 @@ output_indexed_file_key_colseq (const struct cb_file *f, const struct cb_alt_key
...
-	const int	type = cb_tree_type (key, cb_code_field (key));
 	cb_tree		col = NULL;
 
-	/* We only apply a collating sequence if the key is alphanumeric / display */
-	if ((type & COB_TYPE_ALNUM) || (type == COB_TYPE_NUMERIC_DISPLAY)) {
+	/* We only apply a collating sequence if the key is of class alphanumeric;
+	   Warned in `validate_indexed_key_field`. */
+	if (CB_TREE_CLASS (key) == CB_CLASS_ALPHANUMERIC) {
 		col = key_col ? key_col : f->collating_sequence;
 #if 0	/* TODO: this should be done for national, when available */
```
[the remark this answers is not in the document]
GitMensch: Thanks for handling it!

GitMensch (approved):

nberth:
Now upstream.

cobc/tree.c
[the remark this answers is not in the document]
GitMensch: Even more - totally global - so we can include every "ignored" that has no separate warning level (that should be 1 or two warnings currently).

And yes, that's for "sometime in the future".

PR #147, opened 2024-05-29
CI for forward-porting GC3 patches to GC4

ddeclerck:
Follow-up of #146.

PR #178, opened 2024-09-19
ChangeLog and testsuite adjustments

ddeclerck:
As requested [here](https://github.com/OCamlPro/gnucobol/pull/147#discussion_r1763582461) and [here](https://github.com/OCamlPro/gnucobol/pull/147#discussion_r1764849089).

Also it seems at some point recently some changes broke the CI - I'll investigate and update this PR.

### DIFF
--- a/tests/testsuite.src/syn_definition.at
+++ b/tests/testsuite.src/syn_definition.at
@@ -2878,3 +2878,72 @@ prog.cob:6: error: 'DEPTHDR' is not defined
 
 AT_CLEANUP
 
+
+AT_SETUP([POINTER TO typedef-name])
+AT_KEYWORDS([POINTER TYPEDEF])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 data-type PIC X(10) USAGE DISPLAY IS TYPEDEF.
+       01 data-ptr1 USAGE POINTER data-type.
+       01 data-ptr2 USAGE POINTER TO data-type.
+       PROCEDURE DIVISION.
+           STOP RUN.
+       END PROGRAM prog.
+])
+
+AT_CHECK([$COMPILE_ONLY -Wno-pending prog.cob], [0], [], [])
+AT_CLEANUP
+
+
+AT_SETUP([FUNCTION-POINTER TO function-name])
+AT_KEYWORDS([FUNCTION-POINTER])
+
+AT_DATA([prog.cob], [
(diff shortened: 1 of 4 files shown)


GitMensch:
the test looks good, but it seems something is indeed broken in the parser ?!?

ddeclerck:
> the test looks good, but it seems something is indeed broken in the parser ?!?

Yeah but this is unrelated to the conflicts mentionned in the GC3-GC4 merge PR.
It seems `USAGE PROGRAM-POINTER` with a program-prototype argument has never been parsed correctly.

GitMensch:
Are you working on this?

ddeclerck:
> Are you working on this?

I'd rather postpone this issue for now for I fear it would take me too far away from the GC3-GC4 merge (plus I'm already being taken away by PR 109).

ddeclerck:
> Make it an expected failure then and commit upstream, but don't merge it yet - just verify manually that the result with your GC4merge is not worse.

Maybe I can split the 3 cases POINTER/FUNCTION-POINTER/PROGRAM-POINTER so that we can at least "protect" what already works ?

GitMensch:
Good idea. Possibly with a comment to combine again later. Can directly go upstream.

tests/testsuite.src/syn_definition.at
```diff
@@ -2878,3 +2878,76 @@ prog.cob:6: error: 'DEPTHDR' is not defined
...
+       01 data-type PIC X(10) USAGE DISPLAY IS TYPEDEF.
+       01 data-ptr1 USAGE POINTER data-type.
+       01 data-ptr2 USAGE POINTER TO data-type.
+       PROCEDURE DIVISION.
+           STOP RUN.
+       END PROGRAM prog.
+])
+
+AT_CHECK([$COMPILE_ONLY -std=default -Wno-pending prog.cob], [0], [], [])
+
```
GitMensch, line 2900: drop the empty line here as wall as -std=default, only have the -Wno-pending if needed, same for the other tests

tests/testsuite.src/syn_definition.at
```diff
@@ -2878,3 +2878,76 @@ prog.cob:6: error: 'DEPTHDR' is not defined
 
 AT_CLEANUP
 
+
+# Note: the three tests below should be combined in a single one (when fully supported)
```
GitMensch, line 2882: hm. let's drop the comment and rather extend the checks later (using the pointers)

GitMensch (approved):
good to go for upstream, can also be included in the GC4 merge (the CI fixes and it isn't bad to have the tests as-is in there as well), can be a bit improved, see below, and then checked in

PR #180, opened 2024-09-20
Testsuite adjustments (extracted from #109)

ddeclerck:
This PR contains parts of #109 that were unrelated to the feature introduced by the PR.


GitMensch:
Thank you, does it increase the number of working tests on any environment?

If not, why should we include that?

ddeclerck:
> Thank you, does it increase the number of working tests on any environment?
> 
> If not, why should we include that?

I don't know yet, I just extracted those to focus on the actual contents of #109.

PR #181, opened 2024-09-23
Check for incompatible data only when a receiver field is of category numeric in `MOVE` or `SET`

nberth:
This PR alters `MOVE` and `SET` statements to check for numerical data in emitter fields only when at least one receiver field is of category numeric. This essentially permits statements like `MOVE SOME-NUMERIC TO SOME-ALPHANUM` to succeed even when `SOME-NUMERIC` holds invalid data.

---

(Old description:)

On *some* systems (GCOS), checks for non-numeric data in numeric items on MOVE appear to be performed on receiver fields instead of emitters, whereas in GnuCOBOL the checks are always performed on emitters. However, the former behavior seems convenient for encoding "absence" of relevant values (or, in more modern terms, null-ness) by putting `SPACES` in the associated storage area. This notably allows coding patterns where a `MOVE SOME-NUMERIC TO SOME-ALPHANUM` is followed by a check for "nullness" with `SOME-ALPHANUM IS EQUAL TO SPACES`.

Although disabling the (fatal) exception `EC-DATA-INCOMPATIBLE` permits such a coding pattern, it also disables some other checks that are relevant.

Another possible workaround is to use the (somewhat obscure) `-fcorrect-numeric`. Yet, this "solution" currently alters emitter fields on MOVES (to put `ZEROS` instead of `SPACES`), and this is an often undesirable side-effet.

This PR currently identifies changes required to implement the check on receiver fiels instead of emitters. Given that this changes the underlying semantics of `MOVE`/`SET` (albeit in possibly marginal/rare cases — and this fails a NIST test apparently), one could make this subject to a dialect option.

### DIFF
--- a/cobc/typeck.c
+++ b/cobc/typeck.c
@@ -1055,16 +1055,31 @@ cb_emit_list (cb_tree l)
 	return l;
 }
 
+static COB_INLINE COB_A_INLINE int
+cb_tree_is_numeric_ref_or_field (cb_tree x, int include_numeric_edited) {
+	int cat;
+	if (!x || !CB_REF_OR_FIELD_P (x)) {
+		return 0;
+	}
+	cat = CB_TREE_CATEGORY (x);
+	return (cat == CB_CATEGORY_NUMERIC
+	     || (include_numeric_edited && cat == CB_CATEGORY_NUMERIC_EDITED));
+}
+
+static int
+cb_tree_list_has_numeric_ref_or_field (cb_tree l) {
+	for (l;
+	     l && !cb_tree_is_numeric_ref_or_field (CB_VALUE (l), 1);
+	     l = CB_CHAIN (l));
+	return (l != NULL);
+}
+
 static void
 cb_emit_incompat_data_checks (cb_tree x)
 {
 	struct cb_field		*f;
 
-	if (!x || x == cb_error_node) {
-		return;
-	}
-	if (!CB_REF_OR_FIELD_P (x)
-	 || CB_TREE_CATEGORY (x) != CB_CATEGORY_NUMERIC) {
+	if (!cb_tree_is_numeric_ref_or_field (x, 0)) {
 		return;
 	}
 	f = CB_FIELD_PTR (x);
@@ -12884,8 +12899,11 @@ cb_emit_move (cb_tree src, cb_tree dsts)
 		return;
 	}
 
-	/* validate / fix-up source, if requested */
-	cb_emit_incompat_data_checks (src);
+	/* validate / fix-up source, if at least one receiver is of category
+	   numeric */
+	if (cb_tree_list_has_numeric_ref_or_field (dsts)) {
+		cb_emit_incompat_data_checks (src);
+	}
 
 	/* FIXME: this is way to much to cater for sum field */
 	src = cb_check_sum_field (src);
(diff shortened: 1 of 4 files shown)


GitMensch:
Are you sure that GCOS checkls on the receiver side? That would be quite strange.

But one thing that could be useful is to only check on arithmetic statements, and never on `MOVE`/`SET TO`...

nberth:
> Are you sure that GCOS checkls on the receiver side? That would be quite strange.
> 
> But one thing that could be useful is to only check on arithmetic statements, and never on `MOVE`/`SET TO`...

From our tests we could identify that GCOS allows the aforementioned code pattern (`MOVE` from a numeric with invalid data to an alphanumeric item). Still, it errors with `INVALID DECIMAL DATA` on similar moves to a numeric item.
My current interpretation is that checks are done on receiver fields, but another reason could be that checks for numeric are performed on `MOVE`s only when numeric conversions take place.

GitMensch:
yes to the later part, the standard does say that no conversion (only padding/truncation) is done for "same usage", but I'm not 100% sure what that is supposed to mean...

ddeclerck:
Is this okay to be merged @GitMensch ? (this is a critical issue for our customer)

tests/testsuite.src/run_misc.at
```diff
@@ -1172,6 +1172,32 @@ AT_CHECK([$COBCRUN_DIRECT ./prog2], [1], [],
 AT_CLEANUP
 
 
+AT_SETUP([Non-numeric data in numeric items (lax checks)])
```
GitMensch, line 1175: That is not "lax check" but "MOVE of invalid data in numeric item to alphanumeric" or similar like that

GitMensch (approved):
"Minor" thing to do, then good to go upstream is the documentation part, as we change a long-standing way to handle this.

Please add an entry int the "changed" part of the NEWS file during commit.

cobc/typeck.c
```diff
@@ -1055,16 +1055,31 @@ cb_emit_list (cb_tree l)
 	return l;
 }
 
+static COB_INLINE COB_A_INLINE int
+cb_tree_is_numeric_ref_or_field (cb_tree x) {
+	if (x
+	 && x != cb_error_node
+	 && CB_REF_OR_FIELD_P (x)
+	 && CB_TREE_CATEGORY (x) == CB_CATEGORY_NUMERIC) {
```
GitMensch, line 1063: no need to check for error node - that has no numeric category
so possibly just for x, then get the category - it has to be also generated for `CB_CATEGORY_NUMERIC_EDITED`!

nberth: So, this changes the previous behavior of `cb_emit_incompat_data_checks`, right?

nberth: Ignoring `cb_error_node` is fine.
But handling `CB_CATEGORY_NUMERIC_EDITED` in source fields seems to break many tests.
Allowing checks in case numeric-edited receivers makes sense, though: I'll include this (and add that in the test as well).

GitMensch:
Hm, can you please also have a look into the codegen and potentially cob_move?
At least "if the target is alphanumeric, the source should be copied over verbatim (memcpy field->data + memcpy COB_ALPHANUMERIC_SPACE), no?

GitMensch:
NEWS (+potentially Changedoc adjustment for the last push) would be missing

nberth:
> Hm, can you please also have a look into the codegen and potentially cob_move? At least "if the target is alphanumeric, the source should be copied over verbatim (memcpy field->data + memcpy COB_ALPHANUMERIC_SPACE), no?

Yes I think all these cases are handled in `typeck.c:cb_build_move_copy` via `typeck.c:cb_build_move_field`. `cob_move` seems to do the direct copy as well.

GitMensch:
LGTM

nberth:
Now upstream

PR #184, opened 2024-09-27
CI: adding minimal build

GitMensch:
as mentioned in [#177](https://github.com/OCamlPro/gnucobol/pull/177#issuecomment-2378725729)

additional:
* added NIST85 summary and duration files to the NIST result artifact
* adjust artifact names to prevent conflicts

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,26 +1,25 @@
 name: Ubuntu Workflow
 
 on:
+
   pull_request:
     branches: [ gcos4gnucobol-3.x ]
   push:
   # manual run in actions tab - for all branches
   workflow_dispatch:
 
+
 jobs:
+
   build:
     name: Build, test and provide nightly
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest ]
+        os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


GitMensch (commented):
The result LGTM now, so just waiting for a second look.

Note: I haven't checked if we may optimize the long running MSYS workflows (we could, for example, also use the CI tarball generated here, if that helps - but that would need to be a bit adjusted - still possible)

nberth (approved):
Just in case, I had a quick look as well.

ddeclerck:
Was late on this one, but LGTM ;)

> Note: I haven't checked if we may optimize the long running MSYS workflows (we could, for example, also use the CI tarball generated here, if that helps - but that would need to be a bit adjusted - still possible)

Regarding this issue, it occurs randomly (but often) ; could it be a locking issue ?

GitMensch:
I guess that's a locking issue on the underlying MS Windows OS, if you use multiple jobs it is much more easy to reproduce...

Concerning the other current PR with the build system update ... I've just recognized that most of the appveyor CI is broken now as many seem to run with autoconf 2.69 or older, here it is only the MSYS1 version (which I had expected). So I'll inspect that first...

PR #185, opened 2024-09-27
update from upstream

GitMensch:
NIST testsuite and build system update


GitMensch:
As expected: MSY1 cannot (if not manually updated) create the build system. I can have a look at adjusting the CI:

* drop the likely not needed "install" option during bootstrap
* let MSYS1 use the CI dist tarball from the Ubuntu test (we do it similar since years on Appveyor).
* I guess I will push the changes here as well, it likely will be some force-pushes (= also days) until everything is finished.

ddeclerck:
> As expected: MSY1 cannot (if not manually updated) create the build system. I can have a look at adjusting the CI:

That would be very nice ; this CI was quite tricky to get working 😅

PR #186, opened 2024-09-29
Update CI, mostly MSYS1 (now using CI tarball instead of checkout)

GitMensch:
ci adjustments

* ubuntu:
  * only run jobs for "coverage" and "additional warnings" if the main ci build works
    and use its generated tarball in both cases
  * add two new artifacts: test binaries and windows source
* msys1+ubuntu:
  * upload config.log after the build - because we may need it to debug build issues
  * always upload the testsuite.log (additional build documentation)
* msys1:
  * GMP url changes, building it again for performance reasons
  * building BDB with all relevant patches from MSYS2
  * using msys-build instead of building Bison (only necessary for GC4)
  * drop GC install log step and therefore extra prefix
  * drop extra CFLAGS previously necessary for local cJSON (fixed in 3.x)
    and/or ignoring failing NIST) --> as after last upstream update everything works
  * enable NIST85 (+ comment-code in case we ever need to skip something there
    and/or ignoring failing NIST) --> as after last upstream update everything works
  * ci cache adjustment:
    * remove split per matrix
    * split per software, enabling smaller updates
  * use CI tarball like for the minimal build
  * drop workflow specific expected failures that now work fine
  * move env to MSYS job
* integrate msys1.yml into ubuntu.yml, renaming to build_nightly.yml

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,15 @@
 
+2024-10-01  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
+
+	* tree.c (validate_indexed_key_field): warn about ignored collating
+	  sequence for non-alphanumeric keys (considers only primary keys and file
+	  collating sequence for now)
+	* codegen.c (output_indexed_file_key_colseq): assign collating sequence
+	  for any key of alphanumeric class, and preliminary handing of NATIONAL
+	  collations
+	* parser.y: adjust position of messages about unfinished KEY or FILE
+	  COLLATING SEQUENCE
+
 2024-09-29  Simon Sobisch <simonsobisch@gnu.org>
 
 	* cobc.c (cobc_print_info): drop COB_LI_IS_LL
@@ -6,7 +17,7 @@
 
 2024-09-27  Simon Sobisch <simonsobisch@gnu.org>
 
-    * plex.l, scanner.l: use noyywrap option instead of manually
+	* plex.l, scanner.l: use noyywrap option instead of manually
 	  defining related code parts
 	* typeck.c (cb_tree_list_has_numeric_ref_or_field): cleanup
 
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -9341,16 +9341,14 @@ output_indexed_file_key_colseq (const struct cb_file *f, const struct cb_alt_key
 {
 	const cb_tree	key = ak ? ak->key : f->key;
 	const cb_tree	key_col = ak ? ak->collating_sequence_key : f->collating_sequence_key;
-	const int	type = cb_tree_type (key, cb_code_field (key));
 	cb_tree		col = NULL;
 
-	/* We only apply a collating sequence if the key is alphanumeric / display */
-	if ((type & COB_TYPE_ALNUM) || (type == COB_TYPE_NUMERIC_DISPLAY)) {
+	/* We only apply a collating sequence if the key is of class alphanumeric;
+	   Warned in `validate_indexed_key_field`. */
+	if (CB_TREE_CLASS (key) == CB_CLASS_ALPHANUMERIC) {
 		col = key_col ? key_col : f->collating_sequence;
-#if 0	/* TODO: this should be done for national, when available */
-	} else if (type & COB_TYPE_NATIONAL) {
-		col = key_col_n ? key_col_n : f->collating_sequence_n;
-#endif
+	} else if (CB_TREE_CLASS (key) == CB_CLASS_NATIONAL) {
+		col = f->collating_sequence_n;
 	}
 
 	output_prefix ();
--- a/cobc/parser.y
+++ b/cobc/parser.y
@@ -5784,7 +5784,7 @@ collating_sequence_clause:
 	check_repeated ("COLLATING", SYN_CLAUSE_3, &check_duplicate);
 	current_file->collating_sequence = alphanumeric_collation;
 	current_file->collating_sequence_n = national_collation;
-	CB_UNFINISHED ("FILE COLLATING SEQUENCE"); /* only implemented for BDB */
+	CB_UNFINISHED_X (alphanumeric_collation, "FILE COLLATING SEQUENCE"); /* only implemented for BDB */
   }
 ;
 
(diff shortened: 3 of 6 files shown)


GitMensch:
@lefessan @ddeclerck @nberth So this current version does work (or at least should), but it is a rather complicate setup:
* Ubuntu workflow creates the dist tarball after running initial checks
* it starts the MSYS1 workflow "detached", passing its run it
* the MSYS1 workflow downloads the artifact, then uses it

The biggest downside is, that using this approach the MSYS1 workflow doesn't show up in the PR workflow at all, you can only see it in the "actions" tab.

Options to go on:
1. leave as is, merge (and loose MSYS1 out of sight)
2. make the MSYS1 more complicated, by:
 * trigger the second workflow as part of the PR again
 * do a minimal checkout to throw away later, get the git sha out of it
 * poll the GitHub API to check the status of the "Ubuntu workflow" with the git sha , get its run-id and wait until it completes
 * once that happened, fetch the artifact
 * handle timeouts in case the "Ubuntu workflow" takes too long to finis
3. integrate the MSYS1 workflow into the Ubuntu workflow, making it look *a bit* more complicated (and obviously: we should rename it then); this would work as we can set all of build-matrix, runs-on and env on the  job scope; this we we can easily wait for the artifact and get it even without a token

I tend to option 3, the only side effects I see are that we can't only run the Ubuntu or MSYS1 jobs then (as far as I know it is always "all of a single workflow").

.github/workflows/ubuntu.yml
```diff
@@ -55,78 +55,107 @@ jobs:
       #
       - name: Build environment setup
         run: |
-          mkdir _build
+          mkdir build
```
ddeclerck, line 58: Is this necessary ? We like having those build and temporary folders starting with _ to distinguish them easily.

.github/workflows/ubuntu.yml
```diff
@@ -55,78 +55,107 @@ jobs:
...
                          TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"  \
                          --jobs=$(($(nproc)+1)) ||                \
-          make -C _build/gnucobol-$VERSION/_build/sub/tests check \
+          make -C build/gnucobol-$VERSION/build/sub/tests check \
                          TESTSUITEFLAGS="--recheck --verbose"
 
       - name: Upload testsuite.log
         uses: actions/upload-artifact@v4
-        if: failure()
+        # if: failure()   -> always upload as build result documentation
```
ddeclerck, line 94: I believe you have to explicitly specify `always()`

.github/workflows/ubuntu.yml
GitMensch: no, it isn't, it is just what the autogen script suggests and what for example clangd expects as a build directory without configuration - I revert that part

ddeclerck:
As for the MSYS1 workflow : maybe I missed something ; why is it necessary to trigger it from the Ubuntu workflow ?

GitMensch:
If the workflow is separated then we can only get the CI tarball (MSYS1 is too old to build with the new dependencies in GC 3.3) by knowing the run-id of the job; we can either trigger it from Ubuntu to do so (but this misses the output in PRs.. and anywhere outside the "Action" tab) or do some more complicate API calling with timeout in the MSYS1.

... or we just combine them into one, and save all the struggle...

.github/workflows/ubuntu.yml
GitMensch: nope, it is just step in the workflow as all the others, we only don't need to make that conditional

ddeclerck: My memory is vague about it, but I had the impression that if a step fails (and `continue-on-error` is unset or set to `false`), it skips all the subsequent steps ?

GitMensch: Your memory is right, I've adjusted it accordingly.

ddeclerck:
> ... or we just combine them into one, and save all the struggle...

That might be a good idea. But considering you already went through the struggle, might as well keep it as you made it.

GitMensch:
I think I'll give that a shot soonTM, maybe its as easy as I've thought :-)

GitMensch:
It would be really good to know which of those tests hang under MSYS2.

While CI runs take longer that way it could be reasonable doing an explicit `make test` first (multi-job), then a non-parallel run of the internal testsuite - and if there are still hanging results check if we see "regular" issues at the same set of tests ?

ddeclerck:
> It would be really good to know which of those tests hang under MSYS2.

I tried to push the CI a bit ; also checked older logs, looking for missing test results. Looks like this is not tied to any specific test (either this, or most of the tests are affected ?), because each time it was a different one. Here are a few of them (numbers not relevant):
```
58: COPY: recursive
198: ASSIGN to device-name
343: USE FOR DEBUGGING syntax-checks (1) 
514: LISTING directive free-form reference-format
691: 691: LOCAL-STORAGE (2)
950: EXTFH: Indexed with FH--FCD
957: EXTFH: INDEXED with multiple keys
1010: RETURN-CODE moving
```
Most of the time this is accompanied with a WMFO error in the logs, which hints at a locking issue.

As you suggest, we might have resort to runing the testsuite sequentially here (takes ~15-20 min, while in parallel it is more like ~10-12 min). I guess that's acceptable...

GitMensch:
Thanks for checking. I'll leave the single-job test adjustment for MSYS2 to you (at least for now I'd suggest to reorder and have NIST in multi-job run first).

I'll try to combine the other two into one in the meantime, if it works as good as I think, then the result will be better than what we had before.

ddeclerck:
> MSYS1 is too old to build with the new dependencies in GC 3.3

Was curious about it, so I just tried installing autoconf 2.70 from source in MinGW, and it happens to build and pass the testsuite (cf [PR 188](https://github.com/OCamlPro/gnucobol/pull/188/files#diff-254d9246ee9a655a8ec2d9bfb3446158fb21aecae42287d720bb0bdd54a30dc6)). Don't know if that could be useful.

GitMensch:
for some reason the push closed this... so we're going on in #189 now (note: the MSYS1 autoconf update is in as a comment [and split per its "new way] in that PR, but shouldn't be needed).

PR #188, opened 2024-10-04
Fix MSYS CI

ddeclerck:


GitMensch:
included in #189 (so we may should close this, depending on how long that other one takes to merge)

PR #189, opened 2024-10-05
ci adjustments

GitMensch:
ci adjustments

* ubuntu:
  * only run jobs for "coverage" and "additional warnings" if the main ci build works
    and use its generated tarball in both cases
  * add two new artifacts: test binaries and windows source
* ubuntu+msys1+msys2+macos:
  * upload config.log after the build - because we may need it to debug build issues
  * always upload the testsuite.log (additional build documentation)
* ubuntu+msys1+msys2:
  * use --with-pkgversion to mark CI binaries
* msys2+macos:
  * uploading NIST test results
* msys1:
  * GMP url changes, building it again for performance reasons
  * building BDB with all relevant patches from MSYS2
  * drop GC install log step and therefore extra prefix
  * using msys-build instead of building Bison (only necessary for GC4)
  * drop extra CFLAGS previously necessary for local cJSON (fixed in 3.x)
  * enable NIST85 (+ comment-code in case we ever need to skip something there
    and/or ignoring failing NIST) --> as after last upstream update everything works
  * ci cache adjustment:
    * remove split per matrix
    * split per software, enabling smaller updates
  * use CI tarball like for the minimal build, drop flex+bison
  * drop workflow specific expected failures that now work fine
  * move env to MSYS job
  * resolve env vars by build API instead of runner
* integrate msys1.yml into ubuntu.yml, renaming to build_nightly.yml
* msys2:
  * split NIST + internal testsuite and run the later with less jobs to prevent hanging
  * split between "prepare" and "build" job, with the former generating a full distribution
    inclusive documentation, and the later having less packages installed,
    allowing to enable i386 again
  * disable BDB for 32bit build as MSYS2 doesn't provide that any more
  * switch cjson to json-c (more commonly used, MSYS2 still providing 32bit build)
  * explicit list of dependencies for configure
* msvc:
  * testsuite skip adjustments from the generated testsuite
  * enable building binary package and artifact
  * export dependencies and artifact


GitMensch:
@ddeclerck Do you see any reason to _not_ pull this?

I'd merge the changes to 4.x as well (need minor modifications) if that's fine with you.

.github/workflows/windows-msvc.yml
```diff
@@ -78,24 +76,24 @@ jobs:
             ${{ env.VCPKG_ROOT }}/packages
 
       - name: Bootstrap VCPKG
-        if: steps.restore-vcpkg.outputs.cache-hit != 'true'
```
ddeclerck, line 81: Do we want to bootstrap VPKG even if it was found in cache ?

GitMensch: I _think_ that's needed to get the actual updates - several dependencies were out of date (and with the cache available that's super fast).

ddeclerck:
> @ddeclerck Do you see any reason to _not_ pull this?

I guess it's good to merge.

> I'd merge the changes to 4.x as well (need minor modifications) if that's fine with you.

Fine with me ;)

.github/workflows/windows-msvc.yml
ddeclerck: Alright.

GitMensch:
Just FYI: we may be able to let the msys2-mingw32 build include BDB again (it is currently deactivated msys2 upstream) - follow https://github.com/msys2/MINGW-packages/issues/22133 for the details

PR #190, opened 2024-10-07
GC4 ci adjustments

GitMensch:
merging the changes of #189 and make gc4 ci more similar to gc3


ddeclerck:
> @ddeclerck This is nearly done but the MSVC is broken and I have no clue why (also when comparing with the GC3 one). Using cobc to generate something (already for building gcdiff, but otherwise also later in the testsuite) always result in "CL no source file" error message (both with and without setting/unsetting CL).

Welcome to the seventh circle of MSVC hell ;)
Looking at the `cl.exe` command called by `cobc`:
```
executing:	cl.exe /c
		/I"D:\a\gnucobol\gnucobol\build_windows\x64\Release\"
		/I"D:\a\gnucobol\gnucobol\build_windows\x64\Release\include"
		/I"D:\a\gnucobol\gnucobol"
		/I"D:\a\gnucobol\gnucobol\build_windows" /MD
		/Fo"C:\msys64\tmp\cob4720_0.obj" "gcdiff.c"
```
I'd say (~90% sure) this is due to the trailing slash at the end of the first include path (please don't ask..).
Yeah, this sh*t is extremely sensitive to even the slightest inaccurracy.

> 
> If I take the resulting binaries from the artifacts I get something that seems to be a working compiler.
> 
> Any idea? If not I'd suggest to wrap this up - skipping the testsuites (as it was before, just commented out) and package everything "as is" (without gcdiff)... at least we get all other CI environments working (much extended than before).
> 
> Note: the part that took me the most time was the BSD `sed` - which is a bit different than other implementations (I haven't checked, we possibly can use the BSD format in all build definition scripts). Oh, and I've learned that after `-i` you "normally" add an extension which is used for the backup file (if any).

ddeclerck:
(sorry accidental closure 😅)

GitMensch:
> I'd say (~90% sure) this is due to the trailing backslash at the end of the first include path (please don't ask..).
> Yeah, this sh*t is extremely sensitive to even the slightest inaccurracy.

I've rechecked - yes, that trailing slash kicks cl.exe into Nirvana. You know, I'd create a PR to fix that if it would be _at least_ open source... checking what cobc does - it removes any trailing slash (since 2020 - r4051 only for MSVC and since [r4974](https://sourceforge.net/p/gnucobol/code/4974/) in general) via the `remove_trailing_slash()` function.

I'll adjust the CI definition...

PR #191, opened 2024-10-09
Add optional index check

ddeclerck:
This PR attemps to backport SVN commit [4953](https://sourceforge.net/p/gnucobol/code/4953/) to 3.x.

### DIFF
--- a/cobc/flag.def
+++ b/cobc/flag.def
@@ -186,6 +186,9 @@ CB_FLAG (cb_flag_stack_check, 1, "stack-check",
 	_("  -fstack-check         PERFORM stack checking\n"
 	  "                        * turned on by --debug/-g"))
 
+CB_FLAG (cb_flag_check_subscript_set, 1, "opt-check-subscript-set",
+	_("  -fopt-check-subscript-set  check subscript in PERFORM/SET"))
+
 CB_FLAG_OP (1, "memory-check", CB_FLAG_GETOPT_MEMORY_CHECK,
 	_("  -fmemory-check=<scope>     checks for invalid writes to internal storage,\n"
 	  "                        <scope> may be one of: all, pointer, using, none\n"
--- a/cobc/typeck.c
+++ b/cobc/typeck.c
@@ -2710,6 +2710,13 @@ cb_build_identifier (cb_tree x, const int subchk)
 
 				/* Run-time check for all non-literals */
 				if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT)) {
+					if (cb_flag_check_subscript_set
+					 && CB_REF_OR_FIELD_P (sub)) {
+						/* Skip check_subscript; Now done on SET/PERFORM */
+						if (CB_FIELD_PTR (sub)->flag_indexed_by) {
+							continue;
+						}
+					}
 					if (cb_subscript_check != CB_SUB_CHECK_MAX
 					 && p->depending && p->depending != cb_error_node) {
 						cb_tree		e1;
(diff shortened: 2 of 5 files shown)


cobc/typeck.c
```diff
@@ -2710,6 +2710,13 @@ cb_build_identifier (cb_tree x, const int subchk)
...
 				/* Run-time check for all non-literals */
 				if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT)) {
+					if (cb_flag_check_subscript_set
+					 && CB_REF_OR_FIELD_P (sub)) {
+						/* Skip check_subscript; Now done on SET/PERFORM */
+						if (CB_FIELD_PTR (sub)->flag_indexed_by) {
+							continue;
+						}
+					}
 					if (cb_subscript_check != CB_SUB_CHECK_MAX
```
ddeclerck, line 2720: That's where I'm wondering if the added lines above interact badly with this.

tests/testsuite.src/run_misc.at
```diff
@@ -707,7 +707,9 @@ AT_DATA([prog.cob], [
            STOP RUN.
 ])
 
-AT_CHECK([$COMPILE prog.cob], [0], [], [])
+AT_CHECK([$COMPILE -fopt-check-subscript-set prog.cob], [0], [],
```
ddeclerck, line 710: Had to explicitly activate the flag here - but is it what we want ?

GitMensch: Does the result differ otherwise?

As we found that this is a non-standard optimization it should only be enabled by explicit request.

ddeclerck: Actually 4935 added the warning to this test - but to get the warning we have to add the new `-fopt-check-subscript-set` flag.

Maybe we can test both behaviors: without the flag and the warning (as it was before this patch), and with the flag and the warning ?

GitMensch (changes requested):
This is a new cobc option and should have an entry in the NEWS.

Your guess about the broken combination is likely correct and you'd need to adjust the new code to also include the new dialect option added in the meantime - but the (currently missing) extended test case will show if this is the case.

Note: please reword the help in flag.def, maybe something like "for index-names, don't apply EC-BOUND-SUBSCRIPT on subscript use per ISO rules but but on adjusting the index-name via SET or PERFORM statements, to reduce the amount of checks done".
I also wonder if we should place "index" into that option somewhere...

Please include into cobc/Changelog both Ron's original entry (with a remark "merged on YYYY-MM-DD as new option) and your on-top changes (the option and possibly inclusion into the dialect code) with that date.

tests/testsuite.src/run_misc.at
GitMensch: No, one of the point of the back-merge was that the checks and therefore the warnings should always be done, only the codegen depend on the new flag.

... but the warnings belong into a test in syn_occurs.at, not into a runtime check (please move it). And yes, the other runtime test should be done for both cases - w/wo the new option (without is missing and the compile command for the current test uses definitely more options than needed (`$COMPILE -fnew-option prog.cob` should be enough, if not it is reasonable to do minor necessary adjustments to the testcase).

ddeclerck:
Actually I'm not very confident I could do that quickly enough. I'd rather delegate this so I could focus on the GC3/GC4 merge - the clock's ticking and the days allocated to this task are almost depleted.

GitMensch:
I was surprised you took on that backport. The issue that I've now realized is - if you want to merge the dialect option from GC3 to GC4 then this has to be done in any case :-/

ddeclerck:
> I was surprised you took on that backport. The issue that I've now realized is - if you want to merge the dialect option from GC3 to GC4 then this has to be done in any case :-/

True indeed ; that dialect option is commit 5087 (so this will be a problem in ~50 commits). Though I'm just realizing now that feature was implemented by Nicolas ; I'm gonna ask him for help.

cobc/typeck.c
ddeclerck: Actually this was removed from trunk in the subsequent commit (4954).

GitMensch:
MSVC CI should be fixed with updating from upstream (I've worked on that before my vacation).

Apart from that: What is the state of this PR?

ddeclerck:
> Apart from that: What is the state of this PR?


I'd stay "stale" - waiting for someone else to take over.
In the meantime I made PR #194 to (temporarily) disable that check in GC4 so that I could continue the merge (as discussed in https://github.com/OCamlPro/gnucobol/pull/147#issuecomment-2405586694).

GitMensch (changes requested):
... oops, forget to send "submit"... so here is my (6 weeks old?) review, requesting changes.

GitMensch (changes requested):
First: Sorry that the review took so long. I did not submit it before; but that second view now actually brought some more "content" in, so that's possibly not bad.

The non-standard check as possible performance boost while keeping checks enabled are not an actual part of this PR (so far).

Depending on how you want to go on, we could reduce this PR to syntax improvement (parser) and the additional useful warning - and drop the change top flag.def.
Then maybe create a follow-up PR which has the additional move to the non-standard RANGE check (or postpone this for later).

... or: work on this to get this "all in"; but that will definitely be more work for nowö.

GitMensch:
Just stumbled over this, maybe the parser change already handles the issues mentioned in https://sourceforge.net/p/gnucobol/bugs/228/ (or can be extended to that)?

cobc/typeck.c
```diff
@@ -21,6 +21,7 @@
 
 
 #include "config.h"
+#include "libcob/common.h"
```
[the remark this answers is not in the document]
emilienlemaire: I have to say I don't remember adding this include :thinking:

libcob/common.h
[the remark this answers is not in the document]
emilienlemaire: This was a side-effect of the include in typeck.c, it is fixed now

cobc/typeck.c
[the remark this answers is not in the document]
emilienlemaire: Maybe I should check if my editor has an automatic include functionnality

cobc/flag.def
```diff
@@ -186,6 +186,9 @@ CB_FLAG (cb_flag_stack_check, 1, "stack-check",
 	_("  -fstack-check         PERFORM stack checking\n"
 	  "                        * turned on by --debug/-g"))
 
+CB_FLAG (cb_flag_check_subscript_set, 1, "opt-check-subscript-set",
+	_("  -fopt-check-subscript-set  check subscript in PERFORM/SET"))
```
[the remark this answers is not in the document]
nberth: The change in `codegen.c` impacts `PERFORM UNTIL`; I assume that's why `PERFORM` is mentioned here. however, that specific part in code-generation might be about internal `SET`s induced by `PERFORM` statements. If so, then indeed there's not real point in mentioning `PERFORM` here.

cobc/typeck.c
```diff
@@ -13742,10 +13743,53 @@ cb_check_set_to (cb_tree vars, cb_tree x, const int emit_error)
...
+			if (hasval) {
+				if (setval > p->occurs_max
+				 || setval < p->occurs_min) {
+					cb_warning_x (COBC_WARN_FILLER, l,
+						      _("SET %s TO %d is out of bounds"),
+						      f->name, setval);
+					if (emit_exception) {
+						cb_emit (CB_BUILD_FUNCALL_1 ("cob_set_exception",
+									     cb_int (COB_EC_RANGE_INDEX)));
+					}
```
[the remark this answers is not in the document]
nberth: I'm not sure I fully understand this comment… not to mention what change would be needed in the code ;-)

(Apologies for the request for clarification: I'm just discovering this PR).

PR #192, opened 2024-10-09
Allow spaces in COB_CC

ddeclerck:
This (small) PR allows the COB_CC variable to contain spaces, by adding quotes to any `sprintf` that use it.
Otherwise, one would need to escape all the spaces in COB_CC, and while this is acceptable "in general", this is slightly inconvenient, and besides the other variables do not work this way (for instance try escaping spaces in COB_CONFIG_DIR and you'll see it fails finding the config files - if the path contains spaces).

This of course works as long as COB_CC only contains the path to the C compiler and no argument - which should be the case since arguments should go in COB_CFLAGS & friends.

This would also allow us to get rid of the `gcobc` wrapper we have in our Windows installer (cf. your comment: ["I'd like to know what "gcobc" is"](https://github.com/OCamlPro/superbol-artefacts/issues/1)), whose sole purpose is to call `cobc` with a space-escaped COB_CC.

EDIT: probably needs more work, considering the failures on Windows (had only tested on Linux so far)

### DIFF
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -3169,12 +3169,12 @@ process_command_line (const int argc, char **argv)
 				fflush (stdout);
 				process ("link.exe");
 #else
-				cobc_buffer_size = strlen (cobc_cc) + 11;
+				cobc_buffer_size = strlen (cobc_cc) + 13;
 				cobc_buffer = cobc_malloc (cobc_buffer_size);
-				snprintf (cobc_buffer, cobc_buffer_size, "%s --help", cobc_cc);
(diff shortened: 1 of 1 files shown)


GitMensch:
As you also see in the failing tests - that doesn't work this way as we copy over verbatim what is in the CC variable - and things like -std and -march are commonly set that way.

Note: the more common way is to not use full paths for the binaries but instead prefix its path to PATH.

GitMensch:
Should we close this (and you prefix PATH with the directory in gcobc)?

ddeclerck:
> Should we close this (and you prefix PATH with the directory in gcobc)?

Yup, let me close this.

PR #194, opened 2024-10-15
"Fix" PERFORM bounds check in 4.x - but disable it for now as it is not ISO-compliant

ddeclerck:
As per https://github.com/OCamlPro/gnucobol/pull/147#issuecomment-2405586694.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -8104,10 +8104,38 @@ static void
 output_perform_until (struct cb_perform *p, cb_tree l)
 {
 	struct cb_perform_varying	*v;
-	struct cb_field			*f;
 	cb_tree				next;
 
 	if (l == NULL) {
+#if 0 /* FIXME: add back as option, because not conforming to ISO */
+		if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT)) {
+			cb_tree	xn;
+			/* Check all INDEXED BY variables used in VARYING */
+			for (xn = p->varying; xn; xn = CB_CHAIN (xn)) {
+				v = CB_PERFORM_VARYING (CB_VALUE (xn));
+				if (v->name
+				 && CB_REF_OR_FIELD_P (v->name)) {
+					struct cb_field	*f = CB_FIELD_PTR (v->name);
+					if (f->flag_indexed_by
+					 && f->index_qual) {
+						f = f->index_qual;
+						output_prefix ();
+						output ("cob_check_subscript (");
+						output_integer (v->name);
+						output (", ");
+						if (f->depending) {
+							output_integer (f->depending);
+							output (", \"%s\", 1", f->name);
+						} else {
+							output ("%d, \"%s\", 0", f->occurs_max, f->name);
+						}
+						output (");");
+						output_newline ();
+					}
+				}
+			}
+		}
+#endif
 		/* Perform body at the end */
 		output_perform_once (p);
 		return;
@@ -8125,7 +8153,7 @@ output_perform_until (struct cb_perform *p, cb_tree l)
 			     CB_PERFORM_VARYING (CB_VALUE (next))->name);
 		/* DEBUG */
 		if (current_prog->flag_gen_debug) {
-			f = CB_FIELD (cb_ref (CB_PERFORM_VARYING (CB_VALUE (next))->name));
+			struct cb_field	*f = CB_FIELD (cb_ref (CB_PERFORM_VARYING (CB_VALUE (next))->name));
 			if (f->flag_field_debug) {
 				output_stmt (cb_build_debug (cb_debug_name,
 					     (const char *)f->name, NULL));
@@ -8164,30 +8192,6 @@ output_perform_until (struct cb_perform *p, cb_tree l)
 	output (")");
 	output_newline ();
 	output_line ("  break;");
-	if (CB_EXCEPTION_ENABLE (COB_EC_BOUND_SUBSCRIPT)
-	 && next) {
-		cb_tree	xn;
-		/* Check all INDEXED BY variables used in VARYING */
-		for (xn = l; xn; xn = CB_CHAIN (xn)) {
-			struct cb_field		*q;
-			f = CB_FIELD_PTR (CB_PERFORM_VARYING(CB_VALUE (xn))->name);
-			if (!f->flag_indexed_by) continue;
-			if (!f->index_qual) continue;
-			q = f->index_qual;
-			output_prefix ();
-			output ("cob_check_subscript (");
-			output_integer (CB_PERFORM_VARYING(CB_VALUE (xn))->name);
-			output (", ");
-			if (q->depending) {
-				output_integer (q->depending);
-				output (", \"%s\", 1",f->name);
-			} else {
-				output ("%d, \"%s\", 0",q->occurs_max,f->name);
-			}
-			output (");");
-			output_newline ();
-		}
-	}
 
 	if (p->test == CB_BEFORE) {
 		output_perform_until (p, next);
--- a/tests/testsuite.src/run_subscripts.at
+++ b/tests/testsuite.src/run_subscripts.at
@@ -542,6 +542,7 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [Hi, there!])
 
 AT_CLEANUP
 
+
 AT_SETUP([Check Subscripts])
 AT_KEYWORDS([SUBSCRIPT])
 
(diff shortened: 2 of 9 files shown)


tests/testsuite.src/syn_occurs.at
```diff
@@ -501,3 +501,107 @@ AT_CHECK([$COMPILE_ONLY -std=mf -fodoslide prog.cob ], [0], [], [])
...
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       01 G.
+         02 X           PIC X OCCURS 10 INDEXED I.
+       PROCEDURE        DIVISION.
+           SET I TO ZERO.
+           SET I TO 11.
+           STOP RUN.
```
GitMensch, line 518: please add a `SET I TO lvl-78-containing-12` and possibly a `01-constant-as-13` to have those paths tested as well

tests/testsuite.src/syn_occurs.at
```diff
@@ -501,3 +501,107 @@ AT_CHECK([$COMPILE_ONLY -std=mf -fodoslide prog.cob ], [0], [], [])
...
+               IF MYNAME (IB1) = "Fred"
+                  MOVE "Freddy" TO MYNAME (IB1)
+               END-IF
+           END-PERFORM.
+           STOP RUN.
+           END PROGRAM prog.
+])
+
+AT_CHECK([$COMPILE_ONLY -std=mf prog.cob], [0], [], [prog.cob:25: warning: SET TO should be an integer
+prog.cob:26: warning: source is non-numeric - substituting zero
```
GitMensch, line 601: those two warnings should really be unconditional errors (I suggest within the parser), I don't know of any implementation that allows this

tests/testsuite.src/syn_occurs.at
```diff
@@ -501,3 +501,107 @@ AT_CHECK([$COMPILE_ONLY -std=mf -fodoslide prog.cob ], [0], [], [])
...
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [0], [],
+[prog.cob:9: warning: SET I TO 0 is out of bounds
+prog.cob:10: warning: SET I TO 11 is out of bounds
+])
+
+AT_CLEANUP
+
+AT_SETUP([SET index])
+AT_KEYWORDS([SET])
```
GitMensch, line 529: please use a double empty line before the new test (as above) and comment out the unused keyword (possibly also in the test case above "ODO")

tests/testsuite.src/run_subscripts.at
```diff
@@ -642,7 +643,7 @@ AT_DATA([prog.cob], [
        END PROGRAM SUBN.
 ])
 
-AT_CHECK([$COMPILE -x -std=mf -debug -Wall -debug -O prog.cob ], [0], [], [prog.cob:25: warning: SET TO should be an integer
+AT_CHECK([$COMPILE -x -std=mf prog.cob], [0], [], [prog.cob:25: warning: SET TO should be an integer
```
GitMensch, line 646: we want to run this and don't care about the warnings here - disable by whatever option is best matching and drop `-x` and - if possible - also the `-std`

cobc/codegen.c
```diff
@@ -8108,6 +8108,33 @@ output_perform_until (struct cb_perform *p, cb_tree l)
...
+			cb_tree	xn;
+			/* Check all INDEXED BY variables used in VARYING */
+			for (xn = p->varying; xn; xn = CB_CHAIN (xn)) {
+				struct cb_field		*q;
+				v = CB_PERFORM_VARYING (CB_VALUE (xn));
+				if (!v->name) continue;
+				f = CB_FIELD_PTR (v->name);
+				if (!f->flag_indexed_by) continue;
+				if (!f->index_qual) continue;
+				q = f->index_qual;
```
GitMensch, line 8122: we don't need `q` here (just reset `f`);
_maybe_ localize `f`'s definition here and in the debugging code below
and _maybe maybe_ you find a nice way to use less/no `continue` here - but "maybe"s are optional

GitMensch (changes requested):
please change those two warnings to an error and adjust the testsuite - in general this disabling is exactly the right thing and also should ease later merges

tests/testsuite.src/syn_occurs.at
ddeclerck: > comment out the unused keyword (possibly also in the test case above "ODO")

I'm not sure to get this; could you be more specific ?

GitMensch: `AT_KEYWORDS([SET])` is unused because the word SET is already in the "list of words used in AT_SETUP", which are automatically included as keywords. If you don't have another keyword you would want to run this test with just comment the complete AT_KEYWORDS line

tests/testsuite.src/syn_occurs.at
```diff
@@ -501,3 +501,115 @@ AT_CHECK([$COMPILE_ONLY -std=mf -fodoslide prog.cob ], [0], [], [])
...
+       01  TBL2.
+           05  MYMRK  PIC X(3)
+               OCCURS 2 TO 5 DEPENDING ON MAXIDX
+               INDEXED BY IB2.
+       PROCEDURE DIVISION.
+           MOVE 5 TO MAXIDX
+           SET NIDX TO IB1.
+           DISPLAY "Initial value: " NIDX.
+           SET IB2 TO 0.2.
+           SET IB2 TO "fred".
```
GitMensch, line 563: this needs to be an error as well, then this roundup is complete

tests/testsuite.src/run_subscripts.at
```diff
@@ -562,15 +563,15 @@ AT_DATA([prog.cob], [
...
                OCCURS 2 TO 5 DEPENDING ON MAXIDX
                INDEXED BY IB2.
        PROCEDURE DIVISION.
            MOVE 5 TO MAXIDX
            SET NIDX TO IB1.
            DISPLAY "Initial value: " NIDX.
-           SET IB2 TO 0.2.
-           SET IB2 TO "fred".
+      *    SET IB2 TO 0.2.
+      *    SET IB2 TO "fred".
```
GitMensch, line 574: drop those please

GitMensch (changes requested):
Thanks for cleaning this up!

tests/testsuite.src/syn_occurs.at
ddeclerck: The message `warning: numeric value is expected` for  `IB2 TO "fred"` comes from `validate_move`, itself called by `cb_build_move`, called from `cb_emit_set_to`. Guess I should not change that to a warning in `validate_move`. Should I just detect this case earlier in `cb_emit_set_to` to output a proper error ?

GitMensch: yes, before, ideally in the parser already by using a different terminal which includes the check for numeric or integer constant

ddeclerck: I can't find any existing terminal that makes this check - should I create a new one ?

ddeclerck: Oh but actually, the `set_to` rule in `parser.y` seems to be used not only for indices (seems to be used at least for `ADDRESS OF` too). Seems the check must be in `typeck.c`.

GitMensch: But both `set_to`  and `set_up_down` should not use `x` but a numeric or pointer or integer value (which possibly needs a new rule `x_numeric` or similar), which would cover all of this, so that `cb_emit_set_to` has only to check for the pointer part, no?

ddeclerck: Maybe, though, trying to add this check, this raised some questions.

1 / In `syn_misc.at` test "invalid use of condition-name", there is a SET to a level-88 condition, and the following expected output:
```
prog.cob:33: error: condition-name not allowed here: 'val-i1'
```
Since we're now forbiding anything but integer/pointer at parsing, the reported error will be different ; is that okay ?

2 / In `syn_move.at` test "MOVE FIGURATIVE to NUMERIC", we have `SET MYFLD TO HIGH-VALUES` and `SET MYFLD TO SPACES`, with an expected error of "error: invalid SET statement" - again this will change the error message : is it okay ?

GitMensch: changing the error messages is fine
concerning 1: we can make that check as well "integer expected, not condition-name 'val-i1'"  or similar

GitMensch: ... but we may can drop the check "later on" that currently triggers these messages as those would be dead code then

GitMensch:
@ddeclerck I'm confused - what is the state of this PR?

ddeclerck:
> @ddeclerck I'm confused - what is the state of this PR?

This one has been merged already (to trunk).

PR #199, opened 2024-11-22
Fix bug 1008 : regression in move to numeric edited items

ddeclerck:
This patch by @chuck-haatvedt fixes bug [1008](https://sourceforge.net/p/gnucobol/bugs/1008).

### DIFF
--- a/libcob/move.c
+++ b/libcob/move.c
@@ -156,6 +156,19 @@ store_common_region (cob_field *f, const unsigned char *data,
 	const int	hf2 = fsize + lf2;
 	const int	gcf = cob_min_int (hf1, hf2);
 
+#if 0 /* TODO: globally add support for COB_EC_DATA_TRUNCATION */
+	const unsigned char *p = data;
+	const unsigned char *end = data + hf1 - gcf;
+	while (p < end) {
+		if ((COB_FIELD_IS_NUMERIC (f) && (*p != '0'))
+		 || (COB_FIELD_IS_ANY_ALNUM (f) && (*p != ' '))) {
+			cob_set_exception (COB_EC_DATA_TRUNCATION);
+			break;
+		}
+		++p;
+	}
+#endif
+
 	/* the target may have leading/trailing additional zeroes
 	   and in rare cases, we may be out of scale competely;
 	   we pre-set all positions as this saves a bunch of
@@ -1011,18 +1024,15 @@ optimized_move_display_to_edited (cob_field *f1, cob_field *f2)
 	int   suppress_zero = 1;
 	int   have_decimal_point = 0;
 	int   have_check_protect = 0;
-	int   n;
 	int   cntr_currency = 0;
-	int   cntr_sign_neg = 0;
-	int   cntr_sign_pos = 0;
+	int   cntr_sign = 0;
 	unsigned char pad = ' ';
 	unsigned char c;
 	unsigned char float_char = 0x00;
-	const unsigned char dec_symbol = COB_MODULE_PTR->decimal_point == ','
-																 ? ',' : '.';
+	const unsigned char dec_symbol = COB_MODULE_PTR->decimal_point == ',' ? ',' : '.';
 	const unsigned char currency = COB_MODULE_PTR->currency_symbol;
 
-#if 1	/* Sanity check to ensure that the data types of both the fields have the 
+#if 1	/* Sanity check to ensure that the data types of both the fields have the
 		   correct attributes, if not then something is brokend and needs to be fixed  */
 	if (!(COB_FIELD_TYPE (f2) == COB_TYPE_NUMERIC_EDITED
 	 && COB_FIELD_DIGITS (f1) == COB_FIELD_DIGITS (f2)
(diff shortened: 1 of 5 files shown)


GitMensch:
Looks good to me. Can you check that in in behalf of chuck? I'll adjust the metadata on Monday. 

But please say in the changelog also a bit what this bug was about.

GitMensch:
See upstream bug report for updated tests and patch

ddeclerck:
> See upstream bug report for updated tests and patch

I see there have been a few more fixes. I'll try to have a look at this before monday.

ddeclerck:
@GitMensch I've included the updates from Chuck ; is it good for merge ?

ddeclerck:
Note: I've also just removed the extra error message:
```
cob_runtime_error ("optimized_move_display_to_edited: overflow in destination field");
```

This message was not present before, and truncation occured silently (this is another "regression" reported by our customer).

GitMensch:
Please add an example to the testsuite where the removed message previously was returned.

From the first look this should indeed not happen, should it? If it is reasonable, then the `break` should possibly be possibly be changed to a `goto endofouterloop`.

As you adjust this function in any case - please move `n = p->times_repeated;` into the `for` loop and, if possible, localize the `int n` into the loop.

ddeclerck:
> Please add an example to the testsuite where the removed message previously was returned.

Done.

> From the first look this should indeed not happen, should it?

This happens when moving to a shorter edited item.

> If it is reasonable, then the `break` should possibly be possibly be changed to a `goto endofouterloop`.

Done.
 
> As you adjust this function in any case - please move `n = p->times_repeated;` into the `for` loop and, if possible, localize the `int n` into the loop.

Also done ;)

libcob/move.c
```diff
@@ -1047,8 +1045,8 @@ optimized_move_display_to_edited (cob_field *f1, cob_field *f2)
 	c = 0x00;
 
 	for (p = COB_FIELD_PIC (f2); p->symbol; ++p) {
+		int n = p->times_repeated;
 		c = p->symbol;
```
GitMensch, line 1049: that's seldom used - we can "inline" `n = p->times_repeated` in the condition; as `cntr_sign_neg` and `cntr_sign_pos` are only used to `break` out (if there's floating -> at least two) and these are mutually exclusive, we should change that to a single `cntr_sign`, combining the condition for -+

tests/testsuite.src/run_fundamental.at
```diff
@@ -2510,6 +2510,102 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 W-SRC PIC 999V99 VALUE 123.45.
+       01 W-DST PIC ZZZV.
+
+       PROCEDURE DIVISION.
+           MOVE W-SRC TO W-DST.
```
GitMensch, line 2598: please add case of `W-SRC2 99999V9 VALUE 1234.5.` (truncation to the left)

Note: this _should_ raise

```c
/* A non-space or significant-digit was truncated during MOVE (COBOL2025) */
COB_EXCEPTION  (0308, COB_EC_DATA_TRUNCATION,
		"EC-DATA-TRUNCATION", 0)
```

which we don't have yet and also need to add for the other `MOVE` cases... this can be done with a follow-up, but should be kept in mind (at least add as a `#if 0 /* TODO: add in general` + `cob_set_exception (COB_EC_DATA_TRUNCATION);`. please - then again you may also just add it in exception.def...)

... and if/when you add that, please also add the definition for

```c
/* Attempt to use a null valued data item in a context where not permitted (COBOL2025) */
COB_EXCEPTION  (0307, COB_EC_DATA_NULL,
		"EC-DATA-NULL", 1)
```

GitMensch (changes requested):
Please add left truncation in the test - I guess this goes into the same optimized function, no?

ddeclerck:
> Please add left truncation in the test - I guess this goes into the same optimized function, no?

The optimized function deals only with same size/scale source and dest.
Added the test anyways.

GitMensch:
I'm confused now - if both have same amount of digits and scale (which is verified in a `#if 1` block - then how does the right-side truncation occurs?

GitMensch (approved):
Good to commit as soon as a new Changelog entry about

`move.c (optimized_move_display_to_edited): minor refactoring`
`system.def, move.c: added definition for COPBOL2025 COB_EC_DATA_NULL + COB_EC_DATA_TRUNCATION (currently not used)

is added - but I do still find it puzzling, that the "size must match" and we seemingly still get an overflow ?!?

GitMensch:
so... shouldn't the `#if 1` left part be triggered in the forgotten case?

In any case please adjust the test ensuring both parts are included there.

ddeclerck:
> so... shouldn't the `#if 1` left part be triggered in the forgotten case?
> 
> In any case please adjust the test ensuring both parts are included there.

Ah, I understand what's happening. I got confused.

The `optimized_move_display_to_edited` function is indeed meant to be called on fields with same size/scale, and this seems to always be the case.

The right-truncation error message was triggered because we hit a corner case: the target PICTURE was ending with V, but the check for "desination field is full" was made *before* processing this V. Since V does not add any character to the output field, it is in fact okay if the destination is full when processing V. I'll adjust the PR.

GitMensch:
Sounds like a good plan, as well as wrapping that check into `#ifndef NDEBUG` and adding that edge-case to the test (please do so for both the optimized and the "old" function, resulting into 4 test cases in total).

ddeclerck:
> and adding that edge-case to the test (please do so for both the optimized and the "old" function, resulting into 4 test cases in total).

In fact, since the optimized version is only ever called when source and destination size and scale match, the only relevant test is when we have a PICTURE ending with V.
When the size and scale do not match, the move is indirect ; it is made of a call to `cob_move_display_to_display` (that does the truncations), and a call to `optimized_move_display_to_edited`.

chuck-haatvedt:
David,

Are you doing a commit to the 3.x SVN repository for the bug 1008 fix ?

ddeclerck:
Hi @chuck-haatvedt,
Yes, that's the plan, as soon as @GitMensch confirms this PR is OK.

GitMensch (approved):
That's fine with me - you likely want to replace the `#if 1` in line 1025/1035 with the same `#ifndef NDEBUG` - but that is enough on commit upstream.

PR #200, opened 2024-12-02
Dialect option: normalize BCD "on-the-fly" when moving from ALPHANUMERIC to NUMERIC

ddeclerck:
This PR adds a new dialect option to allow on-the-fly "normalization" of BCD when moving from ALPHANUMERIC to NUMERIC. This is needed to properly mimic the GCOS COBOL behavior, for instance moving an ALPHANUMERIC containing "ABC456789}" to a NUMERIC yields "-1234567890" in this implementation (it tries to interpret the source as a PIC S9). See the new test at the end of `run_misc.at`.

EDIT (27/01/2026): we noticed that on GCOS moves from DISPLAY NUMERIC to NUMERIC are subject to this normalization too, so the PR has been updated

### DIFF
--- a/cobc/config.def
+++ b/cobc/config.def
@@ -203,6 +203,9 @@ CB_CONFIG_BOOLEAN (cb_areacheck, "areacheck",
 	  "                         * statements must not start in Area A; and\n"
 	  "                         * separator periods must not be within Area A"))
 
+CB_CONFIG_BOOLEAN (cb_normalize_bcd, "normalize-bcd",
+	_("normalize BCD on-the-fly"))
+
 /* Support flags */
 
 CB_CONFIG_SUPPORT (cb_comment_paragraphs, "comment-paragraphs",
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -3966,6 +3966,25 @@ cob_real_get_sign (cob_field *f, const int adjust_ebcdic)
 	return 0;
 }
 
+/* get the "sign" from an alphanumeric field, as if the field
+   was numeric display with non-separate trailing sign */
+int
+cob_get_sign_from_alnum (cob_field *f)
+{
+	int sign;
+	cob_field_attr attr;
+	cob_field field;
+	COB_FIELD_INIT (COB_FIELD_SIZE (f), COB_FIELD_DATA (f), &attr);
+	COB_ATTR_INIT (COB_TYPE_NUMERIC_DISPLAY, COB_FIELD_SIZE (f), 0, COB_FLAG_HAVE_SIGN, NULL);
+	sign = cob_real_get_sign (&field, 0);
+	if (sign < 0) {
+		return -1;
+	} else if (sign > 0) {
+		return 1;
+	}
+	return 0;
+}
+
 /* store sign to DISPLAY/PACKED fields */
 void
 cob_real_put_sign (cob_field *f, const int sign)
--- a/libcob/move.c
+++ b/libcob/move.c
@@ -189,14 +189,16 @@ store_common_region (cob_field *f, const unsigned char *data,
 
 			while (dst < end) {
 				const char src_data = *src++;
-#if 0		/* seems to be the best result, ..." */
-				/* we don't want to set bad data, so
-				   only take the half byte */
-				*dst = COB_I2D (COB_D2I (src_data));
-#else		/* but does not match the "expected" MF result, which is: */
-				if (src_data == ' ' || src_data == 0) /* already set: *dst = '0'; */ ;
-				else *dst = COB_I2D (src_data - '0');
-#endif
+				if (COB_MODULE_PTR->flag_normalize_bcd) {
+					/* seems to be the best result, ..." */
+					/* we don't want to set bad data, so
+					   only take the half byte */
+					*dst = COB_I2D (COB_D2I (src_data));
+				} else {
+					/* but does not match the "expected" MF result, which is: */
+					if (src_data == ' ' || src_data == 0) /* already set: *dst = '0'; */ ;
+					else *dst = COB_I2D (src_data - '0');
+				}
 				++dst;
 			}
 		}
@@ -309,6 +311,7 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
 	const unsigned char	*e2 = s2 + COB_FIELD_SIZE (f2);
 	const unsigned char	dec_pt = COB_MODULE_PTR->decimal_point;
 	const unsigned char	num_sep = COB_MODULE_PTR->numeric_separator;
+	unsigned char	last;
 	int		sign;
 	int		count;
 	int		size;
@@ -325,21 +328,35 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
 
 	/* Check for sign */
 	sign = 0;
-	if (s1 != e1) {
-		if (*s1 == '+' || *s1 == '-') {
-			sign = (*s1++ == '+') ? 1 : -1;
+	if (!COB_MODULE_PTR->flag_normalize_bcd) {
+		if (s1 != e1) {
+			if (*s1 == '+' || *s1 == '-') {
+				sign = (*s1++ == '+') ? 1 : -1;
+			}
 		}
+	} else {
+		last = f1->data[f1->size - 1];
+		sign = cob_get_sign_from_alnum (f1);
 	}
 
 	/* Count the number of digits before decimal point */
 	count = 0;
 	{
 		register unsigned char	*p;
-		for (p = s1; p < e1 && *p != dec_pt; ++p) {
+		if (!COB_MODULE_PTR->flag_normalize_bcd) {
+			for (p = s1; p < e1 && *p != dec_pt; ++p) {
 			/* note: as isdigit is locale-aware (slower and not what we want),
 			   we use a range check instead */
-			if (*p >= '0' && *p <= '9') {
-				++count;
+				if (*p >= '0' && *p <= '9') {
+					++count;
+				}
+			}
+		} else {
+			for (p = s1; p < e1 && *p != dec_pt; ++p) {
+				const char d = COB_D2I (*p);
+				if (d >= 0 && d <= 9) {
+					++count;
+				}
 			}
 		}
 	}
@@ -349,34 +366,66 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
 	if (count < size) {
 		s2 += size - count;
 	} else {
-		while (count-- > size) {
-			while (*s1 < '0' || *s1 > '9') {
+		if (!COB_MODULE_PTR->flag_normalize_bcd) {
+			while (count-- > size) {
+				while (*s1 < '0' || *s1 > '9') {
+					s1++;
+				}
 				s1++;
 			}
-			s1++;
+		} else {
+			while (count-- > size) {
+				char d;
+				do {
+					d = COB_D2I (*s1++);
+				} while (d < 0 || d > 9);
+			}
 		}
 	}
 
 	/* Move */
 	count = 0;
-	for (; s1 < e1 && s2 < e2; ++s1) {
-		if (*s1 >= '0' && *s1 <= '9') {
-			*s2++ = *s1;
-		} else if (*s1 == dec_pt) {
-			if (count++ > 0) {
+	if (!COB_MODULE_PTR->flag_normalize_bcd) {
+		for (; s1 < e1 && s2 < e2; ++s1) {
+			if (*s1 >= '0' && *s1 <= '9') {
+				*s2++ = *s1;
+			} else if (*s1 == dec_pt) {
+				if (count++ > 0) {
+					goto error;
+				}
+			} else if (!(isspace (*s1) || *s1 == num_sep)) {
 				goto error;
 			}
-		} else if (!(isspace (*s1) || *s1 == num_sep)) {
-			goto error;
+		}
+	} else {
+		for (; s1 < e1 && s2 < e2; ++s1) {
+			const char d = COB_D2I (*s1);
+			if (d >= 0 && d <= 9) {
+				*s2++ = COB_I2D (d);
+			} else if (*s1 == dec_pt) {
+				if (count++ > 0) {
+					goto error;
+				}
+			} else if (!(isspace (*s1) || *s1 == num_sep)) {
+				goto error;
+			}
 		}
 	}
 
 	COB_PUT_SIGN (f2, sign);
+	if (COB_MODULE_PTR->flag_normalize_bcd
+	 && !COB_FIELD_CONSTANT (f1)) {
+		f1->data[f1->size - 1] = last;
+	}
 	return;
 
 error:
 	memset (f2->data, '0', f2->size);
 	COB_PUT_SIGN (f2, 0);
+	if (COB_MODULE_PTR->flag_normalize_bcd
+	 && !COB_FIELD_CONSTANT (f1)) {
+		f1->data[f1->size - 1] = last;
+	}
 }
 
 static void
--- a/tests/testsuite.src/run_misc.at
+++ b/tests/testsuite.src/run_misc.at
@@ -16107,3 +16107,164 @@ callee.cob:8: attempt to reference invalid memory address (signal SIGSEGV)
 ])
 
 AT_CLEANUP
+
+
+AT_SETUP([On-the-fly BCD normalization on MOVE from ALPHANUMERIC])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 SRC-ALNUM     PIC X(10)  VALUE "ABC456789}".
+       01 SRC-NUM REDEFINES
+            SRC-ALNUM   PIC S9(10).
+       01 DST-NUMDISP   PIC S9(10).
+       01 DST-NUMEDIT   PIC +9B999B999B999.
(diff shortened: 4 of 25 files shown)


ddeclerck:
@GitMensch 

This seems to work. Should probably be improved.

MSYS2 CI failure is caused by different output format for floats (eg. where we have '-1.2345679E+9' in most environments, in MSYS2 we have '-1.2345679E+09') - any idea where this could be adjusted ?

ddeclerck:
I see. So it seems BCD is also "normalized" in MF and IBM, except that the sign is mostly ignored (except in MF when mving to NUMERIC-EDITED). Also I don't get why the results for binary in MF have exactly the opposite sign compared to GCOS... And then there are also differences in the way the fields are displayed - but I think this is not related to normalization.

I'm thinking maybe we should have two options, one to normalize the digits, and one for the sign (but then it would have to be more than a YES/NO flag, to account for MF keeping the sign only when moving to NUMEDIC-EDITED).

GitMensch:
If I don't miss something, the most visible thing is that this is NOT about BCD (COMP-3/COMP-6) normalization, but about normalization of `USAGE DISPLAY`, because all environments but ACU show the same data -> normalized and only binary truncated for the binary fields.

I'll do the "review" after some meeting, to comment on the actual changes.

(just FYI - added the ACU results which are different and took too long to get in the first place; I'd ignore that output for now)

ddeclerck:
> If I don't miss something, the most visible thing is that this is NOT about BCD (COMP-3/COMP-6) normalization, but about normalization of `USAGE DISPLAY`

Well, I'd still say it's BCD-related, since we're trying to interpret DISPLAY data as *unpacked* BCD.
But it's hard to come up with a proper description and self-explanatory flag names 😅.

BTW, it could be interesting to try with an ASCII sign ("p") instead of the EBCDIC sign ("}") in SRC-ALNUM.
Also, I'd be curious to know if those environment would consider a sign that is leading/trailing separate (on GCOS, it seems this DISPLAY data that we're moving to numeric is always interpreted as having a non-separate trailing sign).

GitMensch:
Note: I've added the ASCII p variant above as well.

Concerning the PR: that's effectively a general issue. I've thought that I have broken something there, but this code goes back to at least OC 1.1, my changes only produces the same "intended" result with less instructions.
The general issues seen with inspecting the results from other COBOL vendors and the "old" existing implementation that uses `cob_move_alphanum_to_display`:

* the sign is expected to be separate and in the first position; while with "redefined" numeric display data this is normally non-separate trailing
* as soon as the first error is seen, the target is set to all zero
* even with `-fec=all` this raises no exception (with MF it does by default, MF has an option to effectively make that fatal exception [abort] a non-fatal one [data is handled similar to this PR])

The question is how to go on. Using COB_D2I does work (is a bit more expensive, of course), so I tend to use it unconditionally in this function.
This mostly leaves the sign handling. My idea is to first use the current approach (skip all spaces, then check for a sign, if none found check at the last position of the field [effectively what this PR does, but not with an intermediate field]).

Another approach would be to try following COBOL 202x MOVE statement, general rule 7 d 4, which _may_ (I'm not sure) should only be done if the target is neither numeric-display nor numeric-edited (or its national variants if the source is national):

* define an intermediate **unsigned** field with no decimal places
* size+digits = orig_size (or if this is > 31 then 31)
* data=orig_data (or moved to the right to the last 31 positions, if there are more)
* then do a numeric move from that intermediate field to the target one

That _possibly_ yields nearly the same result as this PR (nearly because the data will likely always be handled as unsigned).
It would change the handling of leading separate sign of our current implementation (to an error in ascii in general and an error in ebcdic with "+").

What do you think?

ddeclerck:
> The question is how to go on. Using COB_D2I does work (is a bit more expensive, of course), so I tend to use it unconditionally in this function. This mostly leaves the sign handling. My idea is to first use the current approach (skip all spaces, then check for a sign, if none found check at the last position of the field [effectively what this PR does, but not with an intermediate field]).

Would that imply doing this unconditionally instead of using a flag ?

> Another approach would be to try following COBOL 202x MOVE statement, general rule 7 d 4, which _may_ (I'm not sure) should only be done if the target is neither numeric-display nor numeric-edited (or its national variants if the source is national):
> 
> * define an intermediate **unsigned** field with no decimal places
> * size+digits = orig_size (or if this is > 31 then 31)
> * data=orig_data (or moved to the right to the last 31 positions, if there are more)
> * then do a numeric move from that intermediate field to the target one
> 
> That _possibly_ yields nearly the same result as this PR (nearly because the data will likely always be handled as unsigned). It would change the handling of leading separate sign of our current implementation (to an error in ascii in general and an error in ebcdic with "+").

Would this be configurabled through a dialect flag ? Our customer expects the sign to be preserved, as it is the case on GCOS.

tests/testsuite.src/run_misc.at
```diff
@@ -15131,3 +15131,94 @@ AT_CHECK([COB_PROF_ENABLE=1 COB_PROF_FILE=prof.csv $COBCRUN_DIRECT ./caller], [0
...
+       01 DST-COMPN     PIC 9(10)  COMP-N.
+       01 DST-FLTSHORT             FLOAT-SHORT.
+       01 DST-FLTLONG              FLOAT-LONG.
+       01 DST-INDEX                INDEX.
+
+       PROCEDURE DIVISION.
+       MAIN.
+
+           DISPLAY "SRC-ALNUM        : '" SRC-ALNUM "'".
+
```
GitMensch, line 15164: `IF SRC-ALNUM NOT NUMERIC DISPLAY "NOT NUMERIC".` should be taken (I don't question it does but please add it to the test to ensure it does).

GitMensch:
For both options there would be an option to make this depending on a dialect (or optimization) flag.

If you like the first and go down that route I think there's no option necessary, is it?
Note: in any case we need a NEWS entry (_especially_ for "no option" - I think that can go under "important bug fixes").

It may be good to have that test include a performance check option, as in https://github.com/OCamlPro/gnucobol/blob/675824c1329ed5f1ba98a1760b2775eda94deda7/tests/testsuite.src/run_misc.at#L5609-L5625

This allows to easily verify both correctness and speed if we change that later on.

ddeclerck:
> ## MF VC
> ```
> SRC-ALNUM        : 'ABC456789}'
>  -> DST-NUMDISP  : '123456789=-'
>  -> DST-NUMEDIT  : '-1 234 567 89='
>  -> DST-PACK     : '+123456789='
>  -> DST-BIN      : '+1234567801'
> ...
> SRC-ALNUM        : 'ABC456789p'
>  -> DST-NUMDISP  : '1234567890-'
>  -> DST-NUMEDIT  : '-1 234 567 890'
>  -> DST-PACK     : '+1234567890'
>  -> DST-BIN      : '+1234567890'
> ...
> ```

I do not have the same results with MF VC (ASCII):
```
SRC-ALNUM        : 'ABC456789}'
 -> DST-NUMDISP  : '1234567903-'
 -> DST-NUMEDIT  : '+A BC4 567 89}'
 -> DST-PACK     : '+1234567903'
 -> DST-BIN      : '+1234567903'
 ...
SRC-ALNUM        : 'ABC456789p'
 -> DST-NUMDISP  : '1234567890-'
 -> DST-NUMEDIT  : '+A BC4 567 89p'
 -> DST-PACK     : '+1234567890'
 -> DST-BIN      : '+1234567890'
 ...
```

The two differences I notice are:
- no normalization occur when moving to NUMERIC-EDITED
- "789}" becomes "7903" instead of "789=" (which made more sense to me because '}' = 0x7D and '=' = 0x3D)

Now, I barely know how MF works, so maybe I'm doing something wrong ? Or maybe it's something that behaves differently depending on the version...

GitMensch:
@ddeclerck Given 44c96d20a12e96e0802163fdb9a2d05bb41df578 - what is the current state of this PR (and is it still high-priority)? If it is high, please rebase after upstreaming #204.

ddeclerck:
> @ddeclerck Given [44c96d2](https://github.com/OCamlPro/gnucobol/commit/44c96d20a12e96e0802163fdb9a2d05bb41df578) - what is the current state of this PR

Well, that commit you mention (from PR #201) was about fixing stuff and adding a bit of normalization on moves to NUMERCI-EDITED. This PR is about normalizing when moving from ALPHANUMERIC to NUMERIC.

> (and is it still high-priority)? If it is high, please rebase after upstreaming #204.

It's still kind of important, although not as critical as the other issues reported in the past few days (and a new one that just arrived tonight 😭).

ddeclerck:
> Also note that [this change](https://github.com/OCamlPro/gnucobol/compare/a58b6f1290126785710950e8856eda2597b1e95c..9115c8cce85fac1adba0640f4b316c200b5e4d03), together with a minimal test case, should be upstreamed as well, potentially with a comment /* skip leading zeroes + */.

You'd want this in a separate PR ?

GitMensch:
> > Also note that [this change](https://github.com/OCamlPro/gnucobol/compare/a58b6f1290126785710950e8856eda2597b1e95c..9115c8cce85fac1adba0640f4b316c200b5e4d03), together with a minimal test case, should be upstreamed as well, potentially with a comment /* skip leading zeroes + */.
> 
> You'd want this in a separate PR ?

As you want, either  a separate one or directly upstream.

ddeclerck:
> > > Also note that [this change](https://github.com/OCamlPro/gnucobol/compare/a58b6f1290126785710950e8856eda2597b1e95c..9115c8cce85fac1adba0640f4b316c200b5e4d03), together with a minimal test case, should be upstreamed as well, potentially with a comment /* skip leading zeroes + */.
> > 
> > 
> > You'd want this in a separate PR ?
> 
> As you want, either a separate one or directly upstream.

I went through a PR just to ensure the CI passes on all platforms.
https://github.com/OCamlPro/gnucobol/pull/207

cobc/config.def
GitMensch, line 1: Wasn't our conclusion that we did not need any config option for that as all environments do a normalization and we consider that a fix to a very old issue going back to OpenCOBOL?

Note: for both the dialect option and the general change we need a NEWS entry because of the difference to before.

libcob/common.c
```diff
@@ -3881,6 +3881,25 @@ cob_real_get_sign (cob_field *f, const int adjust_ebcdic)
...
+   was numeric display with non-separate trailing sign */
+int
+cob_get_sign_from_alnum (cob_field *f)
+{
+	int sign;
+	cob_field_attr attr;
+	cob_field field;
+	COB_FIELD_INIT (COB_FIELD_SIZE (f), COB_FIELD_DATA (f), &attr);
+	COB_ATTR_INIT (COB_TYPE_NUMERIC_DISPLAY, COB_FIELD_SIZE (f), 0, COB_FLAG_HAVE_SIGN, NULL);
+	sign = cob_real_get_sign (&field, 0);
```
GitMensch, line 3894: Wouldn't it be more reasonable to duplicate the code of the called functions here? This way we don't need an intermediate field definition, just getting `const char *p last_data = f->data + f->size - 1;` and check `p` as in the function above?

libcob/move.c
```diff
@@ -349,34 +363,69 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
...
+		if (!COB_MODULE_PTR->flag_normalize_bcd) {
+			while (count-- > size) {
+				while (*s1 < '0' || *s1 > '9') {
+					s1++;
+				}
+				s1++;
+			}
+		} else {
+			while (count-- > size) {
+				while (COB_D2I (*s1) < 0 || COB_D2I (*s1) > 9) {
```
GitMensch, line 375: use `const char d = COB_D2I (*s1)` here, then check `d`, same above and below

libcob/move.c
```diff
@@ -309,6 +309,7 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
 	const unsigned char	*e2 = s2 + COB_FIELD_SIZE (f2);
 	const unsigned char	dec_pt = COB_MODULE_PTR->decimal_point;
 	const unsigned char	num_sep = COB_MODULE_PTR->numeric_separator;
+	unsigned char	last;
```
GitMensch, line 312: What do we need that for? The called function doesn't change the data, does it?
(numeric.c (cob_decimal_set_display) may change that, so that's a different thing)

tests/testsuite.src/run_misc.at
GitMensch, line 1: We want an ascii test here for normalization as well.

GitMensch (commented):
Friendly ping @ddeclerck for the current state (a rebase can be done as well ;-)

cobc/config.def
ddeclerck: I'm not sure, since different dialects do not always normalize the same way ; some consider the sign, some ignore it, some don't normalize on some targets (for instance when moving to NUMERIC-EDITED). Depends how close we want to be to the original environment. It's true that a non-normalized result is not really useful in practice, and we also have warnings and/or exceptions to detect those invalid data.

libcob/move.c
ddeclerck: Since this function (indirectly) calls `cob_real_get_sign`, which alters the byte holding the sign (always trailing when normalizing), we have to save/restore it (or maybe we could use `cob_put_sign`).

ddeclerck:
> Friendly ping @ddeclerck for the current state (a rebase can be done as well ;-)

Haven't looked at that in the past few days (focused on the GC3/GC4 merge ;) ), but that's really a PR I'd like to complete.

I added comments to some of yours.

There was also an unanswered questions in a former comment: https://github.com/OCamlPro/gnucobol/pull/200#issuecomment-2546243047

GitMensch (commented):
As noted before I suggest to drop the dialect configuration for now and _always_:
* check for leading separate sign (adjusting the data offset as we did before)
* iterate over the digits using `COB_D2I`
* check for trailing embedded sign (as done with GCOS)

This won't give us exact results for "real" invalid data, but should be good in most cases, no?

It _does_ change the result for "valid" (with embedded sign) to be as with other compilers, preserves the current behavior with "valid" (leading separate) and mostly changes that "invalid" (unexpected) data is only used as zero for real bad data (half-byte > 9, if not a sign).

The part that _seems_ the most tricky to me is how to handle "embedded" spaces; this part may or may not still need a dialect configuration - we'd possibly need to check what compilers do here and if it is anywhere but in GnuCOBOL handled as zero - in this case we could adjust the result (we'll need a NEWS entry in any case).

What do you think?

libcob/move.c
```diff
@@ -349,34 +364,70 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
...
+		}
+	} else {
+		for (; s1 < e1 && s2 < e2; ++s1) {
+			const char d = COB_D2I (*s1);
+			if (d >= 0 && d <= 9) {
+#ifndef	COB_EBCDIC_MACHINE
+				*s2++ = (d | 0x30);
+#else
+				*s2++ = (d | 0xF0);
+#endif
```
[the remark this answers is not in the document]
ddeclerck: Indeed, `COB_I2D` should be used.

libcob/move.c
```diff
@@ -325,21 +326,35 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
```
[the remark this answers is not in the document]
ddeclerck: The thing is, something like `'	-123'` (where the blank before 123 is a tab - code 0x05 in EBCDIC, and considering minus has code 0x60) normalizes as `50123` on GCOS (checked). The same with a space instead of the tab normalizes as `00123`. Depending on whether we use `isspace` or `COB_D2I`, and whether we consider the possibility of a leading sign, we'll get different results. I'm not sure what would be the best thing to do. Maybe we could have an option (not a dialect one) to specify which kind of sign to expect (leading separate, trailing embedded, none...) ?

libcob/move.c
```diff
@@ -325,21 +326,35 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
...
-			if (*p >= '0' && *p <= '9') {
-				++count;
+				if (*p >= '0' && *p <= '9') {
+					++count;
+				}
+			}
+		} else {
+			for (p = s1; p < e1 && *p != dec_pt; ++p) {
+				const char d = COB_D2I (*p);
+				if (d >= 0 && d <= 9) {
```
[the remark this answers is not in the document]
ddeclerck: Actually what happens on GCOS is very different. I had not taken into account what happens when there is a decimal point... While the current GnuCOBOL only moves digits that are before the decimal point, GCOS tries to move and normalize all digits. If it encounters a decimal point, it raises an exception because the decimal point - no matter if it is a comma (0x6B) or a dot (0x4B) - does not normalize to a valid digit...

In fact, GnuCOBOL tries to be smart - skips leading spaces, interprets the sign and the decimal point, while GCOS (and others) more or less boldly convert whatever is there. But how much do we want to keep the original GnuCOBOL behavior ? If we do want to keep it (to not break existing programs), we might as well just have two different normalization functions.

libcob/move.c
```diff
@@ -349,34 +364,70 @@ cob_move_alphanum_to_display (cob_field *f1, cob_field *f2)
...
+#ifndef	COB_EBCDIC_MACHINE
+				*s2++ = (d | 0x30);
+#else
+				*s2++ = (d | 0xF0);
+#endif
+			} else if (*s1 == dec_pt) {
+				if (count++ > 0) {
+					goto error;
+				}
+			} else if (!(isspace (*s1) || *s1 == num_sep)) {
```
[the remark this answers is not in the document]
ddeclerck: True indeed.

libcob/move.c
ddeclerck: Yeah, probably. I'll dive further into this.

ddeclerck:
Hi @GitMensch ,
We'd need to have this merged at some point.
I had replied to most of your comments ; any info to add ?

PR #201, opened 2024-12-09
Fix BLANK WHEN ZERO on signed NUMERIC-EDITED

ddeclerck:
On behalf of Chuck.


libcob/move.c
```diff
@@ -1084,14 +1084,26 @@ optimized_move_display_to_edited (cob_field *f1, cob_field *f2)
...
+				return;
+			}
+		} else {
+			for (; (src <= src_end) ; src++) {
+				if (*src != '0') break;
+			}
+			if (src > src_end) {
+				memset (dst, (int)' ', (size_t)f2->size);
+				return;
+			}
```
GitMensch, line 1106: Please change it to use something like `const unsigned char *check_end = !COB_FIELD_SIGN_SEPARATE (f1) ? src_end : src_end - 1;` instead to reduce the duplication.

Also: What about separate and leading sign?
Note: I _think_ there's no use in explicit casting a char to int, is it? The field->size is size_t already, so that should be dropped in any case.

tests/testsuite.src/run_fundamental.at
```diff
@@ -6977,6 +6977,80 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01  W-X  PIC -ZZZBZZ9,99 BLANK WHEN ZERO.
+       01  W-Y  PIC  ZZZBZZ9,99 BLANK WHEN ZERO.
+       01  W-Z  PIC +ZZZBZZ9,99 BLANK WHEN ZERO.
+       01  J    PIC S9(8)  COMP-5.
```
GitMensch, line 6991: Please replace that source J by four source fields:

```cobolfree
01 L  PIC S9(8) SIGN LEADING.
01 T  PIC S9(8) SIGN TRAILING.
01 LS PIC S9(8) SIGN SEPARATE LEADING
01 LT PIC S9(8) SIGN SEPARATE TRAILING
```

as this then covers all available cases.

GitMensch (changes requested):
Thanks for your work on this again.

Please recheck the leading sign issue by adjusting the test case.

tests/testsuite.src/run_fundamental.at
ddeclerck: Done.

libcob/move.c
ddeclerck: > Also: What about separate and leading sign? And... where are the overpunched positions checked?

There is an invariant here that `f1` either has no sign or has a trailing separate sign (as checked by the block between `#ifndef NDEBUG` block at the top of the function, and also mentionned by Chuck in his message).

> As we just had that in the other PR: What about invalid data - should we check with `COB_D2I() = 0` instead?

We could add that, though the other PR is about moving from an ALPHANUMERIC source, while `optimized_move_display_to_edited` only moves from a NUMERIC-DISPLAY source (either a direct one, or a temporary one).
 
> Note: I _think_ there's no use in explicit casting a char to int, is it?

Absolutely, those character literals should already be of type `int` anyways.

> The field->size is size_t already, so that should be dropped in any case.

Done.

ddeclerck: > Can you expand that to also the leading separate case (_should_ be simple, and we now even have a testcase ready :-) and to the overpunched sign (a bit more complex)? Of course we can postpone that to later (either both or one of those)...

Done for this part.

> Just checked with MF: After `MOVE SPACES TO L(1:3)` before the W-X move, by default we get the abort for non-numeric data in numeric displays, when I disable that at run-time, then " 00000" is blanked to space and " 1234" results in " 12,34" while an additional `MOVE "A" TO L (8:1)` results in
> 
> ```
>  ===>       0,01<===
>  ===>      12,31<===
> ```

Added this to the test case, though the result differs, i.e.:
```
           MOVE 0 TO L.
           MOVE SPACES TO L(1:3).
           MOVE L TO W-X.
           DISPLAY ' ===>' W-X '<==='.

           MOVE "A" TO L(8:1).
           DISPLAY ' ===>' W-X '<==='.

           MOVE 1234 TO L.
           MOVE SPACES TO L(1:3).
           MOVE L TO W-X.
           DISPLAY ' ===>' W-X '<==='.

           MOVE "A" TO L(8:1).
           DISPLAY ' ===>' W-X '<==='.
```

Yields:
```
 ===>           <===
 ===>           <===
 ===>     012,34<===
 ===>     012,34<===
```

Though this is not really related to the "BLANK WHEN ZERO not working anymore" issue.

tests/testsuite.src/run_fundamental.at
```diff
@@ -7203,6 +7221,11 @@ FROM SIGN TRAILING SEPARATE
  ===>+     12,34<===
  ===>-     12,34<===
  ===>-     12,34<===
+EXTRA FROM SIGN LEADING SEPARATE
+ ===>           <===
+ ===>           <===
```
GitMensch, line 7226: I _think_ that should be fine, showing "1" if before `switch (c) {` (around line 1123) you add something like `src_num = COB_D2I(*src)` and use this instead of *src for every part below that does not check for something that's not expected to be not numeric.

GitMensch (changes requested):
I do question if GCOS would output 012,34...

please see some notes and add a Changelog (and yes, that part is a separate one from Chucks fixing something else - still fine for a single commit as that fixes two errors in the same function that are verified by a single test.

... it seems reasonable to check the CI generated coverage as well, to ensure that at least all parts of this function that is now fixed is actually tested (or at least: executed).

ddeclerck:
I think I considered all requested changes.

tests/testsuite.src/run_fundamental.at
GitMensch: Was I wrong about this - as there still is no "1" for the "A"?

ddeclerck: Oh, I was missing two MOVEs in the last test ; now I get the "A"s.

I'd have to check on GCOS (can't do it now, on a business trip), but AFAIR, it does not perform "normalization" when moving to NUMERIC-EDITED (unlike MF, apparently).

libcob/move.c
```diff
@@ -1037,9 +1038,7 @@ optimized_move_display_to_edited (cob_field *f1, cob_field *f2)
 	if (!(COB_FIELD_TYPE (f2) == COB_TYPE_NUMERIC_EDITED
 	 && COB_FIELD_DIGITS (f1) == COB_FIELD_DIGITS (f2)
 	 && COB_FIELD_SCALE (f1) == COB_FIELD_SCALE (f2)
-	 && COB_FIELD_HAVE_SIGN (f1) == COB_FIELD_HAVE_SIGN (f2)
-	 && ((COB_FIELD_HAVE_SIGN (f1) && (!COB_FIELD_SIGN_LEADING (f1) && COB_FIELD_SIGN_SEPARATE (f1)))
-			|| !COB_FIELD_HAVE_SIGN (f1)))) {
+	 && COB_FIELD_HAVE_SIGN (f1) == COB_FIELD_HAVE_SIGN (f2))) {
```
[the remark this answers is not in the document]
ddeclerck: Apparently it does not matter - the current test still passes after removing the matching sign condition (and the function is indeed called a few times with non-matching signs).

libcob/move.c
```diff
@@ -1291,26 +1298,28 @@ optimized_move_display_to_edited (cob_field *f1, cob_field *f2)
...
 						dst++;
 						src++;
 						break;
 				}
-				/* LCOV_EXCL_STOP */
 			} /* END OF SWITCH STATEMENT */
-		}   /* END OF INNER FOR LOOP   */
-	}     /* END OF OUTER FOR LOOP   */
+		} /* END OF INNER FOR LOOP   */
+	} /* END OF OUTER FOR LOOP   */
```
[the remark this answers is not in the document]
ddeclerck: Then let's remove them.

libcob/move.c
```diff
@@ -1686,15 +1695,13 @@ cob_move (cob_field *src, cob_field *dst)
...
 		case COB_TYPE_NUMERIC_EDITED:
-			if (COB_FIELD_DIGITS(src) == COB_FIELD_DIGITS(dst)
-			 && COB_FIELD_SCALE(src) == COB_FIELD_SCALE(dst)
-			 && COB_FIELD_HAVE_SIGN(src) == COB_FIELD_HAVE_SIGN(dst)
-			 && ((COB_FIELD_HAVE_SIGN(src) && (!COB_FIELD_SIGN_LEADING(src) && COB_FIELD_SIGN_SEPARATE (src)))
-				|| COB_FIELD_HAVE_SIGN(src) == 0)) {
-				optimized_move_display_to_edited(src, dst);
+			if (COB_FIELD_DIGITS (src) == COB_FIELD_DIGITS (dst)
+			 && COB_FIELD_SCALE (src) == COB_FIELD_SCALE (dst)
+			 && COB_FIELD_HAVE_SIGN (src) == COB_FIELD_HAVE_SIGN (dst)) {
```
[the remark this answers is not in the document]
ddeclerck: See comment above.

GitMensch:
Sounds I do wait for the next push, right?

ddeclerck:
Yes. Envoyé depuis mon appareil Galaxy
-------- Message d'origine --------De : Simon Sobisch ***@***.***> Date : 13/12/2024  12:20  (GMT+01:00) À : OCamlPro/gnucobol ***@***.***> Cc : OCP David Declerck ***@***.***>, Author ***@***.***> Objet : Re: [OCamlPro/gnucobol] Fix BLANK WHEN ZERO on signed NUMERIC-EDITED (PR #201) 
Sounds I do wait for the next push, right?

tests/testsuite.src/run_fundamental.at
ddeclerck: Okay, works as expected on GCOS.

However, this patch breaks one test, namely "Figurative constants to numeric field" in `run_misc.at:7431`:
```quote
-NUM9 value SPACES is       .
+NUM9 value SPACES is 000000.
```

GitMensch (changes requested):
seems good to me, apart from the Changelog entry we need to fix run_misc.at - the test "Figurative constants to numeric field" (which should get an additional keyword "MOVE") doers not only test what it says, but _additional_ the possible pretty-printing - which it shouldn't.

please change `is " NUM9 "."` to `is " NUM9 (1:) "."`

to fix that test


and yes, we have one additional issue which was seen by this: the pretty-printing of `USAGE DISPLAY` is different in at least those compilers:
acu: no pretty-printing on those *at all*
mf: all signed fields are printed as if they'd have a "SIGN SEPARATE TRAILING"
gc: full pretty-printing (including decimals, normalization, ...)

but _that's_ something to only create a FR on "provide different pretty-display for USAGE DISPLAY" (and I don't consider that a high priority)

ddeclerck:
Here we go.

I also sneaked into this the small testsuite adjustment mentionned in your last comment in #203.

GitMensch (approved):
Good as-is.

ddeclerck:
Merged in SVN @5403.

GitMensch (approved):
haven't checked if the error was triggered by that test case (I'd assume so) - but as the test is already run, please add a check for the expected value in Y (`if Y(1:) NOT = ") as well, then commit.

I'd say it is a follow-up of a previous recent change, so no need for a Changelog entry - but please reference it as "follow-up commit to r...." to point to the other one and give a minimal change note in the svn log.

ddeclerck:
> haven't checked if the error was triggered by that test case (I'd assume so)

At least it was, locally on my machine - and was fixed with the patch. So, let's be confident ;)

> - but as the test is already run, please add a check for the expected value in Y (`if Y(1:) NOT = ") as well, then commit.

Added.

> I'd say it is a follow-up of a previous recent change, so no need for a Changelog entry - but please reference it as "follow-up commit to r...." to point to the other one and give a minimal change note in the svn log.

Alright.

ddeclerck:
Merged in SVN @ 5407.

PR #202, opened 2024-12-10
Update macos.yml

GitMensch:
explicit choose ncurses to let the panel library be found (alternative: install ncurses via brew and extend PKG_CONFIG_DIR)


GitMensch:
.... testing for now, if it doesn't work, then I'll swap to brew install later.

PR #203, opened 2024-12-11
Use global var for `bdb_app_data`

emilienlemaire:
When using `DBT->app_data` to save the collating sequence function, BDB would not copy it to the DBT passed to `bdb_bt_compare` after a number of comparison (it seems like this number depends on the size of the data of the record).

The problems originates from BDB as everything is well set before any call to the BDB functions.

To avoid this problem the collating sequence is now always passed to global variable.


ddeclerck:
Thanks @emilienlemaire .

To add more context, this comes from a client reporting a "bdb_bt_compare was set but no collating sequence was stored in DBT" error (triggered from `bdb_bt_compare`) when using a key collating sequence on an indexed file.

For some reason, the `app_data` field of the DBT seems to be reset to NULL at some point when deleting records. This does not occur when using the global variable alternative version intended for older BDB versions that do not have the `app_data` field. We couldn't figure out if this is a bug in BDB or if we are just missusing this `app_data` field (comments online suggest it is not intended for public use, cf. [this](https://forums.oracle.com/ords/apexds/post/new-field-dbt-app-data-new-flag-db-dbt-usercopy-0521).

I'd suggest we'd just stick with the global variable version. What would you say, @GitMensch ?

libcob/ChangeLog
```diff
@@ -1,3 +1,7 @@
+2024-12-11  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
+
+	* fileio.c: Always use `bdb_app_data` global variable to save the
+	pointer to the collating sequence function
```
GitMensch, line 4: format: please start with a new line (reduces merge conflicts, especially with svn) and indent the second+ line of an entry using two spaces

tests/testsuite.src/run_file.at
```diff
@@ -14653,3 +14653,97 @@ AT_CHECK([$COBCRUN_DIRECT ./prog2 > prog2.txt])
 AT_CHECK([diff expected.txt prog2.txt])
 
 AT_CLEANUP
+
```
GitMensch, line 14656: format: please separate tests with two new lines

libcob/fileio.c
GitMensch, line 1: so... does the given test fail without this change but pass with that?

GitMensch (commented):
I'm not sure yet and need some more details.

libcob/fileio.c
ddeclerck: Yes.

GitMensch:
I can reproduce this as well:

~~~
libcob: error: bdb_bt_compare was set but no collating sequence was stored in DBT
libcob: warning: implicit CLOSE of READDEL ('READDEL')
BDB0616 Closing already-closed cursor
~~~

If you see the last line as well, please check if you can fix that (because that's an unrelated problem, which we normally can't reproduce, but can with the current other implementation).

I'm inspecting the issue that this PR is really about now.

libcob/fileio.c
```diff
@@ -690,15 +690,16 @@ static unsigned int	bdb_lock_id = 0;
 	key.data = fld->data;			\
 	key.size = (cob_dbtsize_t) fld->size
 
-#if (DB_VERSION_MAJOR > 4) || ((DB_VERSION_MAJOR == 4) && (DB_VERSION_MINOR > 0))
```
GitMensch, line 693: change to

```c
#if DB_VERSION_MAJOR >= 6 /* while the fields are part of DBT since 4.0, it wasn't
                             made part of the public API before 6.0 (DB_VERSION_FAMILY 12);
                             older versions overwrite that, see bug NNN */
```

GitMensch (changes requested):
OK, this issue really took 2 hours to check. First I've did some debugging within fileio and BDB (using watchpoints), then I've hunted down different version of the docs and headers.

And this is the result:
* BDB 4.1 added it to the header
* BDB 6.0 (docs from [6.0.19](https://ftp2.osuosl.org/pub/blfs/conglomeration/db/db-6.0.19.tar.gz) added it to the public API _and_ included the ABI break for the function; note: most GNU/Linux (and BSD) distributions have BDB 5.3

So please

* create a bug report with the findings
* adjust the changelog to have something like "fileio.c: make [DBT_SET_APP_DATA] dependent on DB_VERSION_MAJOR >= 6
* adjust the code as noted

I'll check the the updated version of your PR but am positive that the next one is a "pass".

libcob/fileio.c
```diff
@@ -690,15 +690,16 @@ static unsigned int	bdb_lock_id = 0;
...
-#define DBT_SET_APP_DATA(key,data)	((key)->app_data = (data))
-#define DBT_GET_APP_DATA(key)		((key)->app_data)
-#else
+/* #if (DB_VERSION_MAJOR > 4) || ((DB_VERSION_MAJOR == 4) && (DB_VERSION_MINOR > 0))
+ #define DBT_SET_APP_DATA(key,data)	((key)->app_data = (data))
+ #define DBT_GET_APP_DATA(key)		((key)->app_data)
+ #else
+*/
 /* Workaround for older BDB versions that do not have app_data in DBT */
 static void		*bdb_app_data = NULL;
```
GitMensch, line 699: Oh, and please change that to
`COB_TLS void		*bdb_app_data = NULL;`
as this brings _part_ of the newer API benefits to the old API implementation as well.

libcob/fileio.c
emilienlemaire: This does not work, even using the version 6 the test fails.

emilienlemaire: I've also tested with version 18, available on Oracle website, it also failed

GitMensch: That's interesting-  in this case you found a BDB bug.

@emilienlemaire Please change the old `#if` into a `#if 0` explaining the issue in 3-5 lines (possibly keeping the DB6 public note), as this ensures we don't wonder about that in the future and search other solutions if this change bring not-yet-know other issues.

GitMensch:
I do wonder - should we place that test https://github.com/OCamlPro/gnucobol/blob/9afa7a723c08dd55494947a73ee2b4d3f287ce93/libcob/fileio.c#L929-L938 within something like https://github.com/OCamlPro/gnucobol/blob/d5abb19870d0207be58564c4da11e88b954a9de7/libcob/move.c#L1035-L1036 when changing the related code?

libcob/fileio.c
```diff
@@ -690,15 +690,17 @@ static unsigned int	bdb_lock_id = 0;
 	key.data = fld->data;			\
 	key.size = (cob_dbtsize_t) fld->size
 
-#if (DB_VERSION_MAJOR > 4) || ((DB_VERSION_MAJOR == 4) && (DB_VERSION_MINOR > 0))
-#define DBT_SET_APP_DATA(key,data)	((key)->app_data = (data))
+/* #if DB_VERSION_MAJOR >= 6 *//* while the fields are part of DBT since 4.1, it wasn't
+                             made part of the public API before 6.0 (DB_VERSION_FAMILY 12);
+                             older versions overwrite that, see bug NNN */
```
GitMensch, line 695: please reference Bug 1032 here instead of NNN :-)

emilienlemaire: Yes I didn't open it yet when I copied this and then forgot to edit it :sweat_smile:

GitMensch (approved):
good to go in its current state (please mind the updates to integrate it upstream)

ddeclerck:
Merged on behalf of @emilienlemaire  at SVN @ 5398

GitMensch:
Of course I've missed something...

The line `AT_SKIP_IF([test "$COB_HAS_ISAM" != "db"])` should be adjusted - so whoever does a next commit please adjust it (the test will work for any ISAM as the collation is not explicit checked, it is "just used internally").

ddeclerck:
Noted - maybe when merging #201 .

PR #204, opened 2024-12-18
Fix STRING/UNSTRING/INSPECT bug

ddeclerck:
This fixes a bug introduced by PR #137.

The bug causes some source fields in STRING/UNSTRING/INSPECT to be altered under some circumstances, as highlighted by the new test in `run_misc.at`, which produces the following result when not fixed:
```
SRC-Y AFTER:  '  123   '
DST-X AFTER:  'DATA    '
```

This bug was caused by the removal of the `*_copy` fields, that were suspected to be useless (cf. [this comment](https://github.com/OCamlPro/gnucobol/pull/137#discussion_r1643854996) on PR #137).

This PR restores all those `*_copy` fields (which were present since SVN revision 1).

### DIFF
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,4 +1,10 @@
 
+2024-12-18  David Declerck <david.declerck@ocamlpro.com>
+
+	* string.c: fix a bug where the source of STRING/UNSTRING/INSPECT
+	  is overwritten, by restoring the *_copy fields that were removed
+	  with the change on 2024-02-26
+
 2024-12-11  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
 
 	* fileio.c: fixed Bug #1032 by always using global thread-static variable
@@ -296,6 +302,7 @@
 	* common.c: add missing include libxml/parser.h
 
 2024-02-26  Boris Eng <boris.eng@ocamlpro.com>
+
 	FR #488: using state structures instead of state vars for strings
 	* strings.c: moved static variables to structures
 
--- a/libcob/strings.c
+++ b/libcob/strings.c
@@ -63,6 +63,7 @@ struct cob_inspect_state {
 	size_t			      size;
 	cob_u32_t		      replacing;	/* marker about current operation being INSPECT REPLACING */
 	int			          sign;
+	cob_field			  var_copy;	/* use a copy to account for overwritten temporary field, see cob_inspect_init_common_intern */
 	enum inspect_type type;
 };
 
@@ -70,6 +71,9 @@ struct cob_string_state {
 	cob_field		*dst;
 	cob_field		*ptr;
 	cob_field		*dlm;
+	cob_field		dst_copy;	/* use a copy to account for overwritten temporary field, see cob_string_init_intern */
+	cob_field		ptr_copy;	/* use a copy to account for overwritten temporary field, see cob_string_init_intern */
+	cob_field		dlm_copy;	/* use a copy to account for overwritten temporary field, see cob_string_delimited_intern */
 	int		    	offset;	/* value associated with WITH POINTER clauses */
 };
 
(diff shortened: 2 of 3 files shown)


GitMensch:
and... why does it work with those (I see that it does, but copying the fields all the time is of course a performance issue)?

ddeclerck:
> and... why does it work with those

Or, why doesn't it work without those ?
Because the generated code for such STRING statement is as follows:
```
  cob_string_init (COB_SET_FLD (f0, 8, b_20, &a_3), NULL);
  cob_string_delimited (NULL);
  cob_string_append ((cob_field *)&c_3);
  cob_string_append (COB_SET_FLD (f0, 3, b_17 + 0LL, &a_2));
  cob_string_finish ();
```
It uses the same temporary field `f0` both for the destination and one of the sources, so we can not just hold a pointer to `f0`, we have to duplicate it. Note that this generated code is identical to the one that was generated before PR #137.

>  copying the fields all the time is of course a performance issue
I merely restored what was there before PR #137 (except it is now thread-local), so if there was a performance issue, it was there before.

> and... that code is not related to INSPECT at all, is it?
INSPECT also had those copy fields, so I assumed they were here for the same reason.

GitMensch:
I'd like to see the place where the INSPECT (which can also have a multi-part approach) has this - but the _real_ underlying issue is that one of the concepts is broken:

* `output_stmt()` has an internal `stack_id`
* this used for temporary fields and used to keep parts of a statement separate
* using the same id within different parts can have unexpected consequences

I see the following solutions around this:

a) libcob: for _any_ "multi part statements": always use a "per statement execution" copy for cob_fields; this is what seems to have been done for STRING/UNSTRING/INSPECT _before_ (obviously without necessary documentation; a one/two liner about "may be a temporary field which is used in multiple parts, i.e. cob_string_init + cob_string_append, we therefore need a local copy")
b) codegen: ensure that for multi-part statements there is only _one_ stack_id (I'm quite sure that we get into the `output_stmt()` multiple times, a debugging session should provide a hint where to improve that
c) internal AST (during parse): ensure that for anything where we "write" to data, there's no temporary field created, but a "real field" used - but on second thought that's likely not possible because for any ref-mod and subscript we still need that

The issue is seen after 3.2.1 and we want to fix that before 3.3 (=ASAP).
The codegen is identical here (or similar for 2.2), so for everything before 4.x a _required_ codegen change is not possible, ... leaving a)

@ddeclerck can you please:

* review the changes if they ensure that _field_ access happens only to a local copy of the passed fields for multi-parts
* recheck if we need both the pointer and the copy field - I guess replacing "cob_field *dst" by cob_field	dst_copy" would be fine
* add reasonable comments why we don't use a pointer but a copy (minor inline in the struct, referencing the longer version included in the function that setup this copy
* try to add a multi-part INSPECT example for the testsuite - because that's obviously missing

ddeclerck:
> * review the changes if they ensure that _field_ access happens only to a local copy of the passed fields for multi-parts

I'm not sure to get what you mean.

> * recheck if we need both the pointer and the copy field - I guess replacing "cob_field *dst" by cob_field	dst_copy" would be fine

Most of those copied field are tested for NULL at some point or another, so a pointer is convenient. Of course, it could be replaced by a flag to indicate if it is defined or not.

> * add reasonable comments why we don't use a pointer but a copy (minor inline in the struct, referencing the longer version included in the function that setup this copy

Done.

> * try to add a multi-part INSPECT example for the testsuite - because that's obviously missing

It was tricky but I managed to build an example which triggers the bug.

GitMensch:
Mostly FYI - the testcase triggers two issues with MF:

```cobol
           INSPECT DST-X (1) REPLACING
             LEADING "654" BY SRC-X (1)
             TRAILING "321" BY SRC-Y (1).
```

>     15       TRAILING "321" BY SRC-Y (1).

>*1614-S********************                                                  **
**    INSPECT TRAILING match item must be single character item

They do allow that for leading... so, strange regression.

> *1323-S**********************************                                    **
**    Cannot mix TRAILING option with other INSPECT phrases.

The result (apart from displaying the sign in a different place) is identical when changing the code to

```cobol
           INSPECT DST-X (1) REPLACING
             LEADING "654" BY SRC-X (1)
              ALL"321" BY SRC-Y (1).
```

which I _guess_ would also trigger the bug (and is fixed by the changes). If this is the case, please adjust the testcase that way.


As we want to get this upstream "very soon" and the states are only reserved one time per thread (instead of the intermediate version which had the state in each using program... this is otherwise good to go.

GitMensch:
Upstream with [r5406](https://sourceforge.net/p/gnucobol/code/5406).

PR #205, opened 2024-12-18
COMP-X displayed with 3 characters when size is one byte

emilienlemaire:


cobc/tree.h
```diff
@@ -914,6 +914,7 @@ struct cb_field {
 	int			size;		/* Field size */
 	int			level;		/* Level number */
 	int			memory_size;	/* Memory size */
+	int			compx_size;	/* Original COMP-X byte size */
```
GitMensch, line 917: Couldn't we just use f->pic->size for this rarely used definition, instead of adding an extra field to all variables?

GitMensch (changes requested):
I do wonder on the need of the compx_size attribute and the field's alphanumeric category...

emilienlemaire:
Hi Simon,
as mentioned in the description comment, this PR is mainly based on changes from SVN for GC4.
As such I took all changes from one of the commits (which include the addition of the `compx_size` field),
and a (very) small part from another one. I choose to take the whole part from the first so those changes
won't need to be made again when going to GC4.
I can take less from the mentioned commit, if needed.

GitMensch:
Thanks Emilien,

I do understand the approach, but question if the solution in the rw-branch (which was later merged to pangea which later became the new trunk = GC4) was the correct thing to do in the first place.

In any case - please inspect the real commit here - [r614](https://sourceforge.net/p/gnucobol/code/614).
This also allows you to merge the original changelog entry (extending that would be fine) with a possible new Changelog entry about your adjustments, referencing the date of that original entry.

Question: is there a reason to not use the testcase from the original commit instead of the new one (=misses it something)?

emilienlemaire:
I had the tests before editing the code, so I used it to validate the changes and it seemed to me that the other test did more than just check for the display value. Thinking about it again the original test should be used so everything from it is tested again.

emilienlemaire:
This is now taking only changes from r614, once again I think this is the way to go for an easier merge of later commits that uses what r614 changed

PR #206, opened 2024-12-20
Improve handling of partial keys

ddeclerck:
This PR improves handling of partial keys, to ensure BDB always compares keys of identical length (which would trigger an error when using a non-native collating sequence, since the custom comparison function enforce keys of same length).

### DIFF
--- a/tests/testsuite.src/run_file.at
+++ b/tests/testsuite.src/run_file.at
@@ -3098,6 +3098,66 @@ rewriting key: 98651234
 AT_CLEANUP
 
 
+AT_SETUP([INDEXED START with partial key and collating sequence])
+AT_KEYWORDS([runfile READ WRITE START])
+
+AT_SKIP_IF([test "$COB_HAS_ISAM" = "no"])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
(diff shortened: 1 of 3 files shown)


tests/testsuite.src/run_file.at
```diff
@@ -3098,6 +3098,64 @@ rewriting key: 98651234
 AT_CLEANUP
 
 
+AT_SETUP([INDEXED START with partial key and collating sequence])
+AT_KEYWORDS([runfile READ WRITE START])
+
+AT_DATA([prog.cob], [
```
GitMensch, line 3104: Needs to be skipped off tomorrow isn't configured for isam

GitMensch (approved):
That looks clean - let's get that upstream!

ddeclerck:
Just for you to know  I ran some perf test on this, and my findings are as follows:
- I measure similar performance with and without this patch when using BDB's default comparison function
- when using the custom comparison function, this patch (which uses `indexed_key_compare` to compare keys of identical length) yields better performance than a non-patched version that uses a comparison function that considers keys of different lengths (`cob_cmp_strings`)

So this would confirm this patch is the right approach.

As you already approved this, I'll merge upstream in a moment.

ddeclerck:
Merged in SVN @ revision 5412.

PR #209, opened 2025-01-08
Add a COB_LOAD_GLOBAL config option to modify `dlopen` behavior

florianschmidt1994:
This PR adds a config option called `COB_LOAD_GLOBAL` which allows the flags of `dlopen` used in `call.c`  to be changed between `RTLD_GLOBAL` and `RTLD_LOCAL` (see https://linux.die.net/man/3/dlopen)

# Motivation
This flag is motivated by an ambition to be able to write a multi-threaded c program, that then loads `libcob` with `dlmopen` in one namespace per thread, effectively allowing us to call the same libcob / Cobol module from multiple threads in a "thread-safe" way. The image below shows the a very simplified intended call sequence

CanvasAsImage

**Because there is a "bug" in dlopen that stops us from calling `dlopen("module", RTLD_GLOBAL)`, this PR introduces an option to be able to change this flag to `RTLD_LOCAL` at runtime.** 

> As at glibc 2.24, specifying the RTLD_GLOBAL flag when calling dlmopen() generates an error. Furthermore, specifying RTLD_GLOBAL when calling dlopen() results in a program crash (SIGSEGV) if the call is made from any object loaded in a namespace other than the initial namespace.

# Docs
```.cfg
# Environment name:  COB_LOAD_GLOBAL
#   Parameter name:  load_global
#          Purpose:  tell the system loader to provide symbols in CALLed
#                    programs globally, allowing symbols to be found for linked
#                    libraries later
#             Type:  boolean
#             Note:  COBOL CALLs will always find symbols in already CALLed or
#                    pre-modules; this setting is mostly an advise to the system,
#                    not all systems are capable of loading libraries global/local
#          Default:  false
#          Example:  load_global true
```

# Implementation
- Change `lt_dlopen` from a macro definition to an actual function, which internally calls dlopen with the flags based on the configuration
- Add cob_load_global / COB_LOAD_GLOBAL as a configuration
- Add documentation to runtime.cfg about COB_LOAD_GLOBAL


florianschmidt1994:
So basically something like this, right?

```c
static void* cob_dlopen(const char* filename) {
	#ifdef _WIN32
	if (x == NULL) {
		return GetModuleHandle (NULL);
	}
	return LoadLibrary(x);	
	#elif	defined(USE_LIBDL)
	int flags = cobsetptr->cob_load_global 
		? RTLD_LAZY | RTLD_GLOBAL
		: RTLD_LAZY | RTLD_LOCAL;

	return dlopen(filename, flags);
	# else
	// TODO: initialise lt_dladvise
	lt_dlopenadvise(filename, /*advise here*/);
}
```

ChangeLog
```diff
@@ -1,3 +1,12 @@
+2025-01-08  Simon Sobisch <simonsobisch@gnu.org>
+
+	* config/runtime.cfg: Add documentation for COB_LOAD_GLOBAL flag
+
```
GitMensch, line 4: this goes to config/Changelog  (with a leading empty line), preferably with the original date from 2024

ChangeLog
```diff
@@ -1,3 +1,12 @@
...
+2025-01-08  Florian Schmidt <florian.schmidt.1994@icloud.com>
+
+	* libcob/call.c, libcob/coblocal.h, libcob/common.c: Introduce a flag
+	  COB_LOAD_GLOBAL (load_global) to toggle if the symbols of CALLed 
+	  programs should be made available in the global namespace
```
GitMensch, line 10: this goes to libcob/Changelog  (with a leading empty line)

libcob/call.c
```diff
@@ -60,65 +60,82 @@ FILE *fmemopen (void *buf, size_t size, const char *mode);
...
+        lt_dladvise_local(&advise);
+    }
+
+    void* handle = lt_dlopenadvise(filename, advise);
+    lt_dladvise_destroy(&advise); // Clean up advise object
+    return handle;
 #endif
+}
+
+#define lt_dlopen(x) cob_dlopen(x) // TODO: Remove this!
```
GitMensch, line 138: just replace all occurrences of lt_dlopen with the new function in all places, and have the `static cob_dlopen` function down next to the other static functions - which removes the need to change the position of cobsetptr as well, so that can be reverted

libcob/coblocal.h
```diff
@@ -300,6 +300,7 @@ typedef struct __cob_settings {
 
 	/* call.c */
 	int	cob_physical_cancel;	/* 0 "= "logical only" (default), 1 "also unload", -1 "never unload" */
+	unsigned int	cob_load_global; /* Wether dlopen should use the global or local namespace */
```
GitMensch, line 303: that's not only dlopen so something like `/* hint for dynamic linker to make symbols available 0=global, 1=local */` (maybe shorter)

libcob/call.c
```diff
@@ -60,65 +60,82 @@ FILE *fmemopen (void *buf, size_t size, const char *mode);
...
+    lt_dladvise_init(&advise);
+
+    if (cobsetptr->cob_load_global) {
+        lt_dladvise_global(&advise);
+    } else {
+        lt_dladvise_local(&advise);
+    }
+
+    void* handle = lt_dlopenadvise(filename, advise);
+    lt_dladvise_destroy(&advise); // Clean up advise object
```
GitMensch, line 133: please make the advise static and clean it up in call.c's cleanup function (depending on its preprocessor macro, of course), you can also init it in the init function - just set the advise flag here (the setting for both dlopen and lt_dlopen will be moved out to a config function later - you're welcome to add it (see screenio.c, this may be even explained in the video :-) and call it from the config side if this var or COB_LIBRARY_PATH was adjusted

config/runtime.cfg
```diff
@@ -306,6 +306,18 @@
...
+# Environment name:  COB_LOAD_GLOBAL
+#   Parameter name:  load_global
+#          Purpose:  tell the system loader to provide symbols in CALLed
+#                    programs globally, allowing symbols to be found for linked
+#                    libraries later
+#             Type:  boolean
+#             Note:  COBOL CALLs will always find symbols in already CALLed or
+#                    pre-modules; this setting is mostly an advise to the system,
+#                    not all systems are capable of loading libraries global/local
+#          Default:  false
```
GitMensch, line 318: As this changes the current default - please add a NEWS entry at the appropriate place, telling about this change and the option to use the setting for the old behavior

libcob/call.c
```diff
@@ -60,65 +60,82 @@ FILE *fmemopen (void *buf, size_t size, const char *mode);
...
+static cob_settings		*cobsetptr = NULL;
+
 #ifdef	_WIN32
+	#define WIN32_LEAN_AND_MEAN
+	#include <windows.h>
+	#include <io.h>	/* for access */
 
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-#include <io.h>	/* for access */
```
GitMensch, line 67: We only use indents within conditional compilation very rarely- so please revert that part.

GitMensch (changes requested):
nice update, we may be able to push that soon

Note: we still need the testcase addition - does not need to be something that benefit from this as this is hard to make portable - just do a simple CALL and CANCEL. We _should_ have an easy one for COB_LOAD_CASE

config/runtime.cfg
florianschmidt1994: Is this the behavior that we want to have as default? Or should we keep the default as it was before this change and only add this option to change it when needed?

GitMensch: Previously the behaviour was depending on the underlying system (is dlopen there and works, what is the default for its option and otherwise the default for lt_dlopen), with _often_ being global dlopen.
Now we do have an option to change this and yes, default that to local.

This is also the reason that we need a NEWS entry to make the users aware of "previously system specific, now we ask the dynamic link loader to ...".

tests/testsuite.src/run_misc.at
```diff
@@ -4285,6 +4285,57 @@ AT_CHECK([COB_LOAD_CASE=UPPER ./caller], [0], [], [])
...
+       PROGRAM-ID.      caller.
+       PROCEDURE        DIVISION.
+           CALL "callee"
+           END-CALL.
+           CANCEL "callee".
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_MODULE CALLEE.cob], [0], [], [])
+AT_CHECK([$COMPILE caller.cob], [0], [], [])
```
GitMensch, line 4335: After initial testing - please combine those tests into one, possibly named "CALL/CANCEL and  COB_LOAD_GLOBAL"

florianschmidt1994:
When developing, I've noticed some test failures, where some of them might be related to our change. On the latest commit of this repositories "main branch" (`gcos4gnucobol-3.x`), there are some test failures when the build is configured with `--without-dl`. These failures persist also with our changes applied. 
| Commit Hash | Loader | COB_LOAD_GLOBAL | Test Failures |
|--------|--------|--------|--------|
| 731b81a327fe7108ca28d024a5ca13c9457dbcdc | --without-dl | _not applicable_ | 47 49 50 784 failed |
| 731b81a327fe7108ca28d024a5ca13c9457dbcdc | <standard> | _not applicable_ | _no test failure_ |


In the first version of this PR (bd80e8ce601b1663aa69793f086284c66fd2b463), there is an additional test failure (781) when using `COB_LOAD_GLOBAL=false`, i.e. the default, and preloading where symbols are overwritten in an unexpected manner. 
| Commit Hash | Loader | COB_LOAD_GLOBAL | Test Failures |
|--------|--------|--------|--------|
| bd80e8ce601b1663aa69793f086284c66fd2b463 | <standard> | false | **781** failed |
| bd80e8ce601b1663aa69793f086284c66fd2b463 | <standard> | true | _no test failure_ |
| bd80e8ce601b1663aa69793f086284c66fd2b463 | --without-dl | false | 47 49 50 **781** failed |
| bd80e8ce601b1663aa69793f086284c66fd2b463 | --without-dl | true | 47 49 50 failed |

In history up to the latest commit (62de6254d0418b45e142b14f27245b481c1594b1), there exists a fix (see 1f4d60760e3fb908f9818d68ac0044fae7ee520a) that is already applied in WIN32 environment to fix this.
| Commit Hash | Loader | COB_LOAD_GLOBAL | Test Failures |
|--------|--------|--------|--------|
| 62de6254d0418b45e142b14f27245b481c1594b1 | <standard> | false | _no test failure_ |
| 62de6254d0418b45e142b14f27245b481c1594b1 | <standard> | true | _no test failure_ |
| 62de6254d0418b45e142b14f27245b481c1594b1 | --without-dl | false | 47 49 50 failed |
| 62de6254d0418b45e142b14f27245b481c1594b1 | --without-dl | true | 47 49 50 failed |

tests/testsuite.src/run_misc.at
```diff
@@ -4355,6 +4355,32 @@ AT_CHECK([COB_LOAD_CASE=UPPER ./caller], [0], [], [])
...
+           END-CALL.
+           CANCEL "callee".
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_MODULE callee.cob], [0], [], [])
+AT_CHECK([$COMPILE caller.cob], [0], [], [])
+AT_CHECK([COB_LOAD_GLOBAL=true ./caller], [0], [], [])
+AT_CHECK([COB_LOAD_GLOBAL=false ./caller], [0], [], [])
+
```
GitMensch, line 4382: Can you please add both COB_LOAD_GLOBAL options (and its name to the keyword) for the pre-load test to have that covered as well?

config/runtime.cfg
```diff
@@ -306,6 +306,18 @@
...
 #          Example:  PHYSICAL_CANCEL  TRUE
 
+# Environment name:  COB_LOAD_GLOBAL
+#   Parameter name:  load_global
+#          Purpose:  tell the system loader to provide symbols in CALLed
+#                    programs globally, allowing symbols to be found for linked
+#                    libraries later
+#             Type:  boolean
+#             Note:  COBOL CALLs will always find symbols in already CALLed or
+#                    pre-modules; this setting is mostly an advise to the system,
```
GitMensch, line 316: should be "pre-loaded modules"

libcob/call.c
```diff
@@ -289,6 +283,37 @@ static int last_entry_is_working_directory (const char *buff, const char *pstr)
 	return 0;
 }
 
+static void* cob_dlopen(const char* filename) {
+#ifdef _WIN32
```
GitMensch, line 287: because of the `#elif` below, please use `#if defined (_WIN32)` here

libcob/call.c
```diff
@@ -289,6 +283,37 @@ static int last_entry_is_working_directory (const char *buff, const char *pstr)
...
+	}
+	return LoadLibrary(filename);
+#elif	defined(USE_LIBDL)
+	int flags = cobsetptr->cob_load_global 
+		? RTLD_LAZY | RTLD_GLOBAL
+		: RTLD_LAZY | RTLD_LOCAL;
+
+	return dlopen(filename, flags);
+#else
+    if (advise != NULL) {
```
GitMensch, line 299: advise is global, so only set if the "last set" variant is different to the current one - you can store the last used cob_load_global as a static var within this function for that

GitMensch (changes requested):
I hope to reinspect that PR soon, ideally after looking over CANCEL in general.

Can you please do the minor and cosmetic changes outlined in this review and rebase afterwards?

NEWS
```diff
@@ -87,6 +87,12 @@ NEWS - user visible changes				-*- outline -*-
    $b for executable basename, $d for date in YYYYMMDD format, $t for time
    in HHMMSS format (before, only $$ was available for pid)
 
+** introduction of the COB_LOAD_GLOBAL boolean flag, which determines whether
```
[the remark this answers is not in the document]
GitMensch: only sentences should start with an upper-case and be ended with a period, but you can add a _short_ summary of that change in all lower without a period to start it (kind of a headline)

GitMensch:
@florianschmidt1994 thanks for the updates.

Please use the CI generated tarball, to check if your _original_ setup that needed that change (a java starter running the same process more than once) still works. Deep-link to the CI generated source distribution: https://github.com/OCamlPro/gnucobol/actions/runs/16163870452/artifacts/3493326773

Also: can you please create a minimal runner in C that we can add to the testsuite that mimics your java starter?
Currently we only verified that the new setting works with both options, but we don't have a test case that only passes with one of those...

GitMensch:
The "coverage" titled test fails for C89 compat reasons:

~~~

+caller.c:5:22: warning: extra tokens at end of #include directive
+    5 | #include <pthread.h> // For POSIX threads
+      |                      ^
+caller.c:6:22: warning: extra tokens at end of #include directive
+    6 | #include <dlfcn.h>   // For dynamic library loading
+      |                      ^
+caller.c:13:1: error: C++ style comments are not allowed in ISO C90
+   13 | // Function to call COBOL shared library functions
+      | ^
+caller.c:13:1: note: (this will be reported only once per input file)
+caller.c: In function 'callCobol':
+caller.c:34:30: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
+   34 |     RESOLVE_SYMBOL(cob_init, cob_init_func);
+      |                              ^~~~~~~~~~~~~
+caller.c:25:5: note: in definition of macro 'RESOLVE_SYMBOL'
+   25 |     type symbol = (type) dlsym(libCobHandle, #symbol); \
+      |     ^~~~
+caller.c:35:33: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
+   35 |     RESOLVE_SYMBOL(cob_resolve, cob_resolve_func);
+      |                                 ^~~~~~~~~~~~~~~~
+caller.c:25:5: note: in definition of macro 'RESOLVE_SYMBOL'
+   25 |     type symbol = (type) dlsym(libCobHandle, #symbol); \
+      |     ^~~~
+caller.c:36:30: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
+   36 |     RESOLVE_SYMBOL(cob_tidy, cob_tidy_func);
+      |                              ^~~~~~~~~~~~~
+caller.c:25:5: note: in definition of macro 'RESOLVE_SYMBOL'
+   25 |     type symbol = (type) dlsym(libCobHandle, #symbol); \
+      |     ^~~~
+caller.c:38:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
+   38 |     hello_world_func hello_world = cob_resolve("prog");
+      |     ^~~~~~~~~~~~~~~~
+caller.c: In function 'main':
+caller.c:60:5: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
+   60 |     for (long i = 0; i < numThreads; i++) {
+      |     ^~~
+caller.c:60:5: note: use option '-std=c99', '-std=gnu99', '-std=c11' or '-std=gnu11' to compile your code
+caller.c:67:14: error: conflicting types for 'i'; have 'int'
+   67 |     for (int i = 0; i < numThreads; i++) {
+      |              ^
+caller.c:60:15: note: previous definition of 'i' with type 'long int'
+   60 |     for (long i = 0; i < numThreads; i++) {
+      |               ^
+caller.c:67:5: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
+   67 |     for (int i = 0; i < numThreads; i++) {
+      |     ^~~
+cc1: some warnings being treated as errors
~~~

To reproduce that locally compile it with `gcc -std=c89 -Werror=declaration-after-statement` - in general you want variables to start at the begin of a block and use `*/ block comments */` .

GitMensch:
and the MSYS1 entries error because of missing pthread support:

~~~
+caller.c:5:10: fatal error: pthread.h: No such file or directory
+    5 | #include <pthread.h> // For POSIX threads
+      |          ^~~~~~~~~~~
~~~

It is fine to skip this test on MSYS1. To do so add

```m4
# skip test for MSYS1 - we don't have pthread there
# FIXME: make that test more portable
AT_SKIP_IF([test "$OSTYPE" = "msys"])
```

at the start of the test.

tests/testsuite.src/run_misc.at
GitMensch, line 1: note: the issue is that the resolve only works if the file can be found and prog.cob creates prog.dll/prog.so, if that is found it checks next for the entry point, which is written like the PROGRAM-ID.

florianschmidt1994:
Thanks for the hints @GitMensch. I've also disabled the tests on macOS, as there is no `dlmopen` there. 

Test runs on Windows (e.g. https://github.com/OCamlPro/gnucobol/actions/runs/17068765301/job/48392123249?pr=209) say `caller.c not found` when trying to execute.
```
AT_CHECK([$COB_CC $COB_CFLAGS caller.c -ldl $COB_LIBS -o caller], [0], [], [])
```

Do you know if we might need to call the compiler arguments differently to make it more portable across platforms?


Also there's an issue with help2man in the Windows MSYS2 prepare workflow (https://github.com/OCamlPro/gnucobol/actions/runs/17069516851/job/48394605655?pr=209) that I don't quite understand, yet. Any ideas?

GitMensch:
> Thanks for the hints @GitMensch. I've also disabled the tests on macOS, as there is no `dlmopen` there.
> 
> Test runs on Windows (e.g. https://github.com/OCamlPro/gnucobol/actions/runs/17068765301/job/48392123249?pr=209) say `caller.c not found` when trying to execute.
> 
> ```
> AT_CHECK([$COB_CC $COB_CFLAGS caller.c -ldl $COB_LIBS -o caller], [0], [], [])
> ```

Use `AT_CHECK([$COBC -x -ldl -lpthread -caller.c], [0], [], [])` there (drop pthread if you think that's really not necessary).
The Windows tests likely haven't setup those vars correctly for some MSYS2 translation reason.

tests/testsuite.src/run_misc.at
```diff
@@ -4252,13 +4252,125 @@ AT_DATA([caller.cob], [
...
+        }
+    }
+    /* Wait for all threads to finish */
+    for (j = 0; j < numThreads; j++) {
+        pthread_join(threads[[j]], NULL);
+    }
+    return EXIT_SUCCESS;
+}
+])
+AT_CHECK([$COBC -x -ldl -lpthread caller.c], [0], [], [])
```
GitMensch: You may skip the test if that compile fails, this reduces the necessary changes.
Something along:
```
AT_CHECK([$COBC -x -ldl -lpthread caller.c], [0], [], [],
    echo "C compilation failed, skipping rest of group"
    exit 77
)
```
(77 is the standard exit code for "skip")

... or change configure + aclocal as you did, but then also change build_windows/config.h.in

GitMensch:
@florianschmidt1994 Can you please rebase?

oguzcankirmemis:
just adding myself to the thread to get notifications, @GitMensch you can ofc ping me if we need any further changes here

GitMensch:
Now upstream with minor adjustments - let's hope we covered the important parts for portability (which is the possible culprit)

PR #210, opened 2025-01-09
Fix for missing "end verbatim" in doc

ddeclerck:
There was a missing "end verbatim" in the file generated by `cbrunt.tex.gen`. Seems it's been like this for ages, but we only notice now because the MacOS CI began to fail for no apparent reason (in fact brew's `texinfo` version was updated from 7.1 to 7.2, which is probably the reason for this error to be detected).

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,8 @@
 
+2025-01-09  David Declerck <david.declerck@ocamlpro.com>
+
+	* cbrunt.tex.gen: fix for missing "@end verbatim"
+
 2024-10-30  Chuck Haatvedt <chuck.haatvedt@gmail.com>
 
 	* gnucobol.texi: document new multiple window function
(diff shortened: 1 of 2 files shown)


doc/ChangeLog
GitMensch, line 1: As this is the first 2025 entry: please add 2025 to the copyright years found at the end of the file

GitMensch (changes requested):
Comparing the versions from January 3rd and 9th:
the first has a _text_

~~~
@section Call environment
@verbatim
~~~

in Appendix I, the second an actual section in the PDF - so LGTM for upstream

just one requested change in the Changelog - but this can directly go upstream

doc/ChangeLog
ddeclerck: Actually, 2023 and 2024 are also missing though they have some ChangeLog entries. Should those be added too ?

GitMensch: correct, please adjust to 2009-2025

ddeclerck:
Merged in SVN @ 5418.

PR #211, opened 2025-01-10
Set init-justify to no in gcos-strict.conf

ddeclerck:
After testing on GCOS, I confirm we should set init-justify to no.

(yeah, a PR for this is maybe a bit too much ;))

### DIFF
--- a/config/ChangeLog
+++ b/config/ChangeLog
@@ -1,4 +1,8 @@
 
+2025-01-10  David Declerck <david.declerck@ocamlpro.com>
+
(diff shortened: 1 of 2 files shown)


GitMensch (approved):
Nice, GCOS is correct per COBOL then (this is the single place I'm currently aware of where IBM does something explicit specified to be different than the COBOL standard - nearly all other "possible issues" are about undefined behavior).

Your change looks good, but please adjust the copyright year on config/Changelog as well, then that's good for upstream.

GitMensch:
checked in with r5420

PR #213, opened 2025-01-28
[4.x] Close stream first, close fd only if not stream

ddeclerck:
This PR attemps to fix a debug assertion occuring under the MSVC CI (`Assertion failed: (_osfile(fh) & FOPEN)`).

Apparently, when a stream is created from a file descriptor via `fdopen`, it should always be closed by `fclose`. This is not always the case in the 4.x branch. While this seems to fail silently on most OSes, it triggers a debug assertion under MSVC debug.

This proposed fix just always tries to close the stream if it is not NULL, otherwise it just closes the file descriptor.
This fixes 40 (!) tests under MSVC debug.

References:
- Linux: https://linux.die.net/man/3/fdopen
- Windows: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fdopen-wfdopen?view=msvc-170
- BSD: https://man.freebsd.org/cgi/man.cgi?query=fdopen&sektion=3&n=1
- macOS: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/fdopen.3.html
- IBM: https://www.ibm.com/docs/en/i/7.5?topic=functions-fdopen-associates-stream-file-descriptor

### DIFF
--- a/libcob/fileio.c
+++ b/libcob/fileio.c
@@ -1,5 +1,5 @@
 /*
-   Copyright (C) 2002-2012, 2014-2023 Free Software Foundation, Inc.
+   Copyright (C) 2002-2012, 2014-2025 Free Software Foundation, Inc.
    Written by Keisuke Nishida, Roger While, Simon Sobisch, Ron Norman
 
    This file is part of GnuCOBOL.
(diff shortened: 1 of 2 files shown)


libcob/fileio.c
```diff
@@ -4764,32 +4764,17 @@ cob_file_close (cob_file_api *a, cob_file *f, const int opt)
...
-				close (f->fd);
-				f->fd = -1;
-#ifdef _MSC_VER
-				f->file = NULL;
-#endif
-			}
 		}
-#ifndef _MSC_VER /* explicit disallowed there, is that the same for other compilers? */
 		if (f->file != NULL) {
+			/* This should close the underlying fd */
```
GitMensch, line 4769: When we drop all these parts, then `if (f->organization == COB_ORG_LINE_SEQUENTIAL) {` should be dropped completely as the code left is for pipes only, and those are always lsq.

GitMensch, line 4769: Why "should"?

Also: What does valgrind say to this change (`make check VGSUFFIX=close`)?

GitMensch (changes requested):
I wonder why this seems to be no issue in 3.x and if valgrind will report more leaks afterwards.
In any case, please adjust that first "if" (and therefore re-indent the not-yet changed code).

libcob/fileio.c
ddeclerck: > Why "should"?

Because I'm not 200 % sure ?

> Also: What does valgrind say to this change (`make check VGSUFFIX=close`)?

Checking on my computer. This is going to take a while.

ddeclerck:
> I wonder if valgrind will report more leaks afterwards

So, after checking, no change in the number of leaks.

GitMensch:
this means that you can drop your "should" comment along with adjusting that if, then push the result upstream

GitMensch (approved):

ddeclerck:
Merged in SVN @ 5442.

PR #214, opened 2025-01-29
Fix configure.ac : unused command line argument triggers erros

ddeclerck:
As mentionned in a previous comment, unused arguments trigger errors in the configure script :
```
error: argument unused during compilation: '-fstack-clash-protection' [-Werror,-Wunused-command-line-argument]
```
This makes detection of some features (`-fstack-clash-protection` on macOS) fail even if they are available.
This is even worst in 4.x (in the yet-to-be-merged batch), as it fails to detect "gcc pointer sign", which causes many testsuite failures.

This PR simply adds `-Wno-unused-command-line-argument` to the CFLAGS under gcc and Clang (and removes if from the CI workflow). Not sure if this should come earlier in the configure file ?

Should be merged in 4.x once upstream in 3.x.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -461,6 +461,7 @@ AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
 	# error macro not defined
 	#endif]])],
 	[COB_USES_CLANG_ONLY=yes
+	 CFLAGS="$CFLAGS -Wno-unused-command-line-argument"
 	 AC_MSG_RESULT([yes])],
 	[AC_MSG_RESULT([no])])
 
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -52,7 +52,6 @@ jobs:
       - name: configure
         run: |
           cd _build
-          export CFLAGS="-Wno-unused-command-line-argument $CFLAGS"
           ../configure --enable-cobc-internal-checks \
                        --enable-hardening \
                        --with-curses=ncurses \


GitMensch:
Hm, this option was added in clang 3.1, but as expected earlier versions didn't raise a warning if its `Wno` form is used.
It still should not be applied to GCC, but only to clang. So maybe  set it directly after recognizing clang?

ddeclerck:
For some reason, I was under the impression this option also existed under GCC, but now I can't find any info supporting this. I'll adjust accordingly. Envoyé depuis mon appareil Galaxy
-------- Message d'origine --------De : Simon Sobisch ***@***.***> Date : 10/02/2025  20:31  (GMT+01:00) À : OCamlPro/gnucobol ***@***.***> Cc : OCP David Declerck ***@***.***>, Author ***@***.***> Objet : Re: [OCamlPro/gnucobol] Fix configure.ac : unused command line argument triggers erros (PR #214) 
Hm, this option was added in clang 3.1, but as expected earlier versions didn't raise a warning if its Wno form is used.
It still should not be applied to GCC, but only to clang. So maybe  set it directly after recognizing clang?

GitMensch:
> For some reason, I was under the impression this option also existed under GCC, but now I can't find any info supporting this.

I've just asked gcc by adding `-Wunused-command-line-argument` and both new and gcc told me "nope" :-)

ddeclerck:
Updated to apply the change only to Clang.

Now for some reason, the MSYS2 CI stopped working, complaining it can not find tex... That might be because the Windows runner image has been updated since last CI run... I guess I'll have to tackle this in another PR.

GitMensch:
Can't you pin (back) the windows runner image so updating it is a manual procedure (instead of the need to potentially fix it whenever it is updated)?

configure.ac
```diff
@@ -2254,6 +2254,10 @@ else
 	CFLAGS="$curr_cflags"
 fi
 
+if test "$COB_USES_CLANG_ONLY" = yes; then
+	CFLAGS="$CFLAGS -Wno-unused-command-line-argument"
+fi
+
```
GitMensch, line 2260: this still should be set directly when COB_USES_CLANG_ONLY is set, otherwise we can get warnings on previous settings like hardening

GitMensch (changes requested):

ddeclerck:
> Can't you pin (back) the windows runner image so updating it is a manual procedure (instead of the need to potentially fix it whenever it is updated)?

Seems GitHub Actions does not allow that (cf https://github.com/actions/runner/issues/3447)...

ddeclerck:
Merged @ SVN 5445 (3.x - to be merged in 4.x)

PR #215, opened 2025-02-04
Update 4.x CI, mostly following 3.x CI

ddeclerck:
This PR updates the 4.x CI, mostly based on the the 3.x CI and with elements from PR #190, as well as tweaks and cosmetic adjustments.

I'm rather satisfied with these changes. The only remaining problem is the packaging of the MSVC distribution which keeps failing no matter what I tried, so set it to continue-on-error for now.

I would advise to get PR #214 upstream first (targets 3.x), then added to the current 3.x->trunk merge batch, and once that batch is upstream I'll rebase and merge this PR.


GitMensch:
3.x doesn't fail that bad on MSVC, does it?

Checking some errors:

~~~
+  cob6040_2.obj : error LNK2005: DllMain already defined in cob6040_1.obj
+  cob6040_3.obj : error LNK2005: DllMain already defined in cob6040_1.obj
+  cob6040_4.obj : error LNK2005: DllMain already defined in cob6040_1.obj
~~~

--> something that includes `DllMain` needs to be merged (or re-adjusted) to let this pass

~~~
+libcob: prog.cob:23: error: I/O routine BDB cannot be loaded: module libcobdb-1.dll not found
+libcob: prog.cob:23: error: ERROR I/O routine BDB is not present
~~~

Do the MSVC build files include that target (and ideally only build it if BDB is configured)?
If there's nothing like that it _should_ be possible (at least that was the case when Ron worked on this) to have one and only one backend be build directly into fileio (more or less as it was in all versions so far); if that is the case then someone (most likely me, ... or possibly even you while merging) messed that up in 4.x.

~~~
+(0) : Run-Time Check Failure #1 - A cast to a smaller data type has caused a loss of data.  If this was intentional, you should mask the source of the cast with the appropriate bitmask.  For example:  
+	char c = (i & 0xFF);
~~~

yep, we fixed that in 3.x - but don't ask me about the revision number or date...


You noted there are _automated packaging_ issues? Maybe the makedist.cmd needs an updated (merge or manual adjustment)?

GitMensch:
> Yeah, packing just fails, as shown in the CI log:
> 
> ```
> Setting environment for GnuCOBOL.
> Building extras with  
> 
> cobc had unexpected return value -1073741515, running verbose again...
> D:\a\gnucobol\gnucobol\build_windows\distnew\bin_x86\cobc.exe
> Abort!
> ```
> 
> I did investigate, and both cobcrun and cobc terminate with code -1073741515, but I couldn't figure out how to fix that (spent two days on this without anything).

I see, so the version could not be queried https://github.com/OCamlPro/gnucobol/blob/41991806d57b848799e0a90633ed61750a793303/build_windows/makedist.cmd#L393-L399 ,which means cobcrun did terminate before even starting, which means its dependencies cannot be resolved from PATH - which obviously worked during the testsuite.

checking further - all libraries are found and copied as expected, but the previously exported binaries are not used - we'd expect an output for that:
https://github.com/OCamlPro/gnucobol/blob/41991806d57b848799e0a90633ed61750a793303/build_windows/makedist.cmd#L385-L386

Conclusion: VCPKG_EXPORT_DIR is empty and... that part is exactly what got not included from #190 :-)
Add the part from the 3.x CI definition https://github.com/OCamlPro/gnucobol/blob/3fb682569d9e1c69eba2a577d0a4e78bffb877ec/.github/workflows/windows-msvc.yml#L209-L222 and drop continue-on-error.

--------------

> These errors have shown up in the GC4 MSVC CI since the beginning of the merge (SVN revision 5206).
There is no specific MSVC target for those libraries ; the corresponding C files are part of the libcob target.
Note that we specifcally added this TODO entry for this: "Correctly implement delay-loading under MSVC"

These are two different options. While I'm aware that the state is "delay load stuff doesn't work on MSVC", presumably because there's no .vcproj for the library to delay-load [it isn't, just rechecked]; there is (or at least was) another option: don't keep it separate but include it into libcob. This is related to some defines possibly not set or not working anymore.

... further checks on this lead to https://github.com/OCamlPro/gnucobol/blob/41991806d57b848799e0a90633ed61750a793303/libcob/fbdb.c#L24C1-L25C69 - and that file is included in the vcproj (= we already have a static build definition), so the loading https://github.com/OCamlPro/gnucobol/blob/gc4/libcob/fileio.c#L1320-L1324 should work... checking further - the symbols are not correctly exported. blame shows - I've did that back in [June 2022 with r4629 / 3b31b75](https://github.com/OCamlPro/gnucobol/commit/3b31b75dc01dadfd5d4a90d3d045381e56a08fba#diff-c26244c2b09cb26e18c979b995beef4b2d22be76c6f22a76ce08e6369f5fa353) for all backends but BDB(LDMB). I've just fixed that upstream.

ddeclerck:
After rebase to include your fix and the addition of VCPKG_EXPORT_DIR, we're now down from 59 to 29 failures (a bit more on debug, but that should match release once the current GC3/4 batch is merged).

I also included the fix from PR #217 (MSYS2 fix).

GitMensch:
Ci is running again - so back to the merge :-)

ddeclerck:
> Ci is running again - so back to the merge :-)

Yay ! ;)

PR #217, opened 2025-02-11
Fix MSYS2 CI

ddeclerck:
After 8 hours trying to understand what was going on with texi2dvi & texinfo, I resorted to a somewhat hackish `export OSTYPE=msys`, which happened to solve the problem (there is apparently an issue with how texi2dvi deals with the separator (; or :) in the TEXINPUTS variable). Weird that this only occurs now ; probably something changed in MSYS2...

### DIFF
--- a/.github/workflows/windows-msys2.yml
+++ b/.github/workflows/windows-msys2.yml
@@ -40,15 +40,18 @@ jobs:
       - name: Configure GnuCOBOL
         shell: msys2 {0}
         run: |
-          mkdir _build
-          cd _build
-          ../configure --without-db --without-curses --without-xml2 --without-json
+          mkdir _build && cd _build
+          # this is a "prepare" only step for this workflow only, so build
+          # without any optional packages as one-time build speeding up this part
+          ../configure --without-db --without-curses --without-xml2 --without-json \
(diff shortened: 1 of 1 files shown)


GitMensch:
result seems to work (note: I've tried to fix this the same time, I think the real root cause is mixing msys with mingw64; if reducing to msys #216 works that seems preferable)

ddeclerck:
> I think the real root cause is mixing msys with mingw64; if reducing to msys #216 works that seems preferable

I had the same idea, but instead tried using the mingw64 version of all tex-related packages - that wasn't very successful.

Now I can see there are also 3 failures on mingw32... I really wonder what they could possibly have changed in MSYS2 and/or the Windows runner image...

GitMensch:
Note: when adjusting this, you may change the prepare configure line to

```sh
          mkdir _build && cd _build
          # this is a "prepare" only step for this workflow only, so build
          # without any optional packages as one-time build speeding up this part
          ../configure \
              --without-db --without-curses --without-xml2 --without-json --without-iconv \
              --disable-dependency-tracking
```

ddeclerck:
Seems to work when also defining `OSTYPE=msys` when running the testsuite.

PR #223, opened 2025-04-02
Fix depends

ddeclerck:
See [this comment](https://github.com/OCamlPro/gnucobol/pull/109#issuecomment-2758285450) from PR #109 as well a as the comments below [this one](https://github.com/OCamlPro/gnucobol/pull/147#pullrequestreview-2721662495) from PR #147 .


cobc/cobc.c
```diff
@@ -4640,7 +4640,8 @@ process_filename (const char *filename)
 		fn->preprocess = cobc_main_strdup (output_name);
 	} else
 	if (save_all_src || save_temps
-	 || cb_compile_level == CB_LEVEL_PREPROCESS) {
+	 || (cb_compile_level == CB_LEVEL_PREPROCESS
+	 && !cb_depend_output_only)) {
```
GitMensch, line 4644: in this case please make that a single-line expression to have the indent at the || (or indent after its ( and on the next line &&,  but I think the one-liner is clearer to grasp here

GitMensch (changes requested):
please add Changelog and test cases for the fixed issues

GitMensch (approved):
LGTM as is, but we may adjust the doc/help for -fcopybook-deps; If I see it correctly that's for external tools, but not make, but it still only works on dependency file generation for make?!?

Do we really need that for anything? Possibly for copybooks including copybooks, but then it should be likely a different file listing "prog.cob" instead of "prog.o" (=assignment, not target); but again, I'm not sure what we want that for.

ddeclerck:
As far as Fabrice remembers, the difference is that `-fcopybook-deps` returns a list of unresolved dependencies, while `-M` returns the list of resolved dependencies (i.e. considers the include directories). This is to allows external tools to know the unresolved copybook names, possibly trying to resolve them themselves if they want to.

ddeclerck:
> That would be what `-MG` is for... (which can check the dependencies) - obviously that cannot know about an added extension and would take the filenames "as is".

My wording was unclear. The two options are a bit different:
`-fcopybook-deps` returns *all* dependencies as-is without even trying to resolve them.
`-MG` returns the unresolved dependencies as-is (mostly, as it adds a .CPY extension), in addition to already resolved dependencies.

GitMensch:
If you can share more details about this, we possibly can keep that in.
If not - because of the contributing history - I'd be OK to place the related code (flag, help, and with the flag also its code, as well as moving the docs [NEWS, gnucobol.texi] and tests to comments) in an `#ifdef EXPERIMENTAL_COPYBOOK_DEPS_OPTION` block for a later review, which still allows a "local" build (and of course OCamlPro provided binaries) enabling it with `CPPFLAGS=-DEXPERIMENTAL_COPYBOOK_DEPS_OPTION`.

ddeclerck:
> If you can share more details about this, we possibly can keep that in. If not - because of the contributing history - I'd be OK to place the related code (flag, help, and with the flag also its code, as well as moving the docs [NEWS, gnucobol.texi] and tests to comments) in an `#ifdef EXPERIMENTAL_COPYBOOK_DEPS_OPTION` block for a later review, which still allows a "local" build (and of course OCamlPro provided binaries) enabling it with `CPPFLAGS=-DEXPERIMENTAL_COPYBOOK_DEPS_OPTION`.

That does not seem to be very clear yet, so I went with putting this behind an `#ifdef`.

GitMensch:
LGTM, the following may be open:

> For `-MG`: if `-ext` was used it should be taken as extension (either the first or the last invocation - I'd leave that to you), with the default for `-MG` if ` -ext` was _not passed_ being CPY. I very likely have overlooked (=not checked) that for the first MG commit.

ddeclerck:
> > and the two copybooks `stuff1.CPY` and `stuff2.XYZ` (`stuff3` is intentionally missing). Then, the command `cobc -M -MG test.cob` yields:
> > ```
> > test.o: \
> >  test.cob \
> >  stuff1.CPY \
> >  stuff2.CPY \
> >  stuff3.CPY
> > ```
> 
> Then this is an error, it should have stuff2.XYZ if that exists and stuff2.CPY doesn't
> 
> ?
This result is without `-ext`, `stuff2.XYZ` exists and `stuff2.CPY` does not. I don't see how `stuff2.XYZ` could be found.

GitMensch:
Ah, that's fine then -> another for upstream.
Thanks for finishing that (please add an entry on EXPERIMENTAL_COPYBOOK_DEPS_OPTION in cobc/Changelog during upstream, no need for a push to GH up-front)

ddeclerck:
Merged in SVN @ 5489.

PR #224, opened 2025-04-04
Fix test 31

ddeclerck:
This PR attemps to fix test 31, which fails randomly.

It seems the newline appearing at different positions is not caused by libcob nor the COBOL test program. It might be caused by the other commands invoked by the test (sleep/kill). Redirecting the stderr of the COBOL program to a file and checking that against the expected stderr does seem to solve the problem.

### DIFF
--- a/tests/testsuite.src/used_binaries.at
+++ b/tests/testsuite.src/used_binaries.at
@@ -1432,20 +1432,25 @@ AT_DATA([prog.cob], [
 AT_CHECK([$COMPILE prog.cob], [0], [], [])
 
 # note: the stack trace, while enabled per default, is globally disabled in the testsuite
-AT_CHECK([$COBCRUN_DIRECT ./prog & cobpid=$!
+AT_CHECK([$COBCRUN_DIRECT ./prog 2>stderr1 & cobpid=$!
 sleep 6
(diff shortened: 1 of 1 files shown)


GitMensch (approved):
That would explain why I never was able to reproduce that outside of the testsuite... If you think that works, then please check that in upstream!

Note: while you will likely have a good svn log message explaining this commit, it seems useful to also leave a bit of an m4 comment in that test case (doing in upstream commit is enough). I'm dropping the "disable this check" part which I've added to Appveyor CI definition, so please commit "soon".

ddeclerck:
> That would explain why I never was able to reproduce that outside of the testsuite... If you think that works, then please check that in upstream!

Can't be 100 % sure, but I'd say 99 % ;)
Took a while to figure out what was happening.

> Note: while you will likely have a good svn log message explaining this commit, it seems useful to also leave a bit of an m4 comment in that test case (doing in upstream commit is enough). I'm dropping the "disable this check" part which I've added to Appveyor CI definition, so please commit "soon".

Indeed, a comment would be useful, to avoid our future selves "optmizing" this away 😅

ddeclerck:
Merged in SVN @ 5473 (will likely add it to the 3/4 merge right away)

PR #226, opened 2025-04-22
[GC3] Update CIs

ddeclerck:
This PR updates the GC3 CI to make it more similar to the GC4 CI.
It also adds a Clang build on Ubuntu.

Note: as for GC4, MSYS2 builds are now broken because of an update to the diff package...

### DIFF
--- a/.github/workflows/windows-msys2.yml
+++ b/.github/workflows/windows-msys2.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   prepare:
     runs-on: windows-latest
-    timeout-minutes: 30
+    timeout-minutes: 10
 
     steps:
 
@@ -26,16 +26,16 @@ jobs:
         uses: msys2/setup-msys2@v2
         with:
           update: true
-          msystem: mingw64
-          install: autoconf automake libtool make flex bison help2man texinfo texinfo-tex
-                   mingw-w64-x86_64-cc
-                   mingw-w64-x86_64-gmp mingw-w64-x86_64-texlive-core
-                   mingw-w64-x86_64-gettext-runtime gettext-devel
+          msystem: ucrt64
+          install: autoconf automake make libtool flex bison
+                   gettext gettext-devel help2man texinfo texinfo-tex
+                   mingw-w64-ucrt-x86_64-texlive-core
+                   mingw-w64-ucrt-x86_64-cc
+                   mingw-w64-ucrt-x86_64-gmp
 
       - name: Bootstrap GnuCOBOL
         shell: msys2 {0}
-        run: |
-          ./build_aux/bootstrap install
+        run: ./build_aux/bootstrap install
 
       - name: Configure GnuCOBOL
         shell: msys2 {0}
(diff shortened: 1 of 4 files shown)


GitMensch:
Is there an upstream msys2 issue about diffutils?

ddeclerck:
> Is there an upstream msys2 issue about diffutils?

Nope, and actually, the same diff version works fine locally on our Windows machine... And it does work in the "prepare" step of the MSYS workflow. I'd like to be sure the problem comes from diff.

GitMensch:
Is there a reason that we don't have a V[B]ISAM environment, yet - or was it only because of "not setup so far"?
Note: GC41 has VISAM and we possibly could take that over to 3.x.

ddeclerck:
> Is there a reason that we don't have a V[B]ISAM environment, yet - or was it only because of "not setup so far"?

"Not setup so far" would probably be the closest reason; another would be that I'm juggling too many things at once.
Also note that that there are some (3~4) tests that fail with V(B)ISAM under GC3.

GitMensch:
I am aware of the unexpected passes / failures with VBISAM, for the CI we can adjust the results as we did for MSYS.
Thank you for inspecting that one day or another.

.github/workflows/windows-msys2.yml
```diff
@@ -78,93 +81,89 @@ jobs:
...
 
       - name: Configure GnuCOBOL
         shell: msys2 {0}
         run: |
-          tar -xvf gnucobol*.tar.*  --strip-components=1
-          mkdir _build
-          cd _build
-          ../configure $CFGOPT
+          mkdir _build && cd _build
+          PATH="$PATH:/c/Program Files/Git/usr/bin" ../configure $CFGOPT CFLAGS=--std=c17
```
GitMensch, line 138: we should add a note here why PATH is suffixed (... to allow configure finding a "diff" that works, to work around MSYS diffutils issues)

GitMensch, line 138: we should add a note here why PATH is suffixed (... to allow configure finding a "diff" that works, to work around MSYS2 issues); note: I've checked on a fresh and updated MSYS2 installation and there

```sh
diff fx1 fx2 >/dev/null ; res0=$? ; diff fx1 fy >/dev/null ; res1=$?
echo $res0 $res1
```

resulted in the expected "0 1" output in each of msys, mingw64 and mingw32 shells?!?

.github/workflows/windows-msys2.yml
```diff
@@ -78,93 +81,89 @@ jobs:
...
+            echo CFGOPT="--with-pkgversion=GnuCOBOL-CI-MSYS2-Debug --enable-debug --enable-cobc-internal-checks --enable-hardening" >> $env:GITHUB_ENV
           }
           # Common options (dependencies)
           echo CFGOPT="$env:CFGOPT --with-math=gmp --with-curses=ncursesw --with-xml2 --with-json=json-c" >> $env:GITHUB_ENV
           # Conditional inclusion of --with-db (MSYS2 does not provide 32bit builds for BDB any more)
-          if ("${{ matrix.sys }}" -ne "mingw32") {
+          If ("${{ matrix.sys }}" -ne "mingw32") {
             echo CFGOPT="$env:CFGOPT --with-db" >> $env:GITHUB_ENV
-          } else {
+          } Else {
```
GitMensch, line 111: Just out of interest... What was the reason to capitalize if/else?

GitMensch (approved):
LGTM, possibly there is an option for minor adjustments, see below.

.github/workflows/windows-msys2.yml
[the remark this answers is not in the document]
ddeclerck: Tried it (in a separate branch), that did not seem to change anything.

ddeclerck: > resulted in the expected "0 1" output in each of msys, mingw64 and mingw32 shells?!?

Yes, as I mentionned in some other comment (don't remember where), diff works fine on a local MSYS2 installation. It also works fine in the "prepare" step of the workflow, but not in the further steps - and I spent hours trying to figure out why, to no avail.

.github/workflows/windows-msys2.yml
ddeclerck: Oh it is just to match what we have in the GC4 CI. I could as well do the opposite and change the GC4 to lowercase. I don't know what's the usual convention for PowerShell (have seen both used online).

PR #229, opened 2025-05-12
Fixes for C23

ddeclerck:
This PR attempts to fix GnuCOBOL for C23, by adding appropriate function type casts, both in libcob and codegen.

There's still one test failing, due to multiple definition of the same function prototype (I guess this was working before C23 because the first prototype was incomplete ; now it includes all the arguments). Investigation led me to line 1908 in `codegen.c:output_call_cache`:
```
			output_local ("#ifndef %s\n", static_call->call_name);
```
Apparently the intent was to use an `#ifndef` to prevent multiple definition of the same symbol... yet preprocessor directives may not be used to test whether a function is defined...


ddeclerck:
In order to fix the last failing test, I was able to skip generating the "default" prototype when an actual prototype is provided in the COBOL source. However, the test fails later, when testing the >>IMP directive: since this may add function prototypes, they may clash with the ones inferred for static calls, and we have no way to detect that...

libcob/call.c
```diff
@@ -458,7 +458,7 @@ do_cancel_module (struct call_hash *p, struct call_hash **base_hash,
 #pragma warning(suppress: 4113) /* funcint is a generic function prototype */
 	cancel_func = p->module->module_cancel.funcint;
 #else
```
GitMensch, line 460: that pragma should go away and still work fine with the new code

libcob/call.c
```diff
@@ -1290,7 +1290,7 @@ cob_cancel_field (const cob_field *f, const struct cob_call_struct *cs)
 #pragma warning(suppress: 4113) /* funcint is a generic function prototype */
 				cancel_func = s->cob_cstr_cancel.funcint;
 #else
-				cancel_func = s->cob_cstr_cancel.funcint;
+				cancel_func = (int (*)(const int,  void *, void *, void *, void *))s->cob_cstr_cancel.funcint;
 #endif
 				(void)cancel_func (-1, NULL, NULL, NULL,
 						   NULL);
```
GitMensch, line 1296: I have no clue why the cancel_func is setup to have 4 NULL, especially if we have a program with no USING then the cancel_func has only a single parameter `const int entry`, no?
Telling C explicit how many arguments to handle and then ask it to handle too much seems to ask for trouble...

GitMensch (changes requested):
I'm unsure how the code in call.c should work, even more if we don't use the deprecated and in C2x forbidden variable arguments option.

It seems that the best option is to pass only the amount of arguments in a way we can query from the loaded program, with an exported symbol in the shared object... but even then we would need for example libffi (we could include the sources and statically link that per default as GCC does) to do a dynamic call [and if we don't get the number of arguments from the callee, we could use the argc).

... I feel a bit stuck here (still left some comments).

Thoughts?

ddeclerck:
> I'm unsure how the code in call.c should work, even more if we don't use the deprecated and in C2x forbidden variable arguments option.
> 
> It seems that the best option is to pass only the amount of arguments in a way we can query from the loaded program, with an exported symbol in the shared object... but even then we would need for example libffi (we could include the sources and statically link that per default as GCC does) to do a dynamic call [and if we don't get the number of arguments from the callee, we could use the argc).
> 
> ... I feel a bit stuck here (still left some comments).
> 
> Thoughts?

Well I'm missing the "big picture" about how GnuCOBOL handles calling in general, so I might not see the implications, but to me, I don't see much differences between:
- calling without casting a function declared without a prototype
and
- calling with casting a function declared with a prototype 
provided that, in both cases, the actual number and types of arguments passed match the actual number and types of arguments expected by the function (otherwise we're screwed because this is undefined behavior... so the two methods are similarly unsafe).

So when we get to call a function whose prototype we don't know, let's just declare it with a dummy type, and cast it at call site according to the arguments that are passed (and in fact that's more or less what's done in this PR).

But really, users should define proper prototypes...

GitMensch:
> cast it at call site according to the arguments that are passed

That would be the wrong approach. It should be called according to the prototype. Note that cobc deduces the prototype when it processes a program during compilation and if it cannot deduce anything by that, then it still can deduce an internal prototype by the actual CALL statement(s).
I'm not sure how far into codegen this goes, but in general "it is there".

Agreed that the `cob_call()` part - which is "hairy" in any case can be changed to the new version (per gentoo patch and this PR), because that's "just a _possibly_ different undefined behaviour than before".
For the parts that use the cancel func - please adjust it to always use a single `const int` - currently it is different in call.c, common.c (and potentially others), as well as adjusting the `ret_ptr` stuff ... and adding Changelog entries.

The codegen part is something I need to inspect again (did so longer today) - maybe you can have a look how that aligns to the internal prototypes that are for example used in the compile-time checks?

What does the `COB_SYSTEM_GEN` change do exactly?

ddeclerck:
> > cast it at call site according to the arguments that are passed
> 
> That would be the wrong approach. It should be called according to the prototype.

Of course I was only suggesting this for those functions where we don't have any prototype nor we can not infer any.

> Note that cobc deduces the prototype when it processes a program during compilation and if it cannot deduce anything by that, then it still can deduce an internal prototype by the actual CALL statement(s). I'm not sure how far into codegen this goes, but in general "it is there".

> Agreed that the `cob_call()` part - which is "hairy" in any case can be changed to the new version (per gentoo patch and this PR), because that's "just a _possibly_ different undefined behaviour than before". For the parts that use the cancel func - please adjust it to always use a single `const int` - currently it is different in call.c, common.c (and potentially others), as well as adjusting the `ret_ptr` stuff ... and adding Changelog entries.

Will do.

> The codegen part is something I need to inspect again (did so longer today) - maybe you can have a look how that aligns to the internal prototypes that are for example used in the compile-time checks?

Codegen remains tricky, I'm really not confident about this code yet - and I still have one failing test. I need more time to dig into that.

> What does the `COB_SYSTEM_GEN` change do exactly?

It changes the single `void *` argument to a plain `void`, so it makes the function not expecting any argument. I haven't checked further why this was needed - just took this from the Gentoo patch.

GitMensch:
All parts but cobc were adjusted upstream and that seems to build fine with gcc-15 now - but fails a bunch of internal checks, which will likely be solved by the changes here; please rebase.

Note: I'm not sure that casting the funcint in cob_call_field to all possible argument numbers is really a good idea... I _guess_ it may "crash harder" in the case that the function does not take any arguments than specifying none explicit... on some architectures.

GitMensch:
@ddeclerck Thanks for finishing that!

ddeclerck:
> _If_ the user provides a prototype, then it should be "correct". If the user make use of `-fno-gen-c-decl-static-call` then we shouldn't generate them - which is still the case, no?

I had completely forgotten about the `-fno-gen-c-decl-static-call` option. Adding this to the offending test works fine.

Still have to tackle the other comments.

PR #230, opened 2025-05-16
Remove gcdiff and adjust testsuite

ddeclerck:
This PR removes gcdiff and adjusts the testsuite to accomodate for the changes.


GitMensch:
That approach is good, but let's try to get rid of those sed invocations (or make them simpler/less used, for example with SOURCE_DATA_EPOCH) in 3.x first, then merge the change over (I think most of the effected lines will use gcdiff in 4.x).

Oh, and another note: you don't need to use the underscore-sed hack, it may be much easier to use `[[some data    ]]` instead of the current `[some data    _]` (I just learned that too late, and seemingly haven't adjusted all the tests that expect trailing spaces.

Interesting to see that this PR does uncover some tests that have a different result and was not recognized before because of the use of gcdiff.

ddeclerck:
> That approach is good, but let's try to get rid of those sed invocations (or make them simpler/less used, for example with SOURCE_DATA_EPOCH) in 3.x first, then merge the change over (I think most of the effected lines will use gcdiff in 4.x).

I'd do that when time permits.

> Oh, and another note: you don't need to use the underscore-sed hack, it may be much easier to use `[[some data ]]` instead of the current `[some data _]` (I just learned that too late, and seemingly haven't adjusted all the tests that expect trailing spaces.

Ah, I though those underscores were just a visual reminder that the trailing spaces are intentional.

> Interesting to see that this PR does uncover some tests that have a different result and was not recognized before because of the use of gcdiff.

I was as surprised as you are.

GitMensch:
... waits for rebase ...
in any case please also update the references: NOTES + TODO + AUTHORS (side note: feel free to open a PR against 3.x for adding OCamlPro / the contributors, if you see any elements missing) and add Changelog entries for all of the three sub-folders [bin + build_windows may need the (C)2025 update in the footer when doing so].

ddeclerck:
Rebased. Still a bit of work to do.

On test in `syn_file.at` (XFD directive and creation) revealse that the function `output_xfd_file` (sqlxfdgen.c) does not use `SOURCE_DATE_EPOCH`. Should it do so ?

> in any case please also update the references: NOTES + TODO + AUTHORS

What should be changed in NOTES ?

GitMensch:
NOTES is a document we'll handle berfore handling the TODO - you can just delete any elements related to gcdiff in this file as part of this PR (less to glance over).

And yes: SOURCE_DATE_EPOCH _must_ be handled in everything that is generated by the compiler by using the central general time routines that are used otherwise within cobc for timestamps; it _may_ should use the same time than the generated .c file, but that's up to the fixing person (= you :-).
Fixing SOURCE_DATE_EPOCH and the syn_file.at should really be a separate commit - feel free to have both commits (cleanup + fix output_xfd_file) in this PR, or split into two PRs - as long as it ends up with two commits in trunk upstream I don't mind.

tests/testsuite.src/syn_file.at
```diff
@@ -3134,7 +3135,7 @@ AT_CHECK([$DIFF reference tspfilex.ddl], [0], [], [])
 AT_CAPTURE_FILE(./tspfilex.xd)
```
GitMensch, line 3135: while this is unrelated, maybe you can fix that line in this commit as well - the .xd file is expected to be there even if the ddl file check did not work, so it should be moved directly after capture of the ddl (if you do that, it is enough to do it on upstreaming)

tests/testsuite.src/run_file.at
```diff
@@ -6573,7 +6573,8 @@ AT_CHECK([$COMPILE prog.cob], [0], [], [])
 AT_CHECK([COB_CURRENT_DATE="2015/02/06 16:40:52" $COBCRUN_DIRECT ./prog], [0],
 [Normal termination, ending status: 10
 ], [])
-AT_CHECK([gcdiff mini-report reference-report], [0], [], [])
+AT_CHECK([$SED -e 's/[[[:blank:]]][[[:blank:]]]*$//' mini-report > mini-report.sed], [0], [], [])
```
GitMensch, line 6576: What is this needed for? If it is indeed, then please add a comment why (following the idea that we try to always compare what we produce, not only "something similar"), otherwise you may be able to use trigraph or otherwise fix that.

GitMensch (changes requested):
both commits are _nearly_ fine, please adjust (and feel free to do that in the upstream commit directly, if you want to)

tests/testsuite.src/run_file.at
ddeclerck: This comes from a difference between 3.x and trunk. In 3.x the produced output does not have any trailing spaces, while in trunk it does (padding to column 80 or so).

I could add this padding (and the `@&t@` quadrigraph) to each line of the expected output (reference-report), but I thought it was better to keep it identical to the one in 3.x. Hence I stripped those trailing spaces in the produced output instead.

GitMensch: Let's put that to the TODO file and add a summary "this is a difference, for now we change it to 3.x format for comparison purpose" (maybe a _bit_ more in two lines) to the test source - and keep the SED.

Thanks!

GitMensch (approved):

ddeclerck:
Merged in SVN @ 5526 and 5527.

ddeclerck:
> Test failures on 32bit - so we either need a sed script or a conditional test here

Ah, I knew I had forgotten something.
I'll go for the same sed workaround as one of the tests above.

GitMensch:
sure (for GC3 obviously without the odbc/oci part and for configure using `--without-db`) [I'm back in around 3 hours for a possible review, but if you have a working CI only change, then just commit that...]

ddeclerck:
Fix merged in SVN @ 5535.

PR #231, opened 2025-05-19
Limit sed usage

ddeclerck:
This PR attemps to reduce `sed` usage by:

1. using `SOURCE_DATE_EPOCH` to have predictable dates in expected output
2. using `@&t@` quadrigraph to preserve trailing spaces in expected output (note: using `[[...]]` did not work to preserve these spaces)

Note that this uncovered a bug with `SOURCE_DATE_EPOCH` handling when multiple input files are provided to `cobc`. This PR includes a fix for this.


GitMensch (commented):
That looks quite good and I've learned something new today https://www.gnu.org/software/autoconf/manual/autoconf-2.62/html_node/Quadrigraphs.html as well.

Obviously testing SOURCE_DATE_EPOCH indirectly (please add SOURCE_DATE_EPOCH into the keywords of the testcase that now uses this most and would have shown the problem, leaving a comment under the keywords that we use this feature for the time output in the listing and indirectly test it that way.

But for this PR to be really reasonable we'd want tests/listings-sed.sh to be gone and listings.at be adjusted (or at least simplified, as this is the main place where gcdiff was used at the start)

Thank you for having a look at that.

ddeclerck:
> That looks quite good and I've learned something new today https://www.gnu.org/software/autoconf/manual/autoconf-2.62/html_node/Quadrigraphs.html as well.

Learnt that as well today ;)

> Obviously testing SOURCE_DATE_EPOCH indirectly (please add SOURCE_DATE_EPOCH into the keywords of the testcase that now uses this most and would have shown the problem, leaving a comment under the keywords that we use this feature for the time output in the listing/dump and indirectly test it that way as well).

Will do.

> But for this PR to be really reasonable we'd want tests/listings-sed.sh to be gone and listings.at be adjusted (or at least simplified, as this is the main place where gcdiff was used at the start)

Actually I was working on that before seing your comment - check the updated PR.

> Thank you for having a look at that.

No problem ;)

ddeclerck:
Oh, just realizing something must be wrong.
I set the date to Unix epoch 1011960000, which should be the 25th of January, yet GnuCOBOL interprets it as the 24th of January...

tests/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2025-05-19  David Declerck <david.declerck@ocamlpro.com>
+
+	* testsuite.src/run_misc.at, testsuite.src/syn_misc.at:
+	  reduce the use of sed by using SOURCE_DATE_EPOCH when possible
+	  and using @&t@ quadrigraphs in expected output with trailing spaces
+	* listing-sed.sh: removed as no longer useful
```
GitMensch, line 7: be a bit less harsh s'/useful/needed/'

GitMensch (approved):
Nice one! It will be interesting to see what's possibly missing for GC4 after merging that.

tests/ChangeLog
ddeclerck: Done.

GitMensch:
Can you please fix that fix soon? The examples show that there's a bug somewhere, because it sows Saturday while the correct output would be Friday (rest seems fine). Looks like the weekday is possibly incremented twice somewhere?

ddeclerck:
> Can you please fix that fix soon? The examples show that there's a bug somewhere, because it sows Saturday while the correct output would be Friday (rest seems fine). Looks like the weekday is possibly incremented twice somewhere?

I'm on it.
I was so focused on the incorrect month day that I did not realize I broke the week day while fixing it...

GitMensch:
Can you please include an adjustment from Chuck he just sent me?

He pointed out that my original code is dumb (he put it different, of course, but that's what it was) - time_t is defined as "absolute value from epoch", which is what we want here, so "seconds" can be dropped completely and instead `t` be used, dropping the whole manual adjustment after localtime; I _think_ this would also fix the "original problem", in which case we can replace the 2025-05-19 entry by Chucks change and have todays for your fix.

... I really screwed this function up - but thoroughly :-)

ddeclerck:
> Can you please include an adjustment from Chuck he just sent me?
> 
> He pointed out that my original code is dumb (he put it different, of course, but that's what it was) - time_t is defined as "absolute value from epoch", which is what we want here, so "seconds" can be dropped completely and instead `t` be used, dropping the whole manual adjustment after localtime; I _think_ this would also fix the "original problem", in which case we can replace the 2025-05-19 entry by Chucks change and have todays for your fix.
> 
> ... I really screwed this function up - but thoroughly :-)

It would seems it suffice to replace most of the function body with a call to `set_cob_time_from_localtime`.
However, since this is using `localtime`, results may still vary depending on the time zone...
Would it be okay for your if I add `TZ=UTC` in the testsuite for each test that uses `SOURCE_DATE_EPOCH` ?

GitMensch:
yes, the setting of localtime can be simplified as pointed out by Chuck, and the setting of cob_time can then also be simplified...
Please give it a go - and the TZ should be exported in atlocal* where we override LC_ALL/LANG

ddeclerck:
Seems this is working fine now.
It would be tempting to also add `SOURCE_DATE_EPOCH ` to `atlocal*`... or not ?

GitMensch:
No, because we do normally want to check both variants and if one of the other tools in the call process does something unexpected with it (or our code get broken again), then only the specific tests including it will fail, but the last commit in this PR is fine for upstream.

ddeclerck:
Fix merged @ SVN 5531.

libcob/common.c
```diff
@@ -5025,36 +5023,8 @@ cob_set_date_from_epoch (struct cob_time *cb_time, const unsigned char *p)
...
-	tmptr->tm_mday = (int)seconds;
-	tmptr->tm_isdst = -1;
-
-	/* normalize if needed (definitely for epoch, but also for example 30 Feb
-		to be changed to correct march date),
-		set tm_wday, tm_yday and tm_isdst */
-	if (mktime (tmptr) == -1) {
-		return 1;
-	}
+        set_cob_time_from_localtime ((time_t)seconds, cb_time);
```
GitMensch, line 5026: the differences are:

nanosecond  -1 -> 0; while I think that's better for cobc because of (cb_set_intr_when_compiled), it isn't for libcob - so please set it to -1 after set_cob_time_from_localtime here and set it to zero after calling cob_set_date_from_epoch in set_compile_date.

offset_known + utc_offset set; I think the first should be set to zero here as well

GitMensch (changes requested):

libcob/common.c
GitMensch: even different - to following https://reproducible-builds.org/docs/timezones/ we should do it different than before and _always_ do `offset_known = 1; utc_offset = 0;` after calling `set_cob_time_from_localtime` here

ddeclerck: Made the changes in the last commit.

GitMensch: thanks, that looks good to me and it should help for reproducibility and the nano=0 may fix an overflow as well

PR #232, opened 2025-05-21
Add 32-bit Ubuntu workflow

ddeclerck:
This PR adds a 32-bit Ubuntu CI.

This *almost* works, as there is still one failure:
```
## -------------------------------------------- ##
## GnuCOBOL 3.3-dev test suite: GnuCOBOL Tests. ##
## -------------------------------------------- ##
1190. run_extensions.at:2172: testing NUMBER-OF-CALL-PARAMETERS ...
../../tests/run_extensions.at:2210: $COMPILE caller.cob
../../tests/run_extensions.at:2213: $COMPILE_MODULE -fusing-optional=skip callee.cob
../../tests/run_extensions.at:2214: $COBCRUN_DIRECT ./caller
/home/runner/work/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/1190/test-source: line 242: 73572 Segmentation fault      (core dumped) ( $at_check_trace; $COBCRUN_DIRECT ./caller ) >> "$at_stdout" 2>> "$at_stderr" 5>&-
--- /dev/null	2025-05-21 17:38:58.198580733 +0000
+++ /home/runner/work/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/1190/stderr	2025-05-21 17:43:38.941743321 +0000
@@ -0,0 +1,3 @@
+
+caller.cob:12: attempt to reference invalid memory address (signal SIGSEGV)
+
--- -	2025-05-21 17:43:39.021375093 +0000
+++ /home/runner/work/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/1190/stdout	2025-05-21 17:43:39.017744122 +0000
@@ -1,7 +1,2 @@
 +000000000
-+000000001
-+000000002
-+000000003
-+000000004
-+000000004
 
../../tests/run_extensions.at:2214: exit code was 139, expected 0
1190. run_extensions.at:2172:  FAILED (run_extensions.at:2214)```

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -336,7 +336,93 @@ jobs:
             _build/tests/cobol85/**/duration.txt
 
 
-  minmal_build:
+  build_32_bit:
+    name: Build and test on 32-bit
+    needs: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: true
+
+    steps:
+
+      - name: Get CI dist tarball
+        uses: actions/download-artifact@v4
+        with:
+          name: gnucobol-ci source distribution
+
+      - name: Extract CI dist tarball
+        run: tar -xvf gnucobol*.tar.* --strip-components=1
+
+      - name: Setup build environment
+        run: echo "TERM=vt100" >> $GITHUB_ENV
+
+      - name: Install packages
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt -y update
+          sudo apt -y install make libtool gcc-9-multilib \
+                              gcc-9 libgmp-dev:i386 libncurses-dev:i386 \
+                              libdb5.3-dev:i386 libxml2-dev:i386 libcjson-dev:i386
+          sudo apt -y remove flex bison
+
+      - name: Configure GnuCOBOL
+        run: |
+          mkdir _build && cd _build
+          ../configure --with-db \
+                       --enable-cobc-internal-checks \
+                       --enable-hardening \
+                       --prefix=/opt/gnucobol \
+                       CC="gcc-9 -m32" \
+                       CFLAGS="-fno-tree-coalesce-vars"
+          # Note: we pass -fno-tree-coalesce-vars to workaround a bug
+          # occurring with gcc on x86 at optimization level -O1
+
+      - name: Build GnuCOBOL
+        run: make -C _build --jobs=$(($(nproc)+1))
+
+      - name: Upload config.log
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: config-${{ github.job }}.log
+          path: _build/config.log
+
+      - name: Run testsuite
+        run: |
+          make -C _build check TESTSUITEFLAGS="--jobs=$(($(nproc)+1))" || \
+          make -C _build check TESTSUITEFLAGS="--recheck --verbose"
+
+      - name: Upload testsuite.log
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: testsuite-${{ github.job }}.log
+          path: _build/tests/testsuite.log
+
+      - name: Cache newcob.val
+        uses: actions/cache@v4
+        with:
+          path: _build/tests/cobol85/newcob.val
+          key: newcob-val
+          enableCrossOsArchive: true
+
+      - name: Run NIST85 testsuite
+        run: make -C _build test --jobs=$(($(nproc)+1))
+
+      - name: Upload NIST85 Test Suite results
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: NIST85 results on ${{ github.job }}
+          path: |
+            _build/tests/cobol85/summary.*
+            _build/tests/cobol85/**/*.log
+            _build/tests/cobol85/**/*.out
+            _build/tests/cobol85/**/duration.txt
+
+
+  build_minimal:
     name: Build and test with minimal dependencies
     needs: build
     runs-on: ubuntu-latest

### CODE REVIEW

**Comment by GitMensch** [comment]
hm, can you reproduce the error by

```
cd tests
. atconfig && . atlocal
gdb -x run testsuite.dir/1190/caller
```
?

File: .github/workflows/build_nightly.yml [hunk 1]
```diff
@@ -336,7 +336,90 @@ jobs:
...
+          sudo apt -y remove flex bison
+
+      - name: Configure GnuCOBOL
+        run: |
+          mkdir _build && cd _build
+          ../configure --with-db \
+                       --enable-cobc-internal-checks \
+                       --enable-hardening \
+                       --prefix=/opt/gnucobol \
+                       CC="gcc -m32"
```
GitMensch, line 376: Would it be useful to install and use an older gcc here as well?

GitMensch (commented):
LGTM

ddeclerck:
A> hm, can you reproduce the error by
> 
> ```
> cd tests
> make check TESTSUITEFLAGS="--recheck --verbose" COBOL_FLAGS=-g
> . atconfig && . atlocal
> cd testsuite.dir/1190
> gdb -ex run -ex bt ./caller
> ```

Well, passing `-g` makes the error disappear, so I went for `COBOL_FLAGS="-save-temps -A -ggdb"`.
Interrestingly, the error only occurs when configure is called with `--enable-hardening`.

Anyways, here's the backtrace I obtained:
```
0x5655643e in caller_ (entry=0) at caller.c:126
126	  cob_glob_ptr->cob_call_params = 1;
#0  0x5655643e in caller_ (entry=0) at caller.c:126
#1  0x56557b1d in caller () at caller.c:51
#2  0x56557b4d in main (argc=1, argv=0xffffbe64) at caller.c:33
```

And guess what, for some reason, `cob_glob_ptr` is NULL here. So, I'll go check why this happens. More info to come.

ddeclerck:
Something must be very wrong...

In caller.c, line 110, we have:
```
  /* Line: 11        : CALL               : caller.cob */
  module->statement = STMT_CALL;
  cob_glob_ptr->cob_call_params = 0;
  cob_glob_ptr->cob_stmt_exception = 0;
  if (unlikely((cob_glob_ptr->cob_exception_code & 0x0b00) == 0x0b00)) cob_glob_ptr->cob_exception_code = 0;
    if (unlikely(call_callee.funcvoid == NULL || cob_glob_ptr->cob_physical_cancel == 1))
  {
    call_callee.funcvoid = cob_resolve_cobol ("callee", 0, 1);
  }
  b_2 = ((int (*)(void))call_callee.funcint) ();
  cob_check_fence (call_fence_pre, call_fence_post, STMT_CALL, NULL);
```

Before `b_2 = ((int (*)(void))call_callee.funcint) ();`, `cob_glob_ptr` is set. After, it becomes 0.

So I followed the call to `callee`:
```
int
callee (cob_u8_t *b_17, cob_u8_t *b_18, cob_u8_t *b_19, cob_u8_t *b_20)
{
  cob_u8_t *cob_parm_17 = NULL;	/* linkage for W */
  cob_u8_t *cob_parm_18 = NULL;	/* linkage for X */
  cob_u8_t *cob_parm_19 = NULL;	/* linkage for Y */
  cob_u8_t *cob_parm_20 = NULL;	/* linkage for Z */

  /* Get current number of call parameters,
     if the parameter count is unknown, set it to all */
  if (cob_get_global_ptr ()->cob_current_module) {
  	cob_call_params = cob_get_global_ptr ()->cob_call_params;
  } else {
  	cob_call_params = 4;
  }

  /* Set the parameter list */
  if (cob_call_params > 0) {
    cob_parm_17 = b_17;
  }
  if (cob_call_params > 1) {
    cob_parm_18 = b_18;
  }
  if (cob_call_params > 2) {
    cob_parm_19 = b_19;
  }
  if (cob_call_params > 3) {
    cob_parm_20 = b_20;
  }
```

Stepping over the instructions, I noticed the block `if (cob_call_params > 1) { ...` is skipped... And it's exactly at this moment that `cob_glob_ptr` turns NULL (confirmed by a watchpoint). I don't see how the lines around that point could affect that value O_O.

ddeclerck:
Oh my... With `COB_FLAGS="-A -O0"` the error disappear. With ``COB_FLAGS="-A -O1"``, it occurs... Now I'd have to check which optimization exactly causes this... That would give use a clue about what's going wrong.

GitMensch:
I don't know... but if it happens with `-A -O3` but not with`-A -O3 -fno-delayed-branch` and happens only on this machine - then just set it to CFLAGS during configure.

... hm, please check that, but then also consider to use an older GCC in the 32bit CI (whatever the oldest is, that is available) - after all the MSYS2 builds will use the newest 32bit GCC and we have some "backward compatibility" check in the CI as well.
So if we do this, we can then see if we still need the -fno-delayed-branch in the CFLAGS.

.github/workflows/build_nightly.yml
```diff
@@ -336,7 +336,90 @@ jobs:
             _build/tests/cobol85/**/duration.txt
 
 
-  minmal_build:
+  build_32_bit:
+    name: Build and test on 32-but
```
GitMensch, line 340: Possibly "bit"?

GitMensch:
As you seem to have found a bug in gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0 (-m32) it would be good if you can verify that the problem is caused by `-fdelayed-branches` and can provide the output of `cobc --save-temps -A -fdelayed-branches -vv caller.cob` as well as the output of `gcc -E -I /path/to/libcob caller.c` - I'll then bring that up to the GCC folks.

ddeclerck:
> I don't know... but if it happens with `-A -O3` but not with`-A -O3 -fno-delayed-branch` and happens only on this machine - then just set it to CFLAGS during configure.
> 
> ... hm, please check that, but then also consider to use an older GCC in the 32bit CI (whatever the oldest is, that is available) - after all the MSYS2 builds will use the newest 32bit GCC and we have some "backward compatibility" check in the CI as well. So if we do this, we can then see if we still need the -fno-delayed-branch in the CFLAGS.

Hmm, I might have run through some confusion, so I performed more checks with gcc 9 to 12, which all result the same behavior.

Actually, the segfault only only on `-O1` (and not on `-O0` nor `-O2` nor `-O3`).

Passing `-O0` together with ALL flags implied by `-O1` except `-fdelayed-branch` DO avoid the problem, which would suggest `-fdelayed-branch` is the culprit... But we can't check passing `-fdelayed-branch` with `-O0`, as this displays a message indicating delayed branch is not supported.

However passing `-fno-delayed-branch` together with `-O1`, the problem still occurs, which suggest it's something else...

So, I passed `-O1` together with the negation of ALL flags implied by `-O1`, and the problem disappeared. Then I removed those flags bit by bit, until only one remained: `-fno-tree-coalesce-vars`. Removing this one makes the problem reappear. However, using the opposite `-ftree-coalesce-vars` at other optimization levels does not trigger the problem (and according to the doc, this flag is already enabled at all levels). So it might be some bad interaction between this flag and something specific `-O1` does and that is not controlled by flags...

GitMensch:
Is the same true for all of gcc9 to gcc12?
And only on 32bit?

Can you please share the preprocessed C file along with the gcc invocations that (don't) show this error? Then we can let the gcc folks figure the details and fix out.

ddeclerck:
> Is the same true for all of gcc9 to gcc12? And only on 32bit?

As I said, "I performed more checks with gcc 9 to 12, which all result the same behavior".
So this is true for all those gcc versions, on 32-but only.

> Can you please share the preprocessed C file along with the gcc invocations that (don't) show this error? Then we can let the gcc folks figure the details and fix out.

On it (doing 10 things at the same time).

GitMensch:
that's good, so next is to take that over to GC4, possibly adding odbc (the more backends built, the better [that reminds me - the TODO file may miss the entry "testsuite: run with all configured ISAM backends"])

ddeclerck:
[programs.zip](https://github.com/user-attachments/files/20407945/programs.zip)

This archive contains the preprocessed C files corresponding to the two programs `caller` and `callee`.
They were obtained using the following commands:
```
cobc -x -debug -save-temps caller.cob
gcc -m32 -E -I /path/to/libcob caller.c > caller.pp.c

cobc -debug -fusing-optional=skip -save-temps callee.cob
gcc -m32 -E -I /path/to/libcob callee.c > callee.pp.c
```
(note the use of `-debug`: the problem does not occur without this option)

To compile and run, reproducing the bug:
```
gcc -m32 -O1 caller.pp.c -lcob -o caller
gcc -m32 -O1 -shared callee.pp.c -lcob -o callee.so
./caller
```

To compile and run, working around the bug:
```
gcc -m32 -O1 -fno-tree-coalesce-vars caller.pp.c -lcob -o caller
gcc -m32 -O1 -fno-tree-coalesce-vars -shared callee.pp.c -lcob -o callee.so
./caller
```
(assuming `LIBRARY_PATH` and `LD_LIBRARY_PATH` are set appropriately)

GitMensch:
Note: based on your good work I've minimized and cleaned up the test data (which took way to long) and created the bug report at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120417

caarmen:
Hello!

I'm in the middle of investigating a crash in a gnucobol program that occurs due to the following block:

```c
48	  /* Get current number of call parameters,
49	     if the parameter count is unknown, set it to all */
50	  if (cob_get_global_ptr ()->cob_current_module) {
51	  	cob_call_params = cob_get_global_ptr ()->cob_call_params;
52	  } else {
53	  	cob_call_params = 4;
54	  }
``` 

In my case, `cob_call_params` ends up incorrectly set to `1`, when it should be `4`. My cobol procedure expects 4 accessible arguments, but only the 1st one is valid. The last 3 are set to null. When my cobol code tries to access one of these null arguments, the program crashes with a seg fault.

My program is multithreaded (I think 😅 ), and this crash occurs rarely, after thousands of successful passes through the code path.

My analysis isn't developed enough to create a well formulated issue here.

I was just wondering, if by chance, there's some known issue about concurrency here.

If I get further in my analysis, I'll open a new issue with relevant details!

Thanks :)

GitMensch:
there are four possible reasons:

* the caller _really_ passed only one argument
* something trashed the global cob_call_params variable
* you run really multi-threaded with two COBOL threads executing a CALL at nearly the same time - and the second one overwrites the first

The solution for the last one is a codegen change to get the amount of arguments directly from the caller; if those are not yet part of the cob_module structure, then that would be backward incompatilbe (a 4.x change which you could apply locally to a 3.x version as long as you recompile everything afterwards).

PR #233, opened 2025-05-23
[GC4] Add 32-bit Ubuntu workflow

ddeclerck:
Ported from GC3 - #232.

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -38,7 +38,8 @@ jobs:
           sudo apt -y install autoconf automake make libtool flex bison \
                               autopoint gettext help2man texinfo texlive \
                               gcc libgmp-dev libncurses-dev \
-                              libdb5.3-dev libxml2-dev libcjson-dev
+                              libdb5.3-dev libxml2-dev libcjson-dev \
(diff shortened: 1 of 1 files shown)


ddeclerck:
Of course, the only failing tests on 32-bit are those for which the fix in PR #230 should be merged first.

.github/workflows/build_nightly.yml
```diff
@@ -296,6 +296,93 @@ jobs:
...
+      - name: Setup build environment
+        run: echo "TERM=vt100" >> $GITHUB_ENV
+
+      - name: Install packages
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt -y update
+          sudo apt -y install make libtool gcc-9-multilib \
+                              gcc-9 libgmp-dev:i386 libncurses-dev:i386 \
+                              libdb5.3-dev:i386 libxml2-dev:i386 libcjson-dev:i386
```
GitMensch, line 326: What about adding `libodbc2:i386 libodbcinst2:i386 unixodbc-dev:i386` (and potentially `--with-odbc` down there) - also in the existing GC4 one?

ddeclerck:
Works fine with the 32-bit fix from PR #230.
I also added ODBC as suggested.
Is there more to it ?

PR #234, opened 2025-05-27
Fix coverage for recent lcov (>= 2)

### LINKED ISSUE #212: CI for code coverage fails on new Ubuntu images
Since the latest Ubuntu image, lcov fails complaining about negative branch counts.
It suggests passing `fprofile-update=atomic` to gcc, but that does not seem to help either.
So for now I've changed `ubuntu-latest` to `ubuntu-22.04` in the coverage section  CI definition, but it probably needs more investigation.

ddeclerck:
This PR attemps to fix coverage with recent versions of lcov.

This is achieved through different options, each solving a different problem.

```
CODE_COVERAGE_LCOV_RMOPTS_DEFAULT="--ignore-errors unused"
```
This is required because this appears in the logs:
```
lcov: ERROR: 'exclude' pattern '/tmp/*' is unused.
	(use "lcov --ignore-errors unused ..." to bypass this error)
```
Note that the `/tmp/*` pattern is added by `m4/ax_code_coverage.m4`, so maybe an updated version of this M4 script would work with newer versions of lcov ?

```
CODE_COVERAGE_LCOV_OPTIONS_DEFAULT="--ignore-errors source"
```
This is required because this appears in the logs:
```
geninfo: ERROR: unable to open /home/runner/work/gnucobol/gnucobol/_build/cobc/ppparse.c: No such file or directory
	(use "geninfo --ignore-errors source ..." to bypass this error)
```
This error occurs for all generated files (`cobc/pplex.c`, `cobc/scanner.c`, `cobc/ppparse.c`, `cobc/parser.c`), probably because they are in the source tarball and hence missing from the build directory (copying them here would also workaround the error). Since all these files are in `CODE_COVERAGE_IGNORE_PATTERN`, it seems safe to ignore this error anyways, as we're doing here.

Then, we also apply a "fixup" for a warning:
```
CODE_COVERAGE_LCOV_OPTIONS_DEFAULT="--rc geninfo_unexecuted_blocks=1"
```
Needed because we have this:
```
geninfo: WARNING: /home/runner/work/gnucobol/gnucobol/libcob/numeric.c:469: unexecuted block on non-branch line with non-zero hit count.  Use "geninfo --rc geninfo_unexecuted_blocks=1 to set count to zero.
	(use "geninfo --ignore-errors gcov,gcov ..." to suppress this warning)
```

Finally, to be more future-proof, we define those:
```
CODE_COVERAGE_LCOV_SHOPTS_DEFAULT="--rc branch_coverage=1"
CODE_COVERAGE_GENHTML_OPTIONS_DEFAULT="--rc genhtml_branch_coverage=1"
```
Because we have these warnings:
```
geninfo: WARNING: RC option 'lcov_branch_coverage' is deprecated.  Consider using 'branch_coverage. instead.  (Backward-compatible support will be removed in the future
```
```
genhtml: WARNING: RC option 'genhtml_branch_coverage' is deprecated.  Consider using 'branch_coverage. instead.  (Backward-compatible support will be removed in the future
```
But maybe an updated version of `ax_code_coverage.m4` could work ?

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -500,10 +500,7 @@ jobs:
   coverage:
     name: Coverage and Warnings
     needs: build
-# Since the latest Ubuntu image, lcov fails complaining about negative branch counts,
-# and using fprofile-update=atomic as suggested does not help, so use the previous image
-#    runs-on: ubuntu-latest
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     timeout-minutes: 10
 
     steps:
@@ -551,6 +548,8 @@ jobs:
           # make -C _build check-code-coverage # <- (ignores errors)
           make -C _build check TESTSUITEFLAGS="--jobs=$(($(nproc)+1))"
           make -C _build code-coverage-capture \
+                         CODE_COVERAGE_LCOV_SHOPTS_DEFAULT="--rc branch_coverage=1" \
+                         CODE_COVERAGE_GENHTML_OPTIONS_DEFAULT="--rc branch_coverage=1" \
                          CODE_COVERAGE_DIRECTORY="$(realpath .)/_build"
 
       - name: Upload testsuite.log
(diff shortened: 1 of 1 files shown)


GitMensch:
The big question not answered here yet: does your configuration work with lcov 1 (at least 1.16, which is in Debian stable) as well?

ddeclerck:
> Answer to my question: just using `branch_coverage` (without the lcov_/genhtml_ prefixes) raises no warning or error in lcov-1.16... but it does not do the branch coverage any more. The macro should therefore be conditional on the version...

Yeah, I noticed that. Note that `lcov_branch_coverage` and `genhtml_branch_coverage` are still accepted, although deprecated. I made the change just for things to be cleaner.

Also note that all the proposed fixes are only in the workflow definition, which is unlikely to rollback to an older version of lcov.

GitMensch:
> Also note that all the proposed fixes are only in the workflow definition, which is unlikely to rollback to an older version of lcov.

Yes, but the issue is that users run it as well, and we don't want coverage to error then :-)

ddeclerck:
> That's nice. The remaining question is if we can make those two conditional in the code-coverage m4 macro depending on the lcov version - because otherwise this will be broken outside of the CI (either for lcov1 or lcov2 environments).

Are you suggesting we modify `m4/ax_code_coverage.m4` ? I'm not really M4-proficient.

To me it would be okay to *not* use those new options for now and let the deprecation warning occur...

GitMensch:
Friendly ping @ddeclerck

PR #237, opened 2025-07-21
Fix CI

### LINKED ISSUE #236: 3.x CI failures back?
@ddeclerck, maybe you can have a look at the failures we have, not sure if we can fix these or need a conditional skip...

Note: the libxml2 issue was an ABI break and is already fixed upstream (just needs to be "svn mirror'd" / merged to the CI containing main branch of this repo)

ddeclerck:
Recent CI failures have two reasons:

- SVN commit 5549 changed a test in `used_binaries.at`, removing the body of a function: this broke test `27: check include header file` on MacOS, MSVC and MSYS2. I restored the (empty) function body.

- MSYS2 updated the perl package version from 5.38.2 to 5.38.4, yet forgot to update other packages that install perl modules: as a result, those are still installed in the 5.38.2 perl module folder. This broke a perl script invoked when installing tex, causing errors when generating the GnuCOBOL documentation. I opened an [issue](https://github.com/msys2/MINGW-packages/issues/24906) on the MSYS2 repo. Meanwhile, I'm using a very hackish workaround (which I don't intend to merge) to get the CI working. I hope this will be fixed in MSYS2 soon.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -11292,12 +11292,12 @@ output_module_init_function (struct cb_program *prog)
 	if (!prog->nested_level) {
 		output_line ("/* Initialize module structure for %s */",
 			prog->orig_program_id);
-		output_line ("static void %s_module_init (cob_module *module)",
+		output_line ("static void %s_module_init (cob_module *module__)",
 			prog->program_id);
 	} else {
 		output_line ("/* Initialize module structure for %s (nested %d) */",
 			prog->program_id, prog->toplev_count);
-		output_line ("static void %s_%d_module_init (cob_module *module)",
+		output_line ("static void %s_%d_module_init (cob_module *module__)",
 			prog->program_id, prog->toplev_count);
 	}
 	output_block_open ();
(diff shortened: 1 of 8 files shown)


tests/testsuite.src/used_binaries.at
GitMensch, line 1: That gave an error with Fedora 42, not sure what that was...

cobc/codegen.c
```diff
@@ -11751,10 +11751,10 @@ output_internal_function (struct cb_program *prog, cb_tree parameter_list)
 	output_local ("static unsigned int\tinitialized = 0;\n\n");
 	if (prog->flag_recursive) {
 		output_local ("/* Module structure pointer for recursive */\n");
-		output_local ("cob_module\t\t*module = NULL;\n\n");
+		output_local ("cob_module\t\t*module__ = NULL;\n\n");
 	} else {
 		output_local ("/* Module structure pointer */\n");
-		output_local ("static cob_module\t*module = NULL;\n\n");
+		output_local ("static cob_module\t*module__ = NULL;\n\n");
```
GitMensch, line 11757: I'm quite reluctant to change the name here in 3.2 - because that's used in several GDB extensions.
I'm fine with doing that for 4.x though...

Do we _need_ the global name change or is the place you've previously shared (within %s_module_init) enough?
If not, then please change the program name in the testsuite after adding it to cobc.c (cob_csyns) under `P_switch` (we may need to add some more there).

tests/testsuite.src/used_binaries.at
GitMensch: let's see that as solved then...

GitMensch:
all changes but workflow and codegen committed upstream

ddeclerck:
I guess this can be closed, as everything here is now handled by SVN revision [552](https://sourceforge.net/p/gnucobol/code/5552/).

PR #239, opened 2025-07-29
Fix odoslide

ddeclerck:
This is WIP.

There's a piece of code in `cb_emit_move` that is ifdef'ed out in 3.x and active in trunk and that seems to solve the problem. But it does create other problems. Still working on it.

### DIFF
--- a/cobc/codegen.c
+++ b/cobc/codegen.c
@@ -718,7 +718,7 @@ real_field_founder (const struct cb_field *f)
 	return (struct cb_field *)f;
 }
 
-static struct cb_field *
+struct cb_field *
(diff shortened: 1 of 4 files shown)


GitMensch:
So what about trying to merge 2603+2612 in first?

ddeclerck:
> So what about trying to merge 2603+2612 in first?

Didn't realize this came from the reportwriter branch.

Code is a bit different between 3.x and trunk, and some of 2603/2612 seems already merged in.
Anyways, didn't have much success with this one. Guess that'll wait until I'm back. Unless you feel like trying to tackle this.

PR #243, opened 2025-09-09
Fix ORD/CHAR intrinsics

ddeclerck:
This fixes the ORD and CHAR intrisics in presence of collating sequences.

### DIFF
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,4 +1,11 @@
 
+2025-09-09  David Declerck <david.declerck@ocamlpro.com>
+
+	* intrinsic.c (cob_intr_char, cob_intr_ord): consider
+	  the program collating sequence in CHAR and ORD
+	* intrinsic.c (cob_intr_char): raise COB_EC_ARGUMENT_FUNCTION
+	  when calling CHAR with an argument outside the collation range
+
 2025-07-28  Simon Sobisch <simonsobisch@gnu.org>
 
 	* common.h, fileio.c: new externalized typedef EXTFH_FUNC used in
--- a/libcob/intrinsic.c
+++ b/libcob/intrinsic.c
@@ -4318,8 +4318,7 @@ cob_intr_current_date (const int offset, const int length)
 	return curr_field;
 }
 
-/* implementation of FUNCTION CHAR - character from ordinal
-   FIXME: Should use the program's alphanumeric program collating sequence! */
+/* implementation of FUNCTION CHAR - character from ordinal */
 cob_field *
 cob_intr_char (cob_field *srcfield)
 {
@@ -4331,9 +4330,23 @@ cob_intr_char (cob_field *srcfield)
 
 	i = cob_get_int (srcfield);
 	if (i < 1 || i > 256) {
+		cob_set_exception (COB_EC_ARGUMENT_FUNCTION);
 		*curr_field->data = 0;
 	} else {
-		*curr_field->data = (unsigned char)i - 1;
+		const unsigned char *col = COB_MODULE_PTR->collating_sequence;
+		unsigned char chr;
+		cobglobptr->cob_exception_code = 0;
+		--i;
+		if (col == NULL) {
+			chr = (unsigned char)i;
+		} else {
+			for (chr = 0; chr < 255 && col[chr] != i; ++chr);
+			if (chr == 255 && col[chr] != i) {
+				cob_set_exception (COB_EC_ARGUMENT_FUNCTION);
+				chr = 0;
+			}
+		}
+		*curr_field->data = chr;
 	}
 	return curr_field;
 }
@@ -4341,7 +4354,18 @@ cob_intr_char (cob_field *srcfield)
 cob_field *
 cob_intr_ord (cob_field *srcfield)
 {
-	cob_alloc_set_field_uint ((cob_u32_t)(*srcfield->data + 1U));
+	switch (COB_FIELD_TYPE (srcfield)) {
+	case COB_TYPE_NATIONAL:
+		error_not_implemented();
+		break;
+	default: {
+		const unsigned char *col = COB_MODULE_PTR->collating_sequence;
+		unsigned char chr = *srcfield->data;
+		int ord = (col == NULL ? chr : col[chr]) + 1U;
+		cob_alloc_set_field_uint ((cob_u32_t)ord);
+		break;
+	}
+	}
 	return curr_field;
 }
 
--- a/tests/testsuite.src/run_misc.at
+++ b/tests/testsuite.src/run_misc.at
@@ -15436,7 +15436,7 @@ AT_DATA([prog1.cob], [
 
 AT_CHECK([$COMPILE prog1.cob], [0], [], [])
 AT_CHECK([$COBCRUN_DIRECT ./prog1], [0],
-[LOW-VALUE: 063 (064) HIGH-VALUE: 192 (193)], [])
+[LOW-VALUE: 063 (001) HIGH-VALUE: 192 (256)], [])
 
 AT_DATA([prog2.cob], [
        IDENTIFICATION DIVISION.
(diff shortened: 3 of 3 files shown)


libcob/intrinsic.c
```diff
@@ -4333,15 +4332,26 @@ cob_intr_char (cob_field *srcfield)
 	if (i < 1 || i > 256) {
 		*curr_field->data = 0;
 	} else {
-		*curr_field->data = (unsigned char)i - 1;
+		const unsigned char *col = COB_MODULE_PTR->collating_sequence;
```
GitMensch, line 4336: What happens when the collating sequence doesn't have 255 elements?

Note: here and in the "chr == 0" because of loop case below we need to raise an argument exception:

> The value of argument-1 shall be greater than zero and less than or equal to the number of positions 
in the alphanumeric program collating sequence.

The old code should have read:

```c
	if (i < 1 || i > 256) {
		cob_set_exception (COB_EC_ARGUMENT_FUNCTION);
		*curr_field->data = 0;
	} else {
		cobglobptr->cob_exception_code = 0;
		*curr_field->data = (unsigned char)i - 1;
	}
```

please apply it to this change (adding it in the changelog)

libcob/intrinsic.c
```diff
@@ -4333,15 +4332,26 @@ cob_intr_char (cob_field *srcfield)
...
+		unsigned char chr;
+		--i;
+		if (col == NULL) {
+			chr = (unsigned char)i;
+		} else {
+			for (chr = 255; chr > 0 && col[chr] != i; --chr);
```
GitMensch: we need to start from zero:

> If more than one character has the same position in the alphanumeric program collating sequence, 
the  character  returned  is  the  first  character  defined for  that  character position.

libcob/intrinsic.c
```diff
@@ -4333,15 +4332,26 @@ cob_intr_char (cob_field *srcfield)
...
 }
 
 cob_field *
 cob_intr_ord (cob_field *srcfield)
 {
-	cob_alloc_set_field_uint ((cob_u32_t)(*srcfield->data + 1U));
+	const unsigned char *col = COB_MODULE_PTR->collating_sequence;
+	unsigned char chr = *srcfield->data;
+	int ord = (col == NULL ? chr : col[chr]) + 1U;
+	cob_alloc_set_field_uint ((cob_u32_t)ord);
```
GitMensch: Can you add a switch here checking the field's type?

> If the class of argument-1 is national, the returned value is the ordinal position of argument-1 in the 
current national program collating sequence.

It is fine to  use `error_not_implemented()` in the second case.

tests/testsuite.src/run_misc.at
GitMensch, line 1: I _guess_ the 256 is often wrong per standard (see note in intrinsic.c)

libcob/intrinsic.c
GitMensch, line 1: I've just recognized that `cob_intr_ord_min()` and `cob_intr_ord_max` are likely broken - or otherwise too complicated.
They currently use cob_cmp() which has a lot of special cases that may not apply. I _think_ those should use `cob_ord` internally, and comparing its return. Can you have a look at that, please?

In any case it would be good to have a testcase with a non-standard collating sequence for those as well.

GitMensch (changes requested):
Welcome back and thanks for inspecting that. I think I've found some edge cases, please have a look.

libcob/intrinsic.c
ddeclerck: > What happens when the collating sequence doesn't have 255 elements?

It always has. If specifying a custom alphabet with less than 255 characters, the remaining positions are filled with all the characters that were not explicitly specified.

> Note: here and in the "chr == 0" because of loop case below we need to raise an argument exception
> 
> > The value of argument-1 shall be greater than zero and less than or equal to the number of positions
> > in the alphanumeric program collating sequence.

Not sure about the `chr == 0` case ; it would imply the collating sequence is not a bijection, but we know it is by construction (and if it isn't it would probably be an internal error instead of an exception).

libcob/intrinsic.c
ddeclerck: Yet according to the [IBM doc](https://www.ibm.com/docs/en/cobol-zos/6.3.0?topic=section-alphabet-clause):
```quote
Within one alphabet-name clause, a given character must not be specified more than once.
```
So I don't get how two characters would end up at the same position in the collating sequence (GnuCOBOL checks that in `typeck.c:validate_alphabet`).

libcob/intrinsic.c
ddeclerck: Actually I already checked the code, and they end up using  the collation-aware comparison function. Indeed I could add a test for that though.

tests/testsuite.src/run_misc.at
ddeclerck: You mean `ORD (HIGH-VALUE)` ?

libcob/intrinsic.c
ddeclerck: Oh, I've just stumbled upon "ALSO"... But I don't think the way we currently handle collating sequences allows proper support for this...

EDIT: oh, in fact it works ; always new things to learn with COBOL 😅

libcob/intrinsic.c
GitMensch: For the collation that makes sense, for the standard exception... we'd have to add the size.... maybe just add a new cob_intr_char_col that would take the data, collation and collation size (and ~~forward from the old to the new one in 3.x~~ would also remove the need to check for the collation in intr_char)?

libcob/intrinsic.c
GitMensch: If that redirection is useful / not really slower than comparing the `ord()` result then we can leave it as is.

libcob/ChangeLog
```diff
@@ -1,4 +1,11 @@
 
+2025-09-09  David Declerck <david.declerck@ocamlpro.com>
+
+	* intrinsic.c (cob_intr_char, cob_intr_ord): consider
+	  the program collating sequence in CHAR and ORD
+	* intrinsic.c (cob_intr_char): raise COB_EC_ARGUMENT_FUNCTION
+	  when calling CHAR with an argulent outside of the 1-256 range
```
GitMensch: typo: argument
maybe just say "outside of the collation range" (as noted, we may use a new cob_intr_col in this case - and could even do this depending on the setup of a program collation, which removes the checks on this var as well)

GitMensch:
Juist to let you know, as the MSVC runs failed with missing libxml2 headers I've deleted their cache and restarted those with debug logging.

GitMensch:
Same result. If I understood https://github.com/microsoft/vcpkg/pull/47074/files#diff-2e899573a3ba59e3604a2edd75fa4fc6cf6473ff397fc988cb37dacff007a257L83 correctly, it now needs the libxml2 folder be included (PKG_CONFIG and xml2-config provide those includes on other systems)... yes, I do and there's an upstream regression ticket for that https://github.com/microsoft/vcpkg/issues/47155 - the solution (both locally, in GH CI in AppVeyor CII, ... is to pin the version for now). Let me try that.

ddeclerck:
> Same result. If I understood https://github.com/microsoft/vcpkg/pull/47074/files#diff-2e899573a3ba59e3604a2edd75fa4fc6cf6473ff397fc988cb37dacff007a257L83 correctly, it now needs the libxml2 folder be included (PKG_CONFIG and xml2-config provide those includes on other systems)... yes, I do and there's an upstream regression ticket for that [microsoft/vcpkg#47155](https://github.com/microsoft/vcpkg/issues/47155) - the solution (both locally, in GH CI in AppVeyor CII, ... is to pin the version for now). Let me try that.

You might want to try this in this PR : [https://github.com/OCamlPro/gnucobol/pull/244](https://github.com/OCamlPro/gnucobol/pull/244)

GitMensch:
If I not overlooked something on this small mobile screen: That can go upstream can't it?

ddeclerck:
> If I not overlooked something on this small mobile screen: That can go upstream can't it?

I think yes, I'll do that when I'll have a moment.

ddeclerck:
Merged in SVN @ 5589.

PR #244, opened 2025-09-10
Fix CI (MinGW)

ddeclerck:
* IBM Power/Z: a previous bug required us to manually define the HOME variable; since this was fixed upstream, we can remove that manual definition

* Windows/MSYS2: an update of the setup-msys2 action added more checks, which prevents from using the action more than. Adding `release: false` allows to work around the problem

* Windows/MSVC: compilation of GnuCOBOL complains about missing libxml/parser.h and libxml/uri.h, for some reason. I don't know what caused this...

* MacOS: since they updated the latest runner image, we have this failure:
```
869: runtime check: write to internal storage (1)    FAILED (run_misc.at:14410)
```
This does not occur with the previous image.
Note sure what the proper fix would be.

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -604,17 +604,19 @@ jobs:
       MSYS_BIN: C:\MinGW\msys\1.0\bin
 
       MSYSTEM: MINGW32
-      MSYSPKGS: msys-coreutils msys-patch
+      MSYSPKGS: msys-coreutils msys-patch pthreads
 
       MSYS_URL:  https://www.arnoldtrembley.com/MinGW-bkup02.7z
 
       # Note: MSYS1 cannot connect to https://gmplib.org, so using GNU mirror side instead
       GMP_URL:   https://ftp.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz
       PDCM_URL:  https://github.com/Bill-Gray/PDCursesMod/archive/refs/tags/v4.4.0.tar.gz
+      # Note: last version under old sleepycat license, newer is AGPL
       BDB_URL:   https://download.oracle.com/berkeley-db/db-6.0.19.NC.tar.gz
-      BDBP_URL:  https://raw.githubusercontent.com/msys2/MINGW-packages/refs/heads/master/mingw-w64-db
+      # Note: last version on MSYS2 building based on that BDB version
+      BDBP_URL:  https://raw.githubusercontent.com/msys2/MINGW-packages/ba87ffd249bd6021db6baeb60656627985996567/mingw-w64-db
       XML2_URL:  https://github.com/GNOME/libxml2/archive/refs/tags/v2.8.0.tar.gz
-      CJSON_URL: https://raw.githubusercontent.com/DaveGamble/cJSON/v1.7.18
+      CJSON_URL: https://raw.githubusercontent.com/DaveGamble/cJSON/v1.7.19
 
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


ddeclerck:
> * IBM Power/Z: a previous bug required us to manually define the HOME variable; since this was fixed upstream, we can remove that manual definition

Well, for some reason it's not... Cf [https://github.com/IBM/actionspz/issues/2#issuecomment-3259367176](https://github.com/IBM/actionspz/issues/2#issuecomment-3259367176)

GitMensch:
> This does not occur with the previous image.
> Note sure what the proper fix would be.

The proper fix is to adjust the NEWS file entry

> 
>* Known issues in 3.x
>
>** testsuite:
>  * if built with vbisam, cisam or disam, depending on the version used, some
>    tests will lead to UNEXPECTED PASS, while others may fail
>  * possibly failing tests (false positives):
>    * temporary path invalid
>    * compiler outputs (assembler)
>    * compile from stdin
>  * NIST: OBNC1M.CBL false positive (the test runner uses a nonportable way of
>    emulating a program kill)
>  * if build with -fsanitize, then some tests will fail; while we accept patches
>    to improve that, we don't consider the failing tests as bug in GnuCOBOL

```diff
-  * if build with -fsanitize, then some tests will fail; while we accept patches
+  * if build with hardening options like -fsanitize, then some tests will fail; while we accept patches
```

... and skip the test for this CI run case, ideally conditionally (configure checking for options, testsuite skipping).
I'd suggest not to go the most easy route disabling those options - because then we're loosing it for others.

A still clean and easy way would be to adjust the testcase - if it is failing, then re-execute and check again with additional passing `-A "-fno-sanitize -fno-stack-protector-strong -fno-stack-clash-protection -U_FORTIFY_SOURCE"` ... but then we'd also have to create this list in configure because some of those are only available with some compilers.

Can you give that a go? That would be "the patches we accept to improve that" :-)

GitMensch:
Friendly ping @ddeclerck to get at least parts (msys2 config?) in.

Of course fixing the one failing testcase would be nice, but skipping it in this specific test run would be ok as well.

ddeclerck:
Regarding your change the the MSVCI CI: it seems VCPKG does not easily allow to install a specific version (cf https://github.com/microsoft/vcpkg/discussions/39369).

GitMensch:
yes, the most reasonable fix for vcpkg for now is to copy the installed headers to the build_windows subdirectory...

GitMensch:
your include hack works fine; can you check to include everything that this fixes (as a new single commit) directly?

ddeclerck:
> your include hack works fine; can you check to include everything that this fixes (as a new single commit) directly?

Sure, but the old MinGW CI is still broken - do we ignore that for now ?

GitMensch:
Yes, just update all but the mingw parts and when the appveyor CI runs again (I've broken MinGW by a change and was not able to test the fix before my vacation), I'll have a look at the MinGW one here.

You may just pass MinGW even on failure for now.

ddeclerck:
That has been done.

GitMensch:
Do we need the same for the gc4 branch?

ddeclerck:
Probably, I'll check that

ddeclerck:
Do we close this oPRe, or will you use it to experiment with MinGW ?

GitMensch:
I'll use it for that purpose, starting with a rebase.

ddeclerck:
Hmm, after merging upstream GC3, the MSVC CI breaks again.
Apparently due to the changes to `atlocal_win` in [5576](https://sourceforge.net/p/gnucobol/code/5576/).

GitMensch:
I expect to have my change fixed until tomorrow, which should lead to the Appveyor MinGW one and MSCV here working again.

GitMensch:
@ddeclerck you would want to replace the fix-ci branch by the current gcos-gc3 one where macos works again and msys1 is improved; the only open part is that it currently can't link BDB, changing BDB_LIBS for GnuCOBOL's configure should fix this, but I've ran out of time for further inspecting it

ddeclerck:
> @ddeclerck you would want to replace the fix-ci branch by the current gcos-gc3 one where macos works again and msys1 is improved; the only open part is that it currently can't link BDB, changing BDB_LIBS for GnuCOBOL's configure should fix this, but I've ran out of time for further inspecting it

Hi Simon,
I'll have a look ASAP.

PR #245, opened 2025-09-26
Add COB_SIGNAL_REGIME config option to control how runtime signal handler is registered

oguzcankirmemis:
This PR adds a config option `COB_SIGNAL_REGIME` which allows control over how the signal handler of the runtime is registered for a signal in `common.c.`/`cob_set_signal`.

# Motivation
Current implementation overwrites the signal handlers of the main program under which the runtime operates. In some environments, such as the Java Virtual Machine, this can create potential issues where signals intended for the main program are caught by the COBOL runtime, leading to corruption. For reference, Micro Focus compiler provides [a similar flag](https://www.microfocus.com/documentation/visual-cobol/vc100/DevHub/HRRTRHRTCF0J.html).

With this flag, the main program can either completely skip signal registration or skip it for specific signals it uses. Furthermore, this PR also exposes the signal handler of the runtime, allowing the main program to register it manually if desired or to combine it with its own signal handler. The latter mechanism is for instance provided by the `libjsig` [library of the Java Runtime](https://docs.oracle.com/javase/8/docs/technotes/guides/vm/signal-chaining.html).

# Implementation
- export `cob_sig_handler` function in `common.c`
- add `COB_SIGNAL_REGIME` as a configuration
- add documentation to `runtime.cfg` about `COB_SIGNAL_REGIME``
- pass HAVE_SIGNAL_H variable to tests so that the tests for COB_SIGNAL_REGIME only run when the OS provides signal handling

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -29,6 +29,12 @@ NEWS - user visible changes				-*- outline -*-
 ** new runtime configuration COB_HIDE_CURSOR, allows to hide the cursor during
    extended ScreenIO operations
 
+** new runtime configuration COB_SIGNAL_REGIME, allows to skip registration of
+   the signal handler of GnuCOBOL. The registration can be skipped either for
+   all signals or for those that do already have a registered handler. This
+   configuration allows users to preserve their own signal handlers during the
+   initialization of the GnuCOBOL runtime.
+
 ** added multiple window functionality with new system function CBL_GC_WINDOW
 
 ** New system functions
--- a/configure.ac
+++ b/configure.ac
@@ -714,7 +714,7 @@ AC_FUNC_VPRINTF
 AC_CHECK_FUNCS([memmove memset setlocale fcntl strerror strcasecmp
 	strchr strrchr strdup strstr strtol gettimeofday localeconv
 	getexecname canonicalize_file_name popen raise readlink realpath
-	setenv strcoll flockfile])
+	setenv strcoll flockfile kill signal])
 
 # more things to save...
 AC_CACHE_SAVE
--- a/configure.ac
+++ b/configure.ac
@@ -2761,6 +2761,7 @@ AC_SUBST([PROGRAMS_LIBS])
 AC_SUBST([LIBCOB_LIBS])
 AC_SUBST([LIBCOB_CPPFLAGS])
 AC_SUBST([COB_ENABLE_DEBUG], [$enable_debug]) # needed for tests/atlocal
+AC_SUBST([HAVE_SIGNAL_H], ["$ac_cv_header_signal_h"]) # needed for tests/atlocal
 
 dnl was used in bin/Makefile.am - seems not to be needed
 dnl AC_SUBST([COB_EXPORT_DYN])
--- a/configure.ac
+++ b/configure.ac
@@ -714,7 +714,7 @@ AC_FUNC_VPRINTF
 AC_CHECK_FUNCS([memmove memset setlocale fcntl strerror strcasecmp
 	strchr strrchr strdup strstr strtol gettimeofday localeconv
 	getexecname canonicalize_file_name popen raise readlink realpath
-	setenv strcoll flockfile kill signal])
+	setenv strcoll flockfile])
 
 # more things to save...
 AC_CACHE_SAVE
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -1527,7 +1527,7 @@ cob_set_signal (void)
 	struct sigaction	osa;
 
 	signal_regime = getenv ("COB_SIGNAL_REGIME");
-	if (*signal_regime == '2') {
+	if (signal_regime && *signal_regime == '2') {
 		/* Don't set any signal */
 		return;
 	}
@@ -1544,7 +1544,7 @@ cob_set_signal (void)
 
 	for (k = 0; k < NUM_SIGNALS; k++) {
 		if (signals[k].for_set) {
-			if (*signal_regime == '1') {
+			if (signal_regime && *signal_regime == '1') {
 				/* Only take control if no handler is registered (=SIG_DFL) */
 				(void)sigaction (signals[k].sig, NULL, &osa);
 				if (osa.sa_handler == SIG_DFL) {
@@ -1571,14 +1571,14 @@ cob_set_signal (void)
 	void (*ohdlr) (int);
 
 	signal_regime = getenv ("COB_SIGNAL_REGIME");
-	if (*signal_regime == '2') {
+	if (signal && *signal_regime == '2') {
 		/* Don't set any signal */
 		return;
 	}
 
 	for (k = 0; k < NUM_SIGNALS; k++) {
 		if (signals[k].for_set) {
-			if (*signal_regime == '1') {
+			if (signal && *signal_regime == '1') {
 				/* Only take control if no handler is registered (=SIG_DFL) */
 				ohdlr = signal (signals[k].sig, cob_sig_handler);
 				if (ohdlr != SIG_DFL) {
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -531,7 +531,7 @@ static struct config_enum beepopts[] = {{"FLASH", "1"}, {"SPEAKER", "2"}, {"FALS
 static struct config_enum timeopts[] = {{"0", "1000"}, {"1", "100"}, {"2", "10"}, {"3", "1"}, {NULL, NULL}};
 static struct config_enum syncopts[] = {{"P", "1"}, {NULL, NULL}};
 static struct config_enum varseqopts[] = {{"0", "0"}, {"1", "1"}, {"2", "2"}, {"3", "3"}, {NULL, NULL}};
-static struct config_enum sighdlrregopts[] = {{"0", "0"}, {"1", "1"}, {"2", "2"}, {NULL, NULL}};
+static struct config_enum sighdlrregopts[] = {{"DEFAULT", "0"}, {"AVAILABLE", "1"}, {"IGNORE", "2"}, {NULL, NULL}};
 static struct config_enum coeopts[] = {{"0", "0"}, {"1", "1"}, {"2", "2"}, {"3", "3"}, {NULL, NULL}};
 static char	varseq_dflt[8] = "0";
 static unsigned char min_conf_length = 0;
@@ -1175,7 +1175,23 @@ create_dumpfile (void)
 	return ret;
 }
 
+/** \brief default signal handler of the runtime
 
+	In most cases when a signal is received, the signal handler is configured
+	to terminate the program after tidying up the runtime, unless an external
+	signal handler is registered. In such cases, the external handler is
+	invoked before the signal is raised again, which is then handled by the
+	default handler of the host operating system.
+
+	If the signal is received during a \ref cob_call_with_exception_check call,
+	the signal handler aborts the call and notifies that it has been aborted by
+	the signal handler.
+	
+	Additionaly, the signal handler can be configured to create a coredump when
+	a signal occurs. See \ref COB_CORE_ON_ERROR option for more details.
+
+	\param sig number of the signal
+ */
 void
 cob_sig_handler (int sig)
 {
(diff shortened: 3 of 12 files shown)


libcob/common.c
```diff
@@ -1174,10 +1176,10 @@ create_dumpfile (void)
 }
 
 
-#ifdef	HAVE_SIGNAL_H
-static void
+void
 cob_sig_handler (int sig)
 {
+#ifdef	HAVE_SIGNAL_H
```
GitMensch: As the signals have an own check and the signal handler _should_ do something in any case, I think this ifdef needs to be dropped (one of the original reasons to be in was likely that we get unused warnings, which doesn't happen with the signal handler function being exposed now.

As it is "external ABI" now, it would be nice (not needed if you don't feel you can provide something reasonable) to add a doxygen-style function comment on this function.

libcob/common.c
```diff
@@ -531,6 +531,7 @@ static struct config_enum beepopts[] = {{"FLASH", "1"}, {"SPEAKER", "2"}, {"FALS
 static struct config_enum timeopts[] = {{"0", "1000"}, {"1", "100"}, {"2", "10"}, {"3", "1"}, {NULL, NULL}};
 static struct config_enum syncopts[] = {{"P", "1"}, {NULL, NULL}};
 static struct config_enum varseqopts[] = {{"0", "0"}, {"1", "1"}, {"2", "2"}, {"3", "3"}, {NULL, NULL}};
+static struct config_enum sighdlrregopts[] = {{"0", "0"}, {"1", "1"}, {"2", "2"}, {NULL, NULL}};
```
GitMensch: If you can com up with a nice label for those values (maybe "default", "available", "ignore"?) then lpase use them in this tuple.

libcob/common.c
```diff
@@ -1519,10 +1522,17 @@ cob_set_signal (void)
 {
 #if	defined (HAVE_SIGNAL_H)
 	int k;
+	char *signal_regime;
```
GitMensch: please replace by

```c
	char *s = getenv ("COB_SIGNAL_REGIME");
	char signal_regime = s ? *s : '0';
```

this way you can use `signal_regime` directly in the code below and reduce duplicated code

tests/testsuite.src/run_misc.at
```diff
@@ -15682,3 +15682,159 @@ AT_CHECK([$COBCRUN_DIRECT ./prog6], [0],
 [01 02 03 04 05 06 07 08 09 ], [])
 
 AT_CLEANUP
+
+
+AT_SETUP([COB_SIGNAL_REGIME={0,1} registers runtime signal handler])
+AT_KEYWORDS([runmisc COB_SIGNAL_REGIME])
```
GitMensch: cob_call_with_exception_check should also be included in the keywords

tests/testsuite.src/run_misc.at
```diff
@@ -15682,3 +15682,159 @@ AT_CHECK([$COBCRUN_DIRECT ./prog6], [0],
...
+        return 1;
+    }
+    
+    return 0;
+}
+]])
+
+AT_CHECK([$COMPILE caller.c], [0], [], [])
+AT_CHECK([$COMPILE_MODULE callee.cob], [0], [], [])
+AT_CHECK([COB_SIGNAL_REGIME=0 ./caller], [0], [],
```
GitMensch: please use the caller-prefix `$COBCRUN_DIRECT` here and below; that is f.e. used vor valgrind and other test runs

tests/testsuite.src/run_misc.at
```diff
@@ -15682,3 +15682,159 @@ AT_CHECK([$COBCRUN_DIRECT ./prog6], [0],
...
+    /* Check exit by signal handler */
+    if (cob_ret != -3) {
+        return 1;
+    }
+    
+    if (signal_received == 1) {
+        return 0;
+    }
+
+    return 1;
```
GitMensch: Shouldn't that return 2 so we see the difference?

tests/testsuite.src/run_misc.at
```diff
@@ -15682,3 +15682,159 @@ AT_CHECK([$COBCRUN_DIRECT ./prog6], [0],
...
+    if (cob_ret != -3) {
+        return 1;
+    }
+    
+    return 0;
+}
+]])
+
+AT_CHECK([$COMPILE caller.c], [0], [], [])
+AT_CHECK([$COMPILE_MODULE callee.cob], [0], [], [])
```
GitMensch: Here and below, we likely need to explicit disable some runtime check with `-fno-ec=program-arg` or similar, otherwise we don't get to thje signal handler.

tests/atlocal_win
```diff
@@ -167,6 +167,8 @@ if test "$GNUCOBOL_TEST_LOCAL" != "1"; then
 	done
 fi
 
+HAVE_SIGNAL_H="@HAVE_SIGNAL_H@"
```
GitMensch: that should likely be a constant 1 in this file (as it is not autoconfigured but copied over)

GitMensch (changes requested):
Oğuzcan that's a _very_ nice first PR!
Some minor nits below (I'll also see if we can fix the CI...)

libcob/common.c
```diff
@@ -1535,37 +1545,57 @@ cob_set_signal (void)
...
-				/* CHECKME: how should we handle externally registered
-				            error handlers (handler != SIG_DFL)? */
 			}
 		}
 	}
 #else	/* still defined (HAVE_SIGNAL_H) */
+	void (*ohdlr) (int);
+
+	signal_regime = getenv ("COB_SIGNAL_REGIME");
+	if (signal && *signal_regime == '2') {
```
GitMensch: The change above will also fix the typo of checking signal here.

libcob/common.c
oguzcankirmemis: I have added the function comment. Let me know if it is an accurate description of what it does. I am not sure about removing the ifdef. Wouldn't it give a compiler error if the host does not have signal.h, since it also uses symbols from there? I thought that was the use case from the beginning. And what should the signal handler do in an environment where signal.h does not exist?

tests/testsuite.src/run_misc.at
oguzcankirmemis: Included.

tests/testsuite.src/run_misc.at
oguzcankirmemis: Makes sense, thanks for pointing it out!

libcob/common.c
oguzcankirmemis: Resolved by the comment above.

libcob/common.c
oguzcankirmemis: I have added the labels as you suggested, but wouldn't it create confusion because we use `getenv` and directly check if `COB_SIGNAL_REGIME` is either 0,1 or 2? What I mean is for example `COB_SIGNAL_REGIME=default` wouldn't work as expected, which is the intended use case if I am understanding the configuration setup correctly.

oguzcankirmemis:
I am not entirely sure if the remaining failing CI checks are related to the changes in the PR. I would be glad if you could check it.

tests/testsuite.src/run_misc.at
oguzcankirmemis: As of now, the tests work without the flag, i.e. the signal handler is called. Actually, it was the easiest type of error I could find to create a segfault without any additional configuration in the compiler.

libcob/common.c
GitMensch: The only stuff used without a define here would be `SIGABRT` (I guess that's defined if missing at the start) and `kill` (where the ifdef would need to be placed to (two times).

In this case the function would output "unknown signal" then does the abort handling, then `exit(sig)`.

GitMensch:
The single macos CI failure (869) is because that environment sees the error itself before the generated code can check it; that often happens with sanitizers enabled...
The MSVC ones are rooted in vcpkg deciding it is a good idea to break msbuild projects "most use cmake in any case" (it needs a manual downgrade locally, something I had to do on appveyor months ago, I may need to "just test with CI" multiple times or setup a similar environment locally to check for the proper solution (but instead of a downgrade, one could "copy" the libxml folder to the build_windows folder then as well).
The MSYS1 one seems to build a BDB version that won't work and need to be either patched to do so after the download, or use a `CPPFLAGS=-Dsomething=else` defintion (I'd need to check, possibly with a local build as well).
MSYS2 as an install/update issue, possibly something that @ddeclerck fixed but not checked in yet.

Overall the test results look good.

libcob/common.c
```diff
@@ -1519,10 +1538,23 @@ cob_set_signal (void)
...
+	char signal_regime;
 #ifdef	HAVE_SIGACTION
 	struct sigaction	sa;
 	struct sigaction	osa;
+#else
+	void (*ohdlr) (int);
+#endif
 
+	s = getenv ("COB_SIGNAL_REGIME");
+	signal_regime = s ? *s : '0';
```
GitMensch: if you move those directly up to the declaration as init code, you can even  `const char signal_regime` :-)

libcob/common.c
oguzcankirmemis: There are also two calls to the `signal`, which would need to be covered as well (one in the else condition of HAVE_SIGACTION ifdef, the other right after the call to `longjmp`. I can also surround them with their own ifdefs.

GitMensch: sounds good - but only use "own ifdefs" if we already check them in configure (= they are part of config.h then), oitherwise use `HAVE_SIGNAL_H` for those as well (💤 )

libcob/common.c
oguzcankirmemis: That is on me, didn't know what exactly is allowed in ANSI C with variable declarations. It should be good now :)

libcob/common.c
oguzcankirmemis: I have added the symbol checks for signal and kill to configure instead and used "own ifdefs" as you suggested.

configure.ac
```diff
@@ -711,7 +711,7 @@ AC_FUNC_VPRINTF
 AC_CHECK_FUNCS([memmove memset setlocale fcntl strerror strcasecmp
 	strchr strrchr strdup strstr strtol gettimeofday localeconv
 	getexecname canonicalize_file_name popen raise readlink realpath
-	setenv strcoll flockfile])
+	setenv strcoll flockfile kill signal])
```
GitMensch: memo to myself: there are too many functions in here (`strdup` is not used `memset` and others have to be moved to mandatory or removed in the check, `strcasecmp` is used unconditional but possibly does not exist in MSVC2005 which is still "officially supported" [maybe it is time to drop it, it also seem to be the only environment that does not provide `kill` - and not providing signal.h as well])

As noted - it is enough to check signal_h for those, no need to adjust configure (which would mean you'd need to change build_windows/config.h.in otherwise - just use `HAVE_SIGNAL_H` instead of those two checks and changing those two files).

BTW - just in case you wonder where to check things like that:
https://www.gnu.org/software/gnulib/manual/html_node/kill.html
https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering

oguzcankirmemis: Nice, thanks for the tip! Reverted back to `HAVE_SIGNAL_H`.

GitMensch (approved):
looks good, upstream asap

GitMensch:
@oguzcankirmemis I've just realized that we should have a NEWS entry for this (=user-visible change with as less techspeak as possible). Could you rebase and draft one at the right place (there should be more runtime settings referenced for 3.3 already?

If not then I can have a look at that as well, but that may take some days to get to.

oguzcankirmemis:
I have rebased the PR and updated it with the news entry. I also adjusted the ChangeLogs slightly as well.

oguzcankirmemis:
There is a new failing CI (Windows MSVC), will take a look.

NEWS
```diff
@@ -29,6 +29,10 @@ NEWS - user visible changes				-*- outline -*-
 ** new runtime configuration COB_HIDE_CURSOR, allows to hide the cursor during
    extended ScreenIO operations
 
+** new runtime configuration COB_SIGNAL_REGIME, allows to skip registration of
+   the signal handler of GnuCOBOL. The registration can be skipped either for
+   all signals or for those that do already have a registered handler.
```
GitMensch: nice, maybe add something about its "user purpose" (keeping or registering own/library handlers?)

GitMensch:
Concerning the last commit for the failing MSVC debug runtime:

Last time we tackled MSVC's runtime check @ddeclerck explicit setup its handler to not do any popups; that was back with rev5315 / 816bd2be16d88e9ea9f17f2e047fd08844d7f8c2.

The place for skipping them so far was the CI definition itself, see https://github.com/OCamlPro/gnucobol/blame/d28f9fab8e2712918ec0c15ef1d1fc476be4bcb9/.github/workflows/windows-msvc.yml#L180-L193

I'll add the skipping for that test (and also check how using the debug RTS for generated modules may be done/result in and if the "internal memory" part wouldn't only apply to the debug environment.

oguzcankirmemis:
Ah thank you! I didn't have time yesterday to write here, but yes, after setting up a Windows container myself locally, I have determined that the issue is that Windows CRT in Debug mode prevents the signal from reaching to our (libcob's) signal handler because it thinks it is a critical error. I also agree that "skipping" the test on the CI side makes much more sense.

Should I remove my last commit then? Or do you need something else from me for this PR?

GitMensch:
checked in without the last commit upstream, apart from some minor changes and the missing config/Changelog I've adjusted the runtime enum - because previously you could set signal_regime=ignore - but while being shown as expected (runtime config) it still registered everything (check of 0-2).

PR #247, opened 2025-10-07
Fail on `EC-SIZE` exceptions as per the standard

emilienlemaire:


GitMensch:
Why do we need a new flag instead of -fec?
Note: if not implemented yet we should add the cobol2027 ec-all-fatal to enable only those checks (that needs to iterate over the existing table instead of the current "enable the single/group entry".

Note: no changelog needed for new tests

emilienlemaire:
The new flag is introduced because the current behavior is to continue normally when fatal exceptions are raised and are not checked, which is not the behavior described by the standard.

To avoid breaking programs running on GC3 after the PR is merged I added the flag, so now user can choose how these exceptions should behave, i.e do nothing as the default behavior or fail with the flag set.

GitMensch:
I see but that's exactly what -fec is for: giving fine grained control about which exceptions are checked. We have a known defect that -fec=size does not work in every case and that will be fixed with this PR being finished.
If people compile with -fec=all or --debug and don't want the the working check, they can just add -fno-ec=size

(But we only got multiple people reporting that this does not abort, even with those runtime checks enabled).
Emilien Lemaire ***@***.***> schrieb am Dienstag, 7. Oktober 2025 um 20:35:

emilienlemaire:
I edited the tests that shouldn't work with `-fec=EC-SIZE` to add `-fno-ec=EC-SIZE` so they pass

GitMensch:
While I'll need to check the details (@ddeclerck can possibly help with the all failing CI tests):
* tests/Changelog can be reverted
* `EC-SIZE-EXPONENTIATION` should be included
* for the testcases I'd suggest (but not enforce) using the shortened form ` fno-ec=SIZE` (possibly lower-case)
* run_misc should include a test for all checked `EC-SIZE` - that can go all into a single testcase with multiple programs + `AT_CHECK`s
* NEWS should have a user-view entry for this change

emilienlemaire:
I added the tests for all the exceptions that are set in libcob

emilienlemaire:
I also added a configuration option to enable the failure of the exceptions. I figured many could want to run with `--debug`, yet the COBOL dialect they used should not fail on such exceptions, and this avoid adding `-fno-ec` each time it is the case

cobc/config.def
GitMensch, line 1: we don't need the (currently not documented) cb_fatal_exceptions_fail, so can revert this file and all of config - we already have -fec; in a follow-up PR we can consider adding "default" exceptions per language _if_ we find a need, but I think apart from PACKAED-DECIMAL abends on IBM there's like no default exception.

cobc/codegen.c
```diff
@@ -9150,6 +9205,12 @@ output_stmt (cb_tree x)
 				output_debug_stmts (debug_checks);
 			}
 		}
+
+		if ((!p->ex_handler || (p->ex_handler && p->handler_type != SIZE_ERROR_HANDLER))
+			&& cb_fatal_exceptions_fail) {
+			output_size_exception_handler(p);
+		}
```
GitMensch: Is there a very good reason to add that after the tree is finished?
Wouldn't `cb_emit_arithmetic()` be the right place?

cobc/codegen.c
```diff
@@ -9033,6 +9033,61 @@ output_assign (const struct cb_assign *ap)
...
+					}
+				}
+				if (exception_enabled) {
+					output_block_open ();
+					output_line ("if (cob_last_exception_fatal ())");
+					output_block_open ();
+					output_line ("cob_fatal_error(COB_FERROR_FATAL_EC);");
+					output_block_close ();
+					output_block_close ();
+				}
```
GitMensch: Those exceptions are all fatal - is there a reason to not use the current way of checking by adding a call to a new function `cob_call_size_error (enabled_exceptions)` as error handler if no `[NOT] ON SIZE ERROR` is coded?

tests/testsuite.src/run_misc.at
```diff
@@ -14039,7 +14039,7 @@ AT_SETUP([Constant Expressions])
 AT_KEYWORDS([runmisc condition expression])
 
 AT_DATA([prog.cob], [
-       IDENTIFICATION DIVISION.
+       IDENTIFICATION DIVISION
```
GitMensch: that looks like an error

tests/testsuite.src/run_misc.at
GitMensch, line 1: Can you please make the suppressed exception check as specific as possible? I guess that in most cases this will be `-fno-ec=size-overflow`.

... second thought: Most likely the better move in most cases is to add an `ON SIZE ERROR CONTINUE` to those tests.

tests/testsuite.src/run_misc.at
```diff
@@ -15682,3 +15682,176 @@ AT_CHECK([$COBCRUN_DIRECT ./prog6], [0],
...
+])
+
+AT_CHECK([$COMPILE prog7.cob], [0], [], [])
+AT_CHECK([$COBCRUN_DIRECT ./prog7], [0],
+[SIZE ERROR detected
++1
+], [])
+
+
+AT_CLEANUP
```
GitMensch, line 15979: please add a trailing newline

libcob/ChangeLog
```diff
@@ -1,3 +1,13 @@
+2025-10-07  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
```
GitMensch: please add a leading newline instead of the double-trailing one

GitMensch (changes requested):
I think we getting near to the solution :-)

GitMensch:
> I also added a configuration option to enable the failure of the exceptions. I figured many could want to run with `--debug`, yet the COBOL dialect they used should not fail on such exceptions, and this avoid adding `-fno-ec` each time it is the case

`--debug` is something similar (actually more) that you have with other compiler switches on other compilers where you need to enable checks as well. It is fine to keep as-is and if someone ports he can have a one-time conversion of his old options to the new ones (or let a consultant do that or even better, let someone write a wrapper that takes the old command line options (and possibly environment variables) and calls cobc with the appropriate options).

cobc/config.def
emilienlemaire: So should we add `-fno-ec=size` to all the NIST tests? (if yes where should I do that, I get lost in all the config files from automake and stuff)?

cobc/codegen.c
emilienlemaire: I wanted to this first, but we don't have a lot of info about enabled exceptions at runtime, and I figured it would be more efficient to generate a switch at compile time than having a runtime function check for all the enabled exceptions every time it is called

tests/testsuite.src/run_misc.at
emilienlemaire: Yes I missed it when reading my diffs :sweat_smile:

tests/testsuite.src/run_misc.at
emilienlemaire: I thought about adding the `ON SIZE ERROR`, but in my mind it would change the semantics of the tests which is not something I wanted to do

cobc/config.def
GitMensch: That's interesting, you say that those fail?  And do they with all size exceptions or only with overflow?
To answer your question: that's `COBC_FLAGS` in tests/cobol85/Makefile.am directly after what _should_ be `--debug` [there's currently only one hyphen so a good thing to fix at the same time].

tests/testsuite.src/run_misc.at
GitMensch: Depends on the test -  the ones that do want to check for the exception by the exception functions _should_ have been coded with `ON SIZE ERROR CONTINUE` in any case.
But for the review both suppressing the (specific as possible) check or a the conditional `CONTINUE` or a combination are all fine.

cobc/codegen.c
GitMensch: The exception handler will only be generated _if_ the exception was active - and you can do that with a switch-case statement similar to what you did as well.
Side note: the codegen is a _bit_ special here for the other runtime checks as codegen.c has the direct function call but codeoptim.def optimizes most of them to an inline call which is verySoonTM (when I've split that from other local changes) changed to a define instead. The point for this was: you can fast-read the generated C text which has a "direct function call" for the check (also: diffs to older versions are identical), but the actual code via C preprocessor does a "local" check before calling into libcob for no reason.

For this new check of yours the `switch (cob_glob_ptr->cob_exception_code)` for the specific cases as an *implicit* `ON SIZE ERROR` if no handler is coded at all is the good way for the new check. I haven't checked but you may add that as `on_exception` tree from a `cb_build_direct` or similar (doesn't look that pretty in the generated C but I don't mind).

cobc/config.def
emilienlemaire: Some of them do, around 10 or 20 if i recall correctly, I'll test again without the new config to get the exact number

cobc/codegen.c
emilienlemaire: We don't have info about the exception handler in `cb_emit_arithmetics()`, so this would mean adding this information in the function, which I am not sure we want to do, and the checks need to be done after the statement, not during intermediate calculations, where the exceptions are to be set, but not checked yet.

cobc/codegen.c
emilienlemaire: Sorry but I don't really understand what you mean or ask for here (I am not that knowledgeable on codegen yet, but getting there :sweat_smile: )...

Here was my trail of thoughts for implementing the checks this way:

What I meant to do was to keep the code as close as possible as if we had a `ON SIZE ERROR` (the `NOT ON` does not really matter in our case, as even if it is present we need to check for exceptions that were set), or if you will as if the default was an implicit `ON SIZE ERROR` that terminates the program in case of an exception. So I based my code from what the `ON SIZE ERROR` codegen did (and also where it was).

Then I wondered how to make it so there was no code generated if not needed, and how to make it as efficient as possible, because I though this part might be executed quite a lot during a run of any program. So I thought having a switch that only check for the activated exception codes. But I also saw that `COB_EC_SIZE_IMP` (although never set in libcob) is not fatal, so I just added a check for fatal errors.

(I made this recap so you can see why I did it this way)

Having written my recap and re-reading your message, I think I understand what you meant better. Instead of the switch, you'd like me to make a function `cob_call_size_error (enabled_exceptions)` (which would still mean having enabled exceptions registered somewhere to be used at runtime), which can be optimized away using `codeoptim.def` when needed, this codeoptim could itself have a generated switch which would be filled accordingly with enabled exceptions, and then there wouldn't be any need for runtime checks of enabled exceptions?

emilienlemaire:
Hello @GitMensch,
I am trying to implement what I think you asked for in [this comment](https://github.com/OCamlPro/gnucobol/pull/247#discussion_r2417255380), but as I mentioned it is not that easy, as we don't have knowledge of what exceptions have been disabled at runtime.

I made [this commit](https://github.com/OCamlPro/gnucobol/pull/247/commits/25ef5e1336c0888b3aed311964904f753bde3204) (which will be edited), and the runtime knowledge is the last part I have to add, so I was wondering how I should add the information on the exceptions that enabled / disabled ?

GitMensch:
👋 Emilien and thanks again for working on this.

Concerning your quuestion - the information about the checked exceptions is codegen'd into the module - it can even be different "per line" using the TURN directives.
The runtime just always sets the specific exception code and the generated code checks per "what is active":

* if there are tests<, then before the actual statement the same value that is tested for later on (normally checked by its group) is reset
* then the code in the runtime is executed which would set those values
* then the exception values actually active on this line will have a generated test for

that approach _should_ be visible if you now generate with `ON SIZE ERROR` - it "just" needs to be added for the error case without that clause

Your last commit seems to do that quite cleanly, no?

How does the generated code with for example a check for divzero look like?

Did you do anything to the conditions (`IF MY-VAR / MY-OTHER-VAR` may create a divzero) as well or is that open?

What are the parts this PR needs a direction / review on?

Note: instead of a `cob_call_size_error (void)` it seems better top have something more general like `cob_fatal_exception (int exception_code)`.

emilienlemaire:
Hello Simon!

Thank you for the help!

I updated my commits, and hopefully we are getting closer to get it merged :smile: 

Concerning the `IF MY-VAR / MY-OTHER-VAR` I have not yet added any code that can handle that, I'll look into it.

I think I made the changes you were looking for, yet this pose a problem: when any `EC-SIZE` exception is enabled then the check is valid for all `EC-SIZE` errors, this was something that was avoided with my initial `switch` generation. This can also be avoided by generating a check for enabled error at compile time, but I don't think it would be a good idea to generate this inside `parser.y`, so this would mean moving the code generation to codegen.c and do something closer to what I did before (excluding the configs and flags I added, which we can forget about).

GitMensch:
Why do the changes here fail one of the NC tests? did the behavior changed between standards (and we miss a possible distinction between COBOL85 and 2002+) or is there an error in the NIST test or does it depend on undefined behaviour ... or there is an error in the PR itself?

Note: In _all_ cases: please try to "extract" the failure into one of the tests in our own testsuite before working on fixing it,
Thanks for your efforts in this important area (and note that this is one of the missing changes we need this year before the 3.3rc and after that we want to sent the final out asap [=if no breaking tests occur within 2- weeks]).

> Concerning the IF MY-VAR / MY-OTHER-VAR I have not yet added any code that can handle that, I'll look into it.

Thanks. In effect it should have a "post check" generated that inspects the error code - similar as the generation without the SIZE ERROR clause.

For the EC-SIZE error problem - that's no issue as long as the check either:
* generates specific check for the enabled EC-SIZEs
* generates a check that will go into the runtime for any EC-SIZE and passes the EC-SIZE values needed into the function

Just a fair warning: I'm likely not be available for any (or any longer) reviews this week (and also did not checked the current changes either, just the CI results).

emilienlemaire:
There was no `EXCEPTION` in the COBOL85 standard, and I forgot to update the compilation options in my commit, hence the NIST failures (which are remediated by the last commit).

ABOUT-NLS
GitMensch, line 1: change should be reverted, the current version is the replacement link (the file is checked in our current minimal autoconf for its availability so we use that one)

NEWS
```diff
@@ -6,6 +6,19 @@ NEWS - user visible changes				-*- outline -*-
 
 * New GnuCOBOL features
 
+** at runtime, fatal size exception codes now fail when activated (with
+   `-fec=ALL` or `-fec=SIZE or its subordinates). Theses exceptions can be
+   caught with `ON SIZE ERROR` phrases to avoid the program failing when
+   they are set. The affected exceptions codre are:
```
GitMensch: typo, should be: "exception codes"

NEWS
```diff
@@ -6,6 +6,19 @@ NEWS - user visible changes				-*- outline -*-
 
 * New GnuCOBOL features
 
+** at runtime, fatal size exception codes now fail when activated (with
+   `-fec=ALL` or `-fec=SIZE or its subordinates). Theses exceptions can be
```
GitMensch: not sure if we use backticks in that file, if yes, then the closing one for ec=SIZE is missing
As this paragraph is a sentence it should start uppercase

cobc/ChangeLog
```diff
@@ -1,3 +1,10 @@
+2025-11-10  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
```
GitMensch: one empty line before and after the Changelog entry, please (currently it is 0/2) and indent continuation lines by two spaces

cobc/codegen.c
```diff
@@ -8718,6 +8718,47 @@ output_ec_condition_for_handler (const enum cb_handler_type handler_type)
...
+	if (CB_EXCEPTION_ENABLE(COB_EC_SIZE_UNDERFLOW)) {
+		output_line("if (cob_glob_ptr->cob_exception_code == %d) "
+			"cob_fatal_exception (cob_glob_ptr->cob_exception_code);",
+			CB_EXCEPTION_CODE(COB_EC_SIZE_UNDERFLOW));
+	}
+	if (CB_EXCEPTION_ENABLE(COB_EC_SIZE_ZERO_DIVIDE)) {
+		output_line("if (cob_glob_ptr->cob_exception_code == %d) "
+			"cob_fatal_exception (cob_glob_ptr->cob_exception_code);",
+			CB_EXCEPTION_CODE(COB_EC_SIZE_ZERO_DIVIDE));
+	}
```
GitMensch: can't you just have an integer `exception_code = 0` at the start of the function, then || the enabled exception values into it and then finally - if it isn't zero output one single exception check?

cobc/codegen.c
```diff
@@ -9151,6 +9192,15 @@ output_stmt (cb_tree x)
...
 			}
 		}
+
+		if (p->handler_type == SIZE_ERROR_HANDLER && p->ex_handler == NULL) {
+			output_ec_size_handler ();
+		}
+
+		/* if ((!p->ex_handler || (p->ex_handler && p->handler_type != SIZE_ERROR_HANDLER))) {
+			output_size_exception_handler(p);
+		} */
```
GitMensch: if not relevant any more then please delete, otherwise move to `#if 0 /* comment why we may want that */`

cobc/parser.y
```diff
@@ -18134,13 +18134,40 @@ on_size_error_phrases:
...
   {
 	/* no [NOT] ON SIZE ERROR is specified (= no explicit handling) */
+	if (CB_EXCEPTION_ENABLE(COB_EC_SIZE)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_ADDRESS)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_EXPONENTIATION)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_IMP)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_OVERFLOW)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_TRUNCATION)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_UNDERFLOW)
+		|| CB_EXCEPTION_ENABLE(COB_EC_SIZE_ZERO_DIVIDE))
```
GitMensch: style-only:

indent here and below: || by one space, not tab/multiple space, and ideally a space before the macro's (

cobc/parser.y
```diff
@@ -18134,13 +18134,40 @@ on_size_error_phrases:
...
+		current_statement->ex_handler = NULL;
+	}
   }
 | on_size_error _not_on_size_error
 | not_on_size_error _on_size_error
   {
 	if ($2) {
 		cb_verify (cb_not_exception_before_exception,
 			_("NOT SIZE ERROR before SIZE ERROR"));
+	} else if ($1 && $1 != cb_int1) {
```
GitMensch, line 18155: $1 here is always set as `not_on_size_error` is mandatory - you can drop the check for this and the setting to cb_int2 below

Nice and correct addition for that "null handler" in the parser BTW!


... but: it should not be added here by the standard (I think also by previous standards).
To quote: 

> If the EC-SIZE exception condition was raised and a SIZE ERROR phrase is not specified, processing
proceeds as specified in 14.6.13.1.3, Fatal exception conditions. If a NOT SIZE ERROR phrase is specified,
it is ignored.

The point here if one explicit codes `NOT SIZE ERROR` than this is identical to a coded `ON SIZE ERROR CONTINUE`

libcob/call.c
```diff
@@ -19,6 +19,7 @@
 */
 
 
+#include "common.h"
```
GitMensch: possibly added by clangd - can be dropped as it is explicit included via coblocal.h (and should always be included after config.h in any case)

tests/ChangeLog
GitMensch, line 1: duplicated entries, can be dropped

tests/testsuite.src/data_display.at
GitMensch, line 1: please make -fno-ec more specific here and in the other testcases as well

tests/testsuite.src/run_misc.at
```diff
@@ -2689,8 +2689,10 @@ AT_DATA([prog.cob], [
        01  X            PIC 9(2) VALUE 0.
        01  Y            PIC 9(2) VALUE 0.
        PROCEDURE        DIVISION.
-           COMPUTE X = 100. *> this overflows which does not store a result
-           COMPUTE Y = 99.  *> and once in a time this result was not stored
+           COMPUTE X = 100 ON SIZE ERROR CONTINUE.
+           *> this overflows which does not store a result
+           COMPUTE Y = 99 ON SIZE ERROR CONTINUE.
```
GitMensch: Y has no size error, so should not have that clause in

tests/testsuite.src/run_misc.at
```diff
@@ -11733,7 +11737,7 @@ AT_DATA([prog.cob], [
        END PROGRAM prog.
 ])
 
-AT_CHECK([$COMPILE prog.cob], [0], [],
+AT_CHECK([$COMPILE -fno-ec=size-overflow prog.cob], [0], [],
```
GitMensch: With the ON SIZE above there is no need for -fno-ec here any more, is there?

tests/testsuite.src/run_misc.at
```diff
@@ -15735,5 +15743,177 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+                GIVING RESULT.
+           DISPLAY RESULT.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE prog.cob], [0], [], [])
+AT_CHECK([$COBCRUN_DIRECT ./prog], [1],
+[], [libcob: prog.cob:10: error: Fatal exception not handled: EC-SIZE-OVERFLOW
+])
+
```
GitMensch, line 15812: please add a prog0 here that is identical but has NOT ON SIZE ERROR clause - which should pass without an exception

libcob/ChangeLog
```diff
@@ -15,6 +15,16 @@
...
 
+2025-11-10  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
+
+	* common.h, common.c: add `cob_exception_tab_fatal`, which informs
+	whether an exception is fatal or not
+	* common.h, common.c: add `cob_last_exception_fatal`, to check if the
+	last raised exception was fatal
+	* common.h, common.c: add enum member `COB_FERROR_FATAL_EC` for
+	`struct cob_fatal_error` which can be used when terminating the
+	program due to a fatal exception
```
GitMensch, line 61: style: indent by two spaces for continuation lines and  line-break before col81 (it looks like some words may wrap too early)

libcob/common.c
```diff
@@ -468,6 +468,14 @@ static const int		cob_exception_tab_code[] = {
...
 };
 
+#undef	COB_EXCEPTION
+#define COB_EXCEPTION(code, tag, name, critical)	critical,
+static const int		cob_exception_tab_fatal[] = {
+	0,		/* COB_EC_ZERO */
+#include "exception.def"
+	0		/* COB_EC_MAX */
+};
+
```
GitMensch: This and the related vars are not needed any more as there's now a new single function called, right?
So please revert the now not needed changes here and in common.h.

libcob/common.c
```diff
@@ -2502,6 +2511,12 @@ cob_set_exception (const int id)
 	}
 }
 
+int
+cob_last_exception_fatal() {
```
GitMensch: `cob_last_exception_was_fatal (void)`, please - but that isn't used and needed any more, is it?

GitMensch (changes requested):
Just stumbled over an interesting explicit rule in COBOL2027 (before it possibly wasn't specified, would have to check): if the EC-SIZE happens on "item identification", then the ON SIZE phrase is not used but the fatal exception - if enabled should be done.

That is the part that will be covered by the "expression addition" - we should definitely add it to our testsuite - even with an expected failure:


```
MOVE 0 TO SOME
ADD TABLE-VAR (5 / SOME) TO ADD TABLE-VAR (5)
ON SIZE ERROR DISPLAY 'Should not be taken'.
```

The changes look good in general, but apart from some style and other minor issues there's one issue that is logically wrong - see the parser comment about NOT ON SIZE ERROR; I _guess_ that some of the NIST tests don't need the -fno-ec=size flags because of that - but this can be easily checked after adjusting parser.y

cobc/codegen.c
emilienlemaire: We can't really have that, if the exception is not zero, it might still be an exception code that was disable. I've had the problem where I had `-fno-ec=size-overflow` yet it was still raised as other `ec-size` codes were enabled, and the check was a `||` (so if any was enabled, all were checked). This was the simplest way I could ensure that only the enabled exceptions are raised.

cobc/parser.y
emilienlemaire: I think this phrase means that if the exceptions exists and the `NOT SIZE ERROR` phrase is specified, then the `NOT SIZE ERROR` phrase is ignored, not that the exception is ignored

GitMensch: I'm sure this isn't the case. You may try some computers, but the ones I've tested behave like my understanding of the standard. 
Logically: coding only NOT ON SIZE ERROR means you kind of expect but ignore one and execute the "important" part only if there isn't an error. 
Also note that coding it means (per standard) that the target field is unchanged if an error happens, otherwise it is undefined if no checking is on.

emilienlemaire: Sure but the phrasing is `If a NOT SIZE ERROR phrase is specified, it is ignored.`, not `If a NOT SIZE ERROR phrase is specified, the exception is ignored.` so `it` must refer to the subject of the phrase, not something that was in a phrase before.

In addition this phrase appears after the description of what should happen in case of an exception without an `ON SIZE ERROR` phrase, and after specifically saying `and processing proceeds as specified in ...` so logically every steps the standard specify should happen before any consideration for `NOT ON SIZE ERROR`.

Afaik not many (if any) implementers have an exception system in their compiler, so maybe they work differently when having `NOT ON SIZE ERROR`, but if we decide to implement one, we should adhere to the standard as close as possible.

GitMensch: let's take the following approach: we surround that with `#if 0 /* CHECKME: possible dialect configuration */` and get the code upstream, then check what implementations do and also with the standard committee what the intention is

cobc/codegen.c
GitMensch: marked as resolved - then please commit your change solving that

cobc/codegen.c
GitMensch: let's go with it as you have it, along with the minimal -fno-ec in the NIST configuration - I can then have a look if my idea for making it "more direct" works after this code isupstream

cobc/parser.y
```diff
@@ -18176,6 +18203,7 @@ _not_on_size_error:
 not_on_size_error:
   NOT_SIZE_ERROR statement_list
   {
+	$$ = cb_int2;
```
GitMensch: as noted, this should not be needed, even if we enable the handler insertion as you've did

ABOUT-NLS
GitMensch: please commit the resolving change

GitMensch (changes requested):
apart from the things mentioned in the PR this is good for upstream; please commit the missing adjustments and I can make it available

emilienlemaire:
The PR should be OK to merge now, I should have edited everything you asked, and also fixed a few side effects those edits lit up, the main one being to reset the check of the exception before a new statement but still allow checking the exception with intrinsic functions (I checked on the standard, and the exception check should be reset between every statement, but the last exception code and stuff should still be available with intrinsics)

ABOUT-NLS
GitMensch, line 1: this still needs to be reverted

tests/cobol85/report.pl
```diff
@@ -236,6 +236,30 @@
...
+$cobc_flags{NC173A} = "-fno-ec=size-overflow -fno-ec=size-zero-divide";
+$cobc_flags{NC175A} = "-fno-ec=size-overflow";
+$cobc_flags{NC176A} = "-fno-ec=size-overflow";
+$cobc_flags{NC177A} = "-fno-ec=size-overflow";
+$cobc_flags{NC202A} = "-fno-ec=size-overflow";
+$cobc_flags{NC203A} = "-fno-ec=size-zero-divide";
+$cobc_flags{NC205A} = "-fno-ec=size-overflow";
+$cobc_flags{NC251A} = "-fno-ec=size-zero-divide";
+$cobc_flags{NC252A} = "-fno-ec=size-overflow";
+$cobc_flags{NC253A} = "-fno-ec=size-overflow";
```
GitMensch: Are all of those (after the change to NOT SIZE ERROR) still needed?
Are there some missing (as NIST NC currently fails in CI)?

GitMensch (changes requested):
Thanks. Currently NIST NC fails, but I guess you'll push a fix soon.

emilienlemaire:
Yes I'm working on a fix

ABOUT-NLS
emilienlemaire: I thought I did :smiling_face_with_tear:, I'll add it with the fix on NC

tests/cobol85/report.pl
emilienlemaire: I tested again, some of them can be removed, but some errors occurs without a `NOT ON SIZE ERROR`, so we need to keep the one I updated in the last commit

GitMensch:
Something is broken. If we compile NC2032A with -fec=all (or just --debug) it aborts with

~~~
libcob: NC203A.CBL:307: error: Fatal exception not handled: EC-SIZE-ZERO-DIVIDE
libcob: NC203A.CBL:307: warning: implicit CLOSE of PRINT-FILE ('REPORT')

 Last statement of "NC203A" was ADD
        PASS OF SECT-NC203A-001 at NC203A.CBL:307
        DIV-TEST-F4-4-0 OF SECT-NC203A-001 at NC203A.CBL:483
        ENTRY NC203A at NC203A.CBL:295
 Started by ./NC203A
 ~~~
 
 The perform in 403 is

```cobol 
048300     DIVIDE  25COUNT INTO 100 GIVING 25ANS REMAINDER 25REM        NC2034.2
048400             ON SIZE ERROR                                        NC2034.2
048500             PERFORM PASS                                         NC2034.2   *>   here
048600             GO TO DIV-WRITE-F4-4-0.                              NC2034.2
```
 the ADD is 
```cobol
 030900 PASS.  MOVE "PASS " TO P-OR-F.  ADD 1 TO PASS-COUNTER.           NC2034.2
```

so the `DIVIDE` error which is correctly handled by its `ON SIZE ERROR` is evaluated witht the next statement `ADD`.

***Before any change in codegen it would good to have a similar example in the testsuite!***


The reason is that _before_ the statements that check the exceptions after their execution we need to reset it, which is done in codegen.c by those lines:

```c
		if (!p->file && (p->ex_handler || p->not_ex_handler)) {
			output_line ("cob_glob_ptr->cob_exception_code = 0;");
		}
```
which are not taken because no handler is set.

I suggest to add a check `p->handler_type == SIZE_ERROR_HANDLER` (+active ec-size) there and drop `cob_reset_exception`.

This should lead to much more (all?) NC tests not needing `-fno-ec=size`.


For the ec handler output I've meant something along

```c
static void
output_ec_size_handler (void)
{
	int ec_checked = 0;
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ADDRESS)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_ADDRESS);
	}
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_EXPONENTIATION)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_EXPONENTIATION);
	}
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_IMP)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_IMP);
	}
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_OVERFLOW)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_OVERFLOW);
	}
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_TRUNCATION)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_TRUNCATION);
	}
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_UNDERFLOW)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_UNDERFLOW);
	}
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE)) {
		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_ZERO_DIVIDE);
	}
	if (ec_checked) {
		output_line("if ((cob_glob_ptr->cob_exception_code & 0x%04x) == 0x%04x)",
			CB_EXCEPTION_CODE (COB_EC_SIZE), CB_EXCEPTION_CODE (COB_EC_SIZE));
		output_line("\t" "cob_fatal_exception (cob_glob_ptr->cob_exception_code);");
	}
}
```

... but then possibly also have `cob_fatal_exception` not take an argument (it is only called if there is an error and operates on the last error).

emilienlemaire:
> For the ec handler output I've meant something along
> 
> ```c
> static void
> output_ec_size_handler (void)
> {
> 	int ec_checked = 0;
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ADDRESS)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_ADDRESS);
> 	}
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_EXPONENTIATION)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_EXPONENTIATION);
> 	}
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_IMP)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_IMP);
> 	}
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_OVERFLOW)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_OVERFLOW);
> 	}
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_TRUNCATION)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_TRUNCATION);
> 	}
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_UNDERFLOW)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_UNDERFLOW);
> 	}
> 	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE)) {
> 		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_ZERO_DIVIDE);
> 	}
> 	if (ec_checked) {
> 		output_line("if ((cob_glob_ptr->cob_exception_code & 0x%04x) == 0x%04x)",
> 			CB_EXCEPTION_CODE (COB_EC_SIZE), CB_EXCEPTION_CODE (COB_EC_SIZE));
> 		output_line("\t" "cob_fatal_exception (cob_glob_ptr->cob_exception_code);");
> 	}
> }
> ```

On this, I was doing something close to this at one point, but let's say we have only `EC-SIZE-OVERFLOW` enabled, and we compute something that sets a `EC-SIZE-ZERO-DIVIDE`, which is set (among other places), with theses lines:
```c
	if (unlikely (mpz_sgn (d2->value) == 0)) {
		d1->scale = COB_DECIMAL_NAN;
		cob_set_exception (COB_EC_SIZE_ZERO_DIVIDE);
		return;
	}
```
then the generated condition will be true, and the exception will be wrongly raised, even if it was disabled.

As for the exceptions that were wrongly raised in handler statements, the last commit should fix it :smile:

GitMensch:
> On this, I was doing something close to this at one point, but let's say we have only `EC-SIZE-OVERFLOW` enabled, and we compute something that sets a `EC-SIZE-ZERO-DIVIDE`, which is set (among other places), with theses lines:
> 
> ```c
> 	if (unlikely (mpz_sgn (d2->value) == 0)) {
> 		d1->scale = COB_DECIMAL_NAN;
> 		cob_set_exception (COB_EC_SIZE_ZERO_DIVIDE);
> 		return;
> 	}
> ```

Of course my code was wrong, should have been something like 

```c
	if (ec_checked) {
		output_line("if ((cob_glob_ptr->cob_exception_code & 0x%04x) == 0x%04x) && ((cob_glob_ptr->cob_exception_code & 0x%04x) != 0))",
			CB_EXCEPTION_CODE (COB_EC_SIZE), CB_EXCEPTION_CODE (COB_EC_SIZE)), (ec_checked & 0x00FF);
```

--> group is set, and one of the |= exception bits are set

emilienlemaire:
I implemented the check you suggested, but had to edit the exception codes in `exception.def` to avoid overlapping bits in the checks

GitMensch:
As I can't push to your repo - @emilienlemaire please rebase.

emilienlemaire:
Hi @GitMensch, I've rebased the branch :smile:

GitMensch:
Thanks, that's one of my next things to re-inspect.
Would you prefer to upstream it yourself if it is fine or should I do that?

emilienlemaire:
I've mistakenly committed a file with a conflict in my rebase, I'll modify it, you can then upstream it, I'll ping you when its updated

emilienlemaire:
@GitMensch you can upstream it now, or if you wish Fabrice told me he could also upstream it

cobc/ChangeLog
```diff
@@ -122,6 +122,13 @@
 	* ppparse.y: error on extraneous periods at the end of SET
 	  directives; turn to a warning in relaxed syntax mode.
 
+2025-11-24  Emilien Lemaire <emilien.lemaire@ocamlpro.com>
+
+	* codegen.c (output_ec_size_handler): generate a check for a
+	size exception on statements that can raise them
+	* codegen.c (output_stmt): add a check for `ON SIZE ERROR` phrase, and
+	if not present  then generate a check for enabled size exceptions
```
GitMensch, line 130:
```suggestion
	* codegen.c (output_ec_size_handler): generate check for a size exception
	  on statements that can raise them
	* codegen.c (output_stmt): add a check for `ON SIZE ERROR` phrase, and
	  if not present then generate a check for enabled size exceptions
```

cobc/codegen.c
```diff
@@ -8720,6 +8720,44 @@ output_ec_condition_for_handler (const enum cb_handler_type handler_type)
...
+		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_UNDERFLOW);
+	}
+	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE)) {
+		ec_checked |= CB_EXCEPTION_CODE (COB_EC_SIZE_ZERO_DIVIDE);
+	}
+	if (ec_checked) {
+		output_line (
+			"if (((cob_glob_ptr->cob_exception_code & 0x%04x) == 0x%04x)"
+			" && ((cob_glob_ptr->cob_exception_code & 0x%04x) != 0)"
+			" && (cob_glob_ptr->cob_got_exception > 0))",
```
GitMensch, line 8752: you possibly want a \n at the end of those lines, but on the other hand that will likely have bad indent if we're already nested - the alternative is to use output_line multiple times

cobc/codegen.c
```diff
@@ -8720,6 +8720,44 @@ output_ec_condition_for_handler (const enum cb_handler_type handler_type)
...
+			CB_EXCEPTION_CODE (COB_EC_SIZE),
+			CB_EXCEPTION_CODE (COB_EC_SIZE),
+			ec_checked & 0x00FF);
+		output_line ("\t" "cob_fatal_exception (cob_glob_ptr->cob_exception_code);");
+	}
+
+}
```
GitMensch, line 8759:
```suggestion
	}
}
```

GitMensch: maybe end the part early by `if (!ec_checked) return;` above

cobc/codegen.c
```diff
@@ -8728,6 +8766,11 @@ output_handler (const struct cb_statement *stmt)
 		return;
 	}
 
+	if (stmt->ex_handler || stmt->not_ex_handler) {
+		/* We have a handler, so the exceptions should not be raised, and we reset them
+		 * before in case the handling statements can also raise exceptions. */
+		output_line("cob_reset_exception ();");
```
GitMensch, line 8772:
```suggestion
		output_line ("cob_reset_exception ();");
```

cobc/codegen.c
```diff
@@ -9175,6 +9220,13 @@ output_stmt (cb_tree x)
 				output_debug_stmts (debug_checks);
 			}
 		}
+
+		if (p->handler_type == SIZE_ERROR_HANDLER
+		     && p->ex_handler == NULL
+		     && p->not_ex_handler == NULL) {
```
GitMensch, line 9226:
```suggestion
		if (p->handler_type == SIZE_ERROR_HANDLER
		 && p->ex_handler == NULL
		 && p->not_ex_handler == NULL) {
```

cobc/parser.y
```diff
@@ -18338,13 +18338,40 @@ on_size_error_phrases:
...
   {
 	/* no [NOT] ON SIZE ERROR is specified (= no explicit handling) */
+	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ADDRESS)
+	     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_EXPONENTIATION)
+	     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_IMP)
+	     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_OVERFLOW)
+	     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_TRUNCATION)
+	     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_UNDERFLOW)
+	     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE))
+	{
```
GitMensch, line 18348:
```suggestion
	if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ADDRESS)
	 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_EXPONENTIATION)
	 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_IMP)
	 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_OVERFLOW)
	 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_TRUNCATION)
	 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_UNDERFLOW)
	 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE)) {
```
Can''t we test on the whole EC-SIZE group here?

cobc/parser.y
```diff
@@ -18338,13 +18338,40 @@ on_size_error_phrases:
...
+		   check on exceptions */
+#if 0 /* CHECK ME: possible dialect configuration
+		if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ADDRESS)
+		     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_EXPONENTIATION)
+		     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_IMP)
+		     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_OVERFLOW)
+		     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_TRUNCATION)
+		     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_UNDERFLOW)
+		     || CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE))
+		{
```
GitMensch, line 18370:
```suggestion
		if (CB_EXCEPTION_ENABLE (COB_EC_SIZE_ADDRESS)
		 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_EXPONENTIATION)
		 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_IMP)
		 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_OVERFLOW)
		 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_TRUNCATION)
		 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_UNDERFLOW)
		 || CB_EXCEPTION_ENABLE (COB_EC_SIZE_ZERO_DIVIDE)) {
```
if we can check the group above, then also here (and possibly add a note in the TODO), if we expect that this won't be checked soon after upstream

libcob/call.c
```diff
@@ -1095,6 +1095,13 @@ cob_call_error (void)
 	cob_hard_failure ();
 }
 
+void
+cob_fatal_exception (const int exception_code) {
+	cob_runtime_error("Fatal exception not handled: %s",
+			cob_get_last_exception_name());
```
GitMensch, line 1101:
```suggestion
	cob_runtime_error (_("fatal exception not handled: %s"),
			cob_get_last_exception_name ());
```
we commonly start lowercase and need a translation here

libcob/ChangeLog
```diff
@@ -75,11 +75,30 @@
 	* fileio.c (cob_file_close): close file depending on internal state, not
 	  depending on file organization
 
+
```
GitMensch, line 78:
```suggestion
```

libcob/common.h
```diff
@@ -1682,6 +1685,8 @@ COB_EXPIMP void		print_runtime_conf	(void);
 
 COB_EXPIMP void		cob_set_exception	(const int);
 COB_EXPIMP int		cob_last_exception_is	(const int);
+COB_EXPIMP int		cob_last_exception_fatal(void);
```
GitMensch, line 1688:
```suggestion
COB_EXPIMP int		cob_last_exception_fatal (void);
```

libcob/common.c
```diff
@@ -9365,6 +9372,10 @@ cob_fatal_error (const enum cob_fatal_error fatal_error)
 	case COB_FERROR_JSON:
 		cob_runtime_error (_("attempt to use non-implemented JSON I/O"));
 		break;
+	case COB_FERROR_FATAL_EC:
+		cob_runtime_error(_("unhandled fatal exception code: %s"),
```
GitMensch, line 9376:
```suggestion
		cob_runtime_error (_("unhandled fatal exception code: %s"),
```

libcob/common.c
```diff
@@ -466,8 +466,6 @@ static const int		cob_exception_tab_code[] = {
 	0		/* COB_EC_MAX */
 };
 
-#undef	COB_EXCEPTION
-
```
GitMensch, line 470: as you don't use that below, please keep the undefine = "clean space only temporarily needed to import a def file"

libcob/common.c
```diff
@@ -2556,6 +2555,14 @@ cob_set_exception (const int id)
 	}
 }
 
+/* reset exception for the exception checker only */
+void
+cob_reset_exception (void) {
+	if (cobglobptr->cob_got_exception) {
+		cobglobptr->cob_got_exception = -1;
+	}
```
GitMensch, line 2563: that's a small one and `cobglobptr` is ABI - consider to do that in the generated code directly

libcob/common.h
GitMensch, line 1603: please recheck, I _think_ elements of this are used in the generated code - like the call_params ->to not break existing binaries you'd then need to end all fields to the end (with a comment to reorder in GC4)

libcob/ChangeLog
```diff
@@ -75,11 +75,30 @@
...
+	* common.h, common.c: add `cob_reset_exception` which disable an
+	  exception check after one was performed, but still allows the
+	  exception intrinsic functions to check for them
+	* common.h, common.c: add `cob_last_exception_fatal`, to check if the
+	  last raised exception was fatal
+	* common.h, common.c: add enum member `COB_FERROR_FATAL_EC` for
+	  `struct cob_fatal_error` which can be used when terminating the
+	  program due to a fatal exception
+
```
GitMensch, line 94:
```suggestion

	* exception.def: explain thg change
	* common.h, common.c (cob_reset_exception, cob_last_exception_fatal):
	  new functions to disable an exception check after one was performed,
	  but still allowing the exception intrinsic functions to check for them
	  and to check if the last raised exception was fatal
	* common.h, common.c: add enum member `COB_FERROR_FATAL_EC` for
	  `struct cob_fatal_error` which can be used when terminating the
	  program due to a fatal exception
```

libcob/ChangeLog
```diff
@@ -75,11 +75,30 @@
...
+
 2025-11-19  Chuck Haatvedt <chuck.haatvedt+cobol@gmail.com>
 
 	* mlio.c (xml_startDocument, myStructuredErrorHandler, xml_parse):
 	  compat for LIBXML_VERSION < 21400
 
+
```
GitMensch, line 101:
```suggestion
```

tests/cobol85/report.pl
```diff
@@ -236,6 +236,7 @@
 #$cobc_flags{DB203A} = "-fno-ec=data-incompatible";
 #$cobc_flags{DB204A} = "-fno-ec=data-incompatible";
 
+
```
GitMensch, line 239:
```suggestion
```

tests/testsuite.src/data_binary.at
```diff
@@ -2261,7 +2261,7 @@ AT_DATA([prog.cob], [
            STOP RUN.
 ])
 
-AT_CHECK([$COMPILE prog.cob], [0], [], [])
+AT_CHECK([$COMPILE -fno-ec=size-overflow prog.cob], [0], [], [])
```
GitMensch, line 2264: Better: add `ON OVERFLOW CONTINUE` to the ones above we expect to overflow and undo the flag change

GitMensch (changes requested):
I'd tend to add the explicit ON OVERFLOW / ON SIZE ERROR to the computations int he testsuiite instead of the flag.... otherwise some comments below; overall this is in a good state, I'll have to run some local tests once finished.

PR #248, opened 2025-10-12
Print with overflow multiline

rbowler:
### Summary
This pull request introduces an enhancement to the **-ftcmd** functionality of cobc. The changes ensure that when the cobc command line exceeds the page width of the compiler listing, it is properly wrapped and split across as many continuation lines as are necessary.

### Changes Included
- Updated **print_with_overflow** routine to support multiline overflow.
- Added new testcase to verify correct handling of multiline overflow.
- Improved code comments related to multiline printing.

### Motivation
The previous implementation printed the cobc command line with a maximum of one continuation line. The rest of the command line was truncated, as in this example:  
```
command line:
  cobc -fttitle=GnuCOBOL_V.R.P -fno-ttimestamp -q -std=default -Wall
+ -fno-tmessages -fsyntax-only -t- -fno-tsymbols -ftcmd -fdiagnostics-show-opti
```

This update addresses that issue, producing output like this:  
```
command line:
  cobc -fttitle=GnuCOBOL_V.R.P -fno-ttimestamp -q -std=default -Wall
+ -fno-tmessages -fsyntax-only -t- -fno-tsymbols -ftcmd
+ -fdiagnostics-show-option -fdiagnostics-show-line-numbers
+ -fdiagnostics-absolute-paths -fno-remove-unreachable prog
```

### Impact
- Users will be able to see all of the compile options in the compiler listing.
- No breaking changes; existing print functionality remains compatible.

### Testing
- New testcase added to **tests/testsuite.src/listings.at** as part of the **[command line]** test.

---

Please review and provide feedback or suggestions.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,3 +1,4 @@
+
 2025-10-17  Roger Bowler <rbowler@snipix.net>
 
 	* cob.c (print_with_overflow): Print multiple continuation lines on listing
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,3 +1,6 @@
+2025-10-17  Roger Bowler <rbowler@snipix.net>
+
+	* cob.c (print_with_overflow): Print multiple continuation lines on listing
 
 2025-07-30  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -7000,31 +7000,81 @@ static void
 print_with_overflow (const char *prefix, char *content)
 {
 	const unsigned int	max_chars_on_line = cb_listing_wide ? 120 : 80;
-	int offset;
+	size_t			prefix_len_raw;
+	size_t			prefix_len;
+	const char		*remaining;
+	unsigned int		allowed;
+	size_t			rlen;
+	size_t			breakpos;
 
-	offset = snprintf (print_data, max_chars_on_line, "%s%s", prefix, content);
-	if (offset >= 0) {	/* snprintf returns -1 in MS and on HPUX if max is reached */
-		pd_off = offset;
-	} else {
-		pd_off = max_chars_on_line;
-		print_data[max_chars_on_line - 1] = 0;
-	}
-	if (pd_off >= max_chars_on_line) {
-		size_t prefix_offset;
-		/* trim "current line" on last space */
-		pd_off = strlen (print_data) - 1;
-		while (pd_off && !isspace ((unsigned char)print_data[pd_off])) {
-			pd_off--;
-		}
-		print_data[pd_off] = '\0';
+	/* determine visual width for prefix and continuation alignment */
+	prefix_len_raw = strlen (prefix);
+	prefix_len = (prefix_len_raw < 2) ? 2 : prefix_len_raw;
+	remaining = content;
+
+	/* First line: print with original prefix */
+	allowed = (unsigned int)(max_chars_on_line - prefix_len);
+	rlen = strlen (remaining);
+	if (rlen <= allowed) {
+		(void)snprintf (print_data, max_chars_on_line, "%s%s", prefix, remaining);
+		pd_off = (int)strlen (print_data);
 		print_program_data (print_data);
-		prefix_offset = strlen (prefix);
-		pd_off = strlen (print_data) - prefix_offset;
-		if (prefix_offset < 2) prefix_offset = 2;
-		memset (print_data, ' ', prefix_offset - 1);
-		snprintf (print_data + prefix_offset - 2, max_chars_on_line, "%c%s", '+', content + pd_off);
+		return;
+	}
+	/* break at last space strictly within the allowed window */
+	breakpos = (allowed > 0) ? (allowed - 1) : 0;
+	while (breakpos > 0 && !isspace ((unsigned char)remaining[breakpos])) {
+		breakpos--;
+	}
+	if (breakpos == 0) {
+		/* no space found in window (or only at index 0), force break at allowed */
+		breakpos = allowed;
+	}
+	(void)snprintf (print_data, max_chars_on_line, "%s%.*s", prefix, (int)breakpos, remaining);
+	pd_off = (int)strlen (print_data);
+	while (pd_off > 0 && isspace ((unsigned char)print_data[pd_off - 1])) {
+		print_data[--pd_off] = '\0';
 	}
 	print_program_data (print_data);
+	/* advance to continuation without collapsing leading spaces */
+	remaining += breakpos;
+
+	/* Continuation lines: prefix is spaces with leading '+' aligned under content */
+	for (;;) {
+		char *out;
+		unsigned int content_cap;
+		/* build continuation prefix */
+		memset (print_data, ' ', prefix_len - 1);
+		out = print_data + prefix_len - 2;
+		allowed = (unsigned int)(max_chars_on_line - (prefix_len - 1));
+		/* capacity for content excludes the '+' itself */
+		content_cap = (allowed > 0) ? (allowed - 1) : 0;
+
+		rlen = strlen (remaining);
+		if (rlen <= content_cap) {
+			(void)snprintf (out, max_chars_on_line, "%c%s", '+', remaining);
+			pd_off = (int)strlen (print_data);
+			print_program_data (print_data);
+			break;
+		}
+
+		breakpos = (content_cap > 0) ? (content_cap - 1) : 0;
+		while (breakpos > 0 && !isspace ((unsigned char)remaining[breakpos])) {
+			breakpos--;
+		}
+		if (breakpos == 0) {
+			/* no space found in window (or only at index 0), force break at capacity */
+			breakpos = content_cap;
+		}
+		(void)snprintf (out, max_chars_on_line, "%c%.*s", '+', (int)breakpos, remaining);
+		pd_off = (int)strlen (print_data);
+		while (pd_off > 0 && isspace ((unsigned char)print_data[pd_off - 1])) {
+			print_data[--pd_off] = '\0';
+		}
+		print_program_data (print_data);
+		/* advance to next continuation without collapsing leading spaces */
+		remaining += breakpos;
+	}
 }
 
 static void
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -6999,19 +6999,14 @@ print_free_line (const int line_num, char pch, char *line)
 static void
 print_with_overflow (const char *prefix, char *content)
 {
-	const unsigned int	max_chars_on_line = cb_listing_wide ? 120 : 80;
-	size_t			prefix_len_raw;
-	size_t			prefix_len;
+	const size_t		max_chars_on_line = cb_listing_wide ? 120 : 80;
+	const size_t		prefix_len_raw = strlen (prefix);
+	const size_t		prefix_len = (prefix_len_raw < 2) ? 2 : prefix_len_raw;
+	const size_t		allowed = max_chars_on_line - prefix_len;
+	char			*out = print_data + prefix_len - 2;
 	const char		*remaining;
-	unsigned int		allowed;
-	size_t			rlen;
+	size_t			rlen = strlen (content);
 	size_t			breakpos;
-
-	/* determine visual width for prefix and continuation alignment */
-	prefix_len_raw = strlen (prefix);
-	prefix_len = (prefix_len_raw < 2) ? 2 : prefix_len_raw;
-	allowed = (unsigned int)(max_chars_on_line - prefix_len);
-	rlen = strlen (content);
 
 	/* print entire content if it fits within the allowed window */
 	if (rlen <= allowed) {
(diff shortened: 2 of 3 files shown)


GitMensch:
I haven't checked the details yet, but if the missing change-coverage is correct - can you do anything about that?

rbowler:
> if the missing change-coverage is correct - can you do anything about that?

I will have a look.

cobc/cobc.c
```diff
@@ -7000,31 +7000,81 @@ static void
...
+	remaining += breakpos;
+
+	/* Continuation lines: prefix is spaces with leading '+' aligned under content */
+	for (;;) {
+		char *out;
+		unsigned int content_cap;
+		/* build continuation prefix */
+		memset (print_data, ' ', prefix_len - 1);
+		out = print_data + prefix_len - 2;
+		allowed = (unsigned int)(max_chars_on_line - (prefix_len - 1));
```
GitMensch: This line may be done before the loop, no?
If it is then out and content_cap can be set directly in the declaration.

cobc/cobc.c
```diff
@@ -7000,31 +7000,81 @@ static void
...
+		/* capacity for content excludes the '+' itself */
+		content_cap = (allowed > 0) ? (allowed - 1) : 0;
+
+		rlen = strlen (remaining);
```
GitMensch: from a quick glance: not sure if we need strlen or adjusting rlen by breakpos is enough - but that loop will neither be tight nor entered often so I don't really mind...

tests/testsuite.src/listings.at
```diff
@@ -2098,6 +2098,31 @@ command line:
...
+000008         PROCEDURE DIVISION USING PAR-IN RETURNING PAR-OUT.
+000009             ADD 1 TO PAR-IN GIVING PAR-OUT END-ADD.
+000010             GOBACK.
+000011         END FUNCTION WITHPAR.
+
GnuCOBOL V.R.P          prog.cob                                        Page 0002
+
+command line:
+  cobc -fttitle=GnuCOBOL_V.R.P -fno-ttimestamp -q -std=default -Wall
++ -fno-tmessages -fsyntax-only -t- -fno-tsymbols -ftcmd
```
GitMensch: The second line is a directly needed wrap, maybe you can reorder so the third line is a directly non-needed wrap? Most likely you can specify -t as often as you want and for the first time use a target name that leads to the wrap on "-t-" (just place the two behind up front)

GitMensch (changes requested):
looks very good; ideally you can add something where the error prefix is set (instead of "  " only) and there's a line break, but more than one will likely be hard to get (maybe with something that includes a long paragraph name defined in a long section name?)

In any case please add an entry to cobc/Changelog (we only add test Changelogs for non testsuite.src or refactoring in those, so that isn't needed)

Seems that I'm able to upstream that directly afterwards.
[Note: I commonly do that for the first commits and offer direct write access afterwards]

rbowler:
@GitMensch Simon, it looks to me as if the lines that the coverage tool flagged are "belt and braces" checks to avoid buffer overflows. For example one of the lines checks that "prefix" is not longer than the line length, but since the line length is always either 80 or 120, and the function is only ever called with "prefix" set to 2 spaces, or "error" or "warning", there is no way to force cobc to execute these lines. Possibly these coverage alerts may disappear after I have addressed your code review comments. The coverage tool has also flagged files that weren't changed by this branch, not sure why.

GitMensch:
For code that _really_ is unlikely to happen (in this case there would have to be a broken message catalogue creating a too long text for the prefix) **and not easy to test** you can use markers like in the following example:

```c
	/* LCOV_EXCL_START */
	if (content_size >= COB_NORMAL_BUFF) {
		/* in unlikely case of "cob_core_filename is too long" use the simple one */
		sprintf (cmd, "gcore -a -o %s %d", default_name, cob_sys_getpid ());
	}
	/* LCOV_EXCL_STOP */
```

tests/testsuite.src/listings.at
rbowler: Simon, I didn't really understand what you meant by a directly non-needed wrap.
If I add an extra -t and move the last two arguments before the -t it looks like this:
command line:
```
  cobc -fttitle=GnuCOBOL_V.R.P -fno-ttimestamp -q -std=default -Wall
+ -fno-tmessages -fsyntax-only -fno-tsymbols -ftcmd
+ -tcompiler_listing_file_with_long_name.txt -t- -Wimplicit-define
+ -Wstrict-typing multiline_errmsg.cob
```
but I'm not sure that's what you meant.

rbowler:
@GitMensch 
> ideally you can add something where the error prefix is set (instead of " " only) and there's a line break,

There are already some tests in **listings.at** AT_SETUP([**Invalid PICTURE strings]**) which produce multiline error messages:
```
autofonce run -id 131
000037             03  PIC +(5)--.
error: only one of the symbols '-' , '+' or '$' may occur multiple times in a
     + PICTURE string
000038             03  PIC $(4)Z(9).
error: a Z or * which is before the decimal point cannot follow a floating
     + currency symbol string which is before the decimal point
```
I haven't succeeded in finding a way to produce an error or warning which spills over onto a third line.

I did manage to produce a two-line warning:
```
000005         FILE-CONTROL.
000006             SELECT TESTFILE
000007             ASSIGN TO a-very-long-name-which-has-not-been-defined-at-all
000008             ACCESS IS an-invalid-name.
...
000016         PROCEDURE        DIVISION.
warning: variable 'a-very-long-name-which-has-not-been-defined-at-all' will be
       + implicitly defined
```
but I don't know if it is worth adding this as a new test, bearing in mind there are already two-line error tests in the testsuite?

GitMensch:
You can have this in either the same program or another one (multiple AT_TEST in one test are fine), that helps a bit for specific testing.
The one message I saw yesterday was something like "must be defined in WORKING-STORAGE, LOCAL-STORAGE or ...", that's a bit longer but may not wrap to the third line.

For the other question: I will respond when I'm at the PC, possibly tomorrow.
Roger Bowler ***@***.***> schrieb am Samstag, 18. Oktober 2025 um 12:37:

rbowler:
> The one message I saw yesterday was something like "must be defined in WORKING-STORAGE, LOCAL-STORAGE or ...", that's a bit longer but may not wrap to the third line.

Bingo! Good find! That message can be made to wrap onto 3 lines if the item name is long enough. And the three line error can be made into a three-line warning instead by -frelax-syntax-checks but it would require two compilations to test both error and warning, which might be a bit of an overkill. So I will add one new test containing a 2-line warning and a 3-line error, unless you think there should be a test case for a 3-line warning too.

cobc/cobc.c
rbowler: In commit [74830459](https://github.com/OCamlPro/gnucobol/pull/248/commits/748304598882988f3fbd95eeb7fbe191d2eb92c0) I have eliminated the strlen as suggested.

cobc/cobc.c
rbowler: In commit [74830459](https://github.com/OCamlPro/gnucobol/pull/248/commits/748304598882988f3fbd95eeb7fbe191d2eb92c0) I have modified the code as suggested.

rbowler:
In commit https://github.com/OCamlPro/gnucobol/pull/248/commits/ec2b536191f3c51eb4d5664ea6be5a730f71cb31 I have added the new test for multiline error/warning messages.

cobc/ChangeLog
```diff
@@ -1,3 +1,6 @@
+2025-10-17  Roger Bowler <rbowler@snipix.net>
```
GitMensch, line 2: first line should stay empty (the diff works better in several clients then) - but no need for an extra commit only for that

cobc/cobc.c
```diff
@@ -7000,31 +7000,81 @@ static void
...
+	const char		*remaining;
+	unsigned int		allowed;
+	size_t			rlen;
+	size_t			breakpos;
+
+	/* determine visual width for prefix and continuation alignment */
+	prefix_len_raw = strlen (prefix);
+	prefix_len = (prefix_len_raw < 2) ? 2 : prefix_len_raw;
+	allowed = (unsigned int)(max_chars_on_line - prefix_len);
+	rlen = strlen (content);
```
GitMensch: those should be included with the variable declaration and all vars that don't change in the run additionally be `const`ed

cobc/cobc.c
```diff
@@ -7000,31 +7000,81 @@ static void
...
+	}
+	print_program_data (print_data);
+	/* advance to continuation without collapsing leading spaces */
+	remaining = content + breakpos;
+	rlen -= breakpos;
+
+	/* Continuation lines: prefix is spaces with leading '+' aligned under content */
+	allowed = (unsigned int)(max_chars_on_line - (prefix_len - 1));
+	char *out = print_data + prefix_len - 2;
+	unsigned int content_cap = (allowed > 0) ? (allowed - 1) : 0;
```
GitMensch: per C89 this needs either a new block or the vars be moved to the start of the function - prefer whatever is more reasonable; content_cap can be a `const` as it is never changed. as `allowed` is unsigned a check for `allowed ?` or `allowed != 0 ?` is possible (or change to  `allowed > 1`)

cobc/cobc.c
GitMensch: was not sure if that's easily possible, but looks like you did take that optional change - nice

cobc/cobc.c
```diff
@@ -7000,31 +7000,81 @@ static void
...
+		memset (print_data, ' ', prefix_len - 1);
+		/* print entire remaining content if it fits within the allowed window */
+		if (rlen <= content_cap) {
+			(void)snprintf (out, max_chars_on_line, "%c%s", '+', remaining);
+			pd_off = (int)strlen (print_data);
+			print_program_data (print_data);
+			break;
+		}
+		/* look for last space strictly within the allowed window */
+		breakpos = (content_cap > 0) ? (content_cap - 1) : 0;
```
GitMensch: similar as above, possibly change to > 1 or just `content_cap` (= C implied "not = zero") - choose one and use in both places

cobc/cobc.c
```diff
@@ -7000,31 +7000,81 @@ static void
...
+		while (breakpos > 0 && !isspace ((unsigned char)remaining[breakpos])) {
```
GitMensch: breakpos is unsigned so either `breakpos` or `breakpos != 0`

tests/testsuite.src/listings.at
GitMensch: I've meant somehting like

```
GnuCOBOL V.R.P          prog.cob                                        Page 0002

command line:
  cobc -fttitle=GnuCOBOL_V.R.P -fno-ttimestamp -q -std=default -Wall
+ -fno-tmessages -fsyntax-only -fno-tsymbols -fdiagnostics-show-option -t abcde
+ -ftcmd -fdiagnostics-show-line-numbers -fdiagnostics-absolute-paths -t-
+ -fno-remove-unreachable prog.cob
```` 

* `-fno-tmessages` is _exactly_ too much and brings a line break
* `abcde` is exactly enough to fit

GitMensch (changes requested):
the C89 break (should be visible in the CI result, but that's broken for other reasons currently but fixed soon with inclusion of #244 in the GH branch) needs to be fixed

the unsigned >0 is "picky" by me and may be seen more as a design "wish"

hooray for the new message test; for the line break test see my notes

tests/testsuite.src/listings.at
rbowler: In commit https://github.com/OCamlPro/gnucobol/pull/248/commits/d3b3bae4dd735739d71e73677397e57189d56461 I amended the command line test in **listings.at** as you suggested, I then discovered that print_with_overflow was breaking one column too early so I fixed that in the same commit.

cobc/cobc.c
rbowler: Resolved by commit https://github.com/OCamlPro/gnucobol/pull/248/commits/520a105070dc648fdbf9010eede9031cd9970e58

cobc/cobc.c
rbowler: Resolved by commit https://github.com/OCamlPro/gnucobol/pull/248/commits/520a105070dc648fdbf9010eede9031cd9970e58

cobc/ChangeLog
rbowler: Nevertheless I reinstated the blank line in commit https://github.com/OCamlPro/gnucobol/pull/248/commits/be5652b1f13d08072b338700be4f2afcec94382f so that it doesn't get forgotten next time.

cobc/cobc.c
rbowler: Resolved by commit https://github.com/OCamlPro/gnucobol/pull/248/commits/520a105070dc648fdbf9010eede9031cd9970e58

cobc/cobc.c
rbowler: Resolved by commit https://github.com/OCamlPro/gnucobol/pull/248/commits/520a105070dc648fdbf9010eede9031cd9970e58

GitMensch (approved):
That's quite nice. I will get it upstream as soon as the papers are back from FSF.

rbowler:
@GitMensch I have now received the confirmation from the FSF.

GitMensch:
thanks again for working on this, providing the very good PR and also sorry that it took so long to merge...

congrats - you have your first upstream commit https://sourceforge.net/p/gnucobol/code/5608/log/?path and of course your SF profile also correctly points out your contributions https://sourceforge.net/u/rbowler/profile/

The change should be visible in the git mirror soon as well.

GitMensch:
Did we discuss outputting the use of that function also for the warning/error summery?

I _think_ one reason that I didn't do it was that it is easier to parse for external tools (we really should have a "sink" for sarif or json in the first place...) but as the listing is for the user I'd say (with a NEWS entry mentioning the change) it is likely good to correctly wrap at this place as well.

What do you think? Could you have a look at a follow-up PR for that?

rbowler:
@GitMensch : I don't remember it being discussed but it seem reasonable that the summary lines should also wrap if they exceed the listing width. I will prepare a follow-up PR.

PR #249, opened 2025-10-14
comp-5-uses-binary-byteorder dialect configuration option

rbowler:
### Summary
This branch adds a new dialect configuration option which allows COMP-5 fields to be processed using the same byte order as BINARY fields.

### Motivation
Previously GnuCOBOL always stored COMP-5 fields using native endianness, as opposed to BINARY or COMP fields which are held in either big-endian or native format depending on the **binary-byteorder** configuration option. However, COBOL programs ported from the mainframe expect COMP-5 fields to be stored in big-endian format like BINARY fields, the difference between COMP-5 and BINARY being that COMP-5 fields are always treated as no-trunc regardless of the setting of the binary-truncate option. The new configuration option allows the user to request that COMP-5 fields should be stored using the same endianness as BINARY fields.

### Changes
- Added configuration option `comp-5-uses-binary-byteorder` to **config.def**
- Updated all dialect config files to include the statement `comp-5-uses-binary-byteorder: no`
- Modified **field.c** to set the the **flag_binary_swap** flag for COMP-5 fields when:
  -- the comp-5-uses-binary-byteorder option is activated
  -- binary-byteorder=big-endian is specified
  -- the host architecture is little-endian
- Updated test case **configuration.at** to check that all dialect config files contain the new option
- Added a new test case **COMP-5 byteorder** in **data_binary.at** to verify correct behavior both with and without the new configuration option when binary-byteorder=big-endian is specified on little-endian systems.

### Impact
- This change also affects BINARY-SHORT, BINARY-LONG, and BINARY-DOUBLE which GnuCOBOL treats internally as COMP-5.
- Existing programs compiled with `comp-5-uses-binary-byteorder: no` will continue to operate unchanged.

### Testing
- All existing test suites have been run on x86 (little-endian) architecture.
- The new test case produces expected results on x86.

### Related Issues / Discussion
- It may be considered desirable to update **ibm-strict.config** to set `comp-5-uses-binary-byteorder: yes` if that can be done without impacting existing programs.
- This change is compatible with https://github.com/OCamlPro/gnucobol/pull/197

---

Please review and provide feedback or suggestions.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,9 @@
 
+2025-10-21  Roger Bowler <rbowler@snipix.net>
+
+	* field.c (setup_parameters): set flag_binary_swap for COMP-5 fields
+	  when comp-5-uses-binary-byteorder dialect option is set
+
 2025-07-30  Simon Sobisch <simonsobisch@gnu.org>
 
 	* scanner.l, parser.y, reserved.c: use yyless to push part of scanned
--- a/cobc/config.def
+++ b/cobc/config.def
@@ -111,6 +111,9 @@ CB_CONFIG_BOOLEAN (cb_pretty_display, "pretty-display",
 CB_CONFIG_BOOLEAN (cb_binary_truncate, "binary-truncate",
 	_("numeric truncation according to ANSI"))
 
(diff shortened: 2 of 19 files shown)


rbowler:
@GitMensch may I request a review of this pull request? Thanks.

rbowler:
Resolved the conflicts in the changelog; branch can now be cleanly merged.

PR #250, opened 2025-10-21
Fix GC4 CI

ddeclerck:


GitMensch:
Coverage does not work, so maybe the last rebase from 3.x is missing?

ddeclerck:
> Coverage does not work, so maybe the last rebase from 3.x is missing?

Either this or the commit that fixes that in 3.x has not yet been merged to trunk
EDIT: nope, the SVN commits were merged, so it's probably just the GitHub CI part
EDIT2: nope nope, there was no GitHub-specific CI part, it's just that trunk does trigger the negative count error while 3.x does not... I have to check something.

ddeclerck:
Well, it came as a surprise : there were some wrong LCOV terminators (LCOV_EXCL_END instead of LCOV_EXCL_STOP), and even a missing LCOV_EXCL_START).
Then there was also an unused exclusion (statment.def) - which is weird since it appears to be used in 3.x. So I prefered to just request the error to be ignored.
And I aso add to pass an extra option to ignore the negative count.

ddeclerck:
@GitMensch Is it okay if I commit to SVN the non-CI part of this PR ?

.github/workflows/macos.yml
```diff
@@ -133,6 +133,8 @@ jobs:
       - name: Adjust testsuite
         run: |
           cd tests/testsuite.src
+          sed -i '' '/AT_SETUP(\[runtime check: write to internal storage (1)\])/a \
```
GitMensch, line 137: Best is to always leave a note why we exclude something.

cobc/codegen.c
GitMensch, line 1: Good to commit this file's change upstream, possibly also for GC3

libcob/fileio.c
GitMensch, line 1: Also good for upstream

Makefile.am
GitMensch, line 1: Please explain why this is needed, especially if it isn't in 3.x

GitMensch:
What about MinGW? We found the whole BDB part to be broken and the runner using a different bash and buildtools in 3.x CI

Makefile.am
ddeclerck: As stated in my previous comment:
> Then there was also an unused exclusion (statment.def) - which is weird since it appears to be used in 3.x. So I prefered to just request the error to be ignored.
> And I aso add to pass an extra option to ignore the negative count.

So it's just because there are more errors/warnings in 4.x - don't know why.

ddeclerck:
> What about MinGW? We found the whole BDB part to be broken and the runner using a different bash and buildtools in 3.x CI

Haven't touched MinGW in this PR. That would need to be solved in 3.x first.

GitMensch:
three lcov markers fixed upstream (actually one was wrong and was removed)

ddeclerck:
@GitMensch Is this one okay to merge too ?
(mind the change to Makefile.am)

GitMensch:
feel free to push the workspace definitions as those are this mirror's branch only - but I'm hesitate to have a difference beetween GC3 and GC4 in that Makefile - why should we need that?

PR #98, opened 2023-06-16
Allow OPEN EXTEND/I-O to create file if missing

ddeclerck:
This PR adds a `-fio-extend-create` flag that, when set, allows `OPEN EXTEND` and `OPEN I-O` to go on and create the file instead of failing.

PR #252, opened 2025-11-09
Remove strbuff in common.c

oguzcankirmemis:
This PR resolves a use-after-free bug in `libcob` that occurs when the runtime is cleaned up and subsequently reinitialized while using the `COB_PRE_LOAD` environment variable. 

# Issue Description 

`strbuff` is a global cache/variable defined in common.c, which is used by the `cob_strcat` function. After the runtime is initialized via `cob_init`, this variable is expected to be set to null. Normally, `cob_strcat` manages the memory used by `strbuff` internally. However, if the runtime is cleaned up, even though the memory pointed to by `strbuff` is correctly freed, the value of `strbuff` is not set back to null. If the runtime is initialized a second time afterward, `cob_strcat` assumes that `strbuff` already exists and attempts to free it again, causing a double-free situation. This is highly likely to result in a use-after-free during the second usage of the runtime or a segmentation fault during `cob_tidy`, if glibc's heap protection mechanisms detect the issue. 

# Steps to Reproduce 

The easiest way to reproduce the issue is to initialize the runtime and execute a COBOL program while the `COB_PRE_LOAD` environment variable includes one or more libraries. This is because `cob_strcat` is only used in `cob_call` when the `COB_PRE_LOAD `variable is set. After calling `cob_tidy` to clean up the runtime, the `strbuff` variable retains a leftover value, which will be incorrectly used during the second runtime initialization. An example program flow that reproduces the issue is as follows: 

1. `cob_init`
2. `cob_call`
3. `cob_tidy`
4. `cob_init`
5. `cob_call`
6. `cob_tidy`

I have also added this example as a test case to the PR.

# Solution

Since `strbuff` is a global variable and `cob_strcat` is used in also other places. removing it is probably the best solution to also prevent potential future memory corruptions. This naturally implies the removal of `cob_strcat`, followed by `cob_strjoin`, which uses `cob_strcat` itself. Calls to `cob_strcat` is replaced with a simple `strlen`/`malloc`/`snprintf` or fixed-size-buffer/`snprintf` routine.

### DIFF
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,4 +1,10 @@
 
+2025-11-18	Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* common.c: remove strbuff variable and cob_strcat/cob_strjoin utilities
+	which were prone to memory corruption due to strbuff being a global
+	* coblocal.h: remove definitions of cob_strcat and cob_strjoin
+
 2025-11-13  Simon Sobisch <simonsobisch@gnu.org>
 
 	* coblocal.h [__GNUC__]: only use C11's _Thread_local for gcc 5+
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -4,6 +4,8 @@
 	* common.c: remove strbuff variable and cob_strcat/cob_strjoin utilities
 	which were prone to memory corruption due to strbuff being a global
 	* coblocal.h: remove definitions of cob_strcat and cob_strjoin
+	* call.c (add_to_preload): remove usage of cob_strcat and directly 
+	concatenate preload paths instead
 
 2025-11-13  Simon Sobisch <simonsobisch@gnu.org>
 
--- a/libcob/call.c
+++ b/libcob/call.c
@@ -19,7 +19,6 @@
 */
 
 
-#include "common.h"
 #include "config.h"
 
 #ifndef	_GNU_SOURCE
@@ -565,7 +564,7 @@ add_to_preload (const char *path, lt_dlhandle libhandle, struct struct_handle *l
 		cobsetptr->cob_preload_str = cob_strdup(path);
 	} else {
 		/* +1 to len for PATHSEP_CHAR */
-		int len = strlen (path) + strlen (cobsetptr->cob_preload_str) + 2;
+		const size_t len = strlen (path) + strlen (cobsetptr->cob_preload_str) + 2;
 		char *buff = (char *) cob_fast_malloc (len);
 		snprintf (buff, len, "%s%c%s", path, PATHSEP_CHAR,
 			cobsetptr->cob_preload_str);
--- a/libcob/call.c
+++ b/libcob/call.c
@@ -19,6 +19,7 @@
 */
 
 
+#include "common.h"
 #include "config.h"
 
 #ifndef	_GNU_SOURCE
@@ -563,8 +564,13 @@ add_to_preload (const char *path, lt_dlhandle libhandle, struct struct_handle *l
 	if (!cobsetptr->cob_preload_str) {
 		cobsetptr->cob_preload_str = cob_strdup(path);
 	} else {
-		cobsetptr->cob_preload_str = cob_strcat((char*) PATHSEP_STR, cobsetptr->cob_preload_str, 2);
-		cobsetptr->cob_preload_str = cob_strcat((char*) path, cobsetptr->cob_preload_str, 2);
+		/* +1 to len for PATHSEP_CHAR */
+		int len = strlen (path) + strlen (cobsetptr->cob_preload_str) + 2;
+		char *buff = (char *) cob_fast_malloc (len);
+		snprintf (buff, len, "%s%c%s", path, PATHSEP_CHAR,
+			cobsetptr->cob_preload_str);
+		cob_free (cobsetptr->cob_preload_str);
+		cobsetptr->cob_preload_str = buff;
 	}
 }
 
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -7731,8 +7731,11 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format != 0 && val && default_val &&
 		((format != 2 && val[0] == '0') || strcmp (val, default_val) == 0)) {
-		snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
+		int len = snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
 			_("(default)"));
+		if (len < 0 || len >= COB_NORMAL_MAX) {
+			memcpy(val_buff + COB_NORMAL_MAX - 3, "...", 3 + 1);
+		}
 		val = val_buff;
 	} else if (!val && default_val) {
 		val = default_val;
@@ -7752,8 +7755,11 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format == 1 && val && default_val &&
 		(val[0] == '0' || strcmp (val, default_val) == 0)) {
-		snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
+		int len = snprintf (val_buff, COB_NORMAL_MAX, "%s %s", default_val,
 			_("(default)"));
+		if (len < 0 || len >= COB_NORMAL_MAX) {
+			memcpy(val_buff + COB_NORMAL_MAX - 3, "...", 3 + 1);
+		}
 		val = val_buff;
 	} else if (!val && default_val) {
 		val = default_val;
@@ -7765,8 +7771,11 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format != 0 && val && default_val &&
 		((format != 2 && val[0] == '0') || strcmp (val, default_val) == 0)) {
-		snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
+		int len = snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
 			_("(default)"));
+		if (len < 0 || len >= COB_NORMAL_MAX) {
+			memcpy(val_buff + COB_NORMAL_MAX - 3, "...", 3 + 1);
+		}
 		val = val_buff;
 	} else if (!val && default_val) {
 		val = default_val;
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -423,8 +423,6 @@ static char			cob_debug_modules[12][4] = {"", "", "", "", "", "", "", "", "", ""
 static char			*cob_debug_file_name = NULL;
 #endif
 
-static char			*strbuff = NULL;
-
 static int		cob_process_id = 0;
 static int		cob_temp_iteration = 0;
 
@@ -7699,71 +7697,12 @@ cob_int_to_formatted_bytestring (int i, char *number)
 }
 #endif
 
-/* concatenate two strings allocating a new one
-   and optionally free one of the strings
-   set str_to_free if the result is assigned to
-   one of the two original strings
-*/
-char *
-cob_strcat (char *str1, char *str2, int str_to_free)
-{
-	size_t		l;
-	char		*temp1, *temp2;
-
-	l = strlen (str1) + strlen (str2) + 1;
-
-	/*
-	 * If one of the parameter is the buffer itself,
-	 * we copy the buffer before continuing.
-	 */
-	if (str1 == strbuff) {
-		temp1 = cob_strdup (str1);
-	} else {
-		temp1 = str1;
-	}
-	if (str2 == strbuff) {
-		temp2 = cob_strdup (str2);
-	} else {
-		temp2 = str2;
-	}
-
-	if (strbuff) {
-		cob_free (strbuff);
-	}
-	strbuff = (char *) cob_fast_malloc (l);
-
-	sprintf (strbuff, "%s%s", temp1, temp2);
-	switch (str_to_free) {
-		case 1: cob_free (temp1);
-		        break;
-		case 2: cob_free (temp2);
-		        break;
-		default: break;
-	}
-	return strbuff;
-}
-
-char *
-cob_strjoin (char **strarray, int size, char *separator)
-{
-	char	*result;
-	int	i;
-
-	if (!strarray || size <= 0 || !separator) return NULL;
-
-	result = cob_strdup (strarray[0]);
-	for (i = 1; i < size; i++) {
-		result = cob_strcat (result, separator, 1);
-		result = cob_strcat (result, strarray[i], 1);
-	}
-
-	return result;
-}
-
 static void
 var_print (const char *msg, const char *val, const char *default_val,
 		const unsigned int format)
 {
+	char *val_buff;
+
 #if 0 /* currently only format 0/1 used */
 	switch (format) {
 	case 0:
@@ -7792,11 +7731,15 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format != 0 && val && default_val &&
 		((format != 2 && val[0] == '0') || strcmp (val, default_val) == 0)) {
-		char dflt[40];
-		snprintf (dflt, 39, " %s", _("(default)"));
-		val = cob_strcat ((char *) default_val, dflt, 0);
+		char *dflt = _("(default)");
+		/* +1 to len for space */
+		int len = strlen (default_val) + strlen (dflt) + 2;
+		val_buff = (char *) cob_fast_malloc (len);
+		snprintf (val_buff, len, "%s %s", default_val, dflt);
 	} else if (!val && default_val) {
-		val = default_val;
+		val_buff = cob_strdup (default_val);
+	} else {
+		val_buff = cob_strdup (val);
 	}
 #else
 	if (format == 0) {
@@ -7813,11 +7756,15 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format == 1 && val && default_val &&
 		(val[0] == '0' || strcmp (val, default_val) == 0)) {
-		char dflt[40];
-		snprintf (dflt, 39, " %s", _("(default)"));
-		val = cob_strcat ((char *) default_val, dflt, 0);
+		char *dflt = _("(default)");
+		/* +1 to len for space */
+		int len = strlen (default_val) + strlen (dflt) + 2;
+		val_buff = (char *) cob_fast_malloc (len);
+		snprintf (val_buff, len, "%s %s", default_val, dflt);
 	} else if (!val && default_val) {
-		val = default_val;
+		val_buff = cob_strdup (default_val);
+	} else {
+		val_buff = cob_strdup (val);
 	}
 #endif
 
@@ -7826,11 +7773,15 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format != 0 && val && default_val &&
 		((format != 2 && val[0] == '0') || strcmp (val, default_val) == 0)) {
-		char dflt[40];
-		snprintf (dflt, 39, " %s", _("(default)"));
-		val = cob_strcat ((char *) default_val, dflt, 0);
+		char *dflt = _("(default)");
+		/* +1 to len for space */
+		int len = strlen (default_val) + strlen (dflt) + 2;
+		char *val_buff = (char *) cob_fast_malloc (len);
+		snprintf (val_buff, len, "%s %s", default_val, dflt);
 	} else if (!val && default_val) {
-		val = default_val;
+		val_buff = cob_strdup (default_val);
+	} else {
+		val_buff = cob_strdup (val);
 	}
 
 	if (val && strlen (val) <= CB_IVAL_SIZE) {
@@ -7846,7 +7797,7 @@ var_print (const char *msg, const char *val, const char *default_val,
 		char	*token;
 		size_t	n;
 
-		p = cob_strdup (val);
+		p = val_buff;
 
 		n = 0;
 		token = strtok (p, " ");
@@ -10393,7 +10344,6 @@ cob_init (const int argc, char **argv)
 	basext = NULL;
 	share_sort_state = NULL;
 	cob_source_file = NULL;
-	strbuff = NULL;
 	exit_hdlrs = NULL;
 	hdlrs = NULL;
 	commlncnt = 0;
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -18,7 +18,6 @@
    along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-#include "common.h"
 #include "tarstamp.h"
 #include "config.h"
 
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -10393,6 +10393,7 @@ cob_init (const int argc, char **argv)
 	basext = NULL;
 	share_sort_state = NULL;
 	cob_source_file = NULL;
+	strbuff = NULL;
 	exit_hdlrs = NULL;
 	hdlrs = NULL;
 	commlncnt = 0;
--- a/libcob/common.c
+++ b/libcob/common.c
@@ -18,6 +18,7 @@
    along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+#include "common.h"
 #include "tarstamp.h"
 #include "config.h"
 
@@ -7701,7 +7702,7 @@ static void
 var_print (const char *msg, const char *val, const char *default_val,
 		const unsigned int format)
 {
-	char *val_buff;
+	char val_buff[COB_NORMAL_BUFF];
 
 #if 0 /* currently only format 0/1 used */
 	switch (format) {
@@ -7731,15 +7732,11 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format != 0 && val && default_val &&
 		((format != 2 && val[0] == '0') || strcmp (val, default_val) == 0)) {
-		char *dflt = _("(default)");
-		/* +1 to len for space */
-		int len = strlen (default_val) + strlen (dflt) + 2;
-		val_buff = (char *) cob_fast_malloc (len);
-		snprintf (val_buff, len, "%s %s", default_val, dflt);
+		snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
+			_("(default)"));
+		val = val_buff;
 	} else if (!val && default_val) {
-		val_buff = cob_strdup (default_val);
-	} else {
-		val_buff = cob_strdup (val);
+		val = default_val;
 	}
 #else
 	if (format == 0) {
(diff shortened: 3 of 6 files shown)


tests/testsuite.src/run_misc.at
```diff
@@ -15737,3 +15737,90 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+           EXIT PROGRAM.
+])
+
+AT_DATA([caller.c], [[
+#include <libcob.h>
+
+#ifndef NULL
+#define NULL (void*)0
+#endif
+
```
GitMensch: Is there any environment where, after including libcob.h which includes a bunch of headers, there's no definition of `NULL`?

ChangeLog
```diff
@@ -1,4 +1,10 @@
 
+2025-11-09	Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* common.c (cob_init): Explicitly reset the strbuff buffer used by cob_strcat
+	  to null, ensuring that the runtime does not use leftover values from
+	  previous initializations.
```
GitMensch: minor: we commonly don't use full sentences here, so consider starting lowercase and dropping the period - but that's no hard request I'd also merge without the change

tests/testsuite.src/run_misc.at
```diff
@@ -15737,3 +15737,90 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+AT_CHECK([$COMPILE_MODULE -o libdummy1 dummy.cob], [0], [], [])
+AT_CHECK([$COMPILE_MODULE -o libdummy2 dummy.cob], [0], [], [])
+AT_CHECK([$COMPILE_MODULE -o libdummy3 dummy.cob], [0], [], [])
+AT_CHECK([$COMPILE_MODULE callee.cob], [0], [], [])
+AT_CHECK([COB_PRE_LOAD=libdummy1:libdummy2:libdummy3 $COBCRUN_DIRECT ./caller], [0],
+[TEST1
+TEST2
+], [])
+
+AT_CLEANUP
```
GitMensch: please fix the missing newline

tests/testsuite.src/run_misc.at
oguzcankirmemis: Good question, I think I copied this from one of the older tests, I can remove it and retrigger the CI to see what happens

oguzcankirmemis:
~~Some of the Windows CIs are failing because it can't find (preload) the compiled libraries in `COB_PRE_LOAD`. Other environments are able to find them though, do you know what might be the issue?~~

I saw the other tests with `COB_PRE_LOAD` are using the `$PATHSEP` variable, probably this will fix the issue.

tests/testsuite.src/run_misc.at
```diff
@@ -15737,3 +15737,90 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+    }
+
+    return 0;
+}
+]])
+
+AT_CHECK([$COMPILE caller.c], [0], [], [])
+AT_CHECK([$COMPILE_MODULE -o libdummy1 dummy.cob], [0], [], [])
+AT_CHECK([$COMPILE_MODULE -o libdummy2 dummy.cob], [0], [], [])
+AT_CHECK([$COMPILE_MODULE -o libdummy3 dummy.cob], [0], [], [])
```
GitMensch: It seems 2 existing elements are enough, no?

In any case: please copy your short description about why we use the pre-loading as comments in the testcase (starting with a `#` in m4) because this is not clear from the test.

tests/testsuite.src/run_misc.at
oguzcankirmemis: You are most likely right, there isn't any environment where `NULL` isn't defined.

oguzcankirmemis:
Remaining failing CI checks are probably unrelated to the PR.

GitMensch (changes requested):
You may go on with that PR and do the minimal test adjustments...

... but actually I'd like this problem to go different and "once and for all":

for the preload (untested prototyping while in a meeting, so take with salt):

```c
	if (!cobsetptr->cob_preload_str) {
		cobsetptr->cob_preload_str = cob_strdup (path);
	} else {
		char buff[COB_FILE_BUFF];
		int len = snprintf (buff, COB_FILE_MAX,
			"%s%c%s", cobsetptr->cob_preload_str, PATHSEP_CHAR, path);
		if (len >= COB_FILE_MAX) {
			/* silent return if not big enough
			   this is only used to show the resolved var */
			return;
		}
		cob_free (cobsetptr->cob_preload_str);
		cobsetptr->cob_preload_str = cob_strdup (buff);
	}
```

common.c (var_print): looks like a memory leak to me in cases where cob_strcat is used - that should be replaced with a function-local `char val_buff [COB_NORMAL_BUFF]` + snprintf...

cob_strjoin and cob_strcat anbd strbuf can then be removed (common.c+h)

What do you think of that?

Note: the test for two elements in pre-load (just not with the tidy because that makes no difference any more) should already be given in used_binaries.at "cobcrun -M DSO entry argument" - which only misses the keyword COB_PRE_LOAD (all that have "preload" in should have that as well)

Can you do these changes?

tests/testsuite.src/run_misc.at
GitMensch: That sounds so bad like an LLM answer 😁 - but I put the blame on that to the common LLM instructions, not you.

ChangeLog
oguzcankirmemis: adjusted according to the new solution

tests/testsuite.src/run_misc.at
oguzcankirmemis: resolved

tests/testsuite.src/run_misc.at
oguzcankirmemis: I have added a short description and reduced the number of pre-loaded libraries to 2.

oguzcankirmemis:
> You may go on with that PR and do the minimal test adjustments...
> 
> ... but actually I'd like this problem to go different and "once and for all":
> 
> for the preload (untested prototyping while in a meeting, so take with salt):
> 
> ```c
> 	if (!cobsetptr->cob_preload_str) {
> 		cobsetptr->cob_preload_str = cob_strdup (path);
> 	} else {
> 		char buff[COB_FILE_BUFF];
> 		int len = snprintf (buff, COB_FILE_MAX,
> 			"%s%c%s", cobsetptr->cob_preload_str, PATHSEP_CHAR, path);
> 		if (len >= COB_FILE_MAX) {
> 			/* silent return if not big enough
> 			   this is only used to show the resolved var */
> 			return;
> 		}
> 		cob_free (cobsetptr->cob_preload_str);
> 		cobsetptr->cob_preload_str = cob_strdup (buff);
> 	}
> ```
> 
> common.c (var_print): looks like a memory leak to me in cases where cob_strcat is used - that should be replaced with a function-local `char val_buff [COB_NORMAL_BUFF]` + snprintf...
> 
> cob_strjoin and cob_strcat anbd strbuf can then be removed (common.c+h)
> 
> What do you think of that?
> 
> Note: the test for two elements in pre-load (just not with the tidy because that makes no difference any more) should already be given in used_binaries.at "cobcrun -M DSO entry argument" - which only misses the keyword COB_PRE_LOAD (all that have "preload" in should have that as well)
> 
> Can you do these changes?

Good idea, I have implemented a draft. I don't think it matters a lot, but instead of allocating the buffer on the stack, I have chosen to allocate a new one in the heap with the correct length to keep the behavior consistent with `cob_strcat`. I am not sure I am following the suggestion with the test case in `used_binaries.at`, but do ping me if I should adjust the test any further.

libcob/call.c
```diff
@@ -19,6 +19,7 @@
 */
 
 
+#include "common.h"
```
GitMensch: already explicit included via coblocal.h below;
if we want to make that explicit then this would be a separate change / PR (not sure though, but you possibly can convince me :-)

note that config.h is first, then system headers, then own headers

libcob/call.c
```diff
@@ -563,8 +564,13 @@ add_to_preload (const char *path, lt_dlhandle libhandle, struct struct_handle *l
 	if (!cobsetptr->cob_preload_str) {
 		cobsetptr->cob_preload_str = cob_strdup(path);
 	} else {
-		cobsetptr->cob_preload_str = cob_strcat((char*) PATHSEP_STR, cobsetptr->cob_preload_str, 2);
-		cobsetptr->cob_preload_str = cob_strcat((char*) path, cobsetptr->cob_preload_str, 2);
+		/* +1 to len for PATHSEP_CHAR */
+		int len = strlen (path) + strlen (cobsetptr->cob_preload_str) + 2;
```
GitMensch: `const size_t` here (but ok, let's not add artificial restrictions of a fixed-size buffer)

libcob/common.c
```diff
@@ -7699,71 +7697,12 @@ cob_int_to_formatted_bytestring (int i, char *number)
...
-	}
-	return strbuff;
-}
-
-char *
-cob_strjoin (char **strarray, int size, char *separator)
-{
-	char	*result;
-	int	i;
-
```
GitMensch, line 7751: remove from coblocal.h as well (also add to the Changelog)

libcob/common.c
```diff
@@ -7699,71 +7697,12 @@ cob_int_to_formatted_bytestring (int i, char *number)
...
-	}
-
-	return result;
-}
-
 static void
 var_print (const char *msg, const char *val, const char *default_val,
 		const unsigned int format)
 {
+	char *val_buff;
```
GitMensch: the code below is full of memory leaks (that is an older issue, currently made worse) - the "val_buff" is never freed - and in most cases we strdup the existing default for no reason...
either used a single fixed-size buffer (even `COB_LARGE_BUFF` is ok) and assign val to it where we need that, or (more spaghetti) initialize `val_buff = NULL;` and only set it where needed (the len would be again `const size_t`), then replace all possible `return` by a goto to the function end where you free the pointer if it isn't null

tests/testsuite.src/run_misc.at
```diff
@@ -15737,3 +15737,91 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
 
 AT_CLEANUP
+
+
+# This specifically tests a use-after-free issue where COB_PRE_LOAD caused 
+# leftover values from previous initializations to be used due to
+# cob_init/cob_tidy not resetting the buffer of cob_strcat, which is used
+# by cob_call when it preloads more than one library. It is also, however,
+# useful as a general test case to check if there is any leak between
+# cob_init/cob_tidy.
```
GitMensch: the free-after use cannot happen any more, so drop that part; the explanation of leak- and other memory issues is still valid and useful (just include "within" the testgroup = after `AT_KEYWORDS`), but we can drop pre-load and its dummy, no?

It seems that test case would be best moved up below the test group "direct CALL in from C w/wo error; no exit".

GitMensch (changes requested):
changed approach is well, but needs adjustments for `var_print` (if my glancing over the code is not wrong, which could be) and the test

libcob/call.c
oguzcankirmemis: No worries, I also don't remember adding this :). My Clang extension must have thought it was necessary for some reason. Good catch, thank you!

libcob/common.c
oguzcankirmemis: done

libcob/common.c
```diff
@@ -7792,9 +7731,9 @@ var_print (const char *msg, const char *val, const char *default_val,
 		return;
 	} else if (format != 0 && val && default_val &&
 		((format != 2 && val[0] == '0') || strcmp (val, default_val) == 0)) {
-		char dflt[40];
-		snprintf (dflt, 39, " %s", _("(default)"));
-		val = cob_strcat ((char *) default_val, dflt, 0);
+		snprintf (val_buff, COB_NORMAL_BUFF, "%s %s", default_val,
```
oguzcankirmemis: As explained previously, I am not checking the returned length from snprintf here, worst case the user would see a truncated value here. Let me know if I should handle it differently though. Same comment applies for the other two occurences of `val_buff` in `var_print`.

libcob/call.c
oguzcankirmemis: resolved

libcob/common.c
oguzcankirmemis: To be fair, I got confused by the `cob_free(p)` at the end and the early return if `strlen(val) <= CB_IVAL_SIZE` :)

I am now using a fixed-size buffer as you suggested initially with length `COB_NORMAL_BUFF`. I am not sure how to handle if the fixed-size buffer has not enough space. In the end I decided not to handle them, worst case the value would be truncated?

Let me know if we need further adjustments though.

tests/testsuite.src/run_misc.at
oguzcankirmemis: Done, I also removed the explanation regarding COB_PRE_LOAD completely. Let me know if we need more adjustments in the new explanation.

libcob/call.c
oguzcankirmemis: done

libcob/common.c
GitMensch: As this is print only and the elements are unlikely to be that big a truncation is fine.

libcob/common.c
GitMensch: That's fine here in general; in other places we eplicit check for the size and if it is too big memcpy " ..." to the end (including the terminating byte) which has one "hidden" benefit:

We handle C99 breaking libc's (as mingw) which may:
* return -1 on overflow instead of size (... or so I've thought, seems that was not checked in)
* don't write a terminating NULL on overflow -> we do use COB_NORMAL_MAX (= request one byte less) and, if we concat or otherwise string-use it later set the resulting buffer's end to 0 manually

Please use the _MAX instead of the _BUFF and to be safe either do a `val_buff[COB_NORMAL_MAX] = 0` unconditionally afterwards or check the return value and memcpy " ..." to val_buff + COB_NORMAL_BUFF - 5 (or the right number ;-)

tests/testsuite.src/run_misc.at
GitMensch, line 1: nice test and doc

libcob/ChangeLog
```diff
@@ -1,4 +1,10 @@
 
+2025-11-09	Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* common.c: remove strbuff variable and cob_strcat/cob_strjoin utilities
+	which were prone to memory corruption due to strbuff being a global
+	* coblocal.h: remove definitions of cob_strcat and cob_strjoin
```
GitMensch, line 6: An entry for `call.c (add_to_preload)` is missing

GitMensch (changes requested):
minor changes needed, then can go upstream

libcob/ChangeLog
oguzcankirmemis: added

libcob/common.c
oguzcankirmemis: That makes sense, I went for the memcpy "..." approach.

oguzcankirmemis:
The two new failures on MacOS are new, but they also seem unrelated to the PR.

GitMensch (approved):
LGTM (minor spacing but I can handle that when doing the commit)

BTW: as the copyright papers were sent back by FSF this really can go upstream asap.

Also note that your SF user has write access to "contrib" now (for a while someone of the other devs with write access to "code" will handle your commits and I''ll ensure they are properly attributed to you [this also provides the SF "coder" benefits, but I guess you have the compute and website option already otherwise).

GitMensch:
Hi @oguzcankirmemis 

I wanted to merge that PR but got an error when running `make check` (actually: the new test) locally:

~~~
800. run_misc.at:4790: testing can reinitialize cleaned up runtime in C ...
../../tests/run_misc.at:4857: $COMPILE caller.c
../../tests/run_misc.at:4858: $COMPILE_MODULE callee.cob
../../tests/run_misc.at:4859: $COBCRUN_DIRECT ./caller
--- /dev/null   2025-11-17 18:54:58.423082137 +0000
+++ /home/simon/gnucobol-3.x/build/tests/testsuite.dir/at-groups/800/stderr     2025-11-17 22:49:24.083278230 +0000
@@ -0,0 +1,12 @@
+=================================================================
+==142596== ERROR: AddressSanitizer: heap-use-after-free on address 0x60200000ffcc at pc 0x721b45939c63 bp 0x7ffe0ff15940 sp 0x7ffe0ff15930
+READ of size 4 at 0x60200000ffcc thread T0
+    #0 0x721b45939c62 (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x139c62)
+    #1 0x721b45853c70 (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x53c70)
+    #2 0x721b4585a48d (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x5a48d)
+    #3 0x721b45860b09 (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x60b09)
+    #4 0x4009ea (/home/simon/gnucobol-3.x/build/tests/testsuite.dir/0800/caller+0x4009ea)
+    #5 0x721b45422554 (/usr/lib64/libc-2.17.so+0x22554)
+    #6 0x400848 (/home/simon/gnucobol-3.x/build/tests/testsuite.dir/0800/caller+0x400848)
+ASAN:SIGSEGV
+==142596== AddressSanitizer: while reporting a bug found another one.Ignoring.
--- -   2025-11-17 22:49:24.091459169 +0000
+++ /home/simon/gnucobol-3.x/build/tests/testsuite.dir/at-groups/800/stdout     2025-11-17 22:49:24.087278020 +0000
@@ -1,3 +1,2 @@
 TEST1
-TEST2
~~~

This is happening on CentOS7 (should be identical with newer versions) when configured with
`../configure --enable-debug 'CFLAGS=-fsanitize=address -fno-omit-frame-pointer' LDFLAGS=-fsanitize=address COBC_CPPFLAGS=-DCOB_NON_ALIGNED gc_cv_designated_inits=no`

checked with

```sh
make check TESTSUITEFLAGS="--verbose --recheck" COBOL_FLAGS=-g

./pre-inst-env

cd tests/testsuite.dir/0800/

gdb ./caller -ex run -ex bt --batch
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
TEST1
=================================================================
==145228== ERROR: AddressSanitizer: heap-use-after-free on address 0x60200000ffcc at pc 0x7ffff4939c63 bp 0x7fffffffb590 sp 0x7fffffffb580
READ of size 4 at 0x60200000ffcc thread T0
    #0 0x7ffff4939c62 (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x139c62)
    #1 0x7ffff4853c70 (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x53c70)
    #2 0x7ffff485a48d (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x5a48d)
    #3 0x7ffff4860b09 (/home/simon/gnucobol-3.x/build/libcob/.libs/libcob.so.4.2.0+0x60b09)
    #4 0x4009ea (/home/simon/gnucobol-3.x/build/tests/testsuite.dir/0800/caller+0x4009ea)
    #5 0x7ffff4422554 (/usr/lib64/libc-2.17.so+0x22554)
    #6 0x400848 (/home/simon/gnucobol-3.x/build/tests/testsuite.dir/0800/caller+0x400848)

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff4c175a3 in ?? () from /lib64/libasan.so.0
#0  0x00007ffff4c175a3 in ?? () from /lib64/libasan.so.0
#1  0x00007ffff4c15696 in ?? () from /lib64/libasan.so.0
#2  0x00007ffff4c1828b in ?? () from /lib64/libasan.so.0
#3  0x00007ffff4c18e02 in __asan_report_error () from /lib64/libasan.so.0
#4  0x00007ffff4c131e3 in __asan_report_load4 () from /lib64/libasan.so.0
#5  0x00007ffff4939c63 in cob_settings_screenio () at ../../libcob/screenio.c:5186
#6  0x00007ffff4853c71 in set_config_val (value=0x7ffff495d6c0 "0", pos=14) at ../../libcob/common.c:8130
#7  0x00007ffff485a48e in cob_load_config () at ../../libcob/common.c:8866
#8  0x00007ffff4860b0a in cob_init (argc=0, argv=0x0) at ../../libcob/common.c:10414
#9  0x00000000004009eb in main (argc=1, argv=0x7fffffffd958) at caller.c:31
```

Wasn't the idea of this PR to solve this?
Possibly you currently had --without-curses (and that happens only with it?)

Can you please have a look and rebase (which _should_ get the MSVC CIs to pass again)?

oguzcankirmemis:
That is interesting, I can also reproduce the error, will look at it. With your configuration, I have 10 more tests failing, but I assume they are unrelated and normal?

In any case, I have rebased the branch and updated the PR.

GitMensch:
thanks for inspecting it - which other tests fail? I think all should pass (most easy: do `make check TESTSUITEFLAGS=--recheck`, then upload resulting testsuite.log.

oguzcankirmemis:
Relevant output from `make check TESTSUITEFLAGS=--recheck`:

```
## -------------------------------------------- ##
## GnuCOBOL 3.3-dev test suite: GnuCOBOL Tests. ##
## -------------------------------------------- ##

Syntax tests

 81: Option --copy COPYBOOK                          FAILED (syn_copy.at:1113)
132: Invalid PICTURE strings                         FAILED (syn_definition.at:1602)
260: constant-folding out of bounds                  FAILED (syn_refmod.at:148)
343: dangling LINKAGE items                          FAILED (syn_misc.at:4423)
360: Constant Expressions (2)                        FAILED (syn_misc.at:5346)

Run tests

799: direct CALL in from C w/wo error; no exit       FAILED (run_misc.at:4782)
800: can reinitialize cleaned up runtime in C        FAILED (run_misc.at:4859)
836: Trace feature with subroutine                   FAILED (run_misc.at:9018)
841: PICTURE with constant-name                      FAILED (run_misc.at:11737)
972: EXTFH: using ISAM callback                      FAILED (run_file.at:9219)
1281: XML register data after use                     FAILED (run_ml.at:654)

## ------------- ##
## Test results. ##
## ------------- ##

ERROR: All 11 tests were run,
11 failed unexpectedly.
```
You can find the testsuite.log attached below:

[testsuite.log](https://github.com/user-attachments/files/23602968/testsuite.log)

oguzcankirmemis:
Hi @GitMensch 

I have pushed a fix in `screenio.c` that fixes the issue, can you try it on your side as well to see if it works?

GitMensch:
congrats to your first upstream commit :-)

https://sourceforge.net/p/gnucobol/code/5607/log/?path
should be available in git mirror soon

Concerning your other failing tests with active analyzers: please open a bug report for them.

GitMensch:
One error that is most likely part of cobc generation is breaking the Appveyor CI with sanitizers enabled:

~~~
1285. run_ml.at:873: testing JSON GENERATE trimming ...
./run_ml.at:995: $COMPILE prog.cob
./run_ml.at:996: $COBCRUN_DIRECT ./prog
--- /dev/null	2025-11-18 07:52:53.314145400 +0000
+++ /home/appveyor/projects/gnucobol-3-x/tests/testsuite.dir/at-groups/1285/stderr	2025-11-18 08:06:42.661235897 +0000
@@ -0,0 +1,4 @@
+/tmp/cob79363_0.c:221:26: runtime error: store to misaligned address 0x55d719936b1d for type 'int', which requires 4 byte alignment
+0x55d719936b1d: note: pointer points here
+ 00 00 00 00 00 00 00  00 20 20 20 20 20 20 20  20 20 20 30 31 30 30 34  00 00 00 00 00 00 00 00  00
+             ^ 
1285. run_ml.at:873:  FAILED (run_ml.at:996)
~~~

Seen on

Linux x86_64 5.15.0-1089-azure
#98~20.04.1-Ubuntu SMP Fri May 2 20:18:39 UTC 2025

configured with `$ ./configure --enable-debug 'CFLAGS=-fsanitize=address -fsanitize=undefined -fno-omit-frame-pointer' 'LDFLAGS=-fsanitize=address -fsanitize=undefined' --with-json=json-c COBC_CPPFLAGS=-DCOB_NON_ALIGNED gc_cv_designated_inits=no`

cobc is "neuland" for you and it is most likely a minor change - but it would be nice if someone could spent the time to inspect that failure, so we can go green again.

Full log: https://ci.appveyor.com/api/buildjobs/j9s73xdki0bb761f/artifacts/testsuite.log

GitMensch:
Thanks for the make check details. All of those additional ones are from leak-sanitizer and most are bdb related. That issue was triggered by valgrind years ago rand finally ignored - patches welcome (but really not that important as it happens only once per BDB backed INDEXED file and is no leak but a minor burden).

There are two easy ones in the compiler, I keep forgetting to do those - they may be an "easier" entry into cobc:

~~~
+Direct leak of 8 byte(s) in 1 object(s) allocated from:
+    #0 0x720faf4fd340 in calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
+    #1 0x5baa41c9ab8d in cobc_malloc ../../cobc/cobc.c:1001
+    #2 0x5baa41d753d9 in get_number_in_parentheses ../../cobc/tree.c:3403
+    #3 0x5baa41d76056 in cb_build_picture ../../cobc/tree.c:3543
+    #4 0x5baa41d4edd0 in scan_picture .././../cobc/scanner.l:2304
+    #5 0x5baa41d46926 in yylex .././../cobc/scanner.l:1300
+    #6 0x5baa41ceed09 in yyparse cobc/parser.c:15268
+    #7 0x5baa41cba6a6 in process_translate ../../cobc/cobc.c:8204
+    #8 0x5baa41cbe0fc in process_file ../../cobc/cobc.c:9411
+    #9 0x5baa41cbeec7 in main ../../cobc/cobc.c:9633


+Direct leak of 64 byte(s) in 1 object(s) allocated from:
+    #0 0x7fc981afd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
+    #1 0x57c2ba9ae1c5 in ppalloc cobc/pplex.c:8873
+    #2 0x57c2ba9ad205 in pp_create_buffer cobc/pplex.c:8530
+    #3 0x57c2ba9af8e5 in ppopen .././../cobc/pplex.l:1553
+    #4 0x57c2ba97cf77 in preprocess ../../cobc/cobc.c:5404
+    #5 0x57c2ba98fc95 in process_file ../../cobc/cobc.c:9333
+    #6 0x57c2ba990ec7 in main ../../cobc/cobc.c:9633
+
+Indirect leak of 32770 byte(s) in 1 object(s) allocated from:
+    #0 0x7fc981afd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
+    #1 0x57c2ba9ae1c5 in ppalloc cobc/pplex.c:8873
+    #2 0x57c2ba9ad284 in pp_create_buffer cobc/pplex.c:8539
+    #3 0x57c2ba9af8e5 in ppopen .././../cobc/pplex.l:1553
+    #4 0x57c2ba97cf77 in preprocess ../../cobc/cobc.c:5404
+    #5 0x57c2ba98fc95 in process_file ../../cobc/cobc.c:9333
~~~

The first needs a different memory allocation, the second possibly a free.

oguzcankirmemis:
Thanks for looking into the logs. Just a quick clarification: the logs from Appveyor CI is unrelated to the logs I've posted, right? I might take a look at it in my own spare time, but can't really promise that I will fix it.

GitMensch:
Yes, the Appveyor one (which I consider from that the most important one) is separate as noted before.
The leaks above will be possibly easier to fix.

Thanks for considering to fix those.

PR #256, opened 2025-11-20
Fix CI (follow-up of #244)

ddeclerck:
Follow-up of #244.

Opening a new PR, since the target branch was renamed.

However, I couldn't get the MinGW CI to work...

I'm using this opportunity to check whether the custom definition of HOME is still needed for the IBM CI.

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -787,7 +787,7 @@ jobs:
           %MSYS_BIN%\bash -lc "patch -d bdb -p1 -i ../db-tls-m4-fix-pthread.patch"
           %MSYS_BIN%\bash -lc "patch -d bdb -p1 -i ../mingw.patch"
           %MSYS_BIN%\bash -lc "patch -d bdb -p1 -i ../cx11.patch"
-          %MSYS_BIN%\bash -lc "export lt_cv_deplibs_check_method=${lt_cv_deplibs_check_method='pass_all'}; cd bdb/build_unix && ../dist/configure --prefix=/mingw --enable-mingw --enable-debug --disable-static --disable-replication --disable-tcl --without-cryptography LIBCSO_LIBS=-lwsock32 CPPFLAGS='-Wno-incompatible-pointer-types -Wno-old-style-definition'"
+          %MSYS_BIN%\bash -lc "lt_cv_deplibs_check_method=${lt_cv_deplibs_check_method='pass_all'}; cd bdb/build_unix && ../dist/configure --prefix=/mingw --enable-mingw --enable-debug --disable-static --disable-replication --disable-tcl --without-cryptography LIBCSO_LIBS=-lwsock32 CC=/mingw/bin/gcc CPPFLAGS='-Wno-incompatible-pointer-types -Wno-old-style-definition'"
           %MSYS_BIN%\bash -lc "make -C bdb/build_unix --jobs=$(($(nproc)+1))"
 
       - name: Install BDB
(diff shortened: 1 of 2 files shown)


GitMensch:
just to note: you can change the target of a PR to another branch as well :-)

To not do things twice: what did you set BDB_LIBS to and did you already inspected which library files are installed?
Previously BDB was possibly built with a different toolchain (64bit) than the GnuCOBOL one (32 bit), but I _think_ that should be fine with the last adjustments (maybe check twice, just to be sure).

And: yes, HOME can be dropped now for IBM, which is nice.

ddeclerck:
> just to note: you can change the target of a PR to another branch as well :-)

Oh, this sounds new.
The actual reason for the PR being closed is that I forced-pushed here after a rebase while there were no new commits...

> To not do things twice: what did you set BDB_LIBS to and did you already inspected which library files are installed? 

What it is currently set seems correct, there is indeed a file `/mingw/lib/libdb-6.0.dll.a`. But I tried without BDB_LIBS (as it worked before), which didn't help. Comparing with an older working version, I couldn't find why this changed.

> And: yes, HOME can be dropped now for IBM, which is nice.

Yup, finally.

GitMensch:
I think we can use the linking feature of GCC, allowing it to link to dll's directly; not sure if the check run by configure passes that to gcc directly (but the logs should show that).

Just found out another broken issue: MSVC dependency artifacts are empty since a while (that did work before as well).

... CI saves time during development (or at least before a release), but it definitely needs some time on its own.

ddeclerck:
Okay I think I finally found what was wrong.
BDB was compiled in 64-bit mode... Because `configure` "found" a `cc` command in `/mingw64/bin` (don't know where this one comes from...). So I just added `CC=gcc` to BDB's `configure`.

GitMensch:
I guess you have to manually delete the BDB cache.

But the CC for BDB's configure should possibly either point to the full gcc path we want to use or we should `export PATH` directly before the configure step to have the mingw/bin coming first (I'd tend to the second option).

.github/workflows/build_nightly.yml
```diff
@@ -836,7 +836,7 @@ jobs:
       - name: Configure GnuCOBOL
         run: |
           mkdir _build
-          %MSYS_BIN%\bash -lc "cd _build && ../configure --with-db --with-xml2 --with-json=local --with-curses=pdcurses --prefix=/mingw %CFGOPT% LDFLAGS=-lintl BDB_LIBS=/mingw/lib/libdb-6.0.dll.a"
+          %MSYS_BIN%\bash -lc "cd _build && ../configure --with-db --with-xml2 --with-json=local --with-curses=pdcurses --prefix=/mingw %CFGOPT% LDFLAGS=-lintl BDB_LIBS=-ldb-6.0"
```
GitMensch: that's the default, so if it works it would be reasonable to drop BDB_LIBS here

.github/workflows/build_nightly.yml
```diff
@@ -787,7 +787,7 @@ jobs:
           %MSYS_BIN%\bash -lc "patch -d bdb -p1 -i ../db-tls-m4-fix-pthread.patch"
           %MSYS_BIN%\bash -lc "patch -d bdb -p1 -i ../mingw.patch"
           %MSYS_BIN%\bash -lc "patch -d bdb -p1 -i ../cx11.patch"
-          %MSYS_BIN%\bash -lc "export lt_cv_deplibs_check_method=${lt_cv_deplibs_check_method='pass_all'}; cd bdb/build_unix && ../dist/configure --prefix=/mingw --enable-mingw --enable-debug --disable-static --disable-replication --disable-tcl --without-cryptography LIBCSO_LIBS=-lwsock32 CPPFLAGS='-Wno-incompatible-pointer-types -Wno-old-style-definition'"
+          %MSYS_BIN%\bash -lc "export CC=gcc lt_cv_deplibs_check_method=${lt_cv_deplibs_check_method='pass_all'}; cd bdb/build_unix && ../dist/configure --prefix=/mingw --enable-mingw --enable-debug --disable-static --disable-replication --disable-tcl --without-cryptography LIBCSO_LIBS=-lwsock32 CPPFLAGS='-Wno-incompatible-pointer-types -Wno-old-style-definition'"
```
GitMensch: either `export PATH` with the mingw/bin prefixed, or pass CC as configure var - possibly with its full path

ddeclerck: Seems an `export PATH` is not sufficient - it looks for `cc` first, and then for `gcc`, so it would still find `cc`.
So I added it as `CC=/mingws/bin/gcc` as a `configure` argument.

GitMensch:
Nice work! On the the open PRs amnd WIPs :-)

PR #258, opened 2025-11-21
Update name of main GC4 branch in CI workflows

nberth:


nberth:
@GitMensch @ddeclerck IBM's workflow still references `gcos4gnucobol-3.x`. Is it intentional, or an oversight?

ddeclerck:
> @GitMensch @ddeclerck IBM's workflow still references `gcos4gnucobol-3.x`. Is it intentional, or an oversight?

Likely an error, you may replace by `gitside-master`

PR #259, opened 2025-11-24
WIP: allow `OCCURS DEPENDING ON` with `REDEFINES` clauses

emilienlemaire:
As of today this PR only allows the parsing of those cases, the codegen is to be checked

### DIFF
--- a/config/acu-strict.conf
+++ b/config/acu-strict.conf
@@ -285,6 +285,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				" "
 picture-l:				unconformable
+odo-in-redefines:			warning
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		ACU
--- a/config/bs2000-strict.conf
+++ b/config/bs2000-strict.conf
@@ -282,6 +282,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0		# not verified yet, but likely to be as IBM
 picture-l:				unconformable
+odo-in-redefines:			unconformable # to be checked
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		BS2000
--- a/config/gcos-strict.conf
+++ b/config/gcos-strict.conf
@@ -280,6 +280,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	obsolete
 defaultbyte:				0
 picture-l:				ok
+odo-in-redefines:			unconformable # to be checked
 
 # use fixed word list, synonyms and exceptions specified there
 reserved-words:		GCOS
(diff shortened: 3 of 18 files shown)


config/xopen.conf
```diff
@@ -300,6 +300,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	obsolete
 defaultbyte:				none		# "not specifically defined in Standard COBOL"
 picture-l:				unconformable
+odo-in-redefines:			unconformable	# to be checked
```
GitMensch, line 303: xopen is only stricter than cobol85+amendment, so that's unconformable

config/gcos-strict.conf
```diff
@@ -280,6 +280,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	obsolete
 defaultbyte:				0
 picture-l:				ok
+odo-in-redefines:			unconformable # to be checked
```
GitMensch, line 283: checked: unconformable

config/bs2000-strict.conf
```diff
@@ -282,6 +282,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0		# not verified yet, but likely to be as IBM
 picture-l:				unconformable
+odo-in-redefines:			unconformable # to be checked
```
GitMensch, line 285: checked, unconformable

config/acu-strict.conf
```diff
@@ -285,6 +285,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				" "
 picture-l:				unconformable
+odo-in-redefines:			warning
```
GitMensch, line 288: the warning is only raised when compiled with "cautions enabled" - so it should be "ok" here (and a possible cobc_acu wrapper will take the necessary `-a`  parameter and translate it which will include `-fodo-in-redefines=warning`)

... but for larger-redefines it should actually be `warning` - please adjust that when changing this file

config/realia-strict.conf
```diff
@@ -285,6 +285,7 @@ self-call-recursive:			skip
 record-contains-depending-clause:	unconformable
 defaultbyte:				0		# not verified, but likely like IBM
 picture-l:				unconformable
+odo-in-redefines:			unconformable # to be checked
```
GitMensch, line 288: Haven't found anything about it in its COBOL Reference Guide [that's in windows HLP format...] it has some rules on `OCCURS` but doesn't specify anything between that and `REDEFINES`.

In any case that would be `# not verified, but likely like IBM`

GitMensch (changes requested):
I've took 20 minutes to check the dialect support.

Please adjust the comments/valuas as noted and add cobc/Changelog and config/Changelog to your files..

Please also add runtime tests that show what you see on MF - and that GC with the new option has the same result.

GitMensch:
@emilienlemaire Is there any chance to get back to that?

PR #260, opened 2025-11-25
Check for terminating periods at the end of `SET` directives

nberth:


tests/testsuite.src/run_misc.at
GitMensch, line 1: What is the benefit of this test here = why do we ant to run it and how would we see that this did work as expected?

tests/testsuite.src/syn_misc.at
```diff
@@ -5855,11 +5855,28 @@ prog.cob:8: error: PROGRAM-ID header missing
...
+       $SET SOURCEFORMAT"FREE".
+IDENTIFICATION   DIVISION.
+PROGRAM-ID.      prog.
+$SET SOURCEFORMAT "FIXED".
+       PROCEDURE DIVISION.
+             STOP RUN.
+])
+AT_CHECK([$COMPILE_ONLY prog2.cob], [1], [],
+[prog2.cob:2: error: unexpected period
+prog2.cob:2: error: PROGRAM-ID header missing
```
GitMensch: That seems like a missing yyerrok - or we should get rid of YYERROR in that case.

cobc/ppparse.y
```diff
@@ -1377,6 +1377,15 @@ _dot:
 | DOT
 ;
 
+_unexpected_dot:
+| DOT
+  {
+	  if (cb_syntax_check (_("unexpected period")) == 0) {
+		  YYERROR;
```
GitMensch: Why the YYERROR here? The unexpected dot may error/warn via syntax check, but that should have no effect on the processing. For broken $SET the directive should either be "still parsed" like here, or completely ignored - but the following processing should be independent from the directive as _any_ directive is ending at newline

GitMensch (changes requested):
directive "leaking" into parsing needs to be fixed, then good for upstream

GitMensch:
Just two additional notes:

> Some directives must be set at the start of the program; these are known as initial $SET statements. They can precede or follow other $SET statements, but must come before any other statements. 

Concerning $SET that is likely much more important to add (and there should also be a FR ticket for that),

Concerning your intention: isn't -std=mf-strict setting relaxed mode, so there is commonly no error if not made more strict by request? I don't mind for upstream, just saying.

GitMensch:
Also: https://www.microfocus.com/documentation/visual-cobol/vc100/DevHub/GUID-BC811861-EBBF-4389-B53C-A22B547AF17D.html reminds me that it would be useful to have a cobcmf script/binary which checks those environment variables and files and converts the MF ones before finally calling into cobc...

tests/testsuite.src/run_misc.at
nberth: Well, that's where I first added a test before noticing it should be in `syn_misc.at` instead. I left that one as it fit in `SOURCE FIXED/FREE directives`, but probably it is that specific test that needs to be moved, isn't it?

GitMensch (approved):
good for upstream, please commit there

PR #261, opened 2025-11-25
WIP - a branch with multiple commits for MF-AIX

lefessan:
This branch started as a merge of some other branches for MF-AIX, and then some other commits were added:
* Make DOT optional after data descriptions by changing the heuristic to discriminate between LITERAL and LEVEL_NUMBER
* Allow extra DOT between data descriptions

All these commits should probably be submitted through separate pull requests.


GitMensch:
Just a note: Instead of eprintf_token you want to use the (oboslete) [%token-table](https://www.gnu.org/software/bison/manual/html_node/Decl-Summary.html#index-_0025token_002dtable).

You may also want to have a look at https://www.gnu.org/software/bison/manual/html_node/Enabling-Traces.html (though this is not related to your scanner change),

GitMensch:
I highly encourage you to add test cases for what you want to achieve up-front.

This does not only help to validate fast and have the tests "nearly ready" [you may want to adjust them during development}, but also is extra useful if someone has a look at the WIP changes to see its purpose.

PR #262, opened 2025-11-26
Allow inspect with `or` conditions

emilienlemaire:
Allow inspect with `or` conditions in inspect region, when `inspect-with-or` dialect support is enabled


GitMensch (changes requested):
commenting as it isn't a draft...

Please always start with the testcases Here you want a syntax/extension one for the dialect option (seems like that is really only mf-strict.conf allowing it, default and lax would with a warning as it does not conflict with anything) - and of course for _each_ place where after-before-ohrase is applicable also a test case compiling with -finspect-or=ok.

as soon as doing the first change create a Changelog entry - this way you've documented what you did already and don't need to think about it later (which is especially useful when coming back to a change later or if someone takes over),

cobc/parser.y
```diff
@@ -15239,15 +15239,35 @@ inspect_before:
   {
 	$$ = CB_BUILD_FUNCALL_1 ("cob_inspect_before", $3);
   }
+ | BEFORE _initial inspect_or_list
+  {
+	cb_verify(cb_inspect_or, _("INSPECT or"));
```
GitMensch: that string does not need a translation as both INSPECT and OR reference keywords; you possibly can improve that string in any case

emilienlemaire:
Hi Simon!
Thanks for the review, I'll edit the config files to the defaults you suggested.
I'll also mark the PR as draft, as I just forgot to check the box when creating the PR

lefessan:
What would be the better workaround to do in standard GnuCOBOL to do the same thing without INSPECT OR ? Because for me, if there is no better workaround, it should be enabled by default so that everybody can use it, no ?

GitMensch:
@emilienlemaire that looks good, doesn't it? Not sure why we still have that as a draft...
Can you rebase and add the new dialect option to all std's (as unconformable)?

PR #264, opened 2025-12-05
ENTRY statement should begin in area B not area A

rbowler:
When using a dialect configuration which includes `areacheck:yes` (for example, `-std=ibm-strict`), GnuCOBOL rejects an ENTRY statement coded in column 12 with the error message:  
`error: ENTRY should start in Area A`
This is incorrect, as ENTRY statements should begin in Area B (ref: [Enterprise COBOL Language Reference manual](https://www.ibm.com/docs/en/cobol-zos/6.4.0?topic=format-area))

This modification changes `parser.y` so that when `areacheck:yes` is in effect, ENTRY statements must start in Area B not Area A.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,7 @@
 
+2025-12-04  Roger Bowler <rbowler@snipix.net>
+	* parser.y: check that ENTRY statement begins in area B not area A
+
 2025-07-30  Simon Sobisch <simonsobisch@gnu.org>
 
 	* scanner.l, parser.y, reserved.c: use yyless to push part of scanned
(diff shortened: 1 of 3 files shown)


GitMensch (changes requested):
please add a testcase for that (ideally to the existing area check tests) and that's good to merge

GitMensch (approved):
Thanks, looks good - I'll directly check that in upstream.

GitMensch:
Thanks for your contribution - upstream with [r5624](https://sourceforge.net/p/gnucobol/code/5624/)

Oh and just a note if you like to "claim" your [contribution](https://openhub.net/p/11900/contributors/51112258355278?invite=) to GnuCOBOL on Open Hub, then you can do so with the link above [takes some hours/days to update, but still provides some interesting stats].

PR #277, opened 2026-03-02
Change is_test in cob_is_test

lefessan:
Very small change to avoid conflicts, since this variable  `is_test` is external, it should be prefixed with `cob_`.


GitMensch:
That's fine for me - add a Changelog entry and that can go upstream

GitMensch:
Included with [r5654]

PR #278, opened 2026-03-03
Initial addition of CONSTANT SECTION

RamyGeorge:
### **Summary**

This pull request introduces the CONSTANT SECTION as defined for Fujitsu Language reference. Its intended purpose is to introduce VALUES that do not change during program execution. For now, It expects record description entries.

### **Changes**

- Addition of the grammar for CONSTANT SECTION
- New Flag for COBC_HD_CONSTANT_SECTION

### **Motivation**

Specific section for adding CONSTANT values while enforcing proper syntax

### **Tests**

Test for the grammar is added at **tests/testsuite.src/syn_definition.at** 
Test for MOVE with a CONSTANT is expected to fail for now

--------------------------------------------------------------
Please give any feedback and/or any changes required to fulfill current PR goal.


cobc/field.c
```diff
@@ -3408,11 +3408,19 @@ cb_validate_field (struct cb_field *f)
...
 	}
-	if (f->flag_item_78) {
+	if (f->flag_item_78) {	
 		f->flag_is_verified = 1;
 		return;
 	}
 
+	if(f->storage == CB_STORAGE_CONSTANT){
+		if(!f->values){
+			cb_error(_("Items in CONSTANT SECTION must have  VALUE "));
```
GitMensch: formatting: see whitespace as used in other places (and drop the additional trailing spaces in line 3411, please)

msgid (that's hard to get right from the start, you only can get a _clue_ how it might look like when checking the others) should possibly be `"item in CONSTANT SECTION must have a VALUE clause"` (again no trailing space)

cobc/field.c
```diff
@@ -3408,11 +3408,19 @@ cb_validate_field (struct cb_field *f)
...
 		return;
 	}
-	if (f->flag_item_78) {
+	if (f->flag_item_78) {	
 		f->flag_is_verified = 1;
 		return;
 	}
 
+	if(f->storage == CB_STORAGE_CONSTANT){
+		if(!f->values){
```
GitMensch: I'd have to check the rules but it _could_ be that for a record structure only part of it needs a VALUE clause - I really don't know (the Fuji manual should have the rules in).

cobc/parser.y
```diff
@@ -297,7 +297,8 @@ enum cobc_hd {
 	COBC_HD_LINKAGE_SECTION		= (1U << 15),
 	COBC_HD_REPORT_SECTION		= (1U << 16),
 	COBC_HD_SCREEN_SECTION		= (1U << 17),
-	COBC_HD_PROCEDURE_DIVISION	= (1U << 18)
+	COBC_HD_PROCEDURE_DIVISION	= (1U << 18),
+	COBC_HD_CONSTANT_SECTION = (1U << 19)
```
GitMensch: this should have the right order to be able to use > / >= operators;
per Fuji manual CONSTANT SECTION goes before LINKAGE (As you did with the parser later on

cobc/parser.y
```diff
@@ -9077,6 +9079,21 @@ _local_storage_section:
...
 ;
 
+/* CONSTANT SECTION */
+
+constant_section: CONSTANT {check_area_a_of("CONSTANT SECTION");};
+_constant_section:
+
+| constant_section SECTION _dot
+  {
+	check_headers_present(COBC_HD_DATA_DIVISION,0,0,0); // check for Data division before hand
```
GitMensch: the comment is correct but superfluous here, note that per style guide: we only use C89 comment blocks (as GC still compiles on C89 compilers that don't have // inline commments)

cobc/field.c
GitMensch: you want `cb_error_x (CB_TREE(f), ....` here as the current line number is already after the field and you want its starting position

cobc/field.c
```diff
@@ -3408,11 +3408,19 @@ cb_validate_field (struct cb_field *f)
...
+			cb_error(_("Items in CONSTANT SECTION must have  VALUE "));
+		}
+		
+		CB_UNFINISHED ("CONSTANT SECTION code generation");
```
GitMensch: that would come for every line (please add a second var, maybe a record without VALUE to the test to highlight it) - most easy solution would possibly be to add it on the program validation level if constant_section is non-NULL

cobc/parser.y
```diff
@@ -7403,7 +7405,7 @@ constant_entry:
   {
 	cb_tree x;
 	const int level = cb_get_level ($1);
-
+
```
GitMensch: trim trailing whitespace

cobc/tree.h
```diff
@@ -1869,6 +1869,7 @@ struct cb_program {
 	struct cb_field		*linkage_storage;	/* LINKAGE */
 	struct cb_field		*screen_storage;	/* SCREEN */
 	struct cb_field		*report_storage;	/* REPORT */
+	struct cb_field     *constant_storage;  /* CONSTANT */
```
GitMensch, line 1873: that is the right thing to add - but you'd have to add the fields to that as well and then "read" the fields from there later on (have a look how that is done with working_storage)

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,23 @@ AT_DATA([prog.cob], [
...
+    CONSTANT SECTION.
+    01 NUM-CONST PIC 9(3).
+    PROCEDURE DIVISION.
+        STOP RUN.
+
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],[prog.cob:6: error: Items in CONSTANT SECTION must have VALUE
+])
+AT_CLEANUP
```
GitMensch: that wants a line break at the end; please also have a look at how the other tests are formatted (only slightly different)

tests/testsuite.src/syn_move.at
```diff
@@ -752,3 +752,25 @@ prog.cob:7: note: 'BIGFLT' defined here as USAGE FLOAT
 
 AT_CLEANUP
 
+
+
+AT_SETUP([MOVE to CONSTANT SECTION item])
+AT_KEYWORDS([extension constant move])
+
+AT_XFAIL_IF([true])
```
GitMensch, line 759: add a comment why that is an expected fail ("to be implemented soon" :-)

each new AT_SETUP gets two empty lines above, so please drop one

GitMensch (changes requested):
That's a good first take!
Please add cobc/Changelog to your changes as well-

please see https://sourceforge.net/p/gnucobol/wiki/Style%20guide/ for some general hints

The test currently fail (run `make check` and after the first one `make check TESTSUITEFLAGS=--recheck --verbose` locally to see them) because of

~~~
-prog.cob:6: error: Items in CONSTANT SECTION must have VALUE
+prog.cob:2: note: free format detected
+prog.cob:7: error: Items in CONSTANT SECTION must have  VALUE 
+prog.cob:7: warning: handling of CONSTANT SECTION code generation is unfinished; implementation is likely to be changed
~~~

for line two: start DIVISIONs and SECTIONs and paragraphs at column 8 and each "real" code at column 12; line 7 will be line 6 after the suggested change

GitMensch:
follow-up additions after the mentioned issues:
* have a test that MOVEs a constant to a field, then DISPLAYs that field
* have a look how level 78 constants define their constness in their definition - using the same or similar may solve the MOVE TO CONSTANT issue
* optional: in codegen possibly use the constant prefix (which is used for fields built from literals) instead of the field one
* ideally: in codegen generate the data in the same way the constant data (which is used for string literals) is setup
* follow-on:
    * add a new scope "CO" to -fdump option that handles the elements in the constant_storage, along with a test
    * internally set all level 78 constants and 01 cosntants defined in any storage into the constant_storage (so they will dump as well)

cobc/field.c
RamyGeorge: The only place I did find such validation is cb_validate_program_data. I am not exactly sure where to put it within all the logic but i have added a check for cb_storage_constant at the end and called for CB_UNFINISHED)

cobc/tree.h
RamyGeorge: This I'm not exactly sure what should be added. I thought this is mainly for building the AST so I'm not particularly sure what is missing

GitMensch: When producing AST we have a "current storage" where fields are added to (a linked list of 01/77 fields) - I think while the current_storage is setup correctly, the assignment to add fields to that list isn't.
For other storage elements those are already used for the symbol printing (have a look at code for `-tsymbols` - it actually should be added there) and at the end for the codegen.
The new storage will also be added (but just for the -fdump part) into codegen by starting from that linked list.

When you add the new storage to those two places, you'll likely see that they are "empty".

cobc/field.c
GitMensch: yes similar, but instead of the constant you'd check the program's ->constant_storage (once it gets any addition, I don't think the constant section elements currently do get linked there, but into the previous storage).

cobc/ChangeLog
```diff
@@ -1,3 +1,11 @@
+2026-03-03  Ramy George <ramygeorge19@gmail.com>
```
GitMensch, line 2: please start with a newline (that helps with a bunch of diff algrothms for some reason)

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,26 @@ AT_DATA([prog.cob], [
...
+AT_KEYWORDS([extension constant VALUE])
+
+AT_DATA([prog.cob],[
+    IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       CONSTANT SECTION.
+       01  NUM-CONST   PIC 9(3).
+       01  OTHER-CONST PIC 9(3) VALUE 123.
+       PROCEDURE DIVISION.
```
GitMensch, line 2989: please add a third constant with a constant record

tests/testsuite.src/syn_move.at
```diff
@@ -752,3 +752,46 @@ prog.cob:7: note: 'BIGFLT' defined here as USAGE FLOAT
...
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       CONSTANT SECTION.
+       01  NUM-CONST   PIC 9(3) VALUE 123.
+       PROCEDURE DIVISION.
+           MOVE 456 TO NUM-CONST.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE prog.cob], [1], [],[ignore])
```
GitMensch: you may add -Wno-unfinished to the command line - please add the expected error into stderr instead of ignore (see tests above how to split the AT_CHECK into multiple lines for that

tests/testsuite.src/syn_move.at
```diff
@@ -752,3 +752,46 @@ prog.cob:7: note: 'BIGFLT' defined here as USAGE FLOAT
...
+       WORKING-STORAGE SECTION.
+       01  RESULT-FLD  PIC 9(3).
+       PROCEDURE DIVISION.
+           MOVE MY-CONST TO RESULT-FLD.
+           DISPLAY RESULT-FLD.
+           STOP RUN.
+])
+# This is also not implemented yet but should help towards the bigger issue with CONSTANTs being replaced by literals
+AT_CHECK([$COMPILE prog.cob], [0], [], [])
+AT_CLEANUP
```
GitMensch: this test should be moved to run_extensions (with -Wno-unfinished in the compilation) and should include testing the actual result (which I guess will be fine);
please add another runtime test into run_extensions.at that has an alphanumeric constant and uses refererence modification [substring], like `MOVE MY-CONST (2:2) TO RESULT-FLD.` - I guess that will currently fail during compile.

cobc/typeck.c
```diff
@@ -6157,7 +6164,7 @@ enum_explain_storage (const enum cb_storage storage)
 {
 	switch (storage) {
 	case CB_STORAGE_CONSTANT:
-		return "Constants";
+		return "CONSTANT SECTION";
```
GitMensch: Constants may should be lower-cased (not sure), but CONSTANT SECTION is wrong here as other constants have the same storage but are not "living" in that section

GitMensch (changes requested):
mostly testing issues in this review iteration;

please have a look at the already specified changes to -fdump and add a new test to listings.at with -ftsymbol (just edit an existing test and add a CONSTANT SECTION into it, then see if the symbol listing includes those or not

tests/testsuite.src/run_extensions.at
```diff
@@ -6299,3 +6299,31 @@ AT_CHECK([$COMPILE -febcdic-symbolic-characters -febcdic-table=ebcdic500_latin1
...
+       WORKING-STORAGE SECTION.
+       01  RESULT-FLD  PIC X(2).
+       PROCEDURE DIVISION.
+           MOVE MY-CONST (2:2) TO RESULT-FLD.
+           DISPLAY RESULT-FLD NO ADVANCING.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE -Wno-unfinished prog.cob], [0], [], [])
+AT_CHECK([./prog], [0], [BC],[])
```
RamyGeorge: This is one of the weird suggestions i had from an agent?
it suddenly passed when adding this but i have since removed it and still worked so i guess there was a hallucination?! (which is to be expected it is basically the same test from syn_move.at)

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,59 @@ AT_DATA([prog.cob], [
 
 AT_CHECK([$COMPILE_ONLY -Wno-unfinished prog.cob], [0], [], [])
 AT_CLEANUP
+
+
+AT_SETUP([CONSTANT SECTION syntax enforced value])
+AT_KEYWORDS([extension constant VALUE])
+
+#THis fails because codegen is not implemented yet
+AT_XFAIL_IF([true])
```
RamyGeorge: This also fails so i marked with XFAIL since codegen is not implemented

tests/testsuite.src/run_extensions.at
RamyGeorge: 6326**

GitMensch: That should pass without a problem (ref-mod of level 78 constant currently can't work [I need to check if that should be allowed, not sure] and most likely not of 01 CONSTANT AS [which likely should be allowed, also needs to be checked]), so please drop that expected failure (which then shows the result in the CI as well.

GitMensch: This line should be replaced by `AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [BC])`

tests/testsuite.src/syn_definition.at
```diff
@@ -2949,24 +2949,57 @@ AT_CHECK([$COMPILE_ONLY -Wno-unfinished prog.cob], [0], [], [])
...
+       CONSTANT SECTION.
+       01  NUM-CONST   PIC 9(3) VALUE 444.
+       01  OTHER-CONST PIC 9(3) VALUE 123.
+       PROCEDURE DIVISION.
+           DISPLAY NUM-CONST.
+           DISPLAY OTHER-CONST.
+           STOP RUN.
+
+
+
```
GitMensch: drop empty lines

tests/testsuite.src/syn_definition.at
```diff
@@ -2949,24 +2949,57 @@ AT_CHECK([$COMPILE_ONLY -Wno-unfinished prog.cob], [0], [], [])
 AT_CLEANUP
 
 
+AT_SETUP([CONSTANT SECTION syntax enforced value])
+AT_KEYWORDS([extension constant VALUE])
+
+#THis fails because codegen is not implemented yet
```
GitMensch: this should still work in codegen, remove the expected failure and we see where the issue is
I'd expect the return code to be 0, not 1 as it is only a warning.

tests/testsuite.src/syn_definition.at
```diff
@@ -2949,24 +2949,57 @@ AT_CHECK([$COMPILE_ONLY -Wno-unfinished prog.cob], [0], [], [])
...
+
 AT_DATA([prog.cob],[
-    IDENTIFICATION DIVISION.
+       IDENTIFICATION DIVISION.
        PROGRAM-ID. prog.
        DATA DIVISION.
        CONSTANT SECTION.
        01  NUM-CONST   PIC 9(3).
        01  OTHER-CONST PIC 9(3) VALUE 123.
+       01  THIRD-CONST CONSTANT RECORD.
```
GitMensch: The CONSTANT RECORD case should be moved to the WORKING-STORAGE and instead add another item here `bad-record CONSTANT RECORD` that expects a compile-time error (see iso draft for the rules):

> The  CONSTANT  RECORD  clause  may  be  specified  only  in  the  local-storage  or  working-storage  [...].

tests/testsuite.src/syn_move.at
```diff
@@ -770,7 +770,9 @@ AT_DATA([prog.cob],[
            STOP RUN.
 ])
 
-AT_CHECK([$COMPILE prog.cob], [1], [],[ignore])
+AT_CHECK([$COMPILE -Wno-unfinished prog.cob], [1], [],
+[ignore
```
GitMensch: we don't ignore the output - we actually want to see an error here; for the time being please also drop the expected failure, allowing us to see what does not yet work

tests/testsuite.src/syn_move.at
GitMensch: please move that test before the new constant with refmod test to  run_extensions and check the result as well (without an expected failure)

tests/testsuite.src/syn_definition.at
GitMensch: please split the test - CONSTANT RECORD (which doesn't even need a VALUE at all) is not an extension, same for 01 CONSTANT AS

cobc/field.c
```diff
@@ -3413,6 +3413,12 @@ cb_validate_field (struct cb_field *f)
 		return;
 	}
 
+	if (f->storage == CB_STORAGE_CONSTANT) {
+        if (!f->values) {
+            cb_error_x (CB_TREE(f),("item in CONSTANT SECTION must have a VALUE clause"));
+        }
+	}
```
GitMensch: As the `else` branch we want to "internally move" standard-constant (CONSTANT RECORD and CONSTANT AS) into that storage, to use the "normal" code generation we'd have for WORKING-STORAGE - and to find them in the dump and symbol listing there.

cobc/parser.y
GitMensch, line 1: CONSTANT RECORD definition is missing, isn't it?
See iso draft for format / rules (only be specified in level 01, the ANY  LENGTH,  BASED,  BLANK  WHEN  ZERO,  DYNAMIC  LENGTH,  NULLABLE,  select-when, SYNCHRONIZED, and TYPEDEF may not be specified in them or subordinate items)

[site note: NULLABLE is not in yet - you could add it to the tokens, parse it (same place that SYNC may be in), raise a CB_UNSUPPORTED there and set a new flag "nullable" ijn the cb_field]

Note: as you change parser.y this year, please also adjust the year in the file's header (also applies to other files)

cobc/codegen.c
```diff
@@ -11608,6 +11608,27 @@ output_dump_code (struct cb_program *prog, cb_tree parameter_list)
 			}
 		}
 	}
+	if (prog->constant_storage) {
+		if (cb_flag_dump & COB_DUMP_CO) {
+			if (has_field_to_dump(prog->constant_storage)) {
+				has_dump = 1;
+				output_line ("/* Dump CONSTANT SECTION */");
```
GitMensch: The comment should say "Dump constants", also below - as this will also apply to the 01 CONSTANT AS, 01 CONSTANT RECORD and 78-constants.

GitMensch (changes requested):
let's have a look at the actual result by not suppressing the testsuite results and add CONSTANT RECORD and NULLABLE clause as well;
As this PR adds user-visible functionality (extension CONSTANT SECTION and CONSTANT RECORD support + dump + smybol listing for those) there should be 2 entries in the NEWS file as well.

cobc/field.c
RamyGeorge: only flag i found for constant is f->flag_constant 
I am assuming this is it and i am also assuming I'll have to get CONSTANT RECORD to have this flag aswell ?

GitMensch: > only flag i found for constant is f->flag_constant

There is `unsigned int flag_constant	: 1;	/* Is 01 AS CONSTANT */` in tree.h; that is also in one case to skip something (I haven't checked the details)

> I am assuming this is it and i am also assuming I'll have to get CONSTANT RECORD to have this flag aswell ?

Setting it would be reasonable, when doing that the comment needs to be adjusted to `/* Is 01 AS CONSTANT / CONSTANT RECORD (depending on ->children) */` or similar and add the check for the child at the place where currently something is skipped (if it should not apply to CONSTANT RECORD).

RamyGeorge: working on both the NULLABLE and CONSTANT RECORD right now
is there a SYN_CLAUSE for CONSTANT RECORD clause right now or do i add a new one? (no comments on which is which)

GitMensch: that sounds like an understanding issue of our bison use - let's make a 10 minute session out of that ;-)

cobc/parser.y
```diff
@@ -3042,6 +3042,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token NUMERIC
 %token NUMERIC_EDITED		"NUMERIC-EDITED"
 %token NUMVALC_FUNC		"FUNCTION NUMVAL-C"
+%token NULLABLE         "NULLABLE"
```
GitMensch, line 3046: you only need the text version if Bison should use that instead of printing the token name - which is fine as it is identical here :-)

cobc/parser.y
```diff
@@ -7542,6 +7557,10 @@ data_description_clause:
   {
 	CB_PENDING ("VALIDATE");
   }
+| NULLABLE
+  {
+	CB_UNSUPPORTED("NULLABLE");
+  }
```
GitMensch: make a new "nullable_clause" out of that - this also allows to do checks within that

cobc/parser.y
```diff
@@ -7425,6 +7425,21 @@ constant_entry:
 		}
 	}
   }
+| level_number user_entry_name CONSTANT RECORD
```
GitMensch: CONSTANT RECORD will be a "normal" clause, same as `nullable_clause` and also used at the same place; you can then use current_field->level or similar to do the check

cobc/typeck.c
```diff
@@ -1752,6 +1752,9 @@ cb_build_generic_register (const char *name, const char *external_definition,
 	}
 
 	if (current_program) {
+		if (field->storage == CB_STORAGE_CONSTANT) {
+			CB_FIELD_ADD(current_program->constant_storage, field);
+		}
```
GitMensch, line 1759: the code is correct, but it won't be triggered as there is currently no explicit call using CONSTANT STORAGE - so please add an `#ifdef 0 /* currently unreachable */` around that code (and add a space before the opening parenthesis as done in the code below)

cobc/typeck.c
```diff
@@ -5076,6 +5079,10 @@ cb_validate_program_data (struct cb_program *prog)
 			}
 		}
 	}
+
+	if (prog->constant_storage) {
+		CB_UNFINISHED ("CONSTANT SECTION code generation");
+	}
 }
```
GitMensch: let's move that to the parser

tests/testsuite.src/run_extensions.at
```diff
@@ -6299,3 +6299,53 @@ AT_CHECK([$COMPILE -febcdic-symbolic-characters -febcdic-table=ebcdic500_latin1
...
+       01  RESULT-FLD  PIC 9(3).
+       CONSTANT SECTION.
+       01  MY-CONST    PIC 9(3) VALUE 123.
+       PROCEDURE DIVISION.
+           MOVE MY-CONST TO RESULT-FLD.
+           DISPLAY RESULT-FLD.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE -Wunfinished prog.cob], [0], [],
```
GitMensch: just use `AT_CHECK([$COMPILE -Wno-unfinished prog.cob], [0], [], [])` here (it is a runtime check so we don't care about the warning) - and add the missing `AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [123])`

tests/testsuite.src/run_extensions.at
```diff
@@ -6299,3 +6299,53 @@ AT_CHECK([$COMPILE -febcdic-symbolic-characters -febcdic-table=ebcdic500_latin1
...
 
 AT_CLEANUP
+
+
+AT_SETUP([CONSTANT SECTION with reference modification])
+AT_KEYWORDS([extension constant move refmod])
+
+# Expected to fail until reference modification is 
+# fully implemented for CONSTANT literals.
+AT_XFAIL_IF([true])
```
GitMensch, line 6309: as noted before: that should already be fully supported

cobc/parser.y
```diff
@@ -3436,6 +3436,7 @@ set_record_size (cb_tree min, cb_tree max)
 
 %token LEVEL_NUMBER_IN_AREA_A	"level-number (Area A)"
 %token WORD_IN_AREA_A		"Identifier (Area A)"
+%token CONSTANT_RECORD
```
RamyGeorge: Putting this token definition in it lexcographical position broke the whole think 
it particularly lead to a "by 1 off" with reserved c

token WORDS comes before WORKING-STORAGE

declaring WORKING-STORAGE returns the WORDS token

GitMensch: Adding a token adjusts parser.h, if you get an off-by one that likely means some timestamps are screwed.
Try `make -C build/cobc clean && touch cobc/parser.y && make -C build/cobc` to ensure that this isn't a problem.

Also: your latest push nearly passes all CI (where everything is build from scratch).

cobc/ChangeLog
```diff
@@ -1,4 +1,15 @@
 
+2026-03-03  Ramy George <ramygeorge19@gmail.com>
```
GitMensch, line 2: last commit had no Changelog entry; as that wasn't pushed upstream yet I'd suggest to update the date and add the changes to this entry, but you can also create a new one if you prefer

cobc/cobc.c
GitMensch, line 1: the changes in this file miss an update to listings.at, verifying it to work as expected

cobc/cobc.h
```diff
@@ -440,9 +440,10 @@ extern int			cb_flag_dump;
 #define COB_DUMP_RD	0x0004	/* REPORT SECTION */
 #define COB_DUMP_SD	0x0008	/* FILE SECTION -> SORT DESCRIPTION */
 #define COB_DUMP_SC	0x0010	/* SCREEN SECTION */
-#define COB_DUMP_LS	0x0020  /* LINKAGE SECTION */
-#define COB_DUMP_LO	0x0040  /* LOCAL-STORAGE SECTION */
-#define COB_DUMP_ALL	(COB_DUMP_FD|COB_DUMP_WS|COB_DUMP_RD|COB_DUMP_SD|COB_DUMP_SC|COB_DUMP_LS|COB_DUMP_LO)
+#define COB_DUMP_CO 0x0020  /* CONSTANT SECTION*/
+#define COB_DUMP_LS	0x0040  /* LINKAGE SECTION */
+#define COB_DUMP_LO	0x0080  /* LOCAL-STORAGE SECTION */
+#define COB_DUMP_ALL	(COB_DUMP_FD|COB_DUMP_WS|COB_DUMP_RD|COB_DUMP_SD|COB_DUMP_SC|COB_DUMP_CO|COB_DUMP_LS|COB_DUMP_LO)
```
GitMensch: As this changes `-fdump and `-fdump=all` it needs an entry in the NEWS under the cobc options.
We could also think of changing the default (= if none is given) to "most" or similar, or - which is _possibly_ the right thing to do say that "ALL" dumps all storage, the constants are only dumped if explicit requested, for example with `-fdump=all,co` (in this case the COB_DUMP_ALL definition would stay as is).

In any case the behaviour should be clear in the NEWS entry and it should be visible in a test case as well.

cobc/codegen.c
GitMensch: This still applies.

cobc/parser.y
GitMensch: ... the definition of the CONTANT RECORD token below would currently display strange - so that _should_ have an explicit defined token name

cobc/parser.y
```diff
@@ -294,10 +294,11 @@ enum cobc_hd {
 	COBC_HD_WORKING_STORAGE_SECTION	= (1U << 12),
 	COBC_HD_COMMUNICATION_SECTION	= (1U << 13),
 	COBC_HD_LOCAL_STORAGE_SECTION	= (1U << 14),
-	COBC_HD_LINKAGE_SECTION		= (1U << 15),
-	COBC_HD_REPORT_SECTION		= (1U << 16),
-	COBC_HD_SCREEN_SECTION		= (1U << 17),
-	COBC_HD_PROCEDURE_DIVISION	= (1U << 18)
+	COBC_HD_CONSTANT_SECTION = (1U << 15),
```
GitMensch: minor: fix indent before the equal sign

cobc/parser.y
```diff
@@ -7422,7 +7425,7 @@ constant_entry:
 			(void)cb_validate_78_item (CB_FIELD (x), 0);
 		}
 	}
-  }
+  }
```
GitMensch: typo

cobc/parser.y
```diff
@@ -7536,12 +7539,31 @@ data_description_clause:
...
   {
 	CB_PENDING ("VALIDATE");
-  }
+  } 
 ;
 
+nullable_clause:
+  NULLABLE
+  {
+	CB_UNSUPPORTED("NULLABLE");
```
GitMensch, line 7553: setting the new nullable flag here is missing

cobc/scanner.l
```diff
@@ -999,6 +999,10 @@ H#[0-9A-Za-z]+ {
 	count_lines (yytext);
 	RETURN_TOK (LENGTH_OF);
 }
+"CONSTANT"[ \t]+"RECORD" {
+    count_lines(yytext);
```
GitMensch: please add the space before the opening (

cobc/typeck.c
GitMensch: still applies

cobc/typeck.c
GitMensch: still applies

cobc/typeck.c
GitMensch: still applies

tests/testsuite.src/run_extensions.at
GitMensch: still applies

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,70 @@ AT_DATA([prog.cob], [
...
+AT_SETUP([CONSTANT SECTION syntax enforced value])
+AT_KEYWORDS([extension constant VALUE])
+
+AT_DATA([prog.cob],[
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       CONSTANT SECTION.
+       01  NUM-CONST   PIC 9(3) VALUE 444.
+       01  OTHER-CONST PIC 9(3) VALUE 123.
```
GitMensch, line 2961: this test does not check what its title says

tests/testsuite.src/syn_definition.at
GitMensch: That comment of mine was misleading - I did not meant "a CONSTANT RECORD" but "a record that is a constant"; the check where CONSTANT RECORD is possible to use is to be done in its syntax test.

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,70 @@ AT_DATA([prog.cob], [
...
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 good-record CONSTANT RECORD.
+           05  SUB-CONST PIC 9(3) VALUE 123.
+       PROCEDURE DIVISION.
+       STOP RUN.    
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob] , [0] , [], [])
+AT_CLEANUP
```
GitMensch, line 3016: at file end: please add an empty line

tests/testsuite.src/syn_move.at
```diff
@@ -752,3 +752,26 @@ prog.cob:7: note: 'BIGFLT' defined here as USAGE FLOAT
...
+       DATA DIVISION.
+       CONSTANT SECTION.
+       01  NUM-CONST   PIC 9(3) VALUE 123.
+       PROCEDURE DIVISION.
+           MOVE 456 TO NUM-CONST.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE -Wunfinished prog.cob], [1], [],
+[prog.cob:7: warning: handling of CONSTANT SECTION code generation is unfinished; implementation is likely to be changed
```
GitMensch: similar as to above: we don't care about the unfinished here, so you can use -Wno-unfinished for the compile

tests/testsuite.src/syn_move.at
```diff
@@ -752,3 +752,26 @@ prog.cob:7: note: 'BIGFLT' defined here as USAGE FLOAT
...
+prog.cob:8: error: invalid MOVE target: 123
```
GitMensch: you may make that an expected failure and change the expected output to `invalid MOVE target: constant NUM-CONST`.

GitMensch (changes requested):
a bunch of older items are still open + some new - but overall there's visible progress 👍

cobc/parser.y
RamyGeorge: just a clarification on "would display strange"

PS: running ```make -C build/cobc clean && touch cobc/parser.y && make -C build/cobc```

didn't fix the make compiling still explodes with examples like

missing program-id even though it is there?

GitMensch: > just a clarification on "would display strange"

Whenever bison refers the tokens, for example in "expected TOKEN got TOKEN" it uses the token name.
The default is the string version of the enum value of the token "as-is".
This may lead to "Expected CONSTANT got CONSTANT_RECORD"; if you explicit define a name - here "CONSTANT RECORD" it will show that way.

GitMensch: Guess: you have parser.h in cobc and in build/cobc as you've executed bison in the source directory manually.
Am I right? (that guess _may_ be related to me doing that in the past as well...)

tests/testsuite.src/syn_definition.at
RamyGeorge: This means to test for the enforced VALUE caluse
in hindsight the one after with just the syntax has one entry with value and one without so i can possible just erase this

tests/testsuite.src/syn_definition.at
RamyGeorge: drop CONSTANT RECORD.
Add new test for CONSTANT RECORD ON WORKING/LOCAL STORAGE

tests/testsuite.src/syn_definition.at
RamyGeorge: drop a value
expect error for invalid constant section syntax

PR #279, opened 2026-03-16
Added initial EXEC preprocessor handling logic in pplex and ppparse.

utam-1:
### Brief description of PR
The aim of this PR is to add an initial foundation work related to : https://sourceforge.net/p/gnucobol/feature-requests/341/ 

### Changes made
The basic idea behind these changes are to initially detect and raise error for EXEC .... END EXEC block. In order to do so we defined a new state for the compiler <EXEC_STATE>, and necessary grammar rules for parser to handle these blocks, which handle the foreign embedded code.


cobc/pplex.l
```diff
@@ -569,6 +576,16 @@ MAYBE_AREA_A	[ ]?#?
   }
 }
 
+<EXEC_STATE>{
+        "END-EXEC"([ \t]*\.)?      { BEGIN (INITIAL); return END_EXEC; }
+        "INCLUDE"                  { return COPY; }
```
GitMensch: please add a new token INCLUDE handling it in the parser as that will also be used as part of the internal preprocessor's error messages

cobc/pplex.l
```diff
@@ -569,6 +576,16 @@ MAYBE_AREA_A	[ ]?#?
...
+        :[A-Za-z0-9][A-Za-z0-9-]* { pplval.s = cobc_plex_strdup (yytext); return TOKEN; }
+        [A-Za-z0-9][A-Za-z0-9-]*  { pplval.s = cobc_plex_strdup (yytext); return TOKEN; }
```
GitMensch: we only need the "simple" token (effectively a word) to get that into the parser - the rest like bind variables can be consumed silently here as well

cobc/ppparse.y
```diff
@@ -1726,6 +1731,40 @@ replacing_list:
...
+        $$ = NULL;
+  }
+| exec_token_list TOKEN
+  {
+        $$ = ppp_list_add ($1, $2);
+  }
+;
+
+exec_statement:
+  EXEC END_EXEC
```
GitMensch: as already discussed, the "empty" EXEC block is an error (that will be taken care of the error case below), so drop that part

cobc/ppparse.y
```diff
@@ -1726,6 +1731,40 @@ replacing_list:
...
 ;
 
+exec_token_list:
+  /* empty */
+  {
+        $$ = NULL;
+  }
+| exec_token_list TOKEN
+  {
+        $$ = ppp_list_add ($1, $2);
```
GitMensch: as we don't want to handle the token list we can just consume it here, which leads to a small definition that will be referenced below

```yacc
_exec_token_list:
| _exec_token_list TOKEN 
  {
	/* should be handled by preparser, ignored
	   if seen here purely for parsing reasons */
  }
;
````

File: cobc/ppparse.y [hunk 5]
```diff
@@ -1726,6 +1731,40 @@ replacing_list:
...
+;
+
+exec_statement:
+  EXEC END_EXEC
+  {
+        /* empty EXEC block — silently ignored */
+  }
+| EXEC TOKEN COPY copy_source END_EXEC
+  {
+        /* EXEC TAG INCLUDE copybook */
```
GitMensch: the include should be handled as COPY here

GitMensch (changes requested):
that's a good start; missing are: cobc/Changelog and an entry in NEWS with something like

> for parsing purposes, EXEC blocks are now recognized and handed with the "unsupported" warning level (defaulting to error) with INCLUDE statements handled as copybooks

along with test cases in syn_extensions.at (test an empty block, an empty SQL block and missing name after INCLUDE so we have a clue if error handling works, then another that from GixSQL's test cases which uses EXEC SQL INCLUDE with sqlca and another copybook, allowing the validation that the include did work correctly)

tests/testsuite.src/syn_extensions.at
```diff
@@ -0,0 +1,125 @@
...
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           EXEC SQL CONNECT RESET END-EXEC.
+           STOP RUN.
+])
+
+AT_CHECK([$COBC -fsyntax-only -fdiagnostics-plain-output -Wno-error=unsupported prog.cob], [0], [],
```
GitMensch: Should use `$COMPILE_ONLY -Wno-unsupported prog.cob`

tests/testsuite.src/syn_extensions.at
```diff
@@ -0,0 +1,125 @@
...
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           EXEC END-EXEC.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+[prog.cob:5: error: syntax error, unexpected END_EXEC, expecting Word or Literal
```
GitMensch: This shows that the tolen name END_EXEC should get a literal value "END-EXEC"

tests/testsuite.src/syn_extensions.at
```diff
@@ -0,0 +1,125 @@
...
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       EXEC SQL
+           INCLUDE EMPREC
+       END-EXEC.
+       PROCEDURE DIVISION.
+           MOVE 1 TO ENO.
+           STOP RUN.
+])
```
GitMensch: combine those to into a single test

tests/testsuite.src/syn_extensions.at
GitMensch, line 1: tests should be moved to the end of syn_definition.at

tests/test84.c
GitMensch, line 1: this and the other C files are unrelated, should be deleted

cobc/ChangeLog
```diff
@@ -1,4 +1,17 @@
 
+2026-03-16  Uttam Singh Bhadauriya <uttamsinghbhadoriya23@gmail.com>
+
+	* cobc/pplex.l: add EXEC_STATE exclusive start condition to
```
GitMensch: please drop the cobc/ prefix here - as that is the place where that Changelog is

GitMensch (changes requested):
overall a remarkably better commit than the last one; see more notes below

cobc/ppparse.y
```diff
@@ -753,6 +753,9 @@ ppparse_clear_vars (const struct cb_define_struct *p)
 %token <s> VARIABLE_NAME	"Variable"
 %token <s> LITERAL		"Literal"
 
+%token EXEC
+%token END_EXEC "END_EXEC"
```
GitMensch: `%token END_EXEC		"END-EXEC"` -> and adjust testsuite results

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,91 @@ AT_DATA([prog.cob], [
...
+
+
+AT_SETUP([EXEC SQL empty block])
+AT_KEYWORDS([exec sql extensions])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           EXEC SQL END-EXEC.
```
GitMensch: change that to "EXEC empty block", drop the sql from the keywords and use something else here - doesn't really matter what, just use EXEC CICS (which may be seen as well) or EXEC MYPROC (to make more clear we just test for "anything" here).

Benefit: this will show that the parsing works nonetheless and that the diagnostic does print the <lang> the code specifies.

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,91 @@ AT_DATA([prog.cob], [
...
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported prog.cob], [1], [],
```
GitMensch, line 2986: no need for the -Werror, drop it from the command line

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,91 @@ AT_DATA([prog.cob], [
...
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           EXEC SQL CONNECT RESET END-EXEC.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported prog.cob], [1], [],
+[prog.cob:5: error: EXEC SQL statement ignored - no preprocessor configured
+])
```
GitMensch: that is also not SQL specific but a general test, you may drop SQL from the title and keywords and drop -Werror

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,91 @@ AT_DATA([prog.cob], [
...
+       EXEC SQL
+           INCLUDE EMPREC
+       END-EXEC.
+       PROCEDURE DIVISION.
+           DISPLAY SQLCODE.
+           MOVE 1 TO ENO.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -ffold-copy=LOWER prog.cob], [0], [], [])
```
GitMensch: This should output a warning by default, no?

In any case: please add a minimal test for "not found", for example have a second COMPILE_ONLY without the fold-copy after a `[rm -f emprec.cpy]`

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,91 @@ AT_DATA([prog.cob], [
...
+
+AT_CLEANUP
```
GitMensch: please end with an empty line - and add the last missing test "END-EXEC missing" to ensure that is gracefully handled, something like

```m4
AT_SETUP([END-EXEC missing])
AT_KEYWORDS([exec extensions])

AT_DATA([prog.cob], [
       IDENTIFICATION DIVISION.
       PROGRAM-ID. prog.
       PROCEDURE DIVISION.
      * typo added on purpose to check handling that
           EXEC SQL CONNECT RESET ENDEXEC.
           STOP RUN.
])

[
???
])

AT_CLEANUP
```

cobc/pplex.l
```diff
@@ -569,6 +576,14 @@ MAYBE_AREA_A	[ ]?#?
   }
 }
 
+<EXEC_STATE>{
+    "END-EXEC"([ \t]*\.)?     { BEGIN (INITIAL); return END_EXEC; }
+    "INCLUDE"                 { return INCLUDE; }
+    [A-Za-z0-9][A-Za-z0-9-]* { pplval.s = cobc_plex_strdup (yytext); return TOKEN; }
```
GitMensch: Shouldn't we use `{WORD}` here instead of that list?

cobc/ppparse.y
```diff
@@ -1726,6 +1730,34 @@ replacing_list:
...
+  {
+        /* EXEC TAG INCLUDE copybook — handle as COPY */
+        fputc ('\n', ppout);
+        ppcopy ($4, NULL, NULL);
+  }
+| EXEC TOKEN _exec_token_list END_EXEC
+  {
+        /* EXEC TAG ... END-EXEC — warn and ignore */
+        cb_warning (cb_warn_unsupported,
+                    _("EXEC %s statement ignored - no preprocessor configured"), $2);
```
GitMensch: As we currently can't configure a preprocessor: drop the second part (adding it with the preprocessor calling later),

GitMensch (changes requested):
Hi Uttam,

that PR is nearly ready to go upstream. Please have a look at the following comments, push your changes again (should pass CI) and I'll commit it.

This provides the baseline for the preprocessor config file.

Side note: I've added you to "contrib devs" on SF, so you can commit to the contrib svn repo and have the GnuCOBOL project in the associations under your SF profile.

tests/testsuite.src/syn_definition.at
utam-1: I tried dropping -Werror but this was causing the test cases failure as compiler was exiting with code 0, I believe the reason for this is how the macro is defined in `atlocal.in` I think. Due to this I have kept -Werror.

cobc/pplex.l
utam-1: This was good suggestion, it made no sense to use this grep matcher when it is defined already.

tests/testsuite.src/syn_definition.at
GitMensch: fine to go with [0] and warning then

tests/testsuite.src/syn_definition.at
utam-1: This test case was quite useful, as it led me to identify a flaw in the EXEC state block defined in `pplex.l`. I added the following ` \.                        { BEGIN (INITIAL); return END_EXEC; }` in it, as this will treat a `.` as  an implicit END-EXEC terminator, which I believe is justified as periods ( or `.` ) basically end COBOL statements. Without this the test cases failed to run and resulted in a deadlock, possibly due to the fact that the parser waited for END_EXEC token forever.

GitMensch: blocks in END-EXEC may have embedded periods - very likely with SQL referencing table.field and also in literals which your changed rule likely does not cover, so I'd suggest to drop that.

Didn't bison raise something like "END-EXEC expected EOF found"? That would be a nice test result.
Note: you may need to include EOF in the exec parsing state somewhere.

tests/testsuite.src/syn_definition.at
GitMensch: ignore that - atlocal.in explicit disables the warning altogether, so enabling that as you did is the right move

tests/testsuite.src/syn_definition.at
GitMensch, line 3031: This should raise a warning, use `CB_UNSUPPORTED_X` in that case.

GitMensch (changes requested):

utam-1:
Hello Simon, 
I identified the issue related to the typo test cases which stalled the compiler, making it stuck in an infinite loop. To fix the same I used `cb_source_line++` for `\n` which tracks the actual line numbers, and also added a check in the global `<<EOF>>` rule for `<EXEC_STATE>`.
Apart from this I have also added the test case for the scenario in which `emprec.cpy` is deleted. Do let me know if I should make some other modifications.

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,121 @@ AT_DATA([prog.cob], [
...
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       EXEC SQL
+           INCLUDE SQLCA
+       END-EXEC.
+       EXEC SQL
+           INCLUDE EMPREC
```
GitMensch, line 3030: I've missed that before, we also need an example like `INCLUDE "EMPREC"`  -> literal and ideally also `INCLUDE SQLCA.cpy` -> embedded perioid

cobc/pplex.l
```diff
@@ -569,6 +576,14 @@ MAYBE_AREA_A	[ ]?#?
   }
 }
 
+<EXEC_STATE>{
+    "END-EXEC"([ \t]*\.)?     { BEGIN (INITIAL); return END_EXEC; }
+    "INCLUDE"                 { return INCLUDE; }
+    {WORD}                    { pplval.s = cobc_plex_strdup (yytext); return TOKEN; }
```
GitMensch: distinguish between a new `SUBSYSTEM "subsytem"` token and "common" `TOKEN`, which will improve diagnostics;
that may either be done with a new EXEC_STATE_START or with a static variable

GitMensch (changes requested):
That looks nice!

Just a minor thing: replace leading spaces by tab (I could do that on upstream-merge otherwise).

Some more things I've stumbled over now in the review comments below.


Question: Should we start a target branch here + then create one PR against that according to the GSoC proposal - or have a single new one where we do everything until we consider the external preprocessor handling finished?

utam-1:
> Just a minor thing: replace leading spaces by tab (I could do that on upstream-merge otherwise).

Yes that would be nice, I tried to replace them with `sed` but maybe it may not have worked.

> Question: Should we start a target branch here + then create one PR against that according to the GSoC proposal - or have a single new one where we do everything until we consider the external preprocessor handling finished?

I think keeping this as the baseline target branch and then creating a PR against that makes sense.

GitMensch:
sounds like a plan; I'd consider that branch fine now, so you may start with the next steps per your original GSoC plan (I only leave a minor change request, but that's really minor)

cobc/pplex.l
```diff
@@ -569,6 +578,39 @@ MAYBE_AREA_A	[ ]?#?
...
+        return END_EXEC;
+    }
+    "INCLUDE"               { return INCLUDE; }
+    {WORD}\.{WORD}          {
+        /* dotted name e.g. SQLCA.cpy or table.field */
+        pplval.s = cobc_plex_strdup (yytext);
+        if (!exec_subsystem_seen) {
+            exec_subsystem_seen = 1;
+            return SUBSYSTEM;
+        }
```
GitMensch: just always go with TOKEN here, that's no valid SUBSYSTEM

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,173 @@ AT_DATA([prog.cob], [
...
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported prog.cob], [1], [],
+[prog.cob:5: error: EXEC SQL statement ignored
+])
+
+AT_CLEANUP
+
+
+AT_SETUP([EXEC SQL INCLUDE sqlca and copybook])
+AT_KEYWORDS([exec sql include sqlca extensions])
```
GitMensch: all words in the test name are automatically added to the keywords and all keywords are case-senstive, so please cleanup the duplicates in all tests' keyywords; thanks

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,173 @@ AT_DATA([prog.cob], [
...
+
+AT_DATA([emprec.cpy], [
+	01 EMP-TABLE.
+	03 ENO    PIC S9(4) COMP.
+	03 LNAME  PIC X(10).
```
GitMensch: COBOL85 ruleset would not compile that; area A (columns 8-11) should have the level 01+77 + most "boilerplate", column 12+ have higher levels (like these) and all coded statements, see below

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,173 @@ AT_DATA([prog.cob], [
...
+	EXEC SQL
+	INCLUDE SQLCA
+	END-EXEC.
+	EXEC SQL
+	INCLUDE EMPREC
+	END-EXEC.
+	PROCEDURE DIVISION.
+	DISPLAY SQLCODE.
+	MOVE 1 TO ENO.
+	STOP RUN.
```
GitMensch: these are statements and start ideally at col12+(area b); also if there are statements then you should have a section or paragraph definition (those go to area a) - the easiest thing is go with `MAIN.`

Note: this does work "as is" but keeping a portable coding style is a good trait, so I suggest to go over those things for the test.

GitMensch (approved):

utam-1:
> sounds like a plan; I'd consider that branch fine now, so you may start with the next steps per your original GSoC plan (I only leave a minor change request, but that's really minor)

Thank you! 
A small follow-up question from my side: Should this PR be merged into the main branch, so that the baseline be established at first, then I can rebase and make next set of incremental changes for subsequent stages? ( I think it will make the subsequent PRs easier to review and may potentially help in case if a merge conflict arises )

GitMensch:
PR should be merged upstream, then I'll create a rebased branch that you can work from. I still wait for the FSF attorney to send the signed papers (will ping them mid/end of this week if I don't see any response, I'm actually waiting on multiple ones this time).

... but you can already start with that work, just don't push to this branch (either a separate one based on that or push later).

PR #281, opened 2026-03-25
feat: implement intrinsic function BOOLEAN-OF-INTEGER

kmr-srbh:
### Specification

`BOOLEAN-OF-INTEGER` returns a boolean item of usage bit representing the binary value
of `argument-1`. `argument-2` specifies the length of the boolean data item that is returned.

`argument-1` must be a positive integer.
`argument-2` must be a positive non-zero integer

Returned value is a boolean item of usage bit that has the same bit configuration as
the binary representation of the value of `argument-1`, where the rightmost boolean position
is the low-order binary digit. The boolean value is zero-filled or truncated on the left, if
necessary, in order to return a boolean item whose length is specified by `argument-2` in
terms of boolean positions.

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,9 @@
 
+2026-03-25	Saurabh Kumar <developer.saurabh@outlook.com>
+
+	* intrinsic.c, reserved.c, tree.c (cb_build_intrinsic):
+	  implement intrinsic function BOOLEAN-OF-INTEGER
+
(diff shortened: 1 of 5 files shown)


kmr-srbh:
As discussed with @GitMensch, the changes in this PR need to be added as an enhancement over the implementation of `BIT-OF` intrinsic, which does the same work as `BOOLEAN-OF-INTEGER`, but on alphanumeric literals considering full byte lengths instead of N bits. Marking this PR as draft for now.

PR #285, opened 2026-04-17
Add COMP-5 binary size test

rbowler:
This pull request is intended to replace https://github.com/OCamlPro/gnucobol/pull/197.

It adds a COMP-5 binary size test, and fixes a bug where PIC S9 COMP-5 fields were incorrectly generated as one-byte fields when binary-size=2-4-8.


GitMensch:
Boa, that is an old one... so we have some changes in #197 in codegen (referenced here in the Changelog but nothing in the changed files) and all of @rbowler , @ddeclerck and @GitMensch working on it.

The tests here look reasonable.... @lefessan Could you reabase that (here or in another PR) checking if any parts of codegen should be adjusted as well + have a look at the Changelog (strange that only my name was in there)?

Note: I've just verified that the 2-4-8 is definitely correct for acu (no easy micro-focus environment to test against any more) and the 2-4-8 returns the expected values.
Also tested the same on IBM - outputs the same numeric sizes.

Note: For portability reasons `LENGTH OF` should be used in the test instead of the intrinsic function:
* old acu doesn't have it
* IBM does not allow to DISPLAY a binary integer, which is the result of this function

Side note: IBM also does not allow

````cobol
       01 WS-SP17 PIC SVPP9(17) COMP-5.
       01 WS-SP18 PIC SVPP9(18) COMP-5.
````

complaining it includes more than 18 digits.

lefessan:
I have rebased it. It sounds like it's mostly the work of Roger, so maybe I can just squash everything in one of his commits and put the Changelog entry to his name ?

lefessan:
I cleaned the tests (BYTE-LENGTH replaced by LENGTH OF and WS-SP17/18 removed)

PR #287, opened 2026-04-20
feat (parser): add parsing support for all class attributes with "unimplemented" warning

kmr-srbh:
Adds support for the following class attributes:

* AS literal
* IS FINAL
* INHERITS FROM {class-name ...}
* USING {param-name ...}

Also support MF extensions IS STATIC, IS ABSTRACT, IS PARTIAL,
IS PUBLIC, IS INTERNAL.

MF docs: https://docs.rocketsoftware.com/de-DE/bundle/visualcoboldevhub_ug_110/page/pho1742953034730.html

### DIFF
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,13 @@
 
+2026-06-11	Saurabh Kumar <developer.saurabh@outlook.com>
+
+	* parser.y, reserved.c: add support for parsing class attributes - AS literal,
+	  IS FINAL, INHERITS FROM {class-name ...}, USING {param-name ...}
+	  with unimplemented warning; also support MF extensions IS STATIC, IS ABSTRACT,
+	  IS PARTIAL, IS PUBLIC, IS INTERNAL
+	* tree.h, common.h: add COB_MODULE_TYPE_CLASS and bit masks for all 
+	  object-oriented class attributes
+
 2026-06-08  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
 
 	* tree.h, parser.y: change type of cobc_cs_check flags to permit
--- a/cobc/parser.y
+++ b/cobc/parser.y
@@ -123,6 +123,7 @@
 #define	EVAL_DEPTH		32
 #define	PROG_DEPTH		16
 
+
 /* Global variables */
 
 struct cb_program		*current_program = NULL;    /* program in parse/syntax check/codegen */
@@ -1287,10 +1288,17 @@ begin_scope_of_program_name (struct cb_program *program)
 				    elt_name);
 			return;
 		} else if (strcmp (prog_id, elt_id) == 0) {
-			cb_error_x (CB_TREE(program),
-				    _("redefinition of program ID '%s'"),
-				    elt_id);
-			return;
+			if (program->prog_type == COB_MODULE_TYPE_PROGRAM) {
+				cb_error_x (CB_TREE(program),
+						_("redefinition of program ID '%s'"),
+						elt_id);
+				return;
+			} else if (program->prog_type == COB_MODULE_TYPE_CLASS) {
+				cb_error_x (CB_TREE(program),
+						_("redefinition of class ID '%s'"),
+						elt_id);
+				return;
+			}
 		}
 	}
 
@@ -1471,7 +1479,7 @@ setup_program (cb_tree id, cb_tree as_literal, const enum cob_module_type type, 
 			main_flag_set = 1;
 			current_program->flag_main = !!cobc_flag_main;
 		}
-	} else { /* COB_MODULE_TYPE_FUNCTION */
+	} else if (type == COB_MODULE_TYPE_FUNCTION) {
 		current_program->flag_recursive = 1;
 	}
 
@@ -1515,8 +1523,13 @@ decrement_depth (const char *name, const unsigned char type)
 	}
 
 	if (depth != d) {
-		cb_error (_("END PROGRAM '%s' is different from PROGRAM-ID '%s'"),
-			  name, stack_progid[depth]);
+		if (type == COB_MODULE_TYPE_PROGRAM) {
+			cb_error (_("END PROGRAM '%s' is different from PROGRAM-ID '%s'"),
+				  name, stack_progid[depth]);
+		} else if (type == COB_MODULE_TYPE_CLASS) {
+			cb_error (_("END CLASS '%s' is different from CLASS-ID '%s'"),
+				  name, stack_progid[depth]);
+		}
 	}
 }
 
@@ -2548,12 +2561,24 @@ set_record_size (cb_tree min, cb_tree max)
 	}
 }
 
+/* Object-oriented class */
+
+static COB_INLINE void
+set_oo_class_attr(enum cb_oo_class_attribute attr, const char* attr_name)
+{
+	if (current_program->oo_class_attributes & attr) {							
+		emit_duplicate_clause_message (attr_name);						
+	}	
+	current_program->oo_class_attributes |= attr;
+}
+
 %}
 
 %token TOKEN_EOF 0 "end of file"
 
 %token THREEDIMENSIONAL	"3D"
 %token ABSENT
+%token ABSTRACT
 %token ACCEPT
 %token ACCESS
 %token ACTIVEX			"ACTIVE-X"
@@ -2970,6 +2995,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token INDEX
 %token INDEXED
 %token INDICATE
+%token INHERITS
 %token INITIALIZE
 %token INITIALIZED
 %token INITIATE
@@ -2981,6 +3007,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token INSPECT
 %token INSTALLATION		/* remark: not used here */
 %token INTERMEDIATE
+%token INTERNAL
 %token INTO
 %token INTRINSIC
 %token INVALID			/* remark: not used here */
@@ -3167,6 +3194,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token PARAGRAPH
 %token PARENT
 %token PARSE
+%token PARTIAL
 %token PASSWORD
 %token PERFORM
 %token PERMANENT
@@ -3209,6 +3237,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token PROPERTY
 %token PROTECTED
 %token PROTOTYPE
+%token PUBLIC
 %token PURGE
 %token PUSH_BUTTON		"PUSH-BUTTON"
 %token QUERY_INDEX		"QUERY-INDEX"
@@ -3793,12 +3822,10 @@ end_class:
 	check_area_a_of ("END CLASS");
   }
   end_class_name _dot
-  /*
-	TODO
-  	{
-		clean_up_program ($3, COB_MODULE_TYPE_CLASS);
-  	}
-  */
+  {
+	clean_up_program ($3, COB_MODULE_TYPE_CLASS);
+  }
+
 ;
 
 end_function:
@@ -3985,38 +4012,94 @@ class_id_header:
 ;
 
 class_id_name:
-  CLASS_NAME
-  {
-	if (CB_REFERENCE_P ($1) && CB_WORD_COUNT ($1) > 0) {
-		redefinition_error ($1);
-	}
-	$$ = $1;
-  }
+  CLASS_NAME	{ $$ = $1; }
 | LITERAL
   {
 	cb_trim_program_id ($1);
   }
 ;
 
-class_id_paragraph:
-  class_id_header TOK_DOT
+parent_class_name: 
+  WORD
   {
-	CB_PENDING ("CLASS-ID");
+  	current_program->class_inheritance_list = 
+		cb_list_add(current_program->class_inheritance_list, $1);
   }
-  class_id_name _as_literal TOK_DOT
+;
+
+parent_class_name_list:
+  parent_class_name
+| parent_class_name_list parent_class_name
+;
+
+/* Parameterized classes not supported for now. */
+class_param_list:
+  WORD
+| class_param_list WORD
+;
+
+_inherits_phrase:
+  /* empty */
+| INHERITS _from parent_class_name_list
+;
+
+_using_phrase:
+  /* empty */
+| USING class_param_list
+;
+
+class_attribute:
+  _is STATIC		{ set_oo_class_attr(CB_OO_CLASS_ATTR_STATIC, "STATIC"); }
+| _is PARTIAL		{ set_oo_class_attr(CB_OO_CLASS_ATTR_PARTIAL, "PARTIAL"); }
+| _is FINAL			{ set_oo_class_attr(CB_OO_CLASS_ATTR_FINAL, "FINAL"); }
+| _is ABSTRACT		{ set_oo_class_attr(CB_OO_CLASS_ATTR_ABSTRACT, "ABSTRACT"); }
+| _is PUBLIC		{ set_oo_class_attr(CB_OO_CLASS_ATTR_PUBLIC, "PUBLIC"); }
+| _is INTERNAL		{ set_oo_class_attr(CB_OO_CLASS_ATTR_INTERNAL, "INTERNAL"); }
+;
+
+_class_attributes:
+  /* empty */		{ current_program->oo_class_attributes = CB_OO_CLASS_ATTR_NONE; }
+| _class_attributes class_attribute
+;
+
+/*
+ * CLASS-ID paragraph syntax is:
+ *
+ * CLASS-ID. object-class-name-1 [ AS literal-1 ] [ IS STATIC ]
+ * [ IS { PARTIAL, FINAL, ABSTRACT } ... ] [ IS { PUBLIC, INTERNAL } ]
+ * [ INHERITS FROM { object-class-name-2 } ... ]
+ * [ USING { parameter-name-1 } ... ] .
+ * 
+*/
+
+class_id_paragraph:
+  class_id_header TOK_DOT class_id_name _as_literal
   {
-	/* 
-	  TODO: The if block below is added for triggering
-	  a class redefinition error. This is not the correct
-	  way to do it since `current_program` is a dummy AST
-	  node here.
-	  Remove it when adding support for AST generation
-	  through `setup_program()`.
-	*/
-	if (CB_REFERENCE_P ($4)) {
-		cb_define ($4, CB_TREE (current_program));
+	if (setup_program ($3, $4, COB_MODULE_TYPE_CLASS, 1)) {
+		YYABORT;
 	}
+
+	__CS_CLEAR_ALL();
 	cobc_in_id = 0;
+
+	CB_UNSUPPORTED ("object-oriented COBOL");
+  }
+  _inherits_phrase
+  _class_attributes
+  _using_phrase
+  TOK_DOT
+  {
+	/* check consistency of current attribues */
+	if (current_program->oo_class_attributes & CB_OO_CLASS_ATTR_FINAL
+	 && current_program->oo_class_attributes & CB_OO_CLASS_ATTR_ABSTRACT) {
+		emit_conflicting_clause_message ("FINAL", "ABSTRACT");
+		current_program->oo_class_attributes &= ~CB_OO_CLASS_ATTR_FINAL;
+	}
+	if (current_program->oo_class_attributes & CB_OO_CLASS_ATTR_INTERNAL
+	 && current_program->oo_class_attributes & CB_OO_CLASS_ATTR_PUBLIC) {
+		emit_conflicting_clause_message ("INTERNAL", "PUBLIC");
+		current_program->oo_class_attributes &= ~CB_OO_CLASS_ATTR_INTERNAL;
+	}
   }
 ;
 
--- a/cobc/tree.h
+++ b/cobc/tree.h
@@ -514,6 +514,18 @@ enum cb_index_type {
 	CB_STATIC_INT_VARYING
 };
 
+/* Object-oriented class attributes */
+enum cb_oo_class_attribute {
+	CB_OO_CLASS_ATTR_NONE			= 0x00, 	/* No attribute specified */
+	CB_OO_CLASS_ATTR_FINAL			= 0x01,		/* Cannot be subclassed */
+	CB_OO_CLASS_ATTR_ABSTRACT		= 0x02,		/* Can be subclassed, but not instantiated */
+	CB_OO_CLASS_ATTR_PARTIAL		= 0x04,
+	CB_OO_CLASS_ATTR_STATIC			= 0x08,		/* All methods must be marked STATIC */
+	CB_OO_CLASS_ATTR_PUBLIC			= 0x10,
+	CB_OO_CLASS_ATTR_INTERNAL		= 0x20
+};
+
+
 /* Reserved word list structure */
 struct cobc_reserved {
 	const char	*name;		/* Word */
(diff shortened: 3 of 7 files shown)


nberth:
@kmr-srbh You can now rebase and retarget this PR onto `oo4gnucobol-3.x`.

kmr-srbh:
I am moving the tests into `syn_oo.at`.

cobc/parser.y
```diff
@@ -3910,12 +3912,48 @@ class_id_name:
...
+  | parent_class_names class_id_name { $$ = $1; }
+;
+
+class_param_list:
+    WORD { $$ = $1; }
+  | class_param_list WORD { $$ = $1; }
+;
+
+class_attribute:
+    AS LITERAL
```
nberth: Dont forget that you can probably reuse the `_as_literal` rule here.

kmr-srbh: Yes. :+1:

cobc/parser.y
```diff
@@ -3910,12 +3911,67 @@ class_id_name:
...
+| parent_class_names WORD 		{ $$ = $2; }
+;
+
+class_param_list:
+  WORD 							{ $$ = $1; }
+| class_param_list WORD 		{ $$ = $2; }
+;
+
+class_attribute:
+  AS LITERAL					{ $$ = $2; }
```
kmr-srbh: The `_as_literal` rule defines an empty production, which causes conflict with our rule for class attributes, which can also match empty. Since no part of this attribute is optional, I have written it directly here.

kmr-srbh:
I will next add some error tests for the attributes. Then either in this PR, or in a different one, I can work on adding the class specifier in the REPOSITORY paragraph.

cobc/parser.y
```diff
@@ -3910,12 +3911,67 @@ class_id_name:
...
+
+/* 
+ * TODO: Later while constructing AST,
+ * create a new field similar to
+ * `program_spec_list` and add to it. 
+ */
+parent_class_names:
+  WORD 							{ $$ = $1; }
+| parent_class_names WORD 		{ $$ = $2; }
+;
```
nberth, line 4033:
```
parent_class_names:
  parent_class_name
| parent_class_names parent_class_name
;
parent_class_name: WORD { $$ = $1 /* for now */; };
```

GitMensch: you want to do that part correctly from the start -> use a list_add, as it is done in other places

nberth: Maybe `class_name_list` instead, to avoid the `parent` qualifier.

cobc/ChangeLog
```diff
@@ -1,4 +1,10 @@
 
+2026-04-20	Saurabh Kumar <developer.saurabh@outlook.com>
+
+	* parser.y, reserved.c, tree.h: add support for parsing class attributes - `AS literal`,
+	  `IS FINAL`, `INHERITS FROM {class-name ...}`, `USING {param-name ...}`
+	  with unimplemented warning.
```
GitMensch: as this is a plain text file please drop the backticks ` - in your entry it works identical without those;
minor: here and in the entry below: drop the full-stop as the entries are not necessarily sentences

cobc/parser.y
```diff
@@ -2879,6 +2879,7 @@ set_record_size (cb_tree min, cb_tree max)
 %token INDEX
 %token INDEXED
 %token INDICATE
+%token INHERITS			"INHERITS"
```
GitMensch: the token name is identical to the internal definition - you can drop the explicit literal here

cobc/parser.y
```diff
@@ -3910,12 +3911,72 @@ class_id_name:
   }
 ;
 
+/* Keyword `IS` is optional. `FINAL` is required. */
+_is_final:
+  FINAL									{ $$ = $1; }
+| IS FINAL								{ $$ = $2; }
+;
```
GitMensch: totally GnuCOBOL specific:

we start optional items with an underscore; this element _needs_ to be specified (there is no empty rule) so you'd drop the leading underscore
as "optional or not" is already declared that way as optional or not, you can drop the comments saying exactly the same

the above also applies to other places of your changes

please check if that is noted in https://sourceforge.net/p/gnucobol/wiki/Style%20guide/ and suggest edits if it is missing

cobc/parser.y
```diff
@@ -3910,12 +3911,72 @@ class_id_name:
...
+
+class_param_list:
+  class_param_name
+| class_param_list class_param_name
+;
+
+class_attribute:
+  AS LITERAL					{ $$ = $2; }
+| _is_final
+| _inherits_from
```
GitMensch: can't this be 

```
| _is FINAL 	{ $$ = $2; } // note: does returning the FINAL token work here - can it be accessed "somehow useful" in the caller? For parsing _alone_ you don't need any return
| INHERITS _from parent_class_names			{ $$ = $3; }
```

that is the style we commonly use and would drop the additional rules above

cobc/parser.y
```diff
@@ -3910,12 +3911,72 @@ class_id_name:
...
+| USING class_param_list
+;
+
+class_attribute_list:
+  class_attribute
+| class_attribute_list class_attribute 
+;
+
+class_attributes:
```
GitMensch: this gets a leading underscore :-)

cobc/parser.y
```diff
@@ -3910,12 +3911,72 @@ class_id_name:
...
+ *
+*/
+
 class_id_paragraph:
   class_id_header TOK_DOT
   {
 	CB_PENDING ("CLASS-ID");
   }
-  class_id_name _as_literal TOK_DOT
+  class_id_name class_attributes TOK_DOT
```
GitMensch: both `AS literal` and `IS FINAL` should be moved here (using one optional definition for those, if not already existing) returning literal token/NULL something like cb_true/cb_false -> you can then easily access it later in the processing below

That applies similar to the optional INHERITS and USING, we only need a structure like "class_attribute_list" for things that can "occur in any order" - but that is not part of the syntax rules so you can go with direct positions, which commonly also improves bison-generated diagnostics, even if you add attributes to current_program (which, as you've recognized, currently doesn't work as the program would only be setup in the rule below

but: you likely can work around the "not defined yet" part by splitting this into

`class_id_name _as literal {/* definition here */} more_attributes TOK_DOT;`

also note that the referenced $4 below is - even if commented out - now out of bounds

tests/testsuite.src/syn_oo.at
```diff
@@ -85,3 +85,153 @@ prog.cob:6: error: redefinition of 'MyClass'
 prog.cob:3: note: 'MyClass' previously defined here
 ])
 AT_CLEANUP
+
+AT_SETUP([CLASS-ID Attribute Check - INHERITS FROM and USING])
```
GitMensch, line 102: style: two empty lines before AT_SETUP

tests/testsuite.src/syn_oo.at
```diff
@@ -85,3 +85,153 @@ prog.cob:6: error: redefinition of 'MyClass'
...
+AT_KEYWORDS([CLASS-ID INHERITS USING])
```
GitMensch: all words in the title are already inserted into the keywords, so none of those are necessary; you may want to add a "general" keyword for the testsuite though (not sure if you want to go with `oo` - that would be at least an option

cobc/parser.y
```diff
@@ -3910,12 +3911,72 @@ class_id_name:
...
+ * [ USING { parameter-name-1 } ... ] .
+ * 
+ * We allow class attributes in any order.
+ *
+*/
+
 class_id_paragraph:
   class_id_header TOK_DOT
   {
 	CB_PENDING ("CLASS-ID");
```
GitMensch: As the whole OO part is not only pending but won't work at all (that will be changed later on) I'd suggest to change that to `CB_UNSUPPORTED()` using something like "object-oriented COBOL"

tests/testsuite.src/syn_oo.at
```diff
@@ -85,3 +85,153 @@ prog.cob:6: error: redefinition of 'MyClass'
...
+        CLASS-ID. FINAL MySixthClass.
+        END CLASS MySixthClass.
+])
+
+AT_CHECK([$COMPILE_ONLY prog6.cob], [1], [],
+[prog6.cob:3: warning: CLASS-ID is not implemented
+prog6.cob:3: error: syntax error, unexpected Identifier, expecting .
+])
+
+AT_CLEANUP
```
GitMensch: these are all testing bison tokenization in different error cases, I'd commonly only do "good tests" here as the "bad options" are nearly infinite

checking error handing is good for _internally_ raised errors and error-recovery - but those don't apply here (yet)

GitMensch (changes requested):
this is a good time to start learning about not only bison features but also our internal style - as noted below there is most likely documentation missing for this, then _please_ add it to the wiki / make suggestions

... and now I'm keen to see your next iteration

cobc/parser.y
kmr-srbh: Do you mean the attributes need to be ordered as they are declared in the standard? We decided to make the attribute syntax explicitly unordered.

GitMensch: yes, that **was** what I've meant (per standard rules) - but rechecking: the mf docs don't say so (they commonly do) but its compiler accepts that in any order, so stay with the general approach (while still moving to `class-id_name _as_literal { create_code() } _class_attributes TOK_DOT;` - after the create_code that would setup current_program, you can directly access it in the separate attributes , checking for duplicates by non-zero/null values.

tests/testsuite.src/syn_oo.at
kmr-srbh: I agree the incorrect cases are a lot. 👍

cobc/parser.y
kmr-srbh: Thanks for helping with this! :)

cobc/parser.y
```diff
@@ -3910,13 +3911,59 @@ class_id_name:
...
-  class_id_name _as_literal TOK_DOT
+;
+
+parent_class_name_list:
+  parent_class_name
+| parent_class_name_list parent_class_name
+;
+
+/* Parameterized classes not supported for now. */
+class_param_name: WORD { $$ = $1; };
```
GitMensch: this rule is automatic, no need to add it when there is no other part

if you don't need to do an explicit check of the WORD somehow, then I'd suggest to directly insert that instead of adding a rule containing only that token

cobc/parser.y
```diff
@@ -3910,13 +3911,59 @@ class_id_name:
...
+
+class_attribute_list:
+  class_attribute
+| class_attribute_list class_attribute 
+;
+
+_class_attributes:
+  /* empty */ 					{ $$ = NULL; }
+| class_attribute_list
+;
```
GitMensch: you'd directly set the attributes, so no need to have an explicit null return

cobc/parser.y
```diff
@@ -3910,13 +3911,59 @@ class_id_name:
...
+class_param_name: WORD { $$ = $1; };
+
+class_param_list:
+  class_param_name
+| class_param_list class_param_name
+;
+
+class_attribute:
+  _is FINAL									{ $$ = $2; }
+| INHERITS _from parent_class_name_list		{ $$ = $3; }
```
GitMensch: The second is already attached to the program, why not adding and setting an enum for the first (FINAL; MF also has PARTIAL (whatever that is) and ABSTRACT, you can have zero as unset to raise duplication errors if already set) as well?

In any case you then don't need to set up any return tokens here.

cobc/parser.y
```diff
@@ -3910,13 +3911,59 @@ class_id_name:
...
+| USING class_param_list
```
GitMensch, line 4048: May this be attached the same way it is done for program-prototpyes?
Just wondering...

cobc/parser.y
```diff
@@ -3910,13 +3911,59 @@ class_id_name:
...
+/* Parameterized classes not supported for now. */
+class_param_name: WORD { $$ = $1; };
+
+class_param_list:
+  class_param_name
+| class_param_list class_param_name
+;
+
+class_attribute:
+  _is FINAL									{ $$ = $2; }
```
GitMensch: note: MF has as a separate group (so a second enum to the FINAL one) a PUBLIC/INTERNAL attribute, it seems reasonable to add it here as well as checking for duplicate uses of it

tests/testsuite.src/syn_oo.at
```diff
@@ -79,9 +81,82 @@ AT_DATA([prog.cob], [
...
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+[prog.cob:3: error: object-oriented COBOL is not supported
+prog.cob:6: error: object-oriented COBOL is not supported
+prog.cob:10: error: object-oriented COBOL is not supported
+prog.cob:14: error: object-oriented COBOL is not supported
+prog.cob:19: error: object-oriented COBOL is not supported
+])
+AT_CLEANUP
```
GitMensch: end a file with a single emtpy line, please

tests/testsuite.src/syn_oo.at
```diff
@@ -60,14 +61,15 @@ AT_DATA([prog2.cob], [
...
 prog2.cob:4: error: syntax error, unexpected END PROGRAM, expecting END CLASS
 ])
 
 AT_CLEANUP
 
-AT_SETUP([Redefinition of class])
-AT_KEYWORDS([CLASS-ID])
+
+AT_SETUP([Redefinition of CLASS-ID])
+AT_KEYWORDS([])
```
GitMensch: please add a keyword that you use for all oo tests (syntax and runtime) as this allows running all of them with a single keyword, _maybe_ just oo (but that seems to be a strange keyword, so you possibly come up with something else which is short and better)

GitMensch (changes requested):
That looks quite nice, so possibly ready for feature branch upstream before working on the next constructs?

cobc/parser.y
kmr-srbh: Yes. 👍

tests/testsuite.src/syn_oo.at
kmr-srbh: Can we use CLASSES or OOP? I think the keyword CLASS means something else also in COBOL.

cobc/parser.y
kmr-srbh: I have local changes for this, but it seems this is similar to `public` and `protected` class accessibility specifiers like C++, which appear only with the inheritance specifier. How does MF handle this syntax?

cobc/parser.y
kmr-srbh: Isn't the `INTERFACE-ID` paragraph similar to program prototypes? From what I saw in the Standard docs, the `USING` syntax looks like defining args for a class constructor.

tests/testsuite.src/syn_oo.at
GitMensch: I'm fine with OOP.

cobc/tree.h
```diff
@@ -514,6 +514,16 @@ enum cb_index_type {
 	CB_STATIC_INT_VARYING
 };
 
+/* Class inheritance type */
+enum cb_inheritance_type {
+	CB_DEFAULT = 0, 	/* Can be subclassed */
+	CB_FINAL,			/* Cannot be subclassed */
+	/* Types below are non-standard */
+	CB_ABSTRACT,
+	CB_PARTIAL
```
GitMensch: use a prefix, similar to the INTR and PERFORM, so possibly CB_INH_ or CB_INHERITANCE_

cobc/parser.y
```diff
@@ -3910,13 +3915,95 @@ class_id_name:
...
+	/* TODO: Uncomment when fully initializing `current_program`.
+	   		 A duplicate error will get trigerred if used now since
+			 the same dummy `current_program` value is being reused.
+
+		if (current_program->inheritance_type != CB_FINAL) {
+			current_program->inheritance_type = CB_FINAL;
+		} else {
+			(void) cb_syntax_check (_("attribute FINAL already set"));
+		}
+	*/
```
GitMensch: Why not enabling that?
note: you want to compare to "!= DEFAULT" to catch if the attribute was already set and don't distinguish between the diagnostics

cobc/parser.y
```diff
@@ -3910,13 +3915,95 @@ class_id_name:
...
+ * 
+ * We allow class attributes in any order.
+ *
+*/
+
+class_id_paragraph:
+  class_id_header TOK_DOT class_id_name _as_literal
+  {
+	CB_UNSUPPORTED ("object-oriented COBOL");
+
```
GitMensch: you definitely want to add the class definition here (similar as it is done for the prototypes, most likely)

GitMensch (changes requested):
please try to limit review requests outside of your mentoring to one every 1-2 weeks, or when you feel stuck

cobc/parser.y
GitMensch: have a look at their docs https://docs.rocketsoftware.com/de-DE/bundle/visualcoboldevhub_ug_110/page/pho1742953034730.html - if you are then unsure, feel free to ask more specific

cobc/parser.y
```diff
@@ -3910,13 +3915,95 @@ class_id_name:
...
+  _is FINAL
+  {
+	/* TODO: Uncomment when fully initializing `current_program`.
+	   		 A duplicate error will get trigerred if used now since
+			 the same dummy `current_program` value is being reused.
+
+		if (current_program->inheritance_type != CB_FINAL) {
+			current_program->inheritance_type = CB_FINAL;
+		} else {
+			(void) cb_syntax_check (_("attribute FINAL already set"));
```
nberth:
```suggestion
			(void) cb_syntax_check (_("attribute %s already set"), "FINAL");
```
(This reduces the needs for several translations)

GitMensch: "yes" for msgid in general, "no" for this specific case - as the check still needs to be adjusted to "no inheritance type set"; if you want to make that nice, then adjust the check and call `emit_conflicting_clause_message()` or similar.

cobc/tree.h
```diff
@@ -514,6 +514,18 @@ enum cb_index_type {
 	CB_STATIC_INT_VARYING
 };
 
+/* Object-oriented class attributes */
+enum cb_oo_class_attributes {
+	CB_OO_CLASS_ATTR_DEFAULT		= 0x00, 	/* No attributes specified */
```
nberth: I think that should be `NONE` instead of `DEFAULT`. Translated into a default set of flags later during validation.

cobc/parser.y
```diff
@@ -3897,38 +3921,94 @@ class_id_header:
...
-	  Remove it when adding support for AST generation
-	  through `setup_program()`.
-	*/
-	if (CB_REFERENCE_P ($4)) {
-		cb_define ($4, CB_TREE (current_program));
+	if (setup_program ($3, $4, COB_MODULE_TYPE_CLASS, 1)) {
+		YYABORT;
 	}
+
+	cobc_cs_check = 0;
```
nberth: `__CS_CLEAR_ALL ()` now.

cobc/parser.y
```diff
@@ -123,6 +123,17 @@
...
+
+#define SET_OO_CLASS_ATTR(attr)												\
+	do {																	\
+		if (current_program->oo_class_attributes 							\
+			& CB_OO_CLASS_ATTR_##attr) {									\
+			emit_duplicate_clause_message (#attr);							\
+		}																	\
+		current_program->oo_class_attributes |= CB_OO_CLASS_ATTR_##attr;	\
+	} ONCE_COB																\
+
```
GitMensch: This can be easily moved to a function expecting an `enum cb_oo_class_attributes` (the type should be renamed to drop the plural s btw); to get the text for message you add either a lookup function with a switch statement or a constant char*[] lookup table on your own to not have an additional function call.

Reasoning: functions are shown in strack traces, profiles, coverage, while macros aren't. They are much easier to understand and search for compared to name concatenation and also to debug (depending on compile/debug you cannot step through them and `watch`ing elements changed in there can be quite confusing); the compiler is free to inline or not (a macro needs to be inlined).
Only downside: doesn't look as clever as a string concatenation macro, but with the years I've come to the understanding the "more simple" is actually "more clever" for maintenance (including understanding, onboarding, ....) - and there are enough places left where cleverness is necessary, for example good AST and codegen :-)

cobc/parser.y
```diff
@@ -3985,38 +4011,94 @@ class_id_header:
...
+	CB_UNSUPPORTED ("object-oriented COBOL");
+  }
+  _inherits_phrase
+  _class_attributes
+  _using_phrase
+  TOK_DOT
+  {
+	/* check consistency of current attribues */
+	if (current_program->oo_class_attributes & CB_OO_CLASS_ATTR_FINAL 
+		&& current_program->oo_class_attributes & CB_OO_CLASS_ATTR_ABSTRACT) {
```
GitMensch: minor formatting "Ron Norman's indent rule taken over to GnuCOBOL":

```c
	if (current_program->oo_class_attributes & CB_OO_CLASS_ATTR_FINAL 
	 && current_program->oo_class_attributes & CB_OO_CLASS_ATTR_ABSTRACT) {
```

That _should_ be noted in the style guide (wiki on SF), isn't it?

GitMensch (changes requested):
I'm not sure if there's anything left "feature wise" for this PR - overall it looks quite good to me.

cobc/parser.y
kmr-srbh: I am sorry. I created that macro like that just for shorter calls with one arg inside the grammar rule. I will change it to a function.

GitMensch: No need to be sorry, as noted that does look clever - but GSoC is also about learning "different approaches" (and hopefully their reasoning) and "house rules" (we potentially don't have that one in the style guide, I _think_ you should be able to edit the wiki or draft a change - please add the "unwritten" things that come up during GSoC there).

kmr-srbh: Sure, I'll add it. :)

kmr-srbh:
The CI failed as:

```
  cd build_windows
  set CL=/I "%VCPKG_ROOT%\installed\x64-windows\include\libxml2"
  "%MSBUILD%" .\vs2019\GnuCOBOL.sln /m /p:Platform=x64 /p:Configuration=Debug
  shell: C:\Windows\system32\cmd.EXE /D /E:ON /V:OFF /S /C "CALL "{0}""
  env:
    GC_VERSION: GnuCOBOL 3.3-dev
    FLEXBISON: https://github.com/lexxmark/winflexbison/releases/download/v2.5.25/win_flex_bison-2.5.25.zip
    MSBUILD: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\MSBuild.exe
    VCVARS32: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars32.bat
    VCVARS64: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat
    VCPKG_ROOT: C:\vcpkg
    VCPKGS32: mpir:x86-windows pdcurses:x86-windows berkeleydb:x86-windows libxml2:x86-windows cjson:x86-windows
    VCPKGS64: mpir:x64-windows pdcurses:x64-windows berkeleydb:x64-windows libxml2:x64-windows cjson:x64-windows
    MSYS2_ROOT: C:\msys64
    MSYSTEM: UCRT64
    MSYSPKGS: autoconf
    VCVARS: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat
    VCPKGS: mpir:x64-windows pdcurses:x64-windows berkeleydb:x64-windows libxml2:x64-windows cjson:x64-windows
    ARCHDIR: x64
The system cannot find the path specified.
Error: Process completed with exit code 1.
```

The Visual Studio version was bumped to 2026. We recently fixed this same issue at LFortran with https://github.com/lfortran/lfortran/pull/11811/changes. I can push a PR for this if the changes in the linked PR seem fine.

GitMensch:
> The Visual Studio version was bumped to 2026. We recently fixed this same issue at LFortran with https://github.com/lfortran/lfortran/pull/11811/changes. I can push a PR for this if the changes in the linked PR seem fine.

Yes, please create a separate PR to the gitside-3.x branch that replaces all 2022 with the latest option, maybe we don't need the vars where those are used any more).

PR #288, opened 2026-04-27
Fix false warning IMPLIED PIC

RamyGeorge:
Fix for [https://sourceforge.net/p/gnucobol/discussion/help/thread/941d670883/](url)

False IMPLIED PIC

### DIFF
--- a/cobc/field.c
+++ b/cobc/field.c
@@ -1916,6 +1916,7 @@ warn_pic_for_numeric_value_implied (const struct cb_field * const f)
 	cb_tree first_value = get_first_value (f);
 	if (first_value && CB_NUMERIC_LITERAL_P (first_value)) {
 		const cb_tree	x = CB_TREE (f);
+		if (f->pic) return 0;
 		/* TODO: Change to dialect option */
 		cb_warning_x (cb_warn_additional, x,
 			_("'%s' has numeric VALUE without PIC; PIC will be implied"),
--- a/tests/testsuite.src/syn_screen.at
+++ b/tests/testsuite.src/syn_screen.at
@@ -1,4 +1,4 @@
-## Copyright (C) 2014-2023 Free Software Foundation, Inc.
+## Copyright (C) 2014-2023, 2026 Free Software Foundation, Inc.
 ## Written by Simon Sobisch, Edward Hart, Ron Norman
 ##
 ## This file is part of GnuCOBOL.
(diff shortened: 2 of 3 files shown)


cobc/field.c
```diff
@@ -1916,6 +1916,9 @@ warn_pic_for_numeric_value_implied (const struct cb_field * const f)
 	cb_tree first_value = get_first_value (f);
 	if (first_value && CB_NUMERIC_LITERAL_P (first_value)) {
 		const cb_tree	x = CB_TREE (f);
+		if (f->pic){
```
GitMensch: minor formatting; space after `)`

tests/testsuite.src/syn_screen.at
```diff
@@ -905,3 +905,29 @@ AT_DATA([prog.cob], [
 AT_CHECK([$COMPILE_ONLY prog.cob], [0], [], [])
 
 AT_CLEANUP
+
+AT_SETUP([NUMERIC PIC WITH VALUE in SCREEN SECTION])
```
GitMensch, line 909: minor formatting: before AT_SETUP there go two empty lines, after the final AT_CLEANUP there's one empty line

tests/testsuite.src/syn_screen.at
```diff
@@ -905,3 +905,29 @@ AT_DATA([prog.cob], [
...
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. testsection.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01  x PIC X.
+       01  num PIC 9.
+
```
GitMensch: those are not used, so the four lines should be deleted

tests/testsuite.src/syn_screen.at
GitMensch, line 1: that file wasn't changed in a long time - please add `, 2026` to the file's header

GitMensch (changes requested):
there is only 1 part missing: cobc/field.c, then it can go upstream - ideally with the minor things noted here finished

Can you please either do a follow-up commit or push a revised version (doesn't make a difference for upstream) with that?

GitMensch:
@RamyGeorge as copyright papers are fine, can you please rebase (which should fix CI) for a final look before we merge that upstream?

GitMensch (approved):
minimal cleanup can be done when putting upstream - thanks for that fix

lefessan:
Shall I push this one into SVN ?

GitMensch:
thanks for the offer, yes - please do so (but the changelog entry should be moved one down)

lefessan:
Is there something to do in `svn` to attribute the work to RamyGeorge instead of myself ?

lefessan:
Merged into SVN

GitMensch:
> Merged into SVN

Thanks.

> Is there something to do in svn to attribute the work to RamyGeorge instead of myself ?

Nope, we could only add extra properties which you manually have to set and won't be visible in most clients.
I've changed the attributes of the last two commits in svn now (there is a a server-side pre-check allowing that only for brian and me - no need to change that because once we reached codeberg, that all will be moot).

PR #289, opened 2026-05-03
Initial @file

RamyGeorge:
### Summary 
This pull request is intended to introduce the "@FILE" command line option to GnuCOBOL.

### Motivation
Such option would no longer limit the amount of options to that of a given OS's Terminal size.

### Changes

- added initial implementation for "@" detection and reading File contents
- Called new ```process_at_file``` in  ```process_command_line```

### Misc
I Have left a few notes on current changes and hope for feedback


cobc/cobc.c
```diff
@@ -3155,7 +3156,17 @@ process_command_line (const int argc, char **argv)
...
-
+    cobc_compile_mode = 0;
+	int atfile_size;
+	char ** atfile_options;
+	for(int i = 1 ;i <argc; ++i){
+		atfile_options = process_at_file(argc[i], &cobc_compile_mode, &atfile_size);
+		if (atfile_options) break;
+	}
+	if (cobc_compile_mode) {
+
```
RamyGeorge: I initially thought i could "naively" realloc argv with new input but a simple Stack Overflow entry proved this can't be done

libcob/cobgetopt.c
```diff
@@ -476,6 +475,51 @@ _getopt_initialize (const char *optstring)
   return optstring;
 }
 
+char *
+process_at_file(const char* argument_values,int* compile_mode,int* atfile_size){
```
RamyGeorge: I am not sure how people feel about output paramterers in GnuCOBOL and since the latter 2 are i figured i should ask

libcob/cobgetopt.c
```diff
@@ -476,6 +475,51 @@ _getopt_initialize (const char *optstring)
...
+					/* Ptr->endptr carries the directory */
+		char* path_str = malloc(path_size);
+        memcpy(path_str, ptr, path_size - 1);
+        path_str[path_size - 1] = '\0'; 
+		FILE* path = fopen(path_str, "r");
+		if (!path) {
+			free(path_str);
+			fprintf(stderr,"File is non existent or invalid path");
+			return "";
+		}
```
GitMensch: as seen in the GCC "spec" (documentation for @file) -> there is no warning/error to raise, the "option" is just passed as a "normal" file, so cobc will then do its normal error handling later

libcob/cobgetopt.c
```diff
@@ -554,7 +598,11 @@ cob_getopt_long_long (const int argc, char *const *argv, const char *optstring,
     print_errors = 0;  
 
   /* Test whether ARGV[optind] points to a non-option argument.  */
-#define NONOPTION_P (argv[optind][0] != '-' || argv[optind][1] == '\0')
+#define NONOPTION_P ((argv[optind][0] != '-' && argv[optind][0] != '@') || argv[optind][1] == '\0')
+
+  if (argv[optind][0] == '@'){ 
+	return '@';		/*Most definitly is wrong but im messing around*/
+  }
```
GitMensch: as discussed - we need to do it all up-front and build a completely new argv/argc before even going to the getopt call

GitMensch (changes requested):
we'll need a force-push from a new local branch to not includes screen changes, it may or may not cherry-pick the current commit on that)

We then need test cases which also shows what the implementation wants to provide and then can have a look if the proposed changes work that way (= work on getting it there).

cobc/cobc.c
```diff
@@ -3441,7 +3444,7 @@ process_command_line (const int argc, char **argv)
 			res = gentable (stdout, code_ebcdic, code_ascii, reversible);
 			exit (res ? EXIT_FAILURE : EXIT_SUCCESS);
 		}
-
+
```
GitMensch, line 3446: those changes are not relevant any more, so you can get rid of them

cobc/cobc.h
```diff
@@ -453,6 +453,7 @@ extern int			cobc_flag_main;
 extern int			cobc_wants_debug;
 extern int			cb_listing_xref;
 extern int			cobc_seen_stdin;
+extern int			cobc_compile_mode;	/*0 for in-terminal command options, 1 for @file */
```
GitMensch: unused so can be dropped

libcob/cobgetopt.h
```diff
@@ -117,6 +117,9 @@ struct option
    arguments to the option '\0'.  This behavior is specific to the GNU
    `getopt'.  */
 
+char * process_at_file(const char* argument_values,int* compile_mode,int* atfile_size);
+char** expand_processed_at_file(char *options, int *output_argc);
+void rebuild_argv_at_file(int * old_argc, char*** old_argv);
```
GitMensch: we only need to externalize one -> that should have a "cob_" prefix, the others can be static in cobgetopt.c

libcob/cobgetopt.c
```diff
@@ -476,6 +475,115 @@ _getopt_initialize (const char *optstring)
...
+        memcpy(path_str, ptr, path_size - 1);
+        path_str[path_size - 1] = '\0'; 
+		FILE* path = fopen(path_str, "r");
+		size_t capacity = 256;
+ 	 	size_t length = 0;
+ 		char *options = malloc(capacity);
+		free(path_str);
+		
+		
+		int ret;
```
GitMensch: as discussed, that won't build when using `configure CC="gcc -std=c89" CPPFLAGS="-Werror=declaration-after-statement"` (or an old compiler, which is the reason why we check that)

libcob/cobgetopt.c
```diff
@@ -476,6 +475,115 @@ _getopt_initialize (const char *optstring)
...
+	}
+
+	if (*compile_mode) {
+		const char* endptr = ++ptr;
+		while (*endptr != ' ' && *endptr != '\0') {
+			endptr++;
+		}
+		size_t path_size = endptr - ptr + 1;
+					/* Ptr->endptr carries the directory */
+		char* path_str = malloc(path_size);
```
GitMensch: instead of a short-lived malloc on the heap use one of the following two approaches:

1 clean, wasteful: local buffer with local `#defined`d max
2 hacker-way: `const char orig = *endptr; *endptr = 0; /*later* *endptr=orig;`

libcob/cobgetopt.c
```diff
@@ -476,6 +475,115 @@ _getopt_initialize (const char *optstring)
   return optstring;
 }
 
+char *
+process_at_file(const char* argument_values,int* compile_mode,int* atfile_size){
+	char * ptr = strstr(argument_values,"@");
+	if (ptr){
+		*compile_mode = 1;
```
GitMensch: this parameter var seems obsolete altogether

libcob/cobgetopt.c
```diff
@@ -476,6 +475,115 @@ _getopt_initialize (const char *optstring)
...
+		
+		int ret;
+		while ((ret = getc(path)) != EOF) {
+			if(length+1 >= capacity){
+				 capacity *= 4;
+         		char *newoptions = realloc(options, capacity);
+				options = newoptions;
+			}
+			
+			options[length++] = (char)ret;
```
GitMensch: Is there a reason to not tokenize here?

libcob/cobgetopt.c
```diff
@@ -476,6 +475,115 @@ _getopt_initialize (const char *optstring)
...
+		const char* endptr = ++ptr;
+		while (*endptr != ' ' && *endptr != '\0') {
+			endptr++;
+		}
+		size_t path_size = endptr - ptr + 1;
+					/* Ptr->endptr carries the directory */
+		char* path_str = malloc(path_size);
+        memcpy(path_str, ptr, path_size - 1);
+        path_str[path_size - 1] = '\0'; 
+		FILE* path = fopen(path_str, "r");
```
GitMensch: you want to check the return value here and early-exit if the file is not there (stays unchanged in argv)

GitMensch (changes requested):
As seen in CI that currently leads to a SEGFAULT:

~~~
1197. run_extensions.at:1761: testing CALL unusual PROGRAM-ID. ...
../../tests/run_extensions.at:1815: $COMPILE_MODULE A@B.cob
/tmp/_actions-runner-working-dir/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/1197/test-source: line 154: 75721 Segmentation fault      (core dumped) ( $at_check_trace; $COMPILE_MODULE A@B.cob ) >> "$at_stdout" 2>> "$at_stderr" 5>&-
--- /dev/null	2026-04-27 01:59:42.626967408 +0000
+++ /tmp/_actions-runner-working-dir/gnucobol/gnucobol/_build/tests/testsuite.dir/at-groups/1197/stderr	2026-05-18 21:19:44.588745273 +0000
@@ -0,0 +1,5 @@
+
+attempt to reference invalid memory address (signal SIGSEGV)
+
+cobc: aborting
+cobc: Please report this!
../../tests/run_extensions.at:1815: exit code was 139, expected 0
1197. run_extensions.at:1761:  FAILED (run_extensions.at:1815)
~~~

so there is still some stuff to do :-)

Most important: please always add Changelog entries as those document the changes you did - also helps for review - and getting the habit of always documenting at least on git push is a good thing; similar applies for the testsuite (would go into used_binaries.at)

libcob/cobgetopt.c
RamyGeorge: Are you talking about compile_mode because that does get used later 

If it is about ptr, I could just put the strstr as the condition in a If statement and later save it in the IF scope

libcob/cobgetopt.c
GitMensch: still open - you don't need the malloc here and then also don't need to free the string anywhere later

libcob/cobgetopt.c
```diff
@@ -493,18 +496,23 @@ process_at_file(const char* argument_values,int* compile_mode,int* atfile_size){
         memcpy(path_str, ptr, path_size - 1);
         path_str[path_size - 1] = '\0'; 
 		FILE* path = fopen(path_str, "r");
-		size_t capacity = 256;
- 	 	size_t length = 0;
  		char *options = malloc(capacity);
```
GitMensch: declare options here, but malloc after the path check

cobc/cobc.c
```diff
@@ -3120,8 +3120,9 @@ file_replace_extension (const char *file, const char *ext)
 
 /* process command line options */
 static int
-process_command_line (const int argc, char **argv)
+process_command_line (int argc, char **argv)
 {
+
```
GitMensch, line 3125: revert the changes to this function to create no diff in that function

cobc/cobc.c
```diff
@@ -9562,6 +9564,7 @@ main (int argc, char **argv)
 	cb_config_text_column = 72;
 
 	/* Process command line arguments */
+	cob_rebuild_argv_at_file(&argc,&argv);
```
GitMensch: that should get an entry in cobc/changelog

libcob/cobgetopt.c
GitMensch: This parameter still seems obsolete to me, why not dropping it?

Also: just checking for !ptr and return here can make the code easier to follow

libcob/cobgetopt.c
```diff
@@ -476,6 +475,122 @@ _getopt_initialize (const char *optstring)
...
 }
 
+static char *
+process_at_file(const char* argument_values,int* compile_mode,int* atfile_size){
+	char * ptr = strstr(argument_values,"@");
+	if (ptr){
+		*compile_mode = 1;
+	}
+
+	if (*compile_mode) {
```
GitMensch: you can keep the starting scope here without the condition (which still gives you c89 compat)

libcob/cobgetopt.c
GitMensch: toekenize = embedding code of expand_processed_at_file here (but directly wit the token), maybe even recursively call __FUNC__ if the arg seen starts with an @?

libcob/cobgetopt.c
```diff
@@ -556,6 +671,7 @@ cob_getopt_long_long (const int argc, char *const *argv, const char *optstring,
   /* Test whether ARGV[optind] points to a non-option argument.  */
 #define NONOPTION_P (argv[optind][0] != '-' || argv[optind][1] == '\0')
 
+
```
GitMensch: revert this non-change for cleaner diff

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,21 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+    PROCEDURE DIVISION.
+        DISPLAY "Hello, GnuCOBOL!".
+        STOP RUN.   
+])
+
+AT_CHECK([echo "-x" > tempfile.txt
+$COMPILE_MODULE @tempfile.txt prog.cob], [1], [],
+[cobc: error: only one of options 'm', 'x', 'b' may be specified
+])
+AT_CLEANUP()
```
GitMensch, line 1629: end with an empty line, but add more tests here: @something (= not an existing file) should be passed as is

if not yet implemented move to another test: multiple @file (including the error case A -> B -> C -> A)

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,21 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+
+AT_DATA([prog.cob], [
+    IDENTIFICATION DIVISION.
+    PROGRAM-ID. HELLO.
+    PROCEDURE DIVISION.
+        DISPLAY "Hello, GnuCOBOL!".
+        STOP RUN.   
+])
+
+AT_CHECK([echo "-x" > tempfile.txt
```
GitMensch: just use AT_DATA for that

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,21 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
 ])
 
 AT_CLEANUP
+
+
+AT_SETUP(@file command line options)
+AT_KEYWORDS([syntax])
```
GitMensch: that looks like the wrong keyword, "cobc" is likely useful here, as well as "getopt", that will then also allow you to `VGSUFFIX=mytests make check TESTSUITEFLAGS="-k getopt"` to verify your changes via valgrind (also see https://sourceforge.net/p/gnucobol/wiki/Hacking%20GnuCOBOL/)

GitMensch (changes requested):
nice progress

libcob/cobgetopt.c
```diff
@@ -500,11 +498,11 @@ process_at_file(const char* argument_values,int* compile_mode,int* atfile_size){
     	return NULL;
  	}
```
GitMensch: the old code here was better, you just need to add an additional scope {} around that; note that I'd suggest to do the define here and use all upper-case with a name that is unlikely conflict with any system header

libcob/cobgetopt.c
```diff
@@ -500,11 +498,11 @@ process_at_file(const char* argument_values,int* compile_mode,int* atfile_size){
...
 	options = malloc(capacity);
-		
+	char *tmp;
```
GitMensch: this one can go one line down

libcob/cobgetopt.h
```diff
@@ -22,6 +22,7 @@
    Modified for use in GnuCOBOL by Roger While
 */
 
+#include "libcob/common.h"
```
GitMensch, line 25: that change is unrelated and should be done - if at all - in a different commit (possibly added by clangd vscode extension - you can configure automated changes there)

libcob/cobgetopt.c
GitMensch, line 1: file should be reverted -> no change

libcob/argv.c
```diff
@@ -0,0 +1,621 @@
...
+
+Libiberty is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Library General Public License for more details.
+
+You should have received a copy of the GNU Library General Public
+License along with libiberty; see the file COPYING.LIB.  If
+not, write to the Free Software Foundation, Inc., 51 Franklin Street - Fifth Floor,
+Boston, MA 02110-1301, USA.  */
```
GitMensch, line 19: have a look at cobgetopt.c how to adjust the header

libcob/ChangeLog
```diff
@@ -1,4 +1,10 @@
 
+2026-05-29  Ramy George <ramygeorge19@gmail.com>
+
+	* argv.c: extracted the argv.c file from GCC
```
GitMensch, line 4: * argv.c: new file, extracted from libiberty [GCC version NNN], [youradjustments]
* cobgetopt.h (cob_expandargv): added the expandargv function in header
* Makefile.am (SOURCES): include argv.c

libcob/argv.c
```diff
@@ -0,0 +1,621 @@
...
+
+
+/*  Create and destroy argument vectors.  An argument vector is simply an
+    array of string pointers, terminated by a NULL pointer. */
+
+#include "cobc/cobc.h"
```
GitMensch, line 25: there should be no cobc in libcob (only the other way around)

libcob/cobargv.c
```diff
@@ -0,0 +1,621 @@
...
+#include <sys/stat.h>
+#endif
+
+#ifndef NULL
+#define NULL 0
+#endif
+
+#ifndef EOS
+#define EOS '\0'
+#endif
```
GitMensch, line 52: check if the headers above are needed (may be partially in common.h)

libcob/argv.c
```diff
@@ -0,0 +1,621 @@
...
+#include "common.h"
+#include <ctype.h>
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
```
GitMensch, line 30: order: config.h comes first, then system headers, then common.h, then cobgetopt.h

libcob/argv.c
```diff
@@ -0,0 +1,621 @@
...
+	    *p++ = *q;
+	}
+      *p++ = '\'';
+    }
+  *p = '\0';
+  freeargv (argv);
+  return buf;
+}
+
+#ifdef MAIN
```
GitMensch, line 572: that is a good design in general for easier testing (and something I do in writing wrapper-libraries) -> but we need the tests in in any case in used_binaries-at -> you possibly can get them in similar;
remove that here

libcob/cobgetopt.h
```diff
@@ -117,6 +118,7 @@ struct option
    arguments to the option '\0'.  This behavior is specific to the GNU
    `getopt'.  */
 
+COB_EXPIMP void cob_expandargv (int *argcp, char ***argvp);
```
GitMensch, line 121: we use unnamed parameters in our headers

libcob/argv.c
GitMensch, line 1: GCC says:

~~~
../../libcob/argv.c:361:1: warning: 'cob_expandargv' redeclared without dllimport attribute: previous dllimport ignored [-Wattributes]
  361 | cob_expandargv (int *argcp, char ***argvp)
      | ^~~~~~~~~~~~~~
../../libcob/argv.c:512:1: warning: 'expandargstr' defined but not used [-Wunused-function]
  512 | expandargstr (const char *progname, const char *val)
      | ^~~~~~~~~~~~
../../libcob/argv.c:486:1: warning: 'countargv' defined but not used [-Wunused-function]
  486 | countargv (char * const *argv)
      | ^~~~~~~~~
../../libcob/argv.c:300:1: warning: 'writeargv' defined but not used [-Wunused-function]
  300 | writeargv (char * const *argv, FILE *f)
      | ^~~~~~~~~
~~~

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,59 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. HELLO.
+       PROCEDURE DIVISION.
+           DISPLAY "Hello, GnuCOBOL!".
+           STOP RUN.   
+])
+
+AT_DATA([echo "-x @tempfile.txt" > tempfile.txt])
+AT_DATA([echo "-g --free" > otherfile.txt])
```
GitMensch, line 1645: see above how AT_DATA is intended to be used (can still be single line)

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,59 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+AT_CHECK([$COMPILE_MODULE @tempfile.txt @otherfile.txt prog.cob], [0], [], [])
+
+AT_CLEANUP()
+
+AT_SETUP(recursive @file with conflict)
+AT_KEYWORDS([cobc getopt])
+
+
```
GitMensch, line 1653: 2 before AT_SETUP one after AT_KEYWORDS

libcob/argv.c
GitMensch: following cobgetopt I'd also suggest to rename the file

cobc/cobc.c
```diff
@@ -9564,6 +9565,7 @@ main (int argc, char **argv)
 	cb_config_text_column = 72;
 
 	/* Process command line arguments */
+	cob_expandargv(&argc, &argv);
```
GitMensch, line 9568: is it useful to also add that to bin/cobcrun?

GitMensch (changes requested):

GitMensch:
@RamyGeorge please rebase, which should improve the CI part and update per review, so we can have a next look at that

cobc/cobc.c
```diff
@@ -9571,6 +9565,7 @@ main (int argc, char **argv)
...
```
GitMensch:
```suggestion
	cob_expandargv (&argc, &argv);
```

GitMensch (commented):
the open comments still need to be tackled
also, argv should be renamed to cobargv

libcob/cobargv.c
```diff
@@ -0,0 +1,648 @@
...
+
+#define INITIAL_MAXARGC 8	/* Number of args + NULL in initial argv */
+#include "common.h"
+#include <cobgetopt.h>
```
GitMensch, line 56: coibgetopt.h would be used from the non-installed version, so `""` instead of ` <>`

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,59 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. HELLO.
+       PROCEDURE DIVISION.
+           DISPLAY "Hello, GnuCOBOL!".
+           STOP RUN.   
+])
+
+AT_DATA([echo "-x" > tempfile.txt])
```
GitMensch, line 1625: put tempfile into AT_DATA instead, same in the following tests

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,59 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. HELLO.
+       PROCEDURE DIVISION.
+           DISPLAY "Hello, GnuCOBOL!".
+           STOP RUN.   
+])
+
+AT_DATA([tempfile.txt], [-x @tempfile.txt
+])
+AT_CHECK([$COMPILE_MODULE @tempfile.txt @otherfile.txt prog.cob], [1], [], [stderr])
```
GitMensch, line 1664: use actual error message in stderr here (reasonably having a line break before its [

AUTHORS
```diff
@@ -5,6 +5,8 @@ and po/ja.po.
 * lib/getopt1.c, lib/getopt.h, lib/getopt1.c were extracted
 from the GNU C Library 2.3.2.  Distributed under GNU LGPL.
 
+* libcob/argv.c was extracted from GCC.
```
GitMensch, line 8:
```suggestion
* libcob/cobargv.c was extracted from libiberty/GCC (argv) VERSIONHERE.
  Distributed under GNU LGPL.
```

GitMensch (changes requested):
the main thing currently is the presumingly not current base - `git pull` (to get the current changes / merge conflict solution),m followed by `git pull upstream --rebase` possibly fixes that

also have a note at the current non-resolved comments of the PR, please

tests/testsuite.src/used_binaries.at
RamyGeorge: might be a naive issue on my end but the error is a simply a path to file directory + : error recursive @file
i dont want to put my local directory since that will most definitly be used dependent

libcob/cobargv.c
```diff
@@ -0,0 +1,648 @@
...
+You should have received a copy of the GNU Library General Public
+License along with libiberty; see the file COPYING.LIB.  If
+not, write to the Free Software Foundation, Inc., 51 Franklin Street - Fifth Floor,
+Boston, MA 02110-1301, USA.  */
+
+
+/*  Create and destroy argument vectors.  An argument vector is simply an
+    array of string pointers, terminated by a NULL pointer. */
+
+#ifdef HAVE_CONFIG_H
```
GitMensch, line 25: as this is no external file and config.h is used unconditionally in other places -> get it in without the guard

libcob/cobargv.c
```diff
@@ -0,0 +1,648 @@
...
+#ifndef NULL
+#define NULL 0
+#endif
+
+#ifndef EOS
+#define EOS '\0'
+#endif
+
+#define INITIAL_MAXARGC 8	/* Number of args + NULL in initial argv */
+#include "common.h"
```
GitMensch, line 55: before that you want

~~~
   /* Force symbol exports */
#define	COB_LIB_EXPIMP
~~~

PR #291, opened 2026-05-19
Distinguish more grammar contexts, and provide fixes

nberth:
Change the type of the variable used to hold the set of contexts used to determine token sensitivity. This opens the way to cleanup some of the workarounds used in parser rules to implement the later.


GitMensch:
why not - but adjust the current context that have a TOPO overlap / re-use as well, which also provides a first test on working correctly

nberth:
> why not - but adjust the current context that have a TOPO overlap / re-use as well, which also provides a first test on working correctly

At a minimum, the idea is to provide means for @kmr-srbh to specify the proper context-sensitivity for some new keywords that appear in OO-related syntax (the REPOSITORY paragraph at least). I am planning to adjust/fix existing contexts as well, if possible.

GitMensch:
As noted: when you fix some of the current ones to use the higher bits, we can upstream that.
re failing CI: MSYS2 dropped mingw32 now also for json-c https://github.com/msys2/MINGW-packages/commit/4d96359d214d2e827a63f96659f3cddd0b0a50b1 - if it is going away completely very soon we'll just drop that env from our CI, if not we may move json-c to an extra pacman (maybe with bdb as well?) an use cjson for mingw32.

GitMensch:
... and json-c will be back in mingw32 (we should just not expect that environment to work any more in 2 years - and may build some dependencies with upstream PKGBUILD in the last supported version "locally", installing that or mingw32).

cobc/cobc.h
```diff
@@ -165,18 +165,19 @@ enum cb_current_date {
 #define	CB_CS_OPEN			(1U << 30)	/* within OPEN */
 #define	CB_CS_JSON_GENERATE		(1U << 31)
 /* HACK: no more space - using minor one until re-written */
```
GitMensch: that line should be moved

cobc/error.c
```diff
@@ -535,8 +534,6 @@ cb_error_internal (const char *fmt, va_list ap)
 	const enum cb_warn_val	pref = get_warn_opt_value (opt);
 	enum cb_warn_val	ret = pref;
 
-	cobc_in_repository = 0;
```
GitMensch, line 538: not sure about this change - would be missing in changelog

GitMensch (commented):
Failing on 32bit likely means a "size_t" somewhere, no?
Maybe run with `-fsanitize=undefined` (could also be added to the 32bit CI definition)?

cobc/parser.y
```diff
@@ -139,13 +139,20 @@ unsigned int			cobc_repeat_last_token = 0;
...
 unsigned int			cobc_areacheck = 0;
 unsigned int			cobc_in_area_a = 0;
 
+#define __CS_RESET(verbose)						\
+	do {								\
+		if (0 && verbose)					\
+			fprintf (stderr, "%d: manual reset!\n", __LINE__); \
+		cobc_cs_check = 0;					\
+	} while (0);
+
```
GitMensch: if you want to keep this, then please move that to COB_DEBUG_LOG, otherwise revert

cobc/parser.y
```diff
@@ -4463,7 +4472,13 @@ repository_name_list:
 
 special_names:
   special_names_header
+  {
+	cobc_cs_check |= CB_CS_SPECIAL_NAMES;
```
GitMensch: Can't we handle that via reserved.c?
Any idea why the C/S is not setup there? Would that harm?

cobc/parser.y
```diff
@@ -8925,9 +8960,14 @@ _on_choice:
 /* DEFAULT clause (content-validation) */
 
 default_clause:
-  DEFAULT _is x_or_none
+  DEFAULT
```
GitMensch, line 9060: Can we handle that via reserved.c?

cobc/cobc.h
```diff
@@ -132,51 +132,51 @@ enum cb_current_date {
...
+#define	CB_CS_XML_GENERATE		(COB_U64_C(1) << 28)
+#define	CB_CS_XML_PARSE			(COB_U64_C(1) << 29)
+#define	CB_CS_OPEN			(COB_U64_C(1) << 30)	/* within OPEN */
+#define	CB_CS_JSON_GENERATE		(COB_U64_C(1) << 31)
+#define	CB_CS_I_O_CONTROL		(COB_U64_C(1) << 32)
+#define	CB_CS_TYPEDEF			(COB_U64_C(1) << 33)
+#define	CB_CS_EXHIBIT			(COB_U64_C(1) << 34)
+#define	CB_CS_INSPECT			(COB_U64_C(1) << 35)
+#define	CB_CS_CONVERT			(COB_U64_C(1) << 36) /* TODO: within CONVERT intrinsic */
+#define	CB_CS_MODULE_NAME		(COB_U64_C(1) << 37) /* TODO: within MODULE-NAME intrinsic */
```
GitMensch, line 172: So those are currently open?

GitMensch (commented):
looks goods overall - if there are still TODOs left it likely is a good idea to adjust the testsuite to cover those

side note: I'm fine with [your "in repository" change](https://github.com/OCamlPro/gnucobol/compare/256aad61196987b00f4638da1a0d47627dec5b33..038e81a62ed00fbabfd1c90a84b36fd3bc6a8563), if you'd prefer that over the current version; if you add it back then please make that as a separate commit (here) with an additional separate entry in the Changelog

cobc/parser.y
nberth: Here, and with `DEFAULT` as well (maybe others), I changed my approach on my way, and I am still not sure what is the best approach.

At the moment I think that relying on the `token->context_set` mechanism in `reserved.c` should be reserved to cases where a context-sensitive token may directly follow a token that starts that context (due to parser lookahead). Any every/most other cases should be made explicit with assignments to `cobc_cs_check` in the parser rules. The rational being that manipulations of context-sets should be made visible for documentation and maintenance purposes. Doing it this way may help identify cases where contexts may not be "exited" due to syntax errors within a context; the previous approach relied heavily on regular resets of the set of active contexts (like after each statement), which may still be re-applied here if we are still unsure about the current changes.

cobc/cobc.h
nberth: Yes I did not find any mechanism that defines special contexts based on intrinsics…

cobc/error.c
nberth: I couldn't figure out why this variable was reset in this function.

cobc/parser.y
nberth: @GitMensch The change of approach mentioned in my previous comment has some deep implications on how we write and define contexts. Do you have an opinion on this? If you prefer to rely more on the "automated"/token-based approach, I can of course revert some of my changes.

GitMensch: I'm fine with that change and that _can_ also be combined with increasing the size of that var if you are relative sure about it - feel free to reach out to your fellow colleagues for a review of the approach, of course ;-)

The Changelog should definitely be explicit about this "shift" and we should document that approach/rule, possibly at least in reserved.c and maybe in [the internal wiki](https://sourceforge.net/p/gnucobol/wiki/Style%20guide/), you _should_ have write access there.

The most important part for mew on that is that we can set the C/S via dialect (that's currently the *, if I remember correctly) and _may_ also define there later to what constructs this is C/S.

Setting the context "more explicit" won't create an issue with that.

So, maybe that's one of the rare "Simon has no opinion on that as long as it works and is documented".

nberth: The general policy I delineated above applied to some cases, but many contexts required adaptation.  In the end a case-by-case reasoning is always required.  So I resorted to the introduction of some helper macros in `parser.y`, that can be used to check consistency of special context manipulations when `--enable-cobc-internal-checks` is given at configure time.

If "obvious" bugs remain in context manipulation, I think they are likely related to cases where a special context is not left after a syntax error occurs; that should only lead to wrong subsequent syntax errors, where a word is taken for reserved in a wrong context.

Such a bug is likely going to be reported if the aforementioned flag is given.  I am  going to try and check with some codebases I have under the hood; however I imagine you have much more opportunities to check the current implementation.

tests/testsuite.src/syn_misc.at
```diff
@@ -4403,6 +4403,24 @@ prog.cob:10: error: CALL-CONVENTION and WITH LINKAGE are mutually exclusive
...
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION  DIVISION.
+       PROGRAM-ID.     prog.
+       DATA          DIVISION.
+       LOCAL-STORAGE SECTION.
+       77 C PIC X.
+       PROCEDURE     DIVISION.
+           CALL EXTERN 'callee' using C.
+           CALL 'callee' using C.
```
nberth, line 4417: @GitMensch BTW that's a test that does not pass without these changes.

cobc/reserved.c
```diff
@@ -254,7 +254,7 @@ static struct cobc_reserved default_reserved_words[] = {
 				0, 0
   },
   { "ACCEPT",			1, 0, ACCEPT,			/* 2002 */
-				CB_CS_ACCEPT, 0
+				0, 0
```
GitMensch: If I see it correctly that context setting is not _needed with the change of this PR any more.
Can we still leave it in or does this need to be in?

Background: I wanted to get rid of the second and third flags here (possibly moving to a .def if we need to keep the struct intact) and using the CS defines in the next line instead (or, if that is dynamically adjusted, pre-set the struct value depending on the define).
The idea was to have that as a follow-up commit after this current PRs content is applied...

cobc/parser.y
```diff
@@ -300,6 +299,99 @@ enum cobc_hd {
...
+		cobc_cs_check &= ~(cs_mask);	\
+	} ONCE_COB /* LCOV_EXCL_LINE */
+
+/* Leaves every special context. */
+#define __CS_CLEAR_ALL()			\
+	do {					\
+		cobc_cs_check = COB_U64_C(0);	\
+	} ONCE_COB /* LCOV_EXCL_LINE */
+
+#ifdef COB_TREE_DEBUG	 /* (--enable-cobc-internal-checks at configure time) */
```
GitMensch, line 325: Olease double-check if we have on of the CI runners configuring with --enable-cobc-internal-checks - and if this isn't the case add it to one in the gitside branch(es) directly (maybe the one with the coverage?)

tests/testsuite.src/syn_misc.at
```diff
@@ -7516,6 +7534,29 @@ AT_DATA([prog.cob], [
...
+    01  INPUT-RECORD.
+        03  IN-TYPE-DEFAULT-ERROR PIC 99
+   *>intentional syntax error to validate the special context for DEFAULT
+   *>is left in such cases.
+	                        DEFAULT PROCEDURE.
+        03  IN-TYPE-WO-DEFAULT PIC 99
+   *>checks DEFAULT NONE is accepted (NONE is sensitive in the DEFAULT
+   *>clause only).
+	                        DEFAULT NONE.
+    PROCEDURE DIVISION.
```
GitMensch, line 7574: please reorder - first the one that should parse, then the one that doesn't - especially as our error recovery is not as good as it is in other places --> we want to ensure that "passing" comes first

cobc/parser.y
```diff
@@ -13256,36 +13413,37 @@ display_statement:
...
   }
-| id_or_lit UPON_COMMAND_LINE _common_exception_phrases
+| id_or_lit UPON_COMMAND_LINE __leave_display_cs _common_exception_phrases
   {
 	cb_emit_command_line ($1);
   }
-| screen_or_device_display _common_exception_phrases
-| _with CONVERSION screen_or_device_display _common_exception_phrases
+| screen_or_device_display __leave_display_cs _common_exception_phrases
+| _with CONVERSION screen_or_device_display __leave_display_cs _common_exception_phrases
```
GitMensch: please change that to `_display_common_exception_phrases` which first leaves the context, then has the normal `_common_exception_phrases`

GitMensch: ... or even better, split directly by having only two rules here:
`display_body_with_exception_phrases` and another `display_body_without_exception_phrases` (you may find a better wording) which will then have the leave + _common_exception_phrases only once and in the second the leave only once - in both cases directly as a macro, not as a rule

cobc/parser.y
```diff
@@ -13256,36 +13413,37 @@ display_statement:
...
-| display_floating_window
-| display_initial_window
+| display_erase __leave_display_cs	/* note: may also be part of display_pos_specifier */
+| display_pos_specifier __leave_display_cs
+| display_message_box __leave_display_cs
+| display_window __leave_display_cs
+| display_floating_window __leave_display_cs
+| display_initial_window __leave_display_cs
 ;
+__leave_display_cs: { __CS_LEAVE (CB_CS_DISPLAY); };
```
GitMensch: For some reason I really dislike the rule that just executes a macro - can you please use it "directly"? See comment above.

GitMensch (changes requested):

cobc/parser.y
nberth: Yes that option is passed in all/most workflows already.

tests/testsuite.src/syn_misc.at
nberth: This test intentionally checks that the special context that is induced by `DEFAULT` in data definitions is exited before it is re-entered in the second definition.  This is why the syntax error appears first.

I'll insert an additional definition with `DEFAULT NONE` before the intentional error, if that's what you meant.

cobc/parser.y
nberth: Ah yes I was myself not satisfied with these changes; I forgot to come back to that after I finished with the other contexts.

cobc/reserved.c
nberth: I understand.  I will check if these changes are indeed required.

tests/testsuite.src/syn_misc.at
GitMensch: I see, yes please add that, then we're good with this test

cobc/tree.h
```diff
@@ -2280,15 +2280,16 @@ extern int		suppress_data_exceptions;
...
-extern unsigned int	cobc_cs_check;
+extern cob_u64_t	cobc_cs_check;
 extern unsigned int	cobc_allow_program_name;
 extern unsigned int	cobc_in_xml_generate_body;
 extern unsigned int	cobc_in_json_generate_body;
 extern unsigned int	cobc_areacheck;
 extern unsigned int	cobc_in_area_a;	      /* set by scanner only */
 extern unsigned int	cobc_still_in_area_a; /* raised by parser only */
 
+#define cobc_in_repository ((cobc_cs_check & CB_CS_REPOSITORY) != COB_U64_C(0))
```
GitMensch: This is now only used in scanner.l and should be moved there.
Side question: Do we get warnings / errors with the simple `!(cobc_cs_check & CB_CS_REPOSITORY)`?

GitMensch: Also: aren't `cobc_cs_check ==` a bad thing in general?
Do we miss some test cases around those?

cobc/parser.y
nberth: Using a token-based entry for `CB_CS_DEFAULT` has a great potential to trigger spurious bugs in special-contexts elsewhere, as `DEFAULT` appears in other places than just at the beginning of the DEFAULT clause.   Resorting to token-based entry may require resetting contexts in unrelated places in the parser…

cobc/parser.y
```diff
@@ -4030,10 +4118,14 @@ init_or_recurse:
...
-| OPTIONS TOK_DOT
+| OPTIONS TOK_DOT		/* Note: OPTIONS auto-enters CB_CS_OPTIONS */
   _options_clauses
   {
-	cobc_cs_check = 0;
+	  __CS_LEAVE (CB_CS_OPTIONS);
+  }
+| OPTIONS error
+  {
+	  __CS_LEAVE (CB_CS_OPTIONS);
```
GitMensch: according to coverage this error path misses a test

cobc/parser.y
```diff
@@ -6345,18 +6452,18 @@ obsolete_dos_vs_apply_phrase:
 /* MULTIPLE FILE TAPE clause */
 
 multiple_file_tape_clause:
-  MULTIPLE
+  MULTIPLE			/* Note: tested in NIST85 only. */
```
GitMensch: Why not testing that with a small syntax test in the internal testsuite?

cobc/cobc.h
GitMensch: You're right, these two need to directly ENTER in the parser after CONVERT or MODULE-NAME is read and then need to LEAVE explicit at the end of the function's TOK_PARENS.
If you can't add that now then please leave an entry in TODO (but the change itself may be even quicker :-)

GitMensch (changes requested):
Two tests and a minor cleanup missing - you may directly commit the result upstream afterwards.

cobc/tree.h
nberth: For sure some bit-focused checks could be replaced by wider checks of a simpler form like `cobc_cs_check ==`, for instance if we can be sure some clauses or statements or clauses may never be active in conjunction with other contexts.  However, that would remove some bits of local reasoning that the mask-based approach provides, and may make a generalization of the special context mechanism harder to implement and get right.

On a related matter, "properly" supporting contexts that are induced by intrinsic functions may actually prove slightly harder and require implementing a stack of active contexts (which is actually the approach I have chosen for the LSP server in SuperBOL Studio—where we don't support those specific contexts yet, but where context-sensitivity is expressed in a way that tightly matches syntax rules of the parser).  See the new TODO note for a beginning of discussion on this aspect.

GitMensch: checking via GitHub (which may be wrong results) shows that we only do this in the following case - the others seemed to be removed in this PR

https://github.com/OCamlPro/gnucobol/blob/a3accbe7616ca8ffd61662af1072812ba65cd523/cobc/scanner.l#L388


(again: that PR looks fine for merge)

nberth: > checking via GitHub (which may be wrong results) shows that we only do this in the following case - the others seemed to be removed in this PR
> 
> https://github.com/OCamlPro/gnucobol/blob/a3accbe7616ca8ffd61662af1072812ba65cd523/cobc/scanner.l#L388
> 
> (again: that PR looks fine for merge)

Ah thanks for the notice (I forgot to double-check for the remaining references to this variable).  I'll check/fix and then commit upstream.

nberth:
Now upstream.

GitMensch:
In [r5686](https://sourceforge.net/p/gnucobol/code/5686/).

PR #292, opened 2026-05-24
Config option tab-width can receive a list of comma-separated widths

lefessan:
Before this patch, option "tab-width" could only receive a single integer value, indicating the width of all tabulations.
After this patch, option "tab-width" can receive a list of comma-separated widths, each one indicating the width of the corresponding tabulation, the last one being kept for all other tabulations.

For example, `-ftab-width=6,1,4` will give the following behavior:
* a tabulation in colomns < 7 will move the cursor to column 7 (indicator)
* a tabulation in column 7 will move the cursor to column 8
* a tabulation in column < 12 will move the cursor to column 12 (area B)
* any other tabulation will move the cursor to 16,20,24,28, etc.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -107,6 +107,10 @@ NEWS - user visible changes				-*- outline -*-
 ** New option --gentable to build EBCDIC/ASCII translation tables,
    for use with -febcdic-table
 
+** Option -ftab-width can now receive a list of comma-separated widths,
+   such as "6,1,4", indicating the width of each tabulation, the last one
+   being reused for all the next ones
+
 * Listing changes
 
 ** the listing header shows the processed file's basename to prevent
--- a/cobc/ChangeLog
+++ b/cobc/ChangeLog
@@ -1,4 +1,16 @@
 
+2026-05-24  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* config.def: "tab-width" option is changed to a comma-separated
+	  list of tab widths, for example "6,1,4", the last one being reused.
+	  This new meaning is backward compatible.
+	  Implements FR #498 " Adjust tab-width to optionally be a list of
+	  tab-stop positions"
+	* config.c: initialize cb_tab_width as a string,
+	  each position indicating the number of spaces to insert for a tab
+	  at that position
+	* pplex.c,cobc.c: use the new type of cb_tab_width
+
 2025-12-29  Roger Bowler <rbowler@snipix.net>
 
 	* tree.c (finalize_file): if file is EXTFH enabled then don't warn for
--- a/cobc/cobc.c
+++ b/cobc/cobc.c
@@ -6784,9 +6784,11 @@ get_next_listing_line (FILE *fd, char **pline, int fixed)
 
 	for (in_char = in_line; i != CB_LINE_LENGTH && *in_char; in_char++) {
 		if (*in_char == '\t') {
-			out_line[i++] = ' ';
-			while (i % cb_tab_width != 0) {
+			/* See comment on tab-width in config.c */
+			int tab_width = cb_tab_width[i];
+			while (tab_width > 0){
 				out_line[i++] = ' ';
+				tab_width--;
 				if (i == CB_LINE_LENGTH) {
 					break;
 				}
--- a/cobc/config.c
+++ b/cobc/config.c
@@ -732,6 +732,57 @@ cb_config_entry (char *buff, const char *fname, const int line)
 			}
 			break;
 
+		} else if (strcmp (name, "tab-width") == 0) {
+			/* The original implementation tab-width used
+			 * to rely on a unique tab width and
+			 * computation by modulos. Instead, this new
+			 * implementation relies on cb_tab_width being
+			 * an array of offsets, each position in the
+			 * array corresponding to the offset from that
+			 * position to the next tabstop. The array is
+			 * initialized here from the tab-width
+			 * option. */
+
+			/* This option is always initialized, so let's
+			   allocate a static buffer. COB_SMALL_BUFF is
+			   much larger than the maximal line length,
+			   but there is currently no other constant
+			   available for that. */
+			static char cb_tab_width_static[COB_SMALL_BUFF];
+
+			char *token = strtok(val, ",");
+			int pos = 0;
+			unsigned int last_tab_width = 8;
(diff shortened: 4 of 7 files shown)


NEWS
```diff
@@ -107,6 +107,10 @@ NEWS - user visible changes				-*- outline -*-
 ** New option --gentable to build EBCDIC/ASCII translation tables,
    for use with -febcdic-table
 
+** Option --ftab-width can now receive a list of comma-separated widths,
```
GitMensch: single dash for -f options

tests/testsuite.src/run_misc.at
```diff
@@ -97,6 +97,43 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
 AT_CLEANUP
 
 
+AT_SETUP([tabstop tab-width])
+AT_KEYWORDS([tabulation])
```
GitMensch: reasonable good test;
please add config or dialect or so (not sure what we use for that) as keyword and use two empty lines after AT_CLEANUP

cobc/pplex.l
```diff
@@ -1426,11 +1426,7 @@ ppopen_get_file (const char *name)
 					line_pos = 0;
 					break;
 				case '\t':
-					buffer[pos] = ' ';
-					line_pos++;				/* move test from run_extensions.at 1809 to -fformat */
-					while (line_pos % cb_tab_width != 0) {
-						line_pos++;
-					}
+					line_pos+= cb_tab_width[pos];							buffer[pos] = ' ';
```
GitMensch: that looks strange

tests/testsuite.src/run_misc.at
GitMensch: as we actually only check the compile warnings: move to configuration.at

cobc/config.c
```diff
@@ -732,6 +732,34 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+			char *token = strtok(val, ",");
+			int pos = 0;
+			unsigned int last_tab_width = 8;
+			if (cb_tab_width == NULL)
+				cb_tab_width = cob_malloc (COB_SMALL_BUFF);
+			while (token != NULL) {
+				unsigned int tab_width = atoi(token);
+
+				if (tab_width<1 || tab_width>12){
+					invalid_value (fname, line, name, val, "1-12", 0, 0);
```
GitMensch: please add some test cases like -ftab-width=6,13,a (I'd suggest to set a return value instead of direct return) for this to configuration.at

cobc/config.c
```diff
@@ -732,6 +732,34 @@ cb_config_entry (char *buff, const char *fname, const int line)
 			}
 			break;
 
+		} else if (strcmp (name, "tab-width") == 0) {
+
+			char *token = strtok(val, ",");
+			int pos = 0;
+			unsigned int last_tab_width = 8;
+			if (cb_tab_width == NULL)
+				cb_tab_width = cob_malloc (COB_SMALL_BUFF);
```
GitMensch: We will have a value of that in any case, so no need for malloc, just directly static assign something of reasonable size - not sure if we have a constant for max line length already... 
That also removes the need to explicit free that at the end, can be memset 0 her (so passing it more than once - first via std + extra or passed twice) won't keep old values.

While I first thought "that seems broken" (the "n" position), then "ah, that's a full list - seems wasteful"... after all it is not a big additional field, so the general approach is fine.

cobc/ChangeLog
```diff
@@ -1,4 +1,14 @@
 
+2026-05-24  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* config.def: "tab-width" option is changed to a comma-separated
```
GitMensch, line 4: this actually implements https://sourceforge.net/p/gnucobol/feature-requests/498/ which should be referenced

cobc/cobc.c
```diff
@@ -6784,9 +6784,10 @@ get_next_listing_line (FILE *fd, char **pline, int fixed)
 
 	for (in_char = in_line; i != CB_LINE_LENGTH && *in_char; in_char++) {
 		if (*in_char == '\t') {
-			out_line[i++] = ' ';
-			while (i % cb_tab_width != 0) {
+			int tab_width = cb_tab_width[i];
```
GitMensch, line 6788: the way it is setup (full max line with consecutive sizes as char[]) should be included in a comment, possibly here and in pplex.l, it is quite different from the long standing implementation

cobc/pplex.l
```diff
@@ -2383,9 +2379,10 @@ start:
 		}
 		if (unlikely (ipchar == '\t')) {
 			if (likely (line_overflow == 0)) {
-				buff[n++] = ' ';
-				while (n % cb_tab_width != 0) {
+				int tab_width = cb_tab_width[n];
+				while (tab_width>0){
```
GitMensch: here and in other places of this : please surround conditional operators with spaces, so `tab_width > 0` here

tests/testsuite.src/run_misc.at
```diff
@@ -97,6 +97,43 @@ AT_CHECK([$COBCRUN_DIRECT ./prog], [0],
...
+09		 DISPLAY "hello".
+10		  DISPLAY "hello".
+11		   DISPLAY "hello".
+12		    DISPLAY "hello".
+13    * After 3 tabs, the line is at the beginning of Area B
+14			DISPLAY "hello".
+15			STOP RUN.
+])
+
+AT_CHECK([$COMPILE -std=ibm -ftab-width=6,1,4 prog.cob], [0], [],
```
GitMensch: I think you can drop the std here - but should have a _second_ ccompile_only test where you have `-ftab-width=6,1,4 -ftab-width=6` - wich then should only have one warning (it currently most likely hasn't)

GitMensch (changes requested):
nearly good for upstream, maybe also check if gnucobol.texi needs an update

lefessan:
I think I have addressed all of your initial comments. Is the new version ok ?

lefessan:
I didn't see any documentation of tab-width in gnucobol.texi, the documentation of configuration options is not automatically  generated from config.def ?

GitMensch:
New version is good for upstream (I'd just switch the tests to COMPILE_ONLY) - happy commit and thank you :-)

PR #293, opened 2026-06-11
feat: add parsing support for `INTERFACE-ID`, `METHOD-ID`, class body and `INVOKE` statement

kmr-srbh:
* Add support for parsing `INTERFACE-ID`, `METHOD-ID`, `FACTORY`,
  `OBJECT` paragraphs and `INVOKE` statement. Support `CLASS` 
  and `INTERFACE` specifiers in `REPOSITORY` paragraph. Also  support
  keywords `SELF` and `ACTIVE-CLASS`.
* Add syntax validation logic to check OO inheritance list for duplicates, 
  consistency with names specified in the `REPOSITORY` paragraph.


cobc/parser.y
```diff
@@ -4615,6 +4836,8 @@ repository_name:
   {
 	yyerrok;
   }
+| CLASS WORD _as_literal _expands_clause
```
nberth:
```suggestion
| CLASS 
  {
    __CS_ENSURE (CB_CS_CLASS_SPECIFIER);
  }
  WORD _as_literal _expands_clause
  {
    __CS_LEAVE (CB_CS_CLASS_SPECIFIER);
  }
```
That's how to deal with special contexts at the moment.

cobc/parser.y
```diff
@@ -3984,50 +4095,144 @@ class_id_header:
...
+
+/* Parameterized classes not supported for now. */
+class_param_list:
+  WORD
+| class_param_list WORD
+;
+
+_inherits_phrase:
+  /* empty */
+| INHERITS _from parent_class_name_list
```
nberth:
```suggestion
| INHERITS _from class_name_list
  {
        current_program->class_inheritance_list = $3;
  }
```

cobc/parser.y
```diff
@@ -3984,50 +4095,144 @@ class_id_header:
...
   {
 	cb_trim_program_id ($1);
   }
 ;
 
+parent_class_name: 
+  WORD
+  {
+  	current_program->class_inheritance_list = 
+		cb_list_add(current_program->class_inheritance_list, $1);
```
nberth: Cf, eg, `alphabet_literal_list` for how to construct a list  rules.

cobc/parser.y
```diff
@@ -4019,17 +4211,57 @@ class_id_name:
...
+  {
+	cb_tree l;
+
+	if (strcasecmp(current_program->program_name, CB_NAME($2)) == 0) {
+		cb_error(_("cannot inherit from the same %s being declared"),
+			get_cob_module_type_string(current_program->prog_type));
+	}
+
+	/* Check for duplicate parent name */
+	for (l = $1; l; l = CB_CHAIN (l)) {
```
GitMensch: this checks the same multiple times, better place both checks afterwards in typeck,c on something like "finalize_interface", going over the list as here and using cb_error_x to point to the right position

cobc/parser.y
```diff
@@ -4103,12 +4343,46 @@ class_id_paragraph:
...
   {
-	cb_trim_program_id ($1);
+	if (setup_program ($3, $4, COB_MODULE_TYPE_INTERFACE, 1)) {
+		YYABORT;
+	}
+
+	__CS_CLEAR_ALL();
+	cobc_in_id = 0;
+
+	CB_UNSUPPORTED ("interfaces in object-oriented COBOL");
```
GitMensch: I'd suggest to move that to the interface_id_header rule

cobc/parser.y
```diff
@@ -4103,12 +4343,46 @@ class_id_paragraph:
...
+method_id_header:
+  METHOD_ID
+  {
+	cobc_in_id = 1;
+  }
+;
+
+get_or_set:
+  GET
+| SET
```
GitMensch, line 4314: you want to store what type of method you define, for example by returning cb_int0/cb_int1 or using a type return (then possibly an enum) here

cobc/parser.y
```diff
@@ -4675,6 +4949,17 @@ repository_list:
 | repository_list repository_name
 ;
 
+_expands_clause:
+  /* empty */
+| EXPANDS WORD
+{
+	/* Check that WORD is a class when used with the CLASS specifier
```
GitMensch: I guess that's a TODO?

tests/testsuite.src/syn_oo.at
```diff
@@ -250,3 +250,50 @@ prog.cob:20: error: object-oriented COBOL is not supported
...
+        REPOSITORY.
+            CLASS SavingsAccount AS "SavingsAccount".
+        END CLASS SavingsAccount.
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+[prog.cob:4: error: object-oriented COBOL is not supported
+prog.cob:8: error: object-oriented COBOL is not supported
+])
+AT_CLEANUP
```
GitMensch, line 361: you definitely want to add more tests . you do have some missing in your blog post

cobc/parser.y
```diff
@@ -2572,6 +2578,32 @@ set_oo_class_attr(enum cb_oo_class_attribute attr, const char* attr_name)
...
+static const COB_INLINE char *
+get_cob_module_type_string(enum cob_module_type prog_type)
+{ 															
+	switch (prog_type)									
+	{														
+		case COB_MODULE_TYPE_PROGRAM: 		return "program";
+		case COB_MODULE_TYPE_FUNCTION: 		return "function";
+		case COB_MODULE_TYPE_CLASS: 		return "class";			
+		case COB_MODULE_TYPE_INTERFACE: 	return "interface";
+		case COB_MODULE_TYPE_METHOD: 		return "method";
```
nberth: Note these may deserve translations.  Surrounding the string literals with `_(` and `)` should do the job.

cobc/parser.y
```diff
@@ -4019,17 +4185,57 @@ class_id_name:
...
+			get_cob_module_type_string(current_program->prog_type));
+	}
+
+	/* Check for duplicate parent name */
+	for (l = $1; l; l = CB_CHAIN (l)) {
+		if (strcasecmp (CB_NAME (CB_VALUE (l)), CB_NAME ($2)) == 0) {
+			cb_error (_("duplicate parent %s name '%s'"), 
+				get_cob_module_type_string(current_program->prog_type),
+				CB_NAME ($2));
+			break;
```
GitMensch: please add a "defined here" note - see how we normally handle user-defined names and check for duplicates (you may or may not be able to even re-use the functions)

cobc/parser.y
GitMensch: The main question is: do we need `to output the type at all or isn't it enough to have a note about the previous definition?
Note: there is a "general" function that outputs the "type" of a tree, we possibly want to adjust that in any case checking the tree's type there; I guess most of the strings there are not msgid's but string constants as those are taken from the COBOL standard, which is possibly what we want here as well.

GitMensch:
Hi guys, just checking for the timeline.

The original proposal had the first coding period ending today with "Handle complete syntax for each construct as defined by the ISO standard." Not sure how far you are with "complete syntax" or may even need to reduce the amount of syntax parsed.

The current GSoC timeline has the first coding period to July 6th - similar for the end of the second period being a week longer with the final end date August 24th.

If both of you want to do longer, we can still extend the period as well, but only if we request this week, if I remember correctly.

kmr-srbh:
We have completed parser support for all the required paragraphs (with also some AST support and syntax checks) with this PR. For the parser, only `INVOKE` statement is left (`SET` was also proposed, but it will be handled later in the semantics phase). I think I am behind on this one statement, but I think I can complete it this week (29th June - 5th July) along with the general syntax check work. 

Regarding the second phase of the coding period, I actually got it from here - https://developers.google.com/open-source/gsoc/timeline#july_6_-_august_16 where it says the work-period ends on August 16th. I might have made a mistake in understanding it.

Edit: I think I would want to push the `INVOKE` statement work today itself as part of the commit which I was to push addressing the reviews.

cobc/parser.y
nberth: Yes this is a debatable point.  The use of these nouns in the current code is in error messages that were fully translated before, where the type of compilation unit appeared clearly.  The growing amount of elements for this type may however incur a lot of code duplication where such messages are emitted…

kmr-srbh:
Update: I have addressed all the review suggestions locally and also implemented the `INVOKE` statement syntax yesterday. The checks have been moved inside `cb_validate_program_data`. I tried adding a full test from the ISO standard, and found some issues with the added grammar rules, like error with defining variables inside `OBJECT` paragraph, error with `LOCAL-STORAGE SECTION` due to the method being nested inside class, etc. I am fixing these errors now. I will push the changes when done.

kmr-srbh:
Update: We fixed an important error in today's meeting. It turned out to be due to an external state variable affecting the scanner's output. The test from the ISO standard throws some errors related to variable names being undefined now, which is related to AST work and fine for now. I will push the test and update the changelog.

kmr-srbh:
Update: All tests pass locally. I can squash all commits after review.

cobc/cobc.h
```diff
@@ -177,6 +177,10 @@ enum cb_current_date {
 #define	CB_CS_REPOSITORY		(COB_U64_C(1) << 42)
 #define	CB_CS_CALL_USING		(COB_U64_C(1) << 43)	/* within USING phrase of CALL statement */
 #define	CB_CS_READY_OR_RESET		(COB_U64_C(1) << 44)
+#define	CB_CS_CLASS_SPECIFIER		(COB_U64_C(1) << 45)
+#define	CB_CS_INTERFACE_SPECIFIER	(COB_U64_C(1) << 46)
+#define	CB_CS_FACTORY_PARAGRAPH		(COB_U64_C(1) << 47)
+#define	CB_CS_OBJECT_PARAGRAPH		(COB_U64_C(1) << 48)
```
nberth, line 183: These new special contexts are not mentioned in the ChangeLog. You can add an entry like `cobc.h: add new special contexts for OO constructs`.

cobc/codegen.c
nberth: As you have added some `TODO` comments here and there already, here could be one that indicates the new OO kinds of programs are to be handled here.

cobc/tree.h
```diff
@@ -2617,6 +2625,9 @@ extern cb_tree		cb_build_xml_parse (cb_tree, cb_tree,
 extern void		cb_emit_json_generate (cb_tree, cb_tree, cb_tree,
 					       cb_tree, cb_tree);
 
+extern const char * get_cob_module_type_string(enum cob_module_type);
+extern int validate_parent_name_in_spec_list(cb_tree, cb_tree);
```
nberth: I think those functions should be prefixed with `cb_`. (Also a space is missing before each opening parenthesis).

cobc/scanner.l
```diff
@@ -1065,9 +1085,19 @@ H#[0-9A-Za-z]+ {
 			RETURN_TOK (LITERAL);
 		}
 	} else if ((second_last_token == CLASS_ID && last_token == TOK_DOT) 
-		     || last_token == END_CLASS) {
+		    || last_token == END_CLASS) {
 		yylval = cb_build_reference (yytext);
 		RETURN_TOK (CLASS_NAME);
+	} else if ((second_last_token == METHOD_ID && last_token == TOK_DOT
+			&& strcasecmp (yytext, "GET") != 0
+			&& strcasecmp (yytext, "SET") != 0)
```
nberth: Could you describe what are the `GET`/`SET` special cases for?

cobc/tree.c
```diff
@@ -7670,6 +7670,34 @@ cb_build_prof_call (enum cb_prof_call prof_call,
 	return CB_BUILD_FUNCALL_3 (func_name, cb_int (prof_call), cb_int (func_arg1), cb_int (func_arg2));
 }
 
+/* Object-oriented */
+
+const char *
+get_cob_module_type_string(enum cob_module_type prog_type)
```
nberth: Style: space missing before parenthesis.

cobc/tree.c
```diff
@@ -7670,6 +7670,34 @@ cb_build_prof_call (enum cb_prof_call prof_call,
...
+{
+	cb_tree l;
+	
+	for (l = spec_list; l; l = CB_CHAIN (l)) {
+		if (strcasecmp (CB_NAME (CB_VALUE (l)), CB_NAME(parent_name)) == 0) {
+			return 0;
+		}
+	}
+	return 1;
+}
```
nberth, line 7701: Seems to me that this function performs a more general search in a list of "names" that just an OO-specific check: either there is already a function that does this search, or this one should be given a more generic name like `cb_check_in_names_list` or similar.

cobc/typeck.c
```diff
@@ -4817,7 +4817,42 @@ validate_assign_name (cb_tree origin, struct cb_program * const prog)
...
+					cb_error_x(x, _("duplicate parent %s name '%s'"), 
+						get_cob_module_type_string(prog->prog_type),
+						CB_NAME (CB_VALUE(x)));
+					cb_note_x(COB_WARNOPT_NONE, l, _("'%s' defined here"), 
+						CB_NAME (CB_VALUE(l)));
+					break;
+				}
+			}
+		}
+	}
```
nberth, line 4873: Considering this is likely going to grow and check other OO-related stuff *later*, I'd be in favor of moving those checks into a separate function above, like `cb_validate_oo_program_data` or something similar.  With a call possibly guarded by the value of `prog->prog_type`.

cobc/typeck.c
```diff
@@ -4817,7 +4817,42 @@ validate_assign_name (cb_tree origin, struct cb_program * const prog)
...
+	if (prog->oo_inheritance_list) {
+		if (prog->prog_type == COB_MODULE_TYPE_CLASS) {
+			spec_list = prog->class_spec_list;
+		} else if (prog->prog_type == COB_MODULE_TYPE_INTERFACE) {
+			spec_list = prog->interface_spec_list;
+		}
+		for (l = prog->oo_inheritance_list; l; l = CB_CHAIN (l)) {
+			if (strcasecmp(prog->program_id, CB_NAME(CB_VALUE (l))) == 0) {
+				cb_error_x(l, _("cannot inherit from the same %s being declared"),
+					get_cob_module_type_string(prog->prog_type));
```
nberth: Maybe a more direct "cannot inherit from %s"?

nberth: Javac is more general with:
```
X.java:1: error: cyclic inheritance involving X
public class X extends X {
       ^
1 error
```
I think I prefer that "cyclic" and "inheritance" are mentioned; likely to be generalized to indirect inheritance *later*.

cobc/parser.y
```diff
@@ -8225,10 +8466,13 @@ type_to_clause:
...
   _usage_is usage
-| USAGE _is conflict_usage	/* auto-enters CB_CS_USAGE */
+| USAGE _is OBJECT REFERENCE _object_reference_type
   {
+	check_and_set_usage (CB_USAGE_OBJECT);
+	CB_PENDING ("USAGE OBJECT");
+	
 	__CS_LEAVE (CB_CS_USAGE);
-  }
+  }	/* auto-enters CB_CS_USAGE */
```
nberth: Can you leave the "auto-enters" comment on the line where the `USAGE` token is?
(I suggest to break the line after `_is`).

cobc/parser.y
```diff
@@ -11216,34 +11453,16 @@ _procedure_division:
...
-	cb_set_system_names ();
-	last_source_line = cb_source_line;
-
-	cb_prof_procedure_division (
-		current_program,
-		cb_source_file,
-		cb_source_line
-		);
-  }
-  DIVISION
```
nberth, line 11238: You can probably move the `procedure_division_header` rule here to minimize the diff.

cobc/parser.y
```diff
@@ -4127,7 +4335,7 @@ _program_body:
 	cb_validate_program_data (current_program);
 	within_typedef_definition = 0;
   }
-  _procedure_division
+  _procedure_division { cobc_in_procedure = 0; }
```
nberth:
```suggestion
  _procedure_division
  {
	/* Manual reset as some full OO constructs (with DATA DIVISION and co.)
	   appear inside PROCEDURE DIVISIONs.  */
	cobc_in_procedure = 0;
  }
```

cobc/parser.y
```diff
@@ -4103,11 +4277,45 @@ class_id_paragraph:
...
+method_id_header:
+  METHOD_ID
+  {
+	__CS_CLEAR_ALL();
+  }
+;
+
+get_or_set:
+  GET	{ $$ = $1; }
+| SET	{ $$ = $1; }
```
nberth: Why return the token? Maybe returning a value from the new `enum cb_oo_property_accessor` may make sense here.

nberth: Otherwise, assign `current_program->getter_or_setter` directly…

cobc/parser.y
```diff
@@ -4019,17 +4171,36 @@ class_id_name:
...
+oo_parent_name_list:
+  WORD
+  {
+	$$ = CB_LIST_INIT ($1);
+  }
+| oo_parent_name_list WORD
+  {
+	$$ = cb_list_add ($1, $2);	
+  }
 ;
```
nberth, line 4217: In the end this is a list of `WORD`s for which every check is delayed; this rule should be given a more generic name like `word_list`.

nberth (commented):
Thanks that's very good progress!

cobc/typeck.c
GitMensch: javac is a good check here, possibly better is g++ (where msgids are "more near" in general)

cobc/scanner.l
GitMensch: it is to return METHOD_NAME instead of WORD, but I think I'd prefer the parser to parse GET/SET and build the method name there (as bison generated parser then says "... expecting GET or SET" which is more clear.

if we need that in any place, then I'd tend to `&& yylen == 3 && memcpmp...`

cobc/parser.y
```diff
@@ -8225,10 +8466,13 @@ type_to_clause:
 
 usage_clause:
   _usage_is usage
-| USAGE _is conflict_usage	/* auto-enters CB_CS_USAGE */
+| USAGE _is OBJECT REFERENCE _object_reference_type
   {
+	check_and_set_usage (CB_USAGE_OBJECT);
+	CB_PENDING ("USAGE OBJECT");
```
nberth: Here I'd suggest `OBJECT REFERENCE` instead of `OBJECT` only, as the standard(s) seem to always use both words.

GitMensch:
> Hi guys, just checking for the timeline.
> 
> The original proposal had the first coding period ending today with "Handle complete syntax for each construct as defined by the ISO standard." Not sure how far you are with "complete syntax" or may even need to reduce the amount of syntax parsed.
> 
> The current GSoC timeline has the first coding period to July 6th - similar for the end of the second period being a week longer with the final end date August 24th.
> 
> If both of you want to do longer, we can still extend the period as well, but only if we request this week, if I remember correctly.

Just to validate: we want to keep the original plan (if not then @nberth may write to the GNU project GSoC org admins), doing the evaluation next week, right? (@nberth please write the evaluation and feel free to check via Matrix if you have any questions, otherwise drop a note and I'll have a second look afterwards)
With or shortly after the evaluation we'll try to get an upstream commit (likely by @nberth with me adjusting the commiter afterwards) of the current work to gc3 before working on the codegen, right?

cobc/tree.c
kmr-srbh: Yeah this does look generic. There are functions that do searches, but they are scoped for their intended purposes. I have renamed this.

kmr-srbh:
Update: Pushed changes with respect to the code review. Removed getter/setter AST node. Fixed tests for GET/SET syntax based on the MicroFocus docs (https://www.microfocus.com/documentation/reuze/60d/opmeth.htm#u040). Supported `SUPER` keyword in `INVOKE` statement. Tried adding support for inline method invocations as decided in our last meeting, but got conflicts because the rule starts with any identifier (I have pushed the commented out rule though).

kmr-srbh:
Update: Got inline method invocation syntax working and added tests. I supported it as an expression with the already present `function` rule. Finding the correct place to add it required looking around a little. Added relevant tests.

cobc/reserved.c
```diff
@@ -637,7 +637,7 @@ static struct cobc_reserved default_reserved_words[] = {
 				CB_CS_GRAPHICAL_CONTROL, CB_CS_DISPLAY | CB_CS_SCREEN
   },
   { "CLASS",			0, 0, CLASS,			/* 2002 */
-				0, 0
+				CB_CS_CLASS_SPECIFIER, 0
```
nberth, line 640: Now that I look at it a second time, I think there is an issue here as `CLASS` is a token that appears in several constructs of the grammar.  So, entering the class-specifier special context in this way may not be the proper solution.  It is quite possible a properly placed `__CS_ENTER (CB_CS_CLASS_SPECIFIER);` in the parser may be enough.

(This additionally says some more checks may be needed in the parser to detect every context is properly exited — or else that means `CLASS` (for "character"-class) is not used that often in the testsuite.)

cobc/scanner.l
```diff
@@ -1065,9 +1085,25 @@ H#[0-9A-Za-z]+ {
...
+		if (strcasecmp (yytext, "GET") == 0) {
+			yylval = cb_build_reference (yytext);
+			RETURN_TOK (GET);
+		} else if (strcasecmp (yytext, "SET") == 0) {
+			yylval = cb_build_reference (yytext);
+			RETURN_TOK (SET);
+		} else {
+			yylval = cb_build_reference (yytext);
+			RETURN_TOK (METHOD_NAME);
+		}
```
nberth: Ah there is still this issue we need to look into more carefuly.

cobc/parser.y
```diff
@@ -2999,19 +3012,24 @@ set_oo_class_attr(enum cb_oo_class_attribute attr, const char* attr_name)
 %token INITIALIZE
 %token INITIALIZED
 %token INITIATE
+%token INLINE_METHOD_INVOCATION_OP
```
nberth: Note we usually do not name tokens according to their semantics. Here I'd suggest doing like `TOK_EQUAL` and co, and using `TOK_COLON_COLON` or `TOK_DOUBLE_COLON`.  Also adding the `"::"` afterwards for error messages.

cobc/parser.y
```diff
@@ -15271,6 +15560,16 @@ inspect_body:
...
 
+/* INVOKE statement */
+
+invoke_statement:
+  INVOKE
+  id_or_class_name
+  id_or_lit
+  call_using
+  call_returning
+;
```
GitMensch, line 15585: Note: You may check the syntax rules for the invoke statement and try to implement at least those outside of the using/returning args as this will likely show a bunch of missing attributes in the used structures of the AST.
Checking the using/returning parameters for conformance is also useful, this part will show if the classes and methods are correctly stored in the internal "external repository".

Both checks should be done in typeck.c and be called from both `invoke_statement` and `inline_method_invocation`, which may both yield in nearly the same code generation later as well.

If possible I'd suggest to have both in before merging the OO changes upstream.
You guys will see together if that's reasonable and if more parts are necessary before that.

cobc/parser.y
```diff
@@ -4008,28 +4157,60 @@ class_id_header:
...
+  }
+;
+
+interface_id_header:
+  INTERFACE_ID
+  {
+	cobc_in_id = 1;
+	__CS_CLEAR_ALL();
+
+	CB_UNSUPPORTED ("interfaces in object-oriented COBOL");
```
GitMensch: Do you plan to postpone the interfaces / split it from the initial OO implementation (the GSoC project)?
If yes, then change to `object-oriented COBOL (INTERFACE)`, if not, then go with the general `object-oriented COBOL`.

kmr-srbh: AST and codegen support for interfaces were excluded in the proposal actually. I will change the error message.

cobc/scanner.l
kmr-srbh: @nberth and I resolved this in our last meeting.

kmr-srbh:
Update: Added checks for arguments in `INVOKE` statement. The changes do not work yet due to a bug or missing implementation that causes the OO class name to not be populated for a variable of `USAGE OBJECT REFERENCE`. It is currently `NULL`. Debugging this.

On a side note, I think I _probably_ found a bug with the argument checks for the `CALL` statement, where the arguments are not being actually checked from the `program` being called but falling back to default `NULL` arguments. MRE:

```cobol
    IDENTIFICATION DIVISION.
       PROGRAM-ID. prog.

       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01 str PIC X(20) VALUE ALL "X".

       PROCEDURE DIVISION.
           *> No arg provided, but statement allowed
           CALL "subprog".
       END PROGRAM prog.

       IDENTIFICATION DIVISION.
       PROGRAM-ID. subprog.

       DATA DIVISION.
       LINKAGE SECTION.
       01 str PIC X ANY LENGTH.

       PROCEDURE DIVISION USING str.
            DISPLAY str.
           MOVE "abcd" TO str
           DISPLAY FUNCTION TRIM (str)
           MOVE "abcd" TO str (5:)
           DISPLAY FUNCTION TRIM (str)
           MOVE ALL "a" TO str
           DISPLAY FUNCTION TRIM (str)
           .
       END PROGRAM subprog.
``` 

Output:

```console
(base) saurabhkumar@Mac gnucobol % cobc -v         
cobc (GnuCOBOL) 3.3-dev.0
Built     Aug 08 2026 00:28:32  Packaged  Jun 15 2026 14:48:00 UTC
C version "Apple LLVM 21.0.0 (clang-2100.1.1.101)"
loading standard configuration file 'default.conf'
cobc: error: no input files
(base) saurabhkumar@Mac gnucobol % cobc -x prog.cob
(base) saurabhkumar@Mac gnucobol % ./prog

attempt to reference invalid memory address (signal SIGSEGV)


 Last statement of "subprog" unknown
 Last statement of "prog" unknown
 Started by ./prog
zsh: segmentation fault  ./prog
```

I think the correct output should have been a "expecting up to ... arguments, but found ..." error message? Since I was using this test to learn about the behaviour of `cb_check_conformance()` and not getting the correct result, I ended up looking through it and found that the function `cb_find_defined_program_by_name()` does not find the program being called inside `defined_prog_list`. This is not a blocker or directly related to our work on OOP though.

tests/testsuite.src/syn_oo.at
```diff
@@ -244,9 +301,515 @@ AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
...
+                MOVE FUNCTION CURRENT-DATE (1: 8) TO the-date
+                GOBACK.
+            END METHOD initializeAccount.
+        END OBJECT.
+        END CLASS Account.
+])
+
+AT_CHECK([$COMPILE_ONLY prog.cob], [1], [],
+[prog.cob:3: error: object-oriented COBOL is not supported
+prog.cob:7: warning: no definition/prototype seen for CLASS 'Base'
```
nberth, line 623: @GitMensch To fix that one (Base is supposed to be implicitly defined), is there a a good place in the code to define this kind of implicit prototypes?

GitMensch: For the time being: use the official prototype for this, maybe even put it into a base.cpy and `--copy` include it:

```cobolfree
Interface-id. BaseFactoryInterface.     
Procedure division.       
   Method-id. New.       
   Data division.       
   Linkage section.       
   01 outObject usage object reference active-class.      
   Procedure division returning outObject.       
End method New.      
End Interface BaseFactoryInterface.

Interface-id. Interface. 
Procedure division.       
   Method-id. FactoryObject.      
   Data division.       
   Linkage section.       
   01 outFactory usage object reference factory of active-class.      
   Procedure division returning outFactory.       
   End method FactoryObject.      
End Interface BaseInterface.
```

but whate we really _should_ have (in reserved.c) is a static code-definition for prototypes that we can "just load" into our prototype structure.

Note that this would also (at least partially) replace or partially use system.def to get the current checks (name, arguments, ...) for CALL and also intrinsic functions in there, removing the _additional_ checks we have for them, while also allowing to define lists that can be included in the words file (hm, maybe just define the structure and always parse the definitions from a file?)
Concerning performance: neither base class, nor the intrinsic functions and even the prototypes for CBL+GC programs won't take too much resources - if we don't create them as full COBOL sources = not using the parser but something like already done in config,c. ... and later on we can (and should) come back to the open FR of allowing cobc to build a binary version of cfg/words/... files at the end of "make" which are then "only loaded" during compile (maybe putting the binary representation into the config dir and use them as long as they are newer than the target file).

GitMensch:
> Update: Added checks for arguments in `INVOKE` statement. The changes do not work yet due to a bug or missing implementation that causes the OO class name to not be populated for a variable of `USAGE OBJECT REFERENCE`. It is currently `NULL`. Debugging this.

Is this solved already or still open?

PR #294, opened 2026-06-14
Integration of EXEC preprocessor beyond baseline.

utam-1:
#### Description of PR
This is a follow - up to #279.

#### Approach and basic idea
`cb_preparser_entry` is added to `cobc.c`, which handles the associated pre-parser metadata. `cb_preparser_list` and `cb_active_preparser` are created, one being a list to registered preparsers, and other is a pointer for `pplex.l` to identify the tag.

`cb_load_preparser_conf()` and `cb_find_preparser()` are created based on similar mechanism of `cb_config_entry()`. 
In `ppparse.y`  a token ` SUBSYSTEM_FOUND` is declared which is used for identifying EOF, and for gracefully exiting the parsing.

So far only general tests have been performed by using a `test.conf` file and setting up a pseudo - preprocessor via **bash**, which shows the replacement works and flag-injection is performed via **gcc**.


cobc/cobc.h
GitMensch, line 1: Can we move those definitions to tree.h?
In any case, please tab-indent (not space-indent).

cobc/pplex.l
```diff
@@ -569,6 +578,43 @@ MAYBE_AREA_A	[ ]?#?
...
+        /* dotted name e.g. SQLCA.cpy or table.field */
+        pplval.s = cobc_plex_strdup (yytext);
+        return TOKEN;
+    }
+    {WORD}                  {
+        struct cb_preparser_entry *pe;
+        pplval.s = cobc_plex_strdup (yytext);
+        if (!exec_subsystem_seen) {
+            exec_subsystem_seen = 1;
+            pe = cb_find_preparser (yytext);
```
GitMensch: switch those two assignments - which also allows you to directly do `struct cb_preparser_entry *pe = cb_find...`

cobc/cobc.c
```diff
@@ -3118,6 +3122,157 @@ file_replace_extension (const char *file, const char *ext)
...
+cb_load_preparser_conf (const char *name)
+{
+        FILE                            *fp;
+        char                            resolved[COB_NORMAL_BUFF];
+        char                            buff[COB_SMALL_BUFF];
+        struct cb_preparser_entry       *pe;
+
+        fp = open_preparser_conf (name, resolved, sizeof (resolved));
+        if (!fp) {
+                cb_error (_("preparser configuration '%s' not found"), name);
```
GitMensch: just use the same approach as with config files here:
`cb_perror (1, "%s: %s", resolved, cb_get_strerror ());`

cobc/cobc.c
```diff
@@ -3118,6 +3122,157 @@ file_replace_extension (const char *file, const char *ext)
...
+                        pe->command = cobc_main_strdup (val);
+                } else if (strcasecmp (key, "cflags") == 0) {
+                        pe->cflags = cobc_main_strdup (val);
+                } else if (strcasecmp (key, "ldflags") == 0) {
+                        pe->ldflags = cobc_main_strdup (val);
+                } else if (strcasecmp (key, "on-error") == 0) {
+                        pe->on_error = (strcasecmp (val, "warn") == 0) ? 0 : 1;
+                } else {
+                        cb_warning (cb_warn_unsupported,
+                                     _("unknown preparser configuration key '%s' in '%s'"),
```
GitMensch: use same msgid as in the config reading:
`			_("unknown configuration tag '%s'"), key);`
(and, if reasonable, rename key->tag)

cobc/cobc.c
```diff
@@ -3118,6 +3122,157 @@ file_replace_extension (const char *file, const char *ext)
...
+        if (!pe->tag || !pe->command) {
+                cb_error (_("preparser configuration '%s' is missing 'tag' or 'command'"),
+                          resolved);
+                return 1;
+        }
+
+        pe->next = cb_preparser_list;
+        cb_preparser_list = pe;
+        return 0;
+}
```
GitMensch: those functions should be moved to config.c - and, where possible, the existing functions be refactored to not have duplicated code for how the lines are read/split

cobc/cobc.h
```diff
@@ -287,6 +287,17 @@ struct cb_text_list {
 	char			*text;
 };
 
+struct cb_preparser_entry {
+        struct cb_preparser_entry      *next;
+        char                            *tag;      /* e.g. "SQL" (upper-cased) */
```
GitMensch: rename to "subsystem"

cobc/cobc.c
```diff
@@ -5460,6 +5633,56 @@ preprocess (struct filename *fn)
...
+                fflush (stderr);
+        }
+
+        ret_sys = call_system (cmd);
+        cobc_free (cmd);
+
+        if (ret_sys != 0) {
+                cb_source_file = fn->source;
+                if (cb_active_preparser->on_error) {
+                        cb_error (_("external preparser '%s' failed with exit status %d"), cb_active_preparser->tag, ret_sys);
```
GitMensch: here and below: tab-indent instead of space-indent (your editor may be setup to insert spaces instead of tabs)

cobc/cobc.c
```diff
@@ -5460,6 +5633,56 @@ preprocess (struct filename *fn)
 	/* Release flex buffers - After file close */
 	plex_call_destroy ();
 
+	if (cb_active_preparser) {
+        char *cmd;
+        size_t cmd_len;
+        int ret_sys;
+        char *new_preprocess;
+
+        cmd_len = strlen (cb_active_preparser->command) + strlen (fn->source) + strlen (fn->preprocess) + 32;
```
GitMensch: Why `+ 32`? Seems the righjt one is `+ 2 + 1` (two spaces and one low-value).

cobc/cobc.c
```diff
@@ -5460,6 +5633,56 @@ preprocess (struct filename *fn)
...
+        cmd = cobc_malloc (cmd_len);
+        snprintf (cmd, cmd_len, "%s %s %s", cb_active_preparser->command, fn->source, fn->preprocess);
+
+        if (verbose_output) {
+                fprintf (stderr, "invoking external preparser: %s\n", cmd);
+                fflush (stderr);
+        }
```
GitMensch: Isn't that part of `call_system()` alreasdy?

GitMensch (changes requested):
* new changes need a new Changelog entry
* we'll need a NEWS entry for the new --preparser, as well as help.c and gnucobol.texi addition; the later should also document a bit about the file format
* please add your tests (you've seem to have done some) to used_binaries.at; we'd need at least:
    * --preparser for not existing file
    * --preparser for existing, but invalid file
    * --preparser for existing file
    * --preparser for two existing files
    * --preparser, where pseudo preparser runs fine
    * --preparser, where psuedo preparser runs in error

cobc/cobc.c
utam-1: I kept the implementation for the two differently as they follow different kinds of implementation ( `config table` and `Linked list` ) the coupling between these two may make it difficult to maintain in my opinion.

doc/gnucobol.texi
```diff
@@ -788,6 +788,9 @@ Compiler uses the given @var{dialect} to determine the reserved words.
 @item -conf=<file>
 User-defined dialect configuration.
 
+@item --preparser=<file>
+Register external preparser configuration.
```
GitMensch: That's a start, but similar to the note below you need to explain what this is used for and either document the structure here or add an example definition (for esqlOC and/or GixSQL) showing the format with explanations.

cobc/cobc.c
GitMensch: `parse_preparser_line()` should be replaced by re-using the code for the config lines; to do so move the code of cb_config_entry for the parsing out of the function (the new static function gets a pointer to name+value, just as your version) - in this new variant name and value will be parsed first before `cb_config_entry()` checks the name (which is now intermixed).

This way we have a single new function called for both compiler and preparser config, ensuring the format is exactly identical, by keeping the way it is looked up / used out of that code.

I agree that `cb_load_preparser_conf()` itself should be kept separate from the config file loading (though they will look very similar) because of the struct / array approach (which we don't need fo the preparser).

cobc/ChangeLog
```diff
@@ -1,4 +1,16 @@
...
+	* pplex.l: add EXEC_STATE exclusive start condition to
+	  tokenize EXEC <TAG> ... END-EXEC blocks; add AREA_A lookahead
+	  rule for EXEC; return INCLUDE token for INCLUDE keyword,
+	  plain TOKEN for words; consume all other content silently
+	* ppparse.y: add exec_statement grammar rule dispatched
+	  from statement_no_replace; EXEC TAG INCLUDE handled as COPY
+	  (full copybook expansion with -I and -ffold-copy support);
+	  all other EXEC TAG blocks warn at "unsupported" level
+	  (defaulting to error) and are ignored; add _exec_token_list
+	  to consume body tokens for parsing purposes only
```
GitMensch:
```suggestion
	  to consume body tokens for parsing purposes only

```

cobc/ChangeLog
GitMensch, line 1: you want one entry for the baseline (with the old date) and another one for the follow-up work, which then outlines the changes in config.c, tree.h, ...

GitMensch: that whole block should - for the baseline changes - be after the may changes, nomt up-front

cobc/cobc.c
```diff
@@ -5372,7 +5391,9 @@ preprocess (struct filename *fn)
 	const size_t exception_table_size = sizeof (struct cb_exception) * COB_EC_MAX;
 	int			save_source_format, save_fold_copy, save_fold_call,
 		save_ref_mod_zero_length;
-
+	char *orig_source     = cobc_strdup (fn->source);
+    char *orig_preprocess = cobc_strdup (fn->preprocess);
+	struct cb_preparser_entry *p_reset;
```
GitMensch, line 5394: You likely can just assign them (otherwise you'd need to ensure to free those pointers once they are not used any more)

cobc/config.c
```diff
@@ -26,7 +26,7 @@
 #include <stddef.h>
 #include <string.h>
 #include <limits.h>
-
+#include <ctype.h>
```
GitMensch: the existing code has no isspace, so you can get rid of that header when refactoring the existing code to be used for the tag/value split

ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2026-07-31  Uttam Bhadauriya <uttamsinghbhadoriya23@gmail.com>
+
+	* cobc/cobc.h, cobc/tree.h, cobc/cobc.c, cobc/config.c, cobc/pplex.l:
+	Refactor external preparser subsystem, move struct and functions,
+	rename tag to subsystem, improve error messages and indentation.
+	* cobc/help.c, doc/gnucobol.texi: Add --preparser documentation.
+	* tests/testsuite.src/used_binaries.at: Add preparser tests.
+
```
GitMensch: cobc and doc go to their own changelog files, the testsuite sources don't get an entry for new elements

GitMensch (commented):
open from last review:
* new changes need a new Changelog entry (you've added them, just in the wrong place)
* gnucobol.texi should also document a bit about the file format

NEWS
```diff
@@ -6,6 +6,8 @@ NEWS - user visible changes				-*- outline -*-
 
 * New GnuCOBOL features
 
+** cobc provides a --preparser option to register external preparser configurations
```
GitMensch: explain here that EXEC is now supported using external preparsers, registered with --preparser (you can add a hint to check the details in the manual, similar to the existing entries doing so) with a fallback of ignoring any EXEC parts but INCLUDE

NEWS
```diff
@@ -6,6 +6,8 @@ NEWS - user visible changes				-*- outline -*-
...
+
```
GitMensch, line 17: new entries to the NEWS file are added at the end of the relevant chapter

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,71 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           EXEC SQL SELECT 1 END-EXEC.
+           DISPLAY "OK".
+           STOP RUN.
+])
+
+# 1. nonexistent file -> error
+AT_CHECK([$COBC --preparser missing.conf prog.cob], [1], [], [stderr])
+AT_CHECK([$GREP "cannot load preparser configuration 'missing.conf'" stderr], [0], [ignore], [ignore])
```
GitMensch: haven't checked the ones below, but instead of placing that to an stderr where you grep from (which either needs a redirection or I learn something new here) you'd just check the output in that [] directly.

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,71 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
 ])
 
 AT_CLEANUP
+
+AT_SETUP([external preparser --preparser])
+AT_KEYWORDS([preparser])
```
GitMensch:
```suggestion
AT_KEYWORDS([cobc configuration EXEC])
```

you can replace the keyword by cobc + configuration - other words are already included by AT_SETUP content

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,71 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
 ])
 
 AT_CLEANUP
+
+AT_SETUP([external preparser --preparser])
```
GitMensch, line 1614:
```suggestion
AT_CLEANUP


AT_SETUP([external preparser --preparser])
```

nice test with multiple checks - just add two empty lines before each new test group :-)

GitMensch (changes requested):

GitMensch:
@utam-1 please rebase, which should improve the MSVC part and update per review, so we can have a next look at that

utam-1:
> @utam-1 please rebase, which should improve the MSVC part and update per review, so we can have a next look at that

I have pushed my changes, and apologies for the delay I was a bit occupied.

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,78 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+])
+
+# 2. invalid file -> error
+AT_DATA([invalid.conf], [
+# missing subsystem and command
+cflags: -O2
+])
+AT_CHECK([$COBC --preparser ./invalid.conf prog.cob], [1], [],
+[error: preparser configuration './invalid.conf' is missing 'subsystem' or 'command'
+])
```
GitMensch: I suggest to move these parts to configuration.at and add the missing "unknown configuration tag" code path

utam-1:
@GitMensch I think the issue with msvc persists, maybe I made some mistake while doing the rebase. I'm not sure though.

cobc/config.c
```diff
@@ -939,3 +913,159 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+	size_t	i;
+
+	for (i = 0; name[i] != 0 && name[i] != SLASH_CHAR; i++);
+
+	if (name[i] != 0 || access (name, F_OK) == 0) {
+		snprintf (resolved, resolved_size, "%s", name);
+	} else {
+		/* Simplified: Always append .conf for plain names in config dir */
+		snprintf (resolved, resolved_size, "%s%c%s.conf",
+			  cob_config_dir, SLASH_CHAR, name);
```
GitMensch: this simplification is likely the reason for the win32 failures:
*  you want to only do the fopen below if the access worked
* the snprintf should only be done if the access worked as you want to point to the "normal" name if there is neither one with the original name nor with the config dir

GitMensch (changes requested):

cobc/config.c
utam-1: I made this change but it seems to fail again for some reason.

utam-1:
@GitMensch The tests are still failing, I'm not sure what's causing this from the error messages it says about some dialect of C89 / C11 not being met ( for declartion of for-loop ) and some other path related issues are present.

cobc/config.c
GitMensch: you still set resolved_size if the path did not exist and use it afterwards, I'll do an edit you can inspect and we'll see if that helps

GitMensch:
> @GitMensch The tests are still failing, I'm not sure what's causing this from the error messages it says about some dialect of C89 / C11 not being met ( for declartion of for-loop ) and some other path related issues are present.

That's
~~~
../../cobc/config.c: In function ‘cb_conf_split_line’:
../../cobc/config.c:938:17: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode
  938 |                 for (size_t j = strlen (buff);
      |                 ^~~
../../cobc/config.c:938:17: note: use option ‘-std=c99’, ‘-std=gnu99’, ‘-std=c11’ or ‘-std=gnu11’ to compile your code
~~~
and it is nice to see that the message got much better over the years :-)

It means "for" is a statement, and "size_t j" is a declaration. In C89 (the compatibility check done here) declarations need to be at the start of a block, so that rule is broken. The fix here is to move the size_t declaration before the for statement - but we don't need that in any case as we can directly adjust the pointer buff.
I'll include that in my commit.

utam-1:
> > @GitMensch The tests are still failing, I'm not sure what's causing this from the error messages it says about some dialect of C89 / C11 not being met ( for declartion of for-loop ) and some other path related issues are present.
> 
> That's
> 
> ```
> ../../cobc/config.c: In function ‘cb_conf_split_line’:
> ../../cobc/config.c:938:17: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode
>   938 |                 for (size_t j = strlen (buff);
>       |                 ^~~
> ../../cobc/config.c:938:17: note: use option ‘-std=c99’, ‘-std=gnu99’, ‘-std=c11’ or ‘-std=gnu11’ to compile your code
> ```
> 
> and it is nice to see that the message got much better over the years :-)
> 
> It means "for" is a statement, and "size_t j" is a declaration. In C89 (the compatibility check done here) declarations need to be at the start of a block, so that rule is broken. The fix here is to move the size_t declaration before the for statement - but we don't need that in any case as we can directly adjust the pointer buff. I'll include that in my commit.

This is rather interesting; I'll review (and learn) from the commit for both the fix related to MSVC build and this issue once it is applied.

GitMensch:
work will still take some time, so I did an untested (not even compiled so far) commit - if it fails it should still show the idea...

edit: seems you only need to update the expected test result to let it pass;
@utam-1 if you have any questions on the changes we can schedule a quick meeting

cobc/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2026-07-31  Uttam Singh Bhadauriya <uttamsinghbhadoriya23@gmail.com>
+
+	* cobc.c, config.c, tree.h, pplex.l:
+	  refactor external preparser: move struct cb_preparser_entry
+	  to tree.h; rename tag->subsystem; move config loading
+	  functions to config.c; share line-parsing with cb_config_entry;
+	  fix indentation; fix cmd_len calculation
+	* help.c: add --preparser option documentation
 2026-06-08  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```
GitMensch:
```suggestion

2026-06-08  Nicolas Berthier <nicolas.berthier@ocamlpro.com>
```

cobc/ChangeLog
```diff
@@ -25,6 +33,31 @@
...
+
+	* config.def: "tab-width" option is changed to a comma-separated
+	  list of tab widths, for example "6,1,4", the last one being reused.
+	  This new meaning is backward compatible.
+	  Implements FR #498 " Adjust tab-width to optionally be a list of
+	  tab-stop positions"
+	* config.c: initialize cb_tab_width as a string,
+	  each position indicating the number of spaces to insert for a tab
+	  at that position
+	* pplex.c,cobc.c: use the new type of cb_tab_width
```
GitMensch, line 47:
```suggestion
	* pplex.l, cobc.c: use the new type of cb_tab_width
```

... but I'm a but confused where that entry comes from...

doc/ChangeLog
```diff
@@ -1,4 +1,9 @@
 
+2026-07-31  Uttam Singh Bhadauriya <uttamsinghbhadoriya23@gmail.com>
+
+	* gnucobol.texi: document --preparser option and EXEC preprocessing;
+	  add @node EXEC preprocessing section with external preparser
+	  configuration file format and GixSQL integration example
```
GitMensch, line 6:
```suggestion
	  configuration file format and GixSQL integration example

```

cobc/tree.h
```diff
@@ -524,6 +524,18 @@ struct cobc_reserved {
...
 };
 
+/* External preparser configuration entry */
+struct cb_preparser_entry {
+	struct cb_preparser_entry	*next;
+	char				*subsystem;	/* e.g. "SQL" (upper-cased) */
+	char				*command;	/* external preparser command */
+	char				*cflags;	/* extra C compiler flags */
+	char				*ldflags;	/* extra linker flags */
+	int				 on_error;	/* 0=warn, 1=error (default) */
```
GitMensch: please refactor so the name (and behavior) is "warn_only" (so 0 default is obvious)

GitMensch (changes requested):

utam-1:
> work will still take some time, so I did an untested (not even compiled so far) commit - if it fails it should still show the idea...
> 
> edit: seems you only need to update the expected test result to let it pass; @utam-1 if you have any questions on the changes we can schedule a quick meeting

@GitMensch I was able to get an insight on the approach, especially for the fix related to movement of pointer in `buff` instead of relying at `size_t`. However, I believe the issue related to MSVC is still there no?
Edit: The tests are passing successfully now, I modified the tests and changed the path look-up. Let me know if any further changes are needed.

NEWS
GitMensch: that's still a TODO

tests/testsuite.src/used_binaries.at
```diff
@@ -1609,3 +1609,59 @@ D8 D9 DA DB DC DD DE DF E0 E1 E2 7C E3 E4 E5 E6
...
+subsystem: SQL
+command: sh ./badpp.sh
+on-error: warn
+])
+AT_CHECK([$COBC -fno-diagnostics-show-caret -fno-diagnostics-show-option --preparser warn.conf prog.cob], [1], [],
+[prog.cob:5: error: external preparser 'SQL' failed; falling back to baseline
+prog.cob:6: error: EXEC SQL statement ignored
+])
+
+AT_CLEANUP
```
GitMensch, line 1667:
```suggestion
AT_CLEANUP

```

don't ask me why but some diff tools work better if there's an empty line after the cleanup...

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,181 @@ AT_DATA([prog.cob], [
...
+       END-EXEC.
+       PROCEDURE DIVISION.
+       MAIN.
+           MOVE 1 TO ENO.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported -ffold-copy=LOWER prog.cob], [1], [],
+[prog.cob:8: error: EXEC SQL INCLUDE handled as COPY
+])
```
GitMensch, line 3080:
```suggestion
AT_CHECK([$COMPILE_ONLY -Wno-unsupported -ffold-copy=LOWER prog.cob], [0], [], [])
```

we want to verify that the copy is read, so no reason for an error

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,181 @@ AT_DATA([prog.cob], [
...
+       PROCEDURE DIVISION.
+       MAIN.
+           DISPLAY SQLCODE.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported -ffold-copy=LOWER prog.cob], [1], [],
+[prog.cob:8: error: EXEC SQL INCLUDE handled as COPY
+])
+
```
GitMensch, line 3105:
```suggestion
AT_CHECK([$COMPILE_ONLY -Wno-unsupported -ffold-copy=LOWER prog.cob], [0], [], [])

```

same as above

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,181 @@ AT_DATA([prog.cob], [
...
+       MAIN.
+      * typo added on purpose to check handling that
+           EXEC SQL CONNECT RESET ENDEXEC.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported prog.cob], [1], [],
+[prog.cob:9: error: missing END-EXEC at end of file
+prog.cob:9: error: EXEC SQL statement ignored
+])
```
GitMensch, line 3125:
```suggestion
AT_CHECK([$COMPILE_ONLY -Wunsupported prog.cob], [1], [],
[prog.cob:9: error: missing END-EXEC at end of file
prog.cob:9: warning: EXEC SQL statement ignored
])
```
no need to error that explicit - but the missing part should be an error in any case

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,181 @@ AT_DATA([prog.cob], [
...
+prog.cob:11: error: EXEC SQL INCLUDE handled as COPY
+])
+
+AT_CHECK([rm -f emprec.cpy], [0], [], [])
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported -ffold-copy=LOWER prog.cob], [1], [], [prog.cob:8: error: EXEC SQL INCLUDE handled as COPY
+prog.cob:11: error: EXEC SQL INCLUDE handled as COPY
+prog.cob:11: error: emprec: No such file or directory
+prog.cob: in paragraph 'MAIN':
+prog.cob:15: error: 'ENO' is not defined
+])
```
GitMensch, line 3050:
```suggestion
AT_CHECK([$COMPILE_ONLY -Wunsupported -ffold-copy=LOWER prog.cob], [0], [],
[prog.cob:8: warning: EXEC SQL INCLUDE handled as COPY
prog.cob:11: warning: EXEC SQL INCLUDE handled as COPY
])

AT_CHECK([rm -f emprec.cpy], [0], [], [])
AT_CHECK([$COMPILE_ONLY -Wunsupported -ffold-copy=LOWER prog.cob], [1], [], [prog.cob:8: warning: EXEC SQL INCLUDE handled as COPY
prog.cob:11: warning: EXEC SQL INCLUDE handled as COPY
prog.cob:11: error: emprec: No such file or directory
prog.cob: in paragraph 'MAIN':
prog.cob:15: error: 'ENO' is not defined
])
```

tests/testsuite.src/syn_definition.at
```diff
@@ -2947,3 +2947,181 @@ AT_DATA([prog.cob], [
...
+       PROCEDURE DIVISION.
+       MAIN.
+           EXEC SQL CONNECT RESET END-EXEC.
+           STOP RUN.
+])
+
+AT_CHECK([$COMPILE_ONLY -Werror=unsupported prog.cob], [1], [],
+[prog.cob:6: error: EXEC SQL statement ignored
+])
+
```
GitMensch, line 3008:
```suggestion
AT_CHECK([$COMPILE_ONLY -Werror=unsupported prog.cob], [1], [],
[prog.cob:6: error: EXEC SQL statement ignored
])
AT_CHECK([$COMPILE_ONLY -Wunsupported prog.cob], [0], [],
[prog.cob:6: warning: EXEC SQL statement ignored
])

```

tests/testsuite.src/configuration.at
```diff
@@ -1064,3 +1064,66 @@ prog.cob:11: warning: start of statement in Area A
...
+])
+
+AT_DATA([prog.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           STOP RUN.
+])
+
+AT_CHECK([$COBC -fno-diagnostics-show-option -Wno-error=unsupported --preparser ./unknown.conf prog.cob], [0], [],
```
GitMensch, line 1125:
```suggestion
AT_CHECK([$COBC --preparser ./unknown.conf prog.cob], [0], [],
```

doc/gnucobol.texi
```diff
@@ -1949,6 +1996,32 @@ The number of calls to this section/paragraph
...
+library paths (@option{-I}) and case folding.
+@item
+For all other blocks, the compiler looks for an external preparser registered 
+for that specific subsystem tag. 
+@end enumerate
+
+External preparsers must be registered via the compiler configuration or the 
+@ref{--preparser option} command-line switch. If no preparser is registered for 
+a tag, or if a preparser fails in @code{warn} mode, the compiler will issue 
+a diagnostic and ignore the block.
```
GitMensch, line 2022: maybe put that part higher - the "internal" handling of INCLUDE only happens if there is no preparser registered for this subsystem

doc/gnucobol.texi
```diff
@@ -788,6 +789,52 @@ Compiler uses the given @var{dialect} to determine the reserved words.
...
+dialect configuration files. Recognized keys are:
+
+@table @code
+@item subsystem
+The @code{EXEC} tag this preparser handles (e.g. @code{SQL}).
+Automatically converted to upper-case. Required.
+
+@item command
+The external command to invoke. The command is called as:
+@code{@var{command} @var{input_file} @var{output_file}}.
```
GitMensch, line 808: Where would options for the preparser go?
Possibly use a reasonable example for GixSQL here as well, to be similar to the other options.

cobc/ppparse.y
```diff
@@ -51,7 +51,7 @@
 #define	_STDLIB_H 1
 #endif
 
-#define pperror(x)	cb_error_always ("%s", x)
+#define pperror(x)	do { if (!cb_active_preparser) cb_error_always ("%s", x); } while(0)
```
GitMensch, line 54: What do we need that additional check for?

cobc/pplex.l
```diff
@@ -1826,6 +1878,9 @@ ppparse_verify (const enum cb_support tag, const char *feature)
 void
 ppparse_error (const char *err_msg)
 {
+	if (cb_active_preparser) {
+		return;
+	}
```
GitMensch, line 1883: What do we need that additional check for?

cobc/pplex.l
```diff
@@ -569,6 +578,42 @@ MAYBE_AREA_A	[ ]?#?
...
+        return END_EXEC;
+    }
+    "INCLUDE"               { return INCLUDE; }
+    {WORD}\.{WORD}          {
+        /* dotted name e.g. SQLCA.cpy or table.field */
+        pplval.s = cobc_plex_strdup (yytext);
+        return TOKEN;
+    }
+    {WORD}                  {
+        pplval.s = cobc_plex_strdup (yytext);
```
GitMensch, line 593:
```suggestion
        /* note: first seen word is the subsystem to be executed */
        pplval.s = cobc_plex_strdup (yytext);
```
It is "kind of" obvious from the check below, but somehow I feel this is still lost easily. It could be that I or someone else feels different at another time, but wanted to suggest making it even more obvious

cobc/pplex.l
```diff
@@ -569,6 +578,42 @@ MAYBE_AREA_A	[ ]?#?
...
+    "END-EXEC"([ \t]*\.)?  {
+        BEGIN (INITIAL);
+        return END_EXEC;
+    }
+    "INCLUDE"               { return INCLUDE; }
+    {WORD}\.{WORD}          {
+        /* dotted name e.g. SQLCA.cpy or table.field */
+        pplval.s = cobc_plex_strdup (yytext);
+        return TOKEN;
+    }
```
GitMensch, line 591: if that doesn't change the lexer's baviour (quite possible for the first to cases, but not for the third) please move those down

cobc/pplex.l
```diff
@@ -569,6 +578,42 @@ MAYBE_AREA_A	[ ]?#?
...
+    {WORD}                  {
+        pplval.s = cobc_plex_strdup (yytext);
+        if (!exec_subsystem_seen) {
+            struct cb_preparser_entry *pe = cb_find_preparser (yytext);
+            exec_subsystem_seen = 1;
+            if (pe) {
+                pe->used = 1;
+                cb_active_preparser = pe;
+                /* Abort internal preparser cleanly by returning EOF */
+                return 0;
```
GitMensch, line 601: I suggest to return a special "internal" token here, like `SUBSYSTEM_TO EXECUTE`, then parse that explicit in the parser, existing cleanly there (possibly by calling `YYACCEPT;`).

cobc/pplex.l
```diff
@@ -525,6 +528,12 @@ MAYBE_AREA_A	[ ]?#?
 	return REPLACE;
 }
 
+"EXEC"/[ ,;\n] {
+	exec_subsystem_seen = 0;
+    BEGIN(EXEC_STATE);
+    return EXEC;
```
GitMensch, line 534:
```suggestion
	BEGIN EXEC_STATE;
	return EXEC;
```
same formatting as other states

cobc/pplex.l
```diff
@@ -145,6 +145,7 @@ static int			quotation_mark = 0;
 static int			echo_newline = 0;
 static int			listing_line = 0;
 static int			requires_listing_line;
+static int	exec_subsystem_seen = 0;	/* set after first EXEC token returned */
```
GitMensch, line 148:
```suggestion
static int			exec_subsystem_seen;	/* set after first EXEC token returned */
```
same formatting as above, var set already in the code

cobc/pplex.l
```diff
@@ -569,6 +578,42 @@ MAYBE_AREA_A	[ ]?#?
   }
 }
 
+<EXEC_STATE>{
+    "END-EXEC"([ \t]*\.)?  {
+        BEGIN (INITIAL);
```
GitMensch, line 583: please format here and below with leading tab/space as done in other states (you want your editor to not replace tabs by spaces and show whitespace for that)

cobc/config.c
```diff
@@ -488,52 +487,43 @@ cb_load_words (void)
 	return ret;
 }
 
+static int cb_conf_split_line (char *buff, char **name, char **val);
```
GitMensch, line 490: please move the function here instead of declarating it

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+			pe->command = cobc_main_strdup (tag_val);
+		} else if (strcasecmp (tag_name, "cflags") == 0) {
+			pe->cflags = cobc_main_strdup (tag_val);
+		} else if (strcasecmp (tag_name, "ldflags") == 0) {
+			pe->ldflags = cobc_main_strdup (tag_val);
+		} else if (strcasecmp (tag_name, "on-error") == 0) {
+			pe->warn_only = (strcasecmp (tag_val, "warn") == 0) ? 1 : 0;
+		} else {
+			cb_warning (cb_warn_unsupported,
+				    _("unknown configuration tag '%s'"), tag_name);
```
GitMensch, line 1050:
```suggestion
			configuration_error (name, line, 1,
				_("unknown configuration tag '%s'"), tag_name;
```

... and apply the line addition as I've had in the commit I've done (check cb_conf_split_line's return to either continue or break), increment line in any case; and yes, that needs an adjustment in the test output

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+		}
+	}
+	fclose (fp);
+
+	if (!pe->subsystem || !pe->command) {
+		cb_error (_("preparser configuration '%s' is missing 'subsystem' or 'command'"),
+			  resolved);
```
GitMensch, line 1057:
```suggestion
		configuration_error (name, 0, 1,
			_("missing preparser definition 'subsystem' or 'command'"));
```

.. and again, changed test output

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+		}
+	} else if (access (name, F_OK) == 0) {
+		/* plain name exists in current directory */
+		snprintf (resolved, resolved_size, "%s", name);
+	} else {
+		/* plain name: look up in COB_CONFIG_DIR/<name>.conf */
+		snprintf (resolved, resolved_size, "%s%c%s.conf",
+			  cob_config_dir, SLASH_CHAR, name);
+		if (access (resolved, F_OK) != 0) {
+			return NULL;
```
GitMensch, line 995:
```suggestion
			resolved[0] = 0; /* unset for the caller to use original name */
			return NULL;
```

if you add back the code in the caller that outputs the name, you may see the error in this -> if the file does not exists, then we pass a changed "resolved" with the config dir prefixed to the caller and use that in the error message

The initial code I've suggested was a bit more complex (only set resolved if the file can be accessed in the config folder [and is neither a path nor exists in the directory], otherwise use the original name) - but it always pointed to the right name in the caller.

You can still get it relative easy with the current code as noted above.

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+			&& name[i] != '/'
+	#endif
+		; i++);
+
+	if (name[i] != 0) {
+		/* explicit path given — use as-is */
+		snprintf (resolved, resolved_size, "%s", name);
+		if (access (resolved, F_OK) != 0) {
+			return NULL;
+		}
```
GitMensch, line 986:
```suggestion
		return fopen (name, "r");
```

Can't we drop that access and have still the same result (perror set from fopen instead of access)?

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+	} else if (access (name, F_OK) == 0) {
+		/* plain name exists in current directory */
+		snprintf (resolved, resolved_size, "%s", name);
```
GitMensch, line 989:
```suggestion
		resolved[0] = 0; /* unset for the caller to use original name */
		return fopen (name, "r");
```
... actually: better combine both conditions (comments on the right of the condition) in a single one, so we have a single fopen for name

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+		/* plain name: look up in COB_CONFIG_DIR/<name>.conf */
+		snprintf (resolved, resolved_size, "%s%c%s.conf",
+			  cob_config_dir, SLASH_CHAR, name);
+		if (access (resolved, F_OK) != 0) {
+			return NULL;
+		}
+	}
+
+	fp = fopen (resolved, "r");
+	return fp;
```
GitMensch, line 1000:
```suggestion
	return fopen (resolved, "r");
```

then we don't need `fp` here at all... actually: better to move the `else` part of the code above out of that, so "if (...use_name...) [return fopen(name);}  then without an "explicit" else do the snprintf + fopen(resolved)

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+int
+cb_load_preparser_conf (const char *name)
+{
+	FILE				*fp;
+	char				resolved[COB_NORMAL_BUFF];
+	char				buff[COB_SMALL_BUFF];
+	struct cb_preparser_entry	*pe;
+
+	fp = open_preparser_conf (name, resolved, sizeof (resolved));
+	if (!fp) {
```
GitMensch, line 1014:
```suggestion
	fp = open_preparser_conf (name, resolved, sizeof (resolved));
	if (resolved[0]) name = resolved;
	if (!fp) {
```

... and use "name" instead of resolved everywhere below

cobc/cobc.c
```diff
@@ -3654,6 +3667,10 @@ process_command_line (const int argc, char **argv)
 
 		case '$':
 			/* -std=<xx> : Specify dialect */
+		case CB_FLAG_GETOPT_PREPARSER:
+			/* registered in first pass */
+			break;
+
```
GitMensch, line 3673:
```suggestion
		case '$':
				/* -std=<xx> : Specify dialect */
		case CB_FLAG_GETOPT_PREPARSER:
			/* --preparser : register external preparser config */
```

to directly go to the single line "these options were all" + break (only once here)

cobc/cobc.c
```diff
@@ -5460,6 +5481,48 @@ preprocess (struct filename *fn)
...
 	plex_call_destroy ();
 
+	if (cb_active_preparser) {
+		char *cmd;
+		size_t cmd_len;
+		int ret_sys;
+		const char *new_preprocess;
+
+		cmd_len = strlen (cb_active_preparser->command) + strlen (fn->source) + strlen (fn->preprocess) + 3;
+		cmd = cobc_malloc (cmd_len);
```
GitMensch, line 5491: move those two line directly to the declaration (making cmd_len a `const` and reorder it)

cobc/cobc.c
```diff
@@ -5460,6 +5481,48 @@ preprocess (struct filename *fn)
 	/* Release flex buffers - After file close */
 	plex_call_destroy ();
 
+	if (cb_active_preparser) {
+		char *cmd;
+		size_t cmd_len;
+		int ret_sys;
+		const char *new_preprocess;
```
GitMensch, line 5488: drop that var, assigning it directly to fn->preprocess

cobc/cobc.c
```diff
@@ -5460,6 +5481,48 @@ preprocess (struct filename *fn)
...
+		snprintf (cmd, cmd_len, "%s %s %s", cb_active_preparser->command, fn->source, fn->preprocess);
+
+		ret_sys = call_system (cmd);
+		cobc_free (cmd);
+
+		if (ret_sys != 0) {
+			cb_source_file = fn->source;
+			if (!cb_active_preparser->warn_only) {
+				cobc_err_exit (_("external preparser '%s' failed with exit status %d"), cb_active_preparser->subsystem, ret_sys);
+			} else {
```
GitMensch, line 5501: no need for the else branch, the err_exit stops the program so everything in an else can be directly placed in the code after the closing }

cobc/cobc.c
```diff
@@ -5460,6 +5481,48 @@ preprocess (struct filename *fn)
...
+				cb_warning (cb_warn_unsupported,_("external preparser '%s' failed; falling back to baseline"),cb_active_preparser->subsystem);
+				cb_active_preparser->disabled = 1;         /* Don't try this subsystem again for this file */
+				fn->source = orig_source;     /* Restore the original file names */
+				fn->preprocess = orig_preprocess;
```
GitMensch, line 5505: Not checked, just wondering: have we overridden those vars at all? Aren't they still fine?

cobc/cobc.c
```diff
@@ -5460,6 +5481,48 @@ preprocess (struct filename *fn)
...
+
+		fn->source     = cobc_strdup (fn->preprocess);   /* sqlpp.sh's output */
+		fn->preprocess = new_preprocess;                  /* fresh file for pass 2 */
+
+		if (cb_active_preparser->cflags) {
+			COBC_ADD_STR (cobc_cflags, " ", cb_active_preparser->cflags, NULL);
+		}
+		if (cb_active_preparser->ldflags) {
+			COBC_ADD_STR (cobc_ldflags, " ", cb_active_preparser->ldflags, NULL);
+		}
```
GitMensch, line 5521: this can be the `else` branch for the failure above... actually better check `if (ret_sys )= 0) {` and have that in the first branch, the hopefully less likely error case in the else - and finally the goto outside of both

GitMensch (changes requested):
The overall tests and code look good - have you already run it sucessfully with the gixsql examples running gixsql?

If yes, please add the sample configuration for this (and possibly others like esqloc and esql) to the config folder.

If not: is there anything missing but the testing?


There are 3 main issues where I've asked (or suggested) for changes, I hope they are explained well enough - if not, feel free to ask here directly.

cobc/config.c
```diff
@@ -939,3 +913,173 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+		return 1;
+	}
+	/* find separator */
+	s = strpbrk (buff, " \t:=");
+	if (!s) {
+		/* strip trailing CR/LF for error message */
+		e = buff + strlen (buff);
+		while (e > buff && (e[-1] == '\r' || e[-1] == '\n')) {
+			*--e = 0;
+		}
```
GitMensch, line 941: That's slightly confusing - Is there a reason to not subtract 1 when computing e and checking e directly instead of e[-1}?
You can then also just do `e--;` in the loop and end with a final `*(e + 1) = 0;` before the return.

... just wondering.

utam-1: My first thoughts were to use something cleaner and readable as you suggested, but from what I gathered the subtract 1 rule results in an **undefined behaviour**  ( when `buff` is empty ) as that'll make e point to an address before the start of array. Modern architectures do not take this thing into account but maybe for stricter architectures it may result in issues.
Be as it may, I have updated the code, as per your suggestions.

cobc/pplex.l
utam-1: Both the above two guards ```if (cb_active_preparser) {
		return;
	} ```
and ``` #define pperror(x)	do { if (!cb_active_preparser) cb_error_always ("%s", x); } while(0) ``` are redundant now. They were placed because without the `SUBSYSTEM_FOUND` token approach you suggested Bison was generating fake error related to `EOF` because of the `return 0` we had ( which was supposed to be an artificial `EOF`).

cobc/pplex.l
utam-1: Based on my research, this may break the behaviour of lexer. So I believe it shouldn't be disturbed.

utam-1:
> have you already run it sucessfully with the gixsql examples running gixsql?
> 
> If yes, please add the sample configuration for this (and possibly others like esqloc and esql) to the config folder.
> 
> If not: is there anything missing but the testing?

@GitMensch I have performed a crude test with GixSQL, but I believe since it wasn't rigorous enough I didn't added the config file. I am describing what I have found below:
The config file I am using looks something like this :
```
subsystem:  SQL
command:    /tmp/gixpp-wrapper.sh
on-error:   error

```
I have used the following command to invoke cobc --preparser :
```
cobc --preparser /tmp/sql-gixsql.conf \
     --save-temps -x \
     -I /home/user/gixsql-1.0.20b/examples \
     -I /home/user/gixsql-1.0.20b/copy \
     /home/user/gixsql-1.0.20b/examples/TSQL001A.cbl \
     -o /tmp/TSQL001A
```
And from this the output I got was : 
```
GIXSQL*    EXEC SQL
GIXSQL*       CONNECT TO :DATASRC USER :DBUSR USING :DBPWD
GIXSQL*    END-EXEC.      
GIXSQL     CALL "GIXSQLConnect" USING
GIXSQL         BY REFERENCE SQLCA
GIXSQL         BY REFERENCE DATASRC
GIXSQL         BY VALUE 64
GIXSQL         BY REFERENCE x"00"
GIXSQL         BY VALUE 0
GIXSQL         BY REFERENCE x"00"
GIXSQL         BY VALUE 0
GIXSQL         BY REFERENCE DBUSR
GIXSQL         BY VALUE 64
GIXSQL         BY REFERENCE DBPWD
GIXSQL         BY VALUE 64
GIXSQL     END-CALL.
```
I think the translation works, however, if possible, I think it would be better if you would perform more rigorous testing from your end.

Apart from the testing, I don't recall there is anything else, as we have implemented almost everything based on the plan we had outlined.

cobc/config.c
GitMensch: > from what I gathered the subtract 1 rule results in an **undefined behaviour** ( when `buff` is empty )

Just interested to follow your thought: How can that be with the previous code (skipping empty lines, and the separator not found)?

cobc/cobc.c
```diff
@@ -3654,6 +3667,11 @@ process_command_line (const int argc, char **argv)
 
 		case '$':
 			/* -std=<xx> : Specify dialect */
+		case CB_FLAG_GETOPT_PREPARSER:
+			/* --preparser : register external preparser config */
+			/* These options were all processed in the first getopt-run */
+			break;
```
GitMensch, line 3670:
```suggestion
```

cobc/cobc.h
GitMensch: still open

cobc/config.c
```diff
@@ -488,52 +487,91 @@ cb_load_words (void)
...
+	}
+	/* find separator */
+	s = strpbrk (buff, " \t:=");
+	if (!s) {
+		/* strip trailing CR/LF for error message */
+		e = buff + strlen (buff) - 1;
+		while (e >= buff && (*e == '\r' || *e == '\n')) {
+			e--;
+		}
+		*(e + 1) = 0;
```
GitMensch, line 515: interesting side question: what happens if the file has no trailing CR/LF and we get here?

cobc/config.c
```diff
@@ -939,3 +961,134 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+	pe->disabled  = 0;
+	pe->next      = NULL;
+
+	while (fgets (buff, sizeof (buff), fp)) {
+		char *tag_name, *tag_val;
+		int	 split_ret;
+		line++;
+		split_ret = cb_conf_split_line (buff, &tag_name, &tag_val);
+		if (split_ret == 1) {
+			continue;
```
GitMensch:
```suggestion
			continue;		/* blank or comment line */
```

cobc/config.c
```diff
@@ -939,3 +961,134 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+		}
+		if (split_ret == -1) {
+			configuration_error (name, line, 1,
+				_("invalid configuration tag '%s'"), buff);
+			status = 1;
+			continue;
```
GitMensch, line 1037: according to the coverage check, this one misses a test case

cobc/config.c
```diff
@@ -939,3 +961,134 @@ cb_config_entry (char *buff, const char *fname, const int line)
...
+	}
+
+	/* plain name: look in COB_CONFIG_DIR/<name>.conf */
+	snprintf (resolved, resolved_size, "%s%c%s.conf",
+		  cob_config_dir, SLASH_CHAR, name);
+	if (access (resolved, F_OK) != 0) {
+		resolved[0] = 0;	/* unset: caller uses original name */
+		return NULL;
+	}
+	return fopen (resolved, "r");
```
GitMensch, line 993: this part is not covered in the tests, for the second just use a file that does not exists, for the first one temporarily set COB_CONFIG_DIR=tst (and place the config file in a sub-folder tst)

tests/testsuite.src/configuration.at
```diff
@@ -1064,3 +1064,68 @@ prog.cob:11: warning: start of statement in Area A
...
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog.
+       PROCEDURE DIVISION.
+           STOP RUN.
+])
+
+AT_CHECK([$COBC --preparser ./unknown.conf prog.cob], [1], [],
+[configuration error:
+./unknown.conf:4: unknown configuration tag 'badkey'
+])
```
GitMensch, line 1142:
```suggestion
AT_DATA([unknown.conf], [
subsystem: SQL

command: /bin/true # value does not matter, we check for the key below
                    #  (and here for inline parsing of comments)
badkey: something
])

AT_DATA([prog.cob], [
       IDENTIFICATION DIVISION.
       PROGRAM-ID. prog.
       PROCEDURE DIVISION.
           STOP RUN.
])

AT_CHECK([$COBC --preparser ./unknown.conf prog.cob], [1], [],
[configuration error:
./unknown.conf:6: unknown configuration tag 'badkey'
])
```

cobc/cobc.c
```diff
@@ -5460,6 +5480,41 @@ preprocess (struct filename *fn)
...
 
+	if (cb_active_preparser) {
+		int ret_sys;
+
+		const size_t	cmd_len = strlen (cb_active_preparser->command) + strlen (fn->source) + strlen (fn->preprocess) + 3;
+		char	*cmd = cobc_malloc (cmd_len);
+		snprintf (cmd, cmd_len, "%s %s %s", cb_active_preparser->command, fn->source, fn->preprocess);
+
+		ret_sys = call_system (cmd);
+		cobc_free (cmd);
```
GitMensch: I do wonder if/how we should pass include paths explicit given to cobc, maybe just as a third long argument that contains `-I path1 -I /path/two`?
That still allows to do "something else" in the preparser config if `-I` is not supported by the preparser.

cobc/scanner.l
```diff
@@ -107,6 +107,7 @@
 #define YYSTYPE			cb_tree
 #define _PARSER_H	/* work around bad Windows SDK header */
 #include "parser.h"
+#include <stdio.h>
```
GitMensch:
```suggestion
```

We don't need that as there is no new code, do we?

cobc/pplex.l
```diff
@@ -91,6 +91,7 @@ static void insert_copy_arg (void);
 #include "cobc.h"
 #include "tree.h"
 #include "ppparse.h"
+#include <stdio.h>
```
GitMensch: What do we need this for?

doc/gnucobol.texi
```diff
@@ -1949,6 +2004,36 @@ The number of calls to this section/paragraph
...
+@cindex EXEC blocks
+@cindex Preprocessing EXEC
+
+GnuCOBOL supports @code{EXEC} blocks (such as @code{EXEC SQL} or @code{EXEC CICS}) 
+through the use of an internal preprocessor and optional external preparsers.
+
+Note that @code{INCLUDE} statements (e.g., @code{EXEC SQL INCLUDE ...}) are 
+always handled by the internal preprocessor as a standard @code{COPY} statement, 
+with full support for library paths (@option{-I}) and case folding, before any 
+external preparser is considered.
```
GitMensch: I _think_ that does only apply if no preparser is found:

If a preparser is configured correctly then we would get EXEC (EXEC_STATE) WORD + SUBSYSTEM_FOUND = YYACCEPT + restart --> then the configured EXEC preparser would already see and replace INCLUDE, so COBOL would never see that in the second process, no?

GitMensch (changes requested):
please attach your LLM findings about nested preprocessor issues to this PR

Have you tried it locally (for example with a pseudo EXEC CICS which just replaces CICS by SQL and your GixSQL test configuration)?

cobc/config.c
utam-1: You're right the previous lines do ensure that buff doesn't stay empty, but then again there are issues with aggressive compiler optimization if we're using `Undefined Behaviour` code, as they're picked up as warnings of some sort in old school systems, apart from that if someone includes a `static analysis` tools do tend to flag such codes if they're present. I personally would not vouch for encountering issues with compiler optimization when using such kind of code, however I have witnessed issues with static analysis tools complaining if we're moving away from pre-defined standards.
This is nitpicking in my humble opinion, the code as such right now wouldn't create any issues.

cobc/config.c
utam-1: That wouldn't be in an issue I think, the loop would exit and the null terminator would be overwritten. The string wouldn't change.

utam-1:
> please attach your LLM findings about nested preprocessor issues to this PR
> 
> Have you tried it locally (for example with a pseudo EXEC CICS which just replaces CICS by SQL and your GixSQL test configuration)?

@GitMensch I have added a folder called `test_preparser` which details the findings. I have tested this locally. The issue was with name collision in `file_replace_extension` which seems to have been resolved.

utam-1:
Hello @GitMensch and @ddeclerck, I hope you are doing well. I was looking at the implementation plan that we prepared and noticed that we do have some **stretch goals** that are left, I am listing them out here for more clarity and along with that I have a few questions associated with them : 

- CLI override of on-error : Currently we're able to control this with the help of `.conf` file, and most probably we'd be using it on to test the behaviour in separate environment ( say prod vs dev ) without the need of another `.conf` file. So my question is do we really need something like this in the current implementation? The compiler as - is, already has numerous flags and adding this may require managing ordering dependency.
- Source diagnostics attribution: This is something that is worth looking into I think, making compiler point to `.pco` file ( and attribute errors to original source file and line numbers).

Hopefully, the current changes are as expected, and if not do let me know.

Looking forward to hearing from you!

PR #296, opened 2026-07-07
Add COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE runtime options to enable redirection of stdout/stderr output

oguzcankirmemis:
This PR adds runtime options `COB_SET_RUNTIME_STDOUT_FILE` and `COB_SET_RUNTIME_STDERR_FILE` which allows control over the location where the stdout/stderr output is written.

# Motivation

For modules written in C that interact with libcob, it is currently difficult to change the location of stdout/stderr output without disturbing the stdout/stderr of the whole process (i.e., other threads). This is especially useful for thread applications, where it is desired that each thread gets its own stdout/stderr without disturbing the one another.

With these options, the main program can easily redirect the output of the COBOL program (including the messages from the runtime itself) and decide what to do with it. For that reason, the main program is also responsible for managing the lifetime of the passed stdout/stderr files, should they be used.

# Implementation

- Add `COB_SET_RUNTIME_STDOUT_FILE` and `COB_SET_RUNTIME_STDERR_FILE` options to `cob_runtime_option_switch` in `common.h`.
- Extend the `cob_settings` in `coblocal.h` to store the location of configured `stdout/stderr`. By default, they are initialized to stdout/stderr of the process.
- Add support for both options in `cob_set_runtime_option` and `cob_get_runtime_option` in `common.c`.
- Adjust all print statements in across all various files in the runtime, i.e., under `./libcob`, so that they behave accordingly.
- Document both options in `doc/gnucobol.texi`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -55,10 +55,9 @@ NEWS - user visible changes				-*- outline -*-
 
 ** new runtime configurations: COB_STDIN_FILE, COB_STDOUT_FILE,
    COB_STDERR_FILE. They allow redirection of stdin/stdout/stderr input/output
-   to a file. Alternatively, these configurations are also available as runtime
-   options which can be used to set them programmatically:
-   COB_SET_RUNTIME_STDIN_FILE, COB_SET_RUNTIME_STDOUT_FILE,
-   COB_SET_RUNTIME_STDERR_FILE.
+   to a file. These configurations can be set via configuration file,
+   environment variables or programmatically with COB_SET_RUNTIME_STDIN_FILE,
+   COB_SET_RUNTIME_STDOUT_FILE, COB_SET_RUNTIME_STDERR_FILE.
 
 ** added multiple window functionality with new system function CBL_GC_WINDOW
 
--- a/NEWS
+++ b/NEWS
@@ -53,9 +53,12 @@ NEWS - user visible changes				-*- outline -*-
    already have a registered handler, allowing users to preserve their own
    signal handlers during the initialization of the GnuCOBOL runtime.
 
-** new runtime options COB_SET_RUNTIME_STDOUT_FILE/COB_SET_RUNTIME_STDERR_FILE,
-   allows redirection of stdout/stderr output to a file from a module written
-   in C.
+** new runtime configurations: COB_STDIN_FILE, COB_STDOUT_FILE,
+   COB_STDERR_FILE. They allow redirection of stdin/stdout/stderr input/output
+   to a file. Alternatively, these configurations are also available as runtime
+   options which can be used to set them programmatically:
+   COB_SET_RUNTIME_STDIN_FILE, COB_SET_RUNTIME_STDOUT_FILE,
+   COB_SET_RUNTIME_STDERR_FILE.
 
 ** added multiple window functionality with new system function CBL_GC_WINDOW
 
--- a/NEWS
+++ b/NEWS
@@ -53,6 +53,10 @@ NEWS - user visible changes				-*- outline -*-
    already have a registered handler, allowing users to preserve their own
    signal handlers during the initialization of the GnuCOBOL runtime.
 
+** new runtime options COB_SET_RUNTIME_STDOUT_FILE/COB_SET_RUNTIME_STDERR_FILE,
+   allows redirection of stdout/stderr output to a file from a module written
+   in C.
+
 ** added multiple window functionality with new system function CBL_GC_WINDOW
 
 ** New system functions
--- a/config/ChangeLog
+++ b/config/ChangeLog
@@ -1,4 +1,10 @@
 
+2026-07-07	Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* runtime.cfg: add runtime configurations COB_STDIN_FILE, COB_STDOUT_FILE,
+	  COB_STDERR_FILE to allow redirection of stdin/stdout/stderr input/output
+	  to given files
+
 2025-11-19  Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
 
 	* runtime.cfg: add runtime configuration COB_SIGNAL_REGIME to allow
--- a/config/ChangeLog
+++ b/config/ChangeLog
@@ -832,7 +832,7 @@
 	* default.inc, Makefile.am: New files.
 
 
-Copyright 2003,2005-2007-2010,2014-2024 Free Software Foundation, Inc.
+Copyright 2003,2005-2007-2010,2014-2026 Free Software Foundation, Inc.
 
 Copying and distribution of this file, with or without modification, are
 permitted provided the copyright notice and this notice are preserved.
--- a/config/runtime.cfg
+++ b/config/runtime.cfg
@@ -118,6 +118,30 @@
 #          Default:  false
 #          Example:  COB_SET_DEBUG  1
 
+# Environment name:  COB_STDIN_FILE
+#   Parameter name:  stdin_file
+#          Purpose:  to define where COBOL stdin input should be read from
+#             Type:  string (file) ; may use $-sequences
+#             Note:  file is opened for read-only
+#          Default:  stdin
+#          Example:  STDIN_FILE  ${HOME}/stdin.$$
+
+# Environment name:  COB_STDOUT_FILE
+#   Parameter name:  stdout_file
+#          Purpose:  to define where COBOL stdout output should go
+#             Type:  string (file) ; may use $-sequences
+#             Note:  file is opened for append if name starts with "+"
+#          Default:  stdout
+#          Example:  STDOUT_FILE  ${HOME}/stdout.$$
+
+# Environment name:  COB_STDERR_FILE
+#   Parameter name:  stderr_file
+#          Purpose:  to define where COBOL stderr output should go
+#             Type:  string (file) ; may use $-sequences
+#             Note:  file is opened for append if name starts with "+"
+#          Default:  stderr
+#          Example:  STDERR_FILE  ${HOME}/stderr.$$
+
 # Environment name:  COB_SET_TRACE
 #   Parameter name:  set_trace
 #          Purpose:  to enable COBOL trace feature
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,4 +1,9 @@
 
+2026-07-07  Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* gnucobol.texi: document the new COB_SET_RUNTIME_STDOUT_FILE and
+	  COB_SET_RUNTIME_STDERR_FILE runtime options.
+
 2025-03-26  David Declerck <david.declerck@ocamlpro.com>
 
 	* gnucobol.texi: document the new --gentable option
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -336,7 +336,7 @@
 	* open-cobol.texi (Customize): Add COB_LDADD, COB_CONFIG_FILE.
 
 
-Copyright 2002,2005,2009-2025 Free Software Foundation, Inc.
+Copyright 2002,2005,2009-2026 Free Software Foundation, Inc.
 
 Copying and distribution of this file, with or without modification, are
 permitted provided the copyright notice and this notice are preserved.
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,8 +1,9 @@
 
 2026-07-07  Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
 
-	* gnucobol.texi: document the new COB_SET_RUNTIME_STDOUT_FILE and
-	  COB_SET_RUNTIME_STDERR_FILE runtime options.
+	* gnucobol.texi: document the new COB_SET_RUNTIME_STDIN_FILE,
+	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE runtime
+	  options
 
 2025-03-26  David Declerck <david.declerck@ocamlpro.com>
 
--- a/doc/gnucobol.texi
+++ b/doc/gnucobol.texi
@@ -1657,6 +1657,7 @@ enum cob_runtime_option_switch @{
    COB_SET_RUNTIME_DISPLAY_PRINTER_FILE    /* 'p' is  FILE *  */
    COB_SET_RUNTIME_RESCAN_ENV              /* rescan environment variables */
    COB_SET_RUNTIME_DISPLAY_PUNCH_FILE      /* 'p' is  FILE *  */
+   COB_SET_RUNTIME_STDIN_FILE              /* 'p' is  FILE *  */
    COB_SET_RUNTIME_STDOUT_FILE             /* 'p' is  FILE *  */
    COB_SET_RUNTIME_STDERR_FILE             /* 'p' is  FILE *  */
 @};
--- a/doc/gnucobol.texi
+++ b/doc/gnucobol.texi
@@ -1657,6 +1657,8 @@ enum cob_runtime_option_switch @{
    COB_SET_RUNTIME_DISPLAY_PRINTER_FILE    /* 'p' is  FILE *  */
    COB_SET_RUNTIME_RESCAN_ENV              /* rescan environment variables */
    COB_SET_RUNTIME_DISPLAY_PUNCH_FILE      /* 'p' is  FILE *  */
+   COB_SET_RUNTIME_STDOUT_FILE             /* 'p' is  FILE *  */
+   COB_SET_RUNTIME_STDERR_FILE             /* 'p' is  FILE *  */
 @};
 COB_EXT_IMPORT  void  cob_set_runtime_option   (enum cob_runtime_option_switch opt, void *p);
 @end smallexample
@@ -1676,6 +1678,20 @@ cob_set_runtime_option (COB_SET_RUNTIME_DISPLAY_PRINTER_FILE,
 You could also redirect all @code{DISPLAY UPON SYSPUNCH} output to a file by:
 @example
 cob_set_runtime_option (COB_SET_RUNTIME_DISPLAY_PUNCH_FILE,
+                         (void*)((FILE*)myfd));
+@end example
+
+Likewise, you could also redirect all @code{DISPLAY} output that would be normally
+printed to stdout to a file by:
+@example
+cob_set_runtime_option (COB_SET_RUNTIME_STDOUT_FILE,
+                         (void*)((FILE*)myfd));
+@end example
+
+The same option is also possible for stderr, including error outputs directly
+coming from the runtime:
+@example
+cob_set_runtime_option (COB_SET_RUNTIME_STDERR_FILE,
                          (void*)((FILE*)myfd));
 @end example
 
--- a/doc/gnucobol.texi
+++ b/doc/gnucobol.texi
@@ -1681,16 +1681,19 @@ cob_set_runtime_option (COB_SET_RUNTIME_DISPLAY_PUNCH_FILE,
                          (void*)((FILE*)myfd));
 @end example
 
-Likewise, you could also redirect all @code{DISPLAY} output that would be normally
-printed to stdout to a file by:
+You are also able to redirect @code{ACCEPT} so that instead of stdin, it reads
+from a file, by using:
 @example
-cob_set_runtime_option (COB_SET_RUNTIME_STDOUT_FILE,
+cob_set_runtime_option (COB_SET_RUNTIME_STDIN_FILE,
                          (void*)((FILE*)myfd));
 @end example
 
-The same option is also possible for stderr, including error outputs directly
-coming from the runtime:
+Likewise, you could also redirect all @code{DISPLAY} output that would be
+normally printed to stdout/stderr, including outputs directly coming from the
+runtime, to a file by:
 @example
+cob_set_runtime_option (COB_SET_RUNTIME_STDOUT_FILE,
+                         (void*)((FILE*)myfd));
 cob_set_runtime_option (COB_SET_RUNTIME_STDERR_FILE,
                          (void*)((FILE*)myfd));
 @end example
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -42,11 +42,9 @@
 
 	* common.c: add COB_STDIN_FILE, COB_STDOUT_FILE and COB_STDERR_FILE runtime
 	  configurations for redirection of stdin/stdout/stderr to a file
-	* common.h (cob_runtime_option_switch): add COB_SET_RUNTIME_STDIN_FILE,
-	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE enums
-	* common.c (cob_set_runtime_option, cob_get_runtime_option): extend for
-	  COB_SET_RUNTIME_STDIN_FILE, COB_SET_RUNTIME_STDOUT_FILE and
-	  COB_SET_RUNTIME_STDERR_FILE options
+	* common.c, common.h: add COB_SET_RUNTIME_STDIN_FILE,
+	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE runtime
+	  options
 	* coblocal.h (cob_settings): extend the setting to store FILE pointers
 	  for stdin/stdout/stderr
 	* fileio.c: introduce cob_settings_fileio to trigger reset of settings
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -43,9 +43,8 @@
 	* common.c: add COB_STDIN_FILE, COB_STDOUT_FILE and COB_STDERR_FILE runtime
 	  configurations for redirection of stdin/stdout/stderr to a file
 	* common.h (cob_runtime_option_switch): add COB_SET_RUNTIME_STDIN_FILE,
-	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE runtime
-	  options to enable redirection programmatically
-	* common.c (cob_set_runtime_option, cob_get_runtime_option): add logic for
+	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE enums
+	* common.c (cob_set_runtime_option, cob_get_runtime_option): extend for
 	  COB_SET_RUNTIME_STDIN_FILE, COB_SET_RUNTIME_STDOUT_FILE and
 	  COB_SET_RUNTIME_STDERR_FILE options
 	* coblocal.h (cob_settings): extend the setting to store FILE pointers
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -38,6 +38,18 @@
 	  events across END-OF-INPUT boundaries. 
 	  This is due to libxml2 internal details.
 
+2026-07-07  Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* common.h (cob_runtime_option_switch): add COB_SET_RUNTIME_STDOUT_FILE and
+	  COB_SET_RUNTIME_STDERR_FILE runtime options to enable redirection of
+	  stdout/stderr output to a file.
+	* common.c (cob_set_runtime_option, cob_get_runtime_option): add logic for
+	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE options.
+	* coblocal.h (cob_settings): extend the setting to store FILE pointers
+	  for stdout/stderr.
+	* common.c, cobgetopt.c, coblocal.h, fileio.c, profiling.c, screenio.c,
+	  termio.c: use stdout/stderr in cob_settings instead of global.
+
 2026-03-02  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
 	* coblocal.h, common.c, profiling.c: rename is_test to cob_is_test
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -40,15 +40,20 @@
 
 2026-07-07  Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
 
-	* common.h (cob_runtime_option_switch): add COB_SET_RUNTIME_STDOUT_FILE and
-	  COB_SET_RUNTIME_STDERR_FILE runtime options to enable redirection of
-	  stdout/stderr output to a file.
+	* common.c: add COB_STDIN_FILE, COB_STDOUT_FILE and COB_STDERR_FILE runtime
+	  configurations for redirection of stdin/stdout/stderr to a file
+	* common.h (cob_runtime_option_switch): add COB_SET_RUNTIME_STDIN_FILE,
+	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE runtime
+	  options to enable redirection programmatically
 	* common.c (cob_set_runtime_option, cob_get_runtime_option): add logic for
-	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE options.
+	  COB_SET_RUNTIME_STDIN_FILE, COB_SET_RUNTIME_STDOUT_FILE and
+	  COB_SET_RUNTIME_STDERR_FILE options
 	* coblocal.h (cob_settings): extend the setting to store FILE pointers
-	  for stdout/stderr.
+	  for stdin/stdout/stderr
+	* fileio.c: introduce cob_settings_fileio to trigger reset of settings
+	  related configurations
 	* common.c, cobgetopt.c, coblocal.h, fileio.c, profiling.c, screenio.c,
-	  termio.c: use stdout/stderr in cob_settings instead of global.
+	  termio.c: use stdin/stdout/stderr in cob_settings instead of global
 
 2026-03-02  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
--- a/libcob/cobgetopt.c
+++ b/libcob/cobgetopt.c
@@ -249,6 +249,8 @@ process_long_option (const int argc, char * const *argv, const char *optstring,
 		     const int long_only,
 		     int print_errors, const char *prefix)
 {
+  FILE *stderr_f = cob_get_runtime_option (COB_SET_RUNTIME_STDERR_FILE) ?
+	cob_get_runtime_option (COB_SET_RUNTIME_STDERR_FILE) : stderr;
   char *nameend;
   size_t namelen;
   const struct option *p;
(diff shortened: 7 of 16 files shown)


oguzcankirmemis:
@GitMensch Here is the promised PR for stdout/stderr redirection. I opened it in draft for now to clarify a small issue in `libcob/cobgetopt.c`: Since the `cobsetptr` is currently not exposed in this file, my naive way of doing the redirection is not possible without further adjustment. How should we proceed here?

I also feel like implementing the redirection in some places seemed unnecessary to me, with the thought that they were meant for direct usage over the command line, rather than a C module interacting with GnuCOBOL. I would appreciate if you could check in general where it makes sense, or whether there are still places missing, where we need to the redirection.

As always, always glad and available for further questions and discussions. Thanks for the support!

oguzcankirmemis:
I can confirm that the failing tests in CI (should be seven in total) are related to the PR, will investigate.

oguzcankirmemis:
All tests should be fixed now, the issue was `cob_runtime_warning`, `cob_runtime_error` and `cob_runtime_hint`  can be called from other modules (like cobc) without initializing the `cobsetptr` (i.e. `cob_settings`) in `common.c`. Upon looking closer, I also discovered that there is also `cob_runtime_warning_ss` for the signal handler, that directly uses the file descriptor of stderr. I believe, we can't use the `FILE` pointer there since `fileno` would not be signal handler safe.. I am leaving as it is for now, is this fine?

oguzcankirmemis:
Not sure what is going on with Windows CI tests, and the vbisam one, but I believe it is unrelated?

libcob/common.c
```diff
@@ -18,6 +18,7 @@
    along with GnuCOBOL.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+#include "common.h"
```
GitMensch: this is included in line 219 and should only be after the system headers - and that's likely the reason for the MSYS2 errors - just drop this to fix it

libcob/fileio.c
```diff
@@ -3634,9 +3634,9 @@ join_environment (void)
 	bdb_env->set_msgcall (bdb_env, bdb_msgcall_set);
 #endif
 #else
-	bdb_env->set_errfile (bdb_env, stderr);
+	bdb_env->set_errfile (bdb_env, cobsetptr->cob_stderr);
 #if (DB_VERSION_MAJOR > 4) || ((DB_VERSION_MAJOR == 4) && (DB_VERSION_MINOR > 2))
-	bdb_env->set_msgfile (bdb_env, stderr);
+	bdb_env->set_msgfile (bdb_env, cobsetptr->cob_stderr);
```
GitMensch: Should this be re-set if the current thread changes the file* ?

libcob/common.c
```diff
@@ -10582,6 +10601,16 @@ cob_set_runtime_option (enum cob_runtime_option_switch opt, void *p)
...
+	case COB_SET_RUNTIME_STDOUT_FILE:
+		if (p) {
+			cobsetptr->cob_stdout = (FILE *)p;
+		}
+	    break;
+	case COB_SET_RUNTIME_STDERR_FILE:
+		if (p) {
+			cobsetptr->cob_stderr = (FILE *)p;
+		}
+		break;
```
GitMensch, line 10643: Should this be set to stdout/stderr if p is empty (not sure)?
Should there be a flush to the previous stdout/stdin?

tests/testsuite.src/run_misc.at
```diff
@@ -16107,3 +16107,119 @@ callee.cob:8: attempt to reference invalid memory address (signal SIGSEGV)
 ])
 
 AT_CLEANUP
+
+AT_SETUP([COB_SET_RUNTIME_STDOUT_FILE/COB_SET_RUNTIME_STDERR_FILE changes runtime stdout/stderr])
```
GitMensch: please two empty lines before a new test (AT_SETUP) and an empty line after the final AT_CLEANUP

GitMensch: this name is likely too long, move the enum names down instead of the long last one, renaming the test

GitMensch (changes requested):
Please also add those two as "normal" runtime settings `COB_STDOUT_FILE` and `COB_STDERR_FILE` with the common file handling (like + opening the file as append), this allows users to both set the filename within a COBOL thread via `SET ENVIRONMENT` as well as easily having one per PID (which would be your example).

Not sure, but... would it be reasonable to redirect stdin the same way?

libcob/common.c
oguzcankirmemis: For some reason my LSP/IDE insists on adding this include every PR and I forget to delete it. Thanks for pointing it out!

libcob/common.c
oguzcankirmemis: 'Should this be set to stdout/stderr if p is empty?' --> I would say it should give a warning that says it is defaulting back to stdout/stderr in that case. I would expect the programmer to pass /dev/null or something equivalent if they want to ignore the stdout/stderr output.

'Should there be a flush to the previous stdout/stdin?' --> That might be a bad idea. What happens if the programmer closes the stream before setting the next file stream? (I realize this is how we are doing it in JNI at the moment, but we can ofc adjust it too.)

oguzcankirmemis:
> Please also add those two as "normal" runtime settings `COB_STDOUT_FILE` and `COB_STDERR_FILE` with the common file handling (like + opening the file as append), this allows users to both set the filename within a COBOL thread via `SET ENVIRONMENT` as well as easily having one per PID (which would be your example).
> 
> Not sure, but... would it be reasonable to redirect stdin the same way?


For stdin, I don't have a clue whether it would be useful, so you can decide. I think it wouldn't be too much of an effort.

For adding them as normal runtime settings, how does "SET ENVIRONMENT" work right now for IO files? I thought GnuCOBOL already implements changing the stdout/stderr via `SET ENVIRONMENT` inside COBOL, that is why I am a little bit confused. I also think if COBOL is the main process, it is actually relatively easy to redirect stdout/stderr `2> /tmp/stderr-location` etc., at least in a conventional OS/Shell, so I don't see the benefit there. What am I missing?

libcob/fileio.c
oguzcankirmemis: How would I trigger the refresh/re-set in `cob_set_runtime_option` in that case, by simply adding a non-static getter for bdb_env?

GitMensch: the right way would be to do the same as in screenio - see cob_settings_screenio and its callers... but at least add one exported function into fileio.c that you can call when this specific runtime setting changed

libcob/common.c
GitMensch: Closing before setting seems a bit strange - but I agree that we should not flush, the "owner" of the stream  may do that, if there is no such owner any more the C runtime library or OS may flush; otherwise the data is lost, but that that will only be display data, so ... better not crash anything for that - no flush.

To "reset" correctly there would still be an option if we don't do anything: by first get the runtime option externally, store it and set later - that's enough, the warning you've added is also fine, so we can close this thread.

GitMensch:
If stdin is not a big effort, then please add it the same way, covering all internally used streams. We may even use it later for other tests.

For the `SET ENVIRONMENT` part (possibly more used with runtime configuration file): see cob_check_trace_file  - the setting would name a file and it would be internally opened as soon as the runtime setting is set.

libcob/common.c
```diff
@@ -9062,7 +9071,12 @@ void
 cob_runtime_error (const char *fmt, ...)
 {
 	struct handlerlist	*h;
-	va_list			ap;
+	va_list				ap;
+	FILE            	*stderr_f = stderr;
+
+	if (cobsetptr && cobsetptr->cob_stderr) {
+		stderr_f = cobsetptr->cob_stderr;
+	}
```
GitMensch: this is code and needs to be placed after line 9081 (the reason why the c89 compat ci fails)

GitMensch: Note: I'd suggest to make that a conditional within the assignment, making it a bit more clearer and needing no reordering:

```c
const FILE		*stderr_f = cobsetptr && cobsetptr->cob_stderr ? cobsetptr->cob_stderr : stderr;
```

(likely adding a line break somewhere)

tests/testsuite.src/run_misc.at
```diff
@@ -16107,3 +16107,120 @@ callee.cob:8: attempt to reference invalid memory address (signal SIGSEGV)
...
+}
+]])
+
+AT_CHECK([$COMPILE caller.c], [0], [], [])
+AT_CHECK([$COMPILE_MODULE callee.cob], [0], [], [])
+
+AT_CHECK([./caller], [0], [], [])
+AT_CHECK([./caller], [0], [], [])
+
+AT_CLEANUP
```
GitMensch, line 16273: minor: please add an empty line at the testsuite sources, that showed to be an easier diff with some tools.

GitMensch (changes requested):
there seems to be one missing element used when calling libcob from cobc, but only on Win32 - I guess that's the place where the binary mode is set (unix_lf), as that will only happen this way on win32

That and the stderr_f assignment inlined should fix CI; adding stdin redirection + reassignment for BDB would then finalize the changes.

If you feel that you can't get the part done soon that works on those changes and are likely more often used by "normal" users  (three matching runtime settings to assign the streams to a filename) then we can get this in with a follow-up PR or work by someone else.

tests/testsuite.src/run_misc.at
oguzcankirmemis: You want one more blank line right? I wonder why it didn't show up in the Github diffs if it was there before. Anyways it should be there now.

libcob/fileio.c
oguzcankirmemis: Done, let me know if it is missing something.

oguzcankirmemis:
> If stdin is not a big effort, then please add it the same way, covering all internally used streams. We may even use it later for other tests.
> 
> For the `SET ENVIRONMENT` part (possibly more used with runtime configuration file): see cob_check_trace_file - the setting would name a file and it would be internally opened as soon as the runtime setting is set.

> there seems to be one missing element used when calling libcob from cobc, but only on Win32 - I guess that's the place where the binary mode is set (unix_lf), as that will only happen this way on win32
> 
> That and the stderr_f assignment inlined should fix CI; adding stdin redirection + reassignment for BDB would then finalize the changes.
> 
> If you feel that you can't get the part done soon that works on those changes and are likely more often used by "normal" users (three matching runtime settings to assign the streams to a filename) then we can get this in with a follow-up PR or work by someone else.

~~I have added the stdin redirection as well, remaining parts are:~~

- ~~Runtime configuration files~~
- ~~The issue with Win32~~

Apart from TODOs I have commented on, some stylings and docs, all parts should be there now. I have changed the test slightly to make it Windows compatible, let's see if CI passes.

libcob/cobgetopt.c
```diff
@@ -326,6 +326,8 @@ process_long_option (const int argc, char * const *argv, const char *optstring,
 	    {
 	      if (ambig_fallback)
 		  {
+		  /* TODO: Decide how to do stdout/stderr redirection here
```
oguzcankirmemis, line 329: What do you think about here? Solve it similarly by adding an exported init function?

libcob/common.c
```diff
@@ -10412,6 +10417,11 @@ cob_init (const int argc, char **argv)
 	/* Get settings structure */
 	cobsetptr = cob_malloc (sizeof (cob_settings));
 
+	/* TODO: Find a better a place set stdout/stderr initial values */
```
oguzcankirmemis: Is initializing the file pointers here ok?

oguzcankirmemis:
Remaining failing CI tests should be hopefully unrelated.

oguzcankirmemis:
Updated docs/changelogs, will mark the PR non-draft, once all the todos are cleared. The one extra failed CI with Ubuntu+MSYS1 should actually pass (i.e. I haven't touched anything other than docs)? Instead, it got broken pipe in SIGPIPE test for some reason.

config/runtime.cfg
```diff
@@ -118,6 +118,30 @@
...
+#          Purpose:  to define where COBOL stdin input should be read from
+#             Type:  string (file) ; may use $-sequences
+#             Note:  file is opened for read-only
+#          Default:  stdin
+#          Example:  STDIN_FILE  ${HOME}/stdin.$$
+
+# Environment name:  COB_STDOUT_FILE
+#   Parameter name:  stdout_file
+#          Purpose:  to define where COBOL stdout output should go
+#             Type:  string (file) ; may use $-sequences
```
oguzcankirmemis, line 132: I copied this from COB_TRACE_FILE, I assume resolving $-sequences happen out-of-the-box during env rescan?

libcob/fileio.c
GitMensch: the only question here is shouldn't the code be _moved_ to the setting function (and that function either be called in the caller or inside this function)?

tests/ChangeLog
GitMensch, line 1: we only add to that Changelog if there are "general" changes - like moving tests around or adding new test files, so you can revert that part

NEWS
```diff
@@ -35,6 +35,10 @@ NEWS - user visible changes				-*- outline -*-
    already have a registered handler, allowing users to preserve their own
    signal handlers during the initialization of the GnuCOBOL runtime.
 
+** new runtime options COB_SET_RUNTIME_STDOUT_FILE/COB_SET_RUNTIME_STDERR_FILE,
```
GitMensch: stdin is missing here

config/runtime.cfg
GitMensch: yes, they do, as long as the setting is of type file

libcob/common.c
```diff
@@ -9178,6 +9186,13 @@ cob_fatal_error (const enum cob_fatal_error fatal_error)
 	int		status;
 #ifdef	_WIN32
 	char		*p;
+
+	FILE 		*stdin_f = cobsetptr && cobsetptr->cob_stdin
+		? cobsetptr->cob_stdin : stdin;
+	FILE 		*stdout_f = cobsetptr && cobsetptr->cob_stdout
+		? cobsetptr->cob_stdout : stdout;
+	FILE 		*stderr_f = cobsetptr && cobsetptr->cob_stderr
+		? cobsetptr->cob_stderr : stderr;
```
GitMensch: I'd prefer to have those declarations directly before the setmode (that's a new block in any case)

libcob/common.c
```diff
@@ -10234,6 +10249,14 @@ cob_common_init (void *setptr)
 		int use_unix_lf = 0;
 		char *s = getenv ("COB_UNIX_LF");
 
+		FILE *stdin_f = cobsetptr && cobsetptr->cob_stdin
+			? cobsetptr->cob_stdin : stdin;
+		FILE *stdout_f = cobsetptr && cobsetptr->cob_stdout
+			? cobsetptr->cob_stdout : stdout;
+		FILE *stderr_f = cobsetptr && cobsetptr->cob_stderr
+			? cobsetptr->cob_stderr : stderr;
```
GitMensch: same as above please move the declarations to the use_unix_lf block

libcob/ChangeLog
```diff
@@ -1,4 +1,21 @@
 
+2026-07-07  Oğuzcan Kırmemiş <oguzcan.kirmemis@gmail.com>
+
+	* common.c: add COB_STDIN_FILE, COB_STDOUT_FILE and COB_STDERR_FILE runtime
+	  configurations for redirection of stdin/stdout/stderr to a file
```
GitMensch, line 5: this should be added to NEWS, the second part (runtime options) is then only an alternative way to set those

doc/gnucobol.texi
```diff
@@ -1657,6 +1657,8 @@ enum cob_runtime_option_switch @{
    COB_SET_RUNTIME_DISPLAY_PRINTER_FILE    /* 'p' is  FILE *  */
    COB_SET_RUNTIME_RESCAN_ENV              /* rescan environment variables */
    COB_SET_RUNTIME_DISPLAY_PUNCH_FILE      /* 'p' is  FILE *  */
+   COB_SET_RUNTIME_STDOUT_FILE             /* 'p' is  FILE *  */
+   COB_SET_RUNTIME_STDERR_FILE             /* 'p' is  FILE *  */
```
GitMensch, line 1662: while dynamically extracting the value would be best - adding STDIN here is enough

libcob/common.c
GitMensch: ... those vars are not set here, are they?

libcob/common.c
GitMensch: yes, that's fine to do for ensuring that is set in any case

libcob/common.c
```diff
@@ -10467,6 +10495,39 @@ cob_init (const int argc, char **argv)
...
+	}
+
+	if (cobsetptr->cob_stderr_filename) {
+		cobsetptr->cob_stderr =
+			cob_open_logfile (cobsetptr->cob_stderr_filename);
+		if (!cobsetptr->cob_stderr) {
+			cobsetptr->cob_stderr_filename = NULL;
+			cobsetptr->cob_stderr = stderr;
+		}
+	}
```
GitMensch: that part needs to be moved (maybe to termio.c?) and called when one of those runtime settings are changed - and in that case close the old pointer, before opening it again (if the pointer changed and the content differs)

tests/testsuite.src/run_misc.at
```diff
@@ -16107,3 +16107,139 @@ callee.cob:8: attempt to reference invalid memory address (signal SIGSEGV)
...
+   FILE *cob_stdin;
+   FILE *cob_stdout;
+   FILE *cob_stderr;
+   char buffer[128];
+   char *fgets_ret;
+   ssize_t nbytes;
+   int cob_ret;
+
+   /* Create files for stdin/stdout/stderr */
+   cob_stdin = fopen ("cob_stdin", "w+b");
```
GitMensch: *optional*: put the fopen directly into the declaration, only doing the check here

GitMensch (changes requested):
ci failures unrelated, most important adjustment is the handling of the setting of the runtime option - the current version won't work if those come from `SET ENVIRONMENT` (explicit setting in the COBOL main instead of runtime config file / environment)

libcob/fileio.c
oguzcankirmemis: That makes sense, I guess I was trying to be as uninvasive as possible, hence redundant stuff.

oguzcankirmemis: I have moved the unused (probably deprecated) part of the ifdef to settings as well. My gut feeling was telling me they should stay together.

oguzcankirmemis:
> ci failures unrelated, most important adjustment is the handling of the setting of the runtime option - the current version won't work if those come from `SET ENVIRONMENT` (explicit setting in the COBOL main instead of runtime config file / environment)

Makes sense, I will take a look.

oguzcankirmemis:
I think SET ENVIRONMENT should work now, added a test for it. It is kind of messy how I do it in cobsetptr, i.e. having `cob_stdout_filename` and `cob_stdout_filename_set` both to remember the old setting in `termio.c`. I was against the idea of storing them as globals inside `termio.c`. Let me know if you think there is a better way to achieve the functionality.

There is also the `print_version` and `print_info` functions in `common.c`, that still print to process stdout. I haven't converted them, because I don't see the use case to be honest. But for completeness sake, if you want it, I can change them as well.

Otherwise, from my side all that is missing is what we should with `cobgetopt.c`. I will make a final pass for news/changelogs etc. once the other changes are finalized.

libcob/common.c
oguzcankirmemis: Moved it, see comment at the bottom.

libcob/common.c
oguzcankirmemis: I must have overlooked this one, fixed it now.

GitMensch:
@oguzcankirmemis please rebase, which should improve the CI results

oguzcankirmemis:
@GitMensch rebase done, I think the failing two tests in Windows are due to runtime mismatch between libcob and the running tests (similar to CBL_GC_HOSTED). The error doesn't make sense to me really otherwise.

Should we try to fix it, or ignore the test in windows debug builds?

You probably already looked into it, but do you know a fast way to experiment/verify with Windows builds, i.e. some ready to use image or machine I can connect to? Then, I can take a look to make the libraries use the same c runtime maybe. It should be helpful for future tests.

oguzcankirmemis:
@GitMensch I fixed the Windows CI issues with a little help of AI, i.e. it found the `cb_source_debugging`  and the `_DEBUG` macro for me. Surprisingly, fixing the issue there also fixes the signal regime tests. Apparently, the runtime abort before signal handler also was happening due to CRT mismatch. As suspected, the fix also fixes the CBL_GC_HOSTED test.

A minor consequence for now is that the debug builds via cobc always uses `/Od`, ignoring what the user is passing. This created a warning message in the `INITIALIZE big table with VALUE` test, where I had to delete the optimization flag `(-02)`. If you think having the flag there is necessary, we can add an if check for debug builds in cobc so that it does not pass `/0d`, if the user has supplied an optimization flag.

Let me know if there are other consequences we need to think about for the fix.

oguzcankirmemis:
For curiosity, I also asked AI to decouple CRT selection from the -g flag, effectively adding another variable on top of cb_source_debugging. Using cb_source_debugging adds more than the CRT flag as mentioned, hence the reason. I can push it if you find this idea cleaner.

GitMensch:
Yes, CRT selection should be independent from the optimization flags.

To not mix/match your MSVC CI fixes (and - very nice - remove some removement of failing tests) with the rest of this PR: please create a separate one (which then @ddeclerck can also merge here + upstream, while I inspect the "real PR = new feature").

GitMensch:
As the MSVC debug builds break in the single test for this PR.. there seems something is actually broken in the implementation, no?

~~~
901. run_misc.at:16112: testing stdin/stdout/stderr redirection ...
tests/run_misc.at:16266: $COMPILE prog.cob
tests/run_misc.at:16267: $COMPILE caller.c
tests/run_misc.at:16268: $COMPILE_MODULE callee.cob
tests/run_misc.at:16270: $COBCRUN_DIRECT ./prog
tests/run_misc.at:16271: ./caller
--- /dev/null	2026-08-09 15:02:56.000000000 +0000
+++ /d/a/gnucobol/gnucobol/tests/testsuite.dir/at-groups/901/stderr	2026-08-09 15:02:56.829262700 +0000
@@ -0,0 +1 @@
+minkernel\crts\ucrt\src\appcrt\lowio\read.cpp(381) : Assertion failed: _osfile(fh) & FOPEN
tests/run_misc.at:16271: exit code was 127, expected 0
901. run_misc.at:16112:  FAILED (run_misc.at:16271)
~~~

config/ChangeLog
GitMensch, line 1: change is fine - as the file has no other 2026 change: please adjust the copyright year in the footer of the file as welll

doc/ChangeLog
GitMensch, line 1: change is fine - as the file has no other 2026 change: please adjust the copyright year in the footer of the file as welll

libcob/ChangeLog
```diff
@@ -1,4 +1,20 @@
...
+	* common.h (cob_runtime_option_switch): add COB_SET_RUNTIME_STDIN_FILE,
+	  COB_SET_RUNTIME_STDOUT_FILE and COB_SET_RUNTIME_STDERR_FILE enums
+	* common.c (cob_set_runtime_option, cob_get_runtime_option): extend for
+	  COB_SET_RUNTIME_STDIN_FILE, COB_SET_RUNTIME_STDOUT_FILE and
+	  COB_SET_RUNTIME_STDERR_FILE options
```
GitMensch, line 10: you can combine those two into a single one

libcob/common.c
GitMensch, line 10660: please add a check to directly return NULL if cobsetptr is not available (= no init done yet)

libcob/cobgetopt.c
GitMensch: if that's the only place here: call into cob_get_runtime_option once to get the fp - if it is NULL use stderr

oguzcankirmemis:
> As the MSVC debug builds break in the single test for this PR.. there seems something is actually broken in the implementation, no?
> 
> ```
> 901. run_misc.at:16112: testing stdin/stdout/stderr redirection ...
> tests/run_misc.at:16266: $COMPILE prog.cob
> tests/run_misc.at:16267: $COMPILE caller.c
> tests/run_misc.at:16268: $COMPILE_MODULE callee.cob
> tests/run_misc.at:16270: $COBCRUN_DIRECT ./prog
> tests/run_misc.at:16271: ./caller
> --- /dev/null	2026-08-09 15:02:56.000000000 +0000
> +++ /d/a/gnucobol/gnucobol/tests/testsuite.dir/at-groups/901/stderr	2026-08-09 15:02:56.829262700 +0000
> @@ -0,0 +1 @@
> +minkernel\crts\ucrt\src\appcrt\lowio\read.cpp(381) : Assertion failed: _osfile(fh) & FOPEN
> tests/run_misc.at:16271: exit code was 127, expected 0
> 901. run_misc.at:16112:  FAILED (run_misc.at:16271)
> ```

I am confused, the CI fix also fixes this issue, no?

libcob/common.c
```diff
@@ -8311,6 +8314,11 @@ set_config_val (char *value, int pos)
 			cob_new_trace_file ();
 		}
 
+		if (data_loc == offsetof (cob_settings, cob_stdin_filename)
+			|| data_loc == offsetof (cob_settings, cob_stdout_filename)
+			|| data_loc == offsetof (cob_settings, cob_stderr_filename)) {
```
GitMensch, line 8319:
```suggestion
		if (data_loc == offsetof (cob_settings, cob_stdin_filename)
		 || data_loc == offsetof (cob_settings, cob_stdout_filename)
		 || data_loc == offsetof (cob_settings, cob_stderr_filename)) {
```

special GnuCOBOL formatting :-)

oguzcankirmemis:
> Yes, CRT selection should be independent from the optimization flags.
> 
> To not mix/match your MSVC CI fixes (and - very nice - remove some removement of failing tests) with the rest of this PR: please create a separate one (which then @ddeclerck can also merge here + upstream, while I inspect the "real PR = new feature").

Gotcha, I will revert the fix here once I opened the other PR.

GitMensch:
> I am confused, the CI fix also fixes this issue, no?

hm, now I see all passing... not sure what I've seen before, guess that was an error from "it is a file pointer, but not from this runtime"

libcob/termio.c
```diff
@@ -1098,9 +1098,113 @@ cob_accept (cob_field *f)
 	cob_move (&temp, f);
 }
 
+static void
+cob_stdin_termio (void)
+{
+	if (!cobsetptr->cob_stdin_filename) {
```
GitMensch, line 1104: Here and below: What happens if that points to an empty string?

What should happen if there's a previous file set and we now get an empty string (or a nullptr)?

libcob/common.c
```diff
@@ -9198,9 +9211,16 @@ cob_fatal_error (const enum cob_fatal_error fatal_error)
...
+			FILE *stdin_f = cobsetptr && cobsetptr->cob_stdin
+				? cobsetptr->cob_stdin : stdin;
+			FILE *stdout_f = cobsetptr && cobsetptr->cob_stdout
+				? cobsetptr->cob_stdout : stdout;
+			FILE *stderr_f = cobsetptr && cobsetptr->cob_stderr
+				? cobsetptr->cob_stderr : stderr;
+
+			(void)_setmode (_fileno (stdin_f), _O_BINARY);
+			(void)_setmode (_fileno (stdout_f), _O_BINARY);
+			(void)_setmode (_fileno (stderr_f), _O_BINARY);
```
GitMensch, line 9223: I _think_ there is no option to get to this code "no init done" with the settings initialized, is there?

If not use the old code here, adding a comment above pointing that issue out.

libcob/common.c
```diff
@@ -10247,9 +10267,16 @@ cob_common_init (void *setptr)
...
```
GitMensch, line 10279: Not sure: wouldn't we need that in the termio inits after opening the file?
Maybe move that whole part there?

libcob/common.c
GitMensch, line 10580: Please add an explicit check for cobsetptr here and call `cob_fatal_error (COB_FERROR_INITIALIZED)` if it isn't set.

libcob/coblocal.h
```diff
@@ -385,6 +385,16 @@ typedef struct __cob_settings {
...
+
+	char 		*cob_stdin_filename;		/* Filename to redirect reads to stdin */
+	char		*cob_stdout_filename;		/* Filename to redirect writes to stdout */
+	char		*cob_stderr_filename;		/* Filename to redirect writes to stderr */
+	char		*cob_stdin_filename_set;	/* Current filename which is replacing stdin */
+	char 		*cob_stdout_filename_set;	/* Current filename which is replacing stdout */
+	char 		*cob_stderr_filename_set;	/* Current filename which is replacing stderr */
+	FILE		*cob_stdin;					/* FILE* to redirect reads to stdin */
+	FILE		*cob_stdout; 				/* FILE* to redirect writes to stdout */
+	FILE		*cob_stderr; 				/* FILE* to redirect writes to stderr */
```
GitMensch, line 397: Just an idea (this doesn't need to be done!) we _may_ `#define COB_STDIN	cobsetpr->cob_stdin` and use this in all places - and _maybe even_ `#define COB_STDERR_OR_DEFAULT	cobsetpr ? cobsetpr->cob_stderr ? cobsetpr->cob_stderr : stderr` for the places we need that.

I'll leave that completely to you to decide.

tests/testsuite.src/run_misc.at
```diff
@@ -16107,3 +16107,168 @@ callee.cob:8: attempt to reference invalid memory address (signal SIGSEGV)
 ])
 
 AT_CLEANUP
+
+
+AT_SETUP([stdin/stdout/stderr redirection])
```
GitMensch, line 16112: Just seen that the first time after you've added stdin.

That test is super clever and I like it - the only _possible_ issue: if the stdin part does not work (what we want to ensure with this test) this test would hang forever, no?

But that can possibly apply to every test, especially file related... so leaving it in is a good way to catch if that does not work :-)

NEWS
```diff
@@ -35,6 +35,13 @@ NEWS - user visible changes				-*- outline -*-
    already have a registered handler, allowing users to preserve their own
    signal handlers during the initialization of the GnuCOBOL runtime.
 
+** new runtime configurations: COB_STDIN_FILE, COB_STDOUT_FILE,
+   COB_STDERR_FILE. They allow redirection of stdin/stdout/stderr input/output
+   to a file. Alternatively, these configurations are also available as runtime
+   options which can be used to set them programmatically:
```
GitMensch, line 41: to set them via configuration file, environment variables, or programatically

GitMensch (changes requested):
We're nearly done on this PR. Thanks for your efforts!

Some notes and mostly minor things to adjust in the review below.

oguzcankirmemis:
I have reverted the CI fix, and opened up this PR for it here: #300

tests/testsuite.src/run_misc.at
oguzcankirmemis: How about running the test with a timeout?, simply like `AT_CHECK([timeout 10s $COBCRUN_DIRECT ./prog], [0], [], [])`

oguzcankirmemis: I would even suggest adding timeout to both, the other one can also hang.

libcob/common.c
oguzcankirmemis: ah, good old tabs vs spaces, I need to check my editor's configuration

libcob/common.c
oguzcankirmemis: Yes, I think we need it also in termio. I would suggest that aside from setting stdout/stderr/stdin in `cob_init`, no further initialization should be done at common, and the rest should happen in termio.

I find even setting the initials in cobsetptr in common is uglyish, but I guess it is an ok compromise unless we we want to introduce a different init routine in termio just for the sake of redirection.

oguzcankirmemis: This ties back to your review in termio with what happens if the string is null/empty etc.: It should ignore and simply work with whatever set in the cobsetptr, which would be the process stdin/stdout/stderr, and do the configuration for it.

oguzcankirmemis: Ok, moving the whole part there sounds problematic, because `cob_init_termio` expects that settings is initialized, whereas the whole point of `cob_common_init` is to leave settings unintialized. But as you said, we also need this logic in the termio. Since calling _setmode twice with the same parameter shouldn't hurt, for now, I would just duplicate the relevant part and also put a comment in cob_common_init hinting at the issue.

libcob/common.c
GitMensch: actually GC tab-indents, but space-indents within conditions that have multiple lines :-)

libcob/termio.c
oguzcankirmemis: I have refactored this quite a bit. Now, if the filename provided is invalid, it is ignored. Otherwise, if there was a previous file set and owned by libcob, it is closed. It doesn't close files that are set programmatically, via `cob_set_runtime_option` though, they are the responsibility of the programmer.

oguzcankirmemis: As you will see, there are a lot of redundancies between stdout/stderr/stdin. But to be honest I find it better this way, instead of implementing something that would accept generic parameters.

libcob/cobgetopt.c
oguzcankirmemis: There are two functions, but I think adding it to both doesn't harm too much.

oguzcankirmemis: By the way, cob_get_runtime_option actually assumes settings is initialized, I have added an if check to bypass it for stdin/stdout/stderr. I can add the same logic to others to make it uniform among all runtime options.

libcob/coblocal.h
oguzcankirmemis: Done, I think it looks nicer.

tests/testsuite.src/run_misc.at
oguzcankirmemis: I had to revert `timeout`, apparently it does not come pre-installed in MacOS. Disappointing really, it should be part of the standard posix tools if I am not misunderstanding. There is a perl alternative we can probably use, but it looks ugly to me. Or one can install coreutils in Mac and alias `gtimeout` with `timeout`, if we want to go there.

GitMensch: not sure about the portability of that... but at least on GNU/Linux and one checked Solaris that's available.

oguzcankirmemis:
Rebased after the merge of #300

libcob/common.c
```diff
@@ -8311,6 +8314,11 @@ set_config_val (char *value, int pos)
 			cob_new_trace_file ();
 		}
 
+		if (data_loc == offsetof (cob_settings, cob_stdin_filename)
+		    || data_loc == offsetof (cob_settings, cob_stdout_filename)
+		    || data_loc == offsetof (cob_settings, cob_stderr_filename)) {
```
GitMensch:
```suggestion
		if (data_loc == offsetof (cob_settings, cob_stdin_filename)
		 || data_loc == offsetof (cob_settings, cob_stdout_filename)
		 || data_loc == offsetof (cob_settings, cob_stderr_filename)) {
```

minor formatting, can also be applied during upstream commit

libcob/termio.c
```diff
@@ -1098,9 +1103,149 @@ cob_accept (cob_field *f)
...
+static void
+cob_stdin_termio (void)
+{
+	if (cobsetptr->cob_stdin_filename) {
+		FILE *stdin_f = NULL;
+		if (cobsetptr->cob_stdin_filename_set &&
+		    strcmp (cobsetptr->cob_stdin_filename,
+	    cobsetptr->cob_stdin_filename_set) == 0) {
+			return;
+		}
```
GitMensch:
```suggestion
		if (cobsetptr->cob_stdin_filename_set
		 && strcmp (cobsetptr->cob_stdin_filename,
		            cobsetptr->cob_stdin_filename_set) == 0) {
			return;
		}
```

minor formatting here and similar below in the new code, can also be done during usptream commit

GitMensch (approved):
LGTM - thanks for your work on this, ready for upstream when the CI passes

If neither @lefessan nor @ddeclerck get to commit that upstream, then I'll do so not later than this weekend.

lefessan:
Merged in SVN as r5709.

PR #300, opened 2026-08-09
Pass C runtime that is linked with libcob in cobc for MSVC

oguzcankirmemis:
The current version does not ensure that the C runtimes of libcob and compiled target match. This can cause undefined behavior if there is a communication between the two C runtimes, like passing a FILE pointer from one to another. This also fixes the issue with two failing CI tests (CBL_GC_HOSTED, COB_SIGNAL_REGIME) that is otherwise failing in MSVC builds.

### DIFF
--- a/.github/workflows/windows-msvc.yml
+++ b/.github/workflows/windows-msvc.yml
@@ -188,18 +188,15 @@ jobs:
           sed -i '/run_misc/{N;/write to internal storage (1)/{N;N;N;N;s/traceon/traceon; echo "workflow:1">"$at_check_line_file"; at_fn_check_skip 77/;}}' tests/testsuite
 
       # Fail tests that behave differently under MSVC Debug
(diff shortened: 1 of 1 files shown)


GitMensch:
the commit needs to be split - I'll to the cobc part upstream now

ddeclerck:
Non-CI parts merged in SVN @ 5691.
I'll update the CI parts separately.

Thanks for this PR !

GitMensch:
Can be closed as soon as the CI parts are checked in, ideally with a clean merge :-)

GitMensch:
@ddeclerck is there anything open here?

lefessan:
I have rebased to keep only the part that was not already merged into the SVN, if it's ok, I can merge it in gitside-gnucobol-3.x

GitMensch:
I'd suggest to keep one of the xfail no/yes changes as a comment, in case it needs to be added later for other test cases - but the change is fine in general and can be merged to gitside

PR #301, opened 2026-08-11
Fix some Windows bugs in 4.x

ddeclerck:
This PR fixes 20 failing tests in the Windows MSVC CI (9 tests under MSYS2).

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -872,22 +872,8 @@ jobs:
         run: |
           cd tests\testsuite.src
 
-          rem Skip test as it sometimes works and sometimes not...
-          rem sed -i '/AT_SETUP(\[temporary path invalid\])/a AT_SKIP_IF(\[true\])' used_binaries.at
-
-          rem The following 13 fail (identical to MSYS2)
+          rem The following 2 fail (identical to MSYS2)
           sed -i '/AT_SETUP(\[Test dump feature (4)\])/a AT_SKIP_IF(\[true\])' run_misc.at
-          sed -i '/AT_SETUP(\[OUTPUT on INDEXED file to missing directory\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[INDEXED undeclared keys\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[PIPE I\/O\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[STDOUT STDERR I\/O\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[LINE SEQUENTIAL one Record\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[INDEXED COMMIT\/ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[SEQUENTIAL\/RELATIVE COMMIT\/ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[LINE SEQUENTIAL COMMIT \/ ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[OPEN OUTPUT COMMIT \/ ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[MF io status (indexed)\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[Test MF Status (sequential)\])/a AT_SKIP_IF(\[true\])' run_file.at
           sed -i '/AT_SETUP(\[LINE SEQUENTIAL\])/a AT_SKIP_IF(\[true\])' run_file.at
 
       - name: Run testsuite
--- a/.github/workflows/windows-msys2.yml
+++ b/.github/workflows/windows-msys2.yml
@@ -150,7 +150,7 @@ jobs:
           # command it provides, to workaround a diff bug under the MSYS2 CI ;
           # this no longer seems necessary today
           # PATH="$PATH:/c/Program Files/Git/usr/bin"
-          ../configure $CFGOPT CFLAGS=-std=c17
+          ../configure $CFGOPT CFLAGS=-std=gnu17
 
       - name: Upload config.log
         if: ${{ !cancelled() }}
@@ -190,23 +190,8 @@ jobs:
         run: |
           cd tests/testsuite.src
 
-          # Skip test as it sometimes works and sometimes not...
-          # sed -i '/AT_SETUP(\[temporary path invalid\])/a AT_SKIP_IF(\[true\])' used_binaries.at
-
-          # The following 13 fail (identical to MSYS1)
-          # Note LINE SEQUENTIAL one Record works with the clang64 environment
+          # The following 2 fail (identical to MSYS1)
           sed -i '/AT_SETUP(\[Test dump feature (4)\])/a AT_SKIP_IF(\[true\])' run_misc.at
-          sed -i '/AT_SETUP(\[OUTPUT on INDEXED file to missing directory\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[INDEXED undeclared keys\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[PIPE I\/O\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[STDOUT STDERR I\/O\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[LINE SEQUENTIAL one Record\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[INDEXED COMMIT\/ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[SEQUENTIAL\/RELATIVE COMMIT\/ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[LINE SEQUENTIAL COMMIT \/ ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[OPEN OUTPUT COMMIT \/ ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[MF io status (indexed)\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[Test MF Status (sequential)\])/a AT_SKIP_IF(\[true\])' run_file.at
           sed -i '/AT_SETUP(\[LINE SEQUENTIAL\])/a AT_SKIP_IF(\[true\])' run_file.at
 
       - name: Run testsuite
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -4,6 +4,20 @@
 	* fileio.c (cob_file_open, cob_file_close) [_WIN32]:
 	  implement bidirectional pipes under Windows
 
+2026-08-14  David Declerck <david.declerck@ocamlpro.com>
+
+	* fileio.c (isdirvalid): implement Windows counterpart
+	* fileio.c (cob_open_qbl): fix flags and permission for Windows
+	* fileio.c (cob_file_sync, cob_file_unlock): call fdcobsync
+	  only if file is open in mode other than COB_OPEN_INPUT
+	  and fflush only is last operation was a write
+	* fileio.c (pathcmp) [_WIN32]: new function to compare paths
+	  while ignoring slash differences
+	* fileio.c (cob_get_filename_print) [_WIN32]: use pathcmp
+	  instead of strcmp to compare paths
+	* fbdb.c (ix_bdb_open): call ix_bdb_close to cleanup on
+	* cobcapi.c (cob_get_buff): protect call to cob_free
+
 2025-07-28  Simon Sobisch <simonsobisch@gnu.org>
 
 	* common.h, fileio.c: new externalized typedef EXTFH_FUNC used in
--- a/libcob/Makefile.am
+++ b/libcob/Makefile.am
@@ -50,7 +50,7 @@ INDEXED_LD_FLAGS = -version-info 1:0:0
 if COB_MAKE_CISAM_LIB
 lib_ci = libcobci.la
 libcobci_la_SOURCES = libcobci.c fsqlxfd.c  # includes fisam.c
-libcobci_la_LIBADD = libcob.la $(CISAM_LIBS)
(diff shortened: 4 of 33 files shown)


ddeclerck:
@GitMensch Note that at least two more tests could be easily fixed (opinion needed).

* INDEXED undeclared keys

Difference in error message :
```
-libcob: prog.cob:82: error: unknown file error (status = 39) for file file2 ('./fileX') on OPEN
+libcob: prog.cob:82: error: unknown file error (status = 39) for file file2 ('./fileX' => .\fileX) on OPEN
```

This is because `cob_get_filename_print` considers `./fileX` and `.\fileX` to be different.
I see two options here:
- using a dedicated path comparison that ignores separator differences
- normalizing paths to use unix-style spearators

Any preference ?

* trace feature with indexed EXTFH

The error is as follows:
```
fatal error C1189: #error:  Macro definition of snprintf conflicts with Standard Library function declaration
```

This should already be taken care of by this definition in `common.h`:
```
#if !COB_USE_VC2015_OR_GREATER
/* VC2015+ provides standard function with plain
   name, old posix emulation with underscore */
#define snprintf		_snprintf
#endif
```

However, `COB_USE_VC2015_OR_GREATER` is defined in `cobinternal.h`, which is not included in `common.h`. Should we move these `COB_USE_VC*` to `common.h` ? Or since this is the only occurrence, just inline the definition of `COB_USE_VC2015_OR_GREATER` in the definition of `snprintf` ?

libcob/fileio.c
```diff
@@ -1916,7 +1928,7 @@ cob_file_sync (cob_file *f)
 		if (f->file) {
 			fflush ((FILE *)f->file);
 		}
-		if (f->fd >= 0) {
+		if (f->fd >= 0 && f->open_mode != COB_OPEN_INPUT) {
```
GitMensch: switch this condition, please, same below

libcob/fileio.c
```diff
@@ -419,12 +427,17 @@ isdirvalid (char *filename)
...
-	/* TODO: This code needs to be tested on Windows and adjusted as needed */
-	/* For now it is effectively disabled on Windows */
-		if (tmp[ln] == SLASH_CHAR) {
+		if (isslash(tmp[ln]))  {
+			while (ln > 1 && isslash(tmp[ln-1])) { --ln; }
 			tmp[ln] = 0;
 			errno = 0;
+#ifdef _WIN32
+			/* Under Windows, a slash is required after a drive letter, else stat fails */
+			if (ln == 2 && tmp[1] == ':' && isalpha((unsigned char)tmp[0])) {
```
GitMensch: isaplha actually has too much elements in if the C locale is not active; I _guess_ `isalnum_l`is not portable, so would suggest to use a plain `>= 'A' && <= 'Z' || `>= 'a' && <= 'z' here

libcob/fileio.c
```diff
@@ -399,12 +399,20 @@ void cob_seqra_init_fileio (cob_file_api *);
 
 /* Local functions */
 
+static int
+isslash (char c)
```
GitMensch, line 403: while the compiler most likely does that: use the `A_INLINE` macros here, please

tests/testsuite.src/run_misc.at
```diff
@@ -3701,22 +3701,28 @@ AT_DATA([prog.cob], [
 
 AT_DATA([cmod.c], [[
 #include <stdio.h>
-int dump (unsigned char *data)
+#include <libcob.h>
+
+COB_EXT_EXPORT int
+dump (unsigned char *data)
```
GitMensch, line 3707: If possible at all, we should prefer replacing most dump calls to execute CBL_OC_DUMP instead...

are those tests fixed in 3.x (possibly just in a different place there?

GitMensch (commented):
Wouldn't be the changes to build_windows be relevant for 3.x already (or did we just not merge them yet)?

Can you please commit https://github.com/OCamlPro/gnucobol/pull/300/changes upstream? I guess that will fix at least part of the MSVC issues here as well, no?

... why don't we have that much failures on MSYS2 also on 3.x? (the fileio parts are clear, these are new in 4.x and not that well tested on Windows)

libcob/fileio.c
ddeclerck: What do you mean by "switching" ?

Note that an identical condition already exists in `cob_file_close` to protect a call to `fdcobsync` - I merely copied that.

tests/testsuite.src/run_misc.at
ddeclerck: All the tests in 3.x that use C functions correctly use `COB_EXT_EXPORT`. Some tests specific to 4.x did not, hence the fixes here.

As for CBL_OC_DUMP, note that it "does not work yet" in 4.x (cf tests "LINE SEQUENTIAL" (run_file.at) failing because CBL_OC_DUMP can not be found). I'd have to dig into that, but since I'm going for vacation this afternoon, it's unlikely I'll have time to tackle this issue before.

ddeclerck:
> Wouldn't be the changes to build_windows be relevant for 3.x already (or did we just not merge them yet)?

Yes, I'm already preparing a PR for that - just my focus was on 4.x for now (takes time to context-switch my brain between the two branches). It will also include the fixes that also apply to 3.x (protecting the call to `fdcobsync` and `cob_free` in `cob_get_buff`).

> Can you please commit https://github.com/OCamlPro/gnucobol/pull/300/changes upstream? I guess that will fix at least part of the MSVC issues here as well, no?

Yes, will do that right now. And this has to be merged in 4.x too (I'll add it to the GC3/GC4 merge PR).

> ... why don't we have that much failures on MSYS2 also on 3.x? (the fileio parts are clear, these are new in 4.x and not that well tested on Windows)

Many failures were just because of new code in 4.x. Also, some new tests in 4.x "accidentally" revealed bugs that are present in 3.x but not triggered.

P.S : any suggestion regarding the two bugs mentionned in https://github.com/OCamlPro/gnucobol/pull/301#issuecomment-5266260174 ?

libcob/fileio.c
GitMensch:
```c
		if (f->open_mode != COB_OPEN_INPUT
		 && f->fd >= 0) {

**Comment by GitMensch** [comment]
> This is because `cob_get_filename_print` considers `./fileX` and `.\fileX` to be different. I see two options here:
> 
>     * using a dedicated path comparison that ignores separator differences

that's reasonable for [_WIN32] - but then this should likely include /c -> [cC]: (I _tend_ to not ignore file case, as FAT32 and NTFS case-sensitive exists on both Windows and Linux)...


This should already be taken care of by this definition in common.h.... in cobinternal.h

OK, that's a "new" bug in GC4 then; I understand Ron's idea with cobinternal.h, but ... am not 100% sure what to do with this and other headers splits. Please think about the general approach and then share your thoughts.

For that specific test where a C test file is compiled: do we need snprintf there in first case?
Isn't there a function in libcob that we could call instead (when then also ensures that the runtime from libcob is used)?

**Comment by GitMensch** [comment]
> Also, some new tests in 4.x "accidentally" revealed bugs that are present in 3.x but not triggered.

I hope we get testcases for those together with the bugfix to 3.x :-)

File: libcob/fbdb.c [hunk 5]
```diff
@@ -1527,6 +1527,7 @@ ix_bdb_open (cob_file_api *a, cob_file *f, char *filename, const enum cob_open_m
 		if (p->data.data != NULL
 		 && p->data.size > 0
 		 && p->data.size > f->record_max) {
+			ix_bdb_close (a ,f, 0);
```
GitMensch:
```suggestion
			ix_bdb_close (a, f, 0);
```

libcob/cobcapi.c
GitMensch, line 26: unrelated, just wondering - shouldn't this be in config.h as well?

GitMensch (approved):
LGTM

libcob/cobcapi.c
ddeclerck: Just checked, the generated `config.h` already has the above `_GNU_SOURCE` definition (whatever the target - Linux, MSYS2 or MacOS). The static `config.h` for MSVC also has the same definition.

GitMensch: please drop it here then

tests/testsuite.src/run_file.at
```diff
@@ -29833,3 +29835,75 @@ AT_CHECK([$COMPILE -fdefault-file-colseq=EBCDIC prog.cob])
 AT_CHECK([$COBCRUN_DIRECT ./prog])
 
 AT_CLEANUP
+
+
+AT_SETUP([Check for potential COB_SYNC bugs])
```
GitMensch, line 29840: those two cases should also be an issue for GC 3.x, no?
In any case I'd suggest to have these tests (and, if necessary fixing code) placed into 3.x, then merged.

tests/testsuite.src/run_file.at
GitMensch, line 26959: this one definitely should be included in the TODO, if it isn't in there already

libcob/cobcapi.c
ddeclerck: FYI, `_GNU_SOURCE` is defined in 3 libcob files: `numeric.c`, `cobcapi.c` and `call.c`. Should it be removed in all three ?

tests/testsuite.src/run_file.at
GitMensch, line 26508: If we don't have that test in twice (once with , once without -std=mf) then this test should be also included in 3.x (renamed, with adjusted test result) [otherwise only the "non mf" test can be copied over and/or this test be copied, but skipped with a note that dialect specific io status is not implemented yet.

libcob/fileio.c
```diff
@@ -9846,7 +9912,7 @@ cob_get_filename_print (cob_file* file, const int show_resolved_name)
 	offset += len;
 
 	if (show_resolved_name
```
GitMensch, line 10012: Shouldn't we have that in GC3.x (first)?

libcob/ChangeLog
```diff
@@ -1,4 +1,16 @@
...
+	* fileio.c (isdirvalid): implement Windows counterpart
+	* fileio.c (cob_open_qbl): fix flags and permission for Windows
+	* fileio.c (cob_file_sync, cob_file_unlock): call fdcobsync
+	  only if file is open in mode other than COB_OPEN_INPUT
+	  and fflush only is last operation was a write
+	* fileio.c (pathcmp) [_WIN32]: new function to compare paths
+	  while ignoring slash differences
+	* fileio.c (cob_get_filename_print) [_WIN32]: use pathcmp
+	  instead of strcmp to compare paths
+
```
GitMensch, line 20: cobcapi and fbdb are missing here

.github/workflows/build_nightly.yml
```diff
@@ -872,22 +872,10 @@ jobs:
         run: |
           cd tests\testsuite.src
 
-          rem Skip test as it sometimes works and sometimes not...
-          rem sed -i '/AT_SETUP(\[temporary path invalid\])/a AT_SKIP_IF(\[true\])' used_binaries.at
-
-          rem The following 13 fail (identical to MSYS2)
+          rem The following 4 fail (identical to MSYS2)
```
GitMensch: Are those failing because of the path output in the result?

GitMensch (changes requested):
only request so far: Changelog... and possibly implementing changes in 3.x first

libcob/cobcapi.c
GitMensch: yes (that's 4.x only, right?), and all of those should have a very early include of config.h

.github/workflows/windows-msys2.yml
```diff
@@ -150,7 +150,7 @@ jobs:
           # command it provides, to workaround a diff bug under the MSYS2 CI ;
           # this no longer seems necessary today
           # PATH="$PATH:/c/Program Files/Git/usr/bin"
-          ../configure $CFGOPT CFLAGS=-std=c17
+          ../configure $CFGOPT CFLAGS=-std=c17 LDFLAGS=-lintl
```
ddeclerck: Note: adding `-lintl` still needed, check latest failure
https://github.com/OCamlPro/gnucobol/actions/runs/33495844200/job/99819552455?pr=301

GitMensch: Looks like the Makefile.am / configure.ac in 4.x just does not pass the right flags to the separate indexed libraries (which are passed when building libcob) - please check that.

Note: instead of -std=c17 that should be -std=gnu17

libcob/ChangeLog
ddeclerck: Accidentally removed... Restored.

.github/workflows/build_nightly.yml
ddeclerck: No, these are more complex. I started investigting them, but I'd rather fix them in dedicated PRs.

.github/workflows/windows-msys2.yml
ddeclerck: Note that it used to work in the past, and stopped working "at some point" (likely after MSYS2 updated some package).

Question: why using gnu17 if it works with a more "standard" c17 ?

tests/testsuite.src/run_file.at
ddeclerck: Yes, I had a PR on 3.x for this, which was merged while I was away... I do intend to have this in 3.x first.

libcob/cobcapi.c
ddeclerck: Same in 3.x

tests/testsuite.src/run_file.at
ddeclerck: I don't see which test this comment refers to ?

libcob/fileio.c
ddeclerck: Probably.

.github/workflows/windows-msys2.yml
GitMensch: we create dist packages here (or should), so "best code possible" is the goal - especially as MSYS2 is commonly our most up-to-date env

The only reason to downgrade at all is the changed standard behaviour of (void) functions.
There is a work "in progress" to test if we need gnu/c - and if yes add it automatically during configure (only if not explicit configured, of course).

tests/testsuite.src/run_file.at
GitMensch: When clicking on that link it shows me

> # TODO: have a single dialect option to change io stats (also for ohter dialects)

oops, typo in there

libcob/Makefile.am
```diff
@@ -50,7 +50,7 @@ INDEXED_LD_FLAGS = -version-info 1:0:0
 if COB_MAKE_CISAM_LIB
 lib_ci = libcobci.la
 libcobci_la_SOURCES = libcobci.c fsqlxfd.c  # includes fisam.c
-libcobci_la_LIBADD = libcob.la $(CISAM_LIBS)
+libcobci_la_LIBADD = libcob.la $(CISAM_LIBS) $(LIBCOB_LIBS)
```
GitMensch, line 53: I guess that works... and then we need to split those in configure.ac/Makefile.am; maybe LIBCOB_LIBS_COMMON (otherwise we'll have curses, libxml, gmp, ... here as well)?

... but in general we should have most of the others separate (as the plan is to have a split libcob-xml, libcob-json libcob-curses before the 4.x release candidate (following the io library split).

While Ron originally planned to only have that for allowing multiple libraries with the same entry point based on the same source file for "multi indexed backends" - this is something that will allow to spit libcob packaging and therefore heavily reduce the dependencies - no curses used, none needed (and also none dlopen'd - distinguished for each process).

tests/testsuite.src/run_file.at
ddeclerck: Note we have two MF-specific tests here : "MF io status (indexed)" and "Test MF Status (sequential)" (below). Both only compiled with `-std=mf`.

I'm fine with the idea of backporting relevant tests from 4.x to 3.x, but if we do so, wouldn't it be better to have a dedicated PR for this and do this globally - instead of sneaking in a few tests from time to time ?

GitMensch: PRs always welcomed :-)

PR #303, opened 2026-08-14
Various GC3 fixes

ddeclerck:
Backported from https://github.com/OCamlPro/gnucobol/pull/301.

### DIFF
--- a/libcob/fileio.c
+++ b/libcob/fileio.c
@@ -1402,7 +1402,7 @@ cob_sync (cob_file *f)
 #endif
 		return;
 	}
-	if (f->organization != COB_ORG_SORT) {
+	if (f->organization != COB_ORG_SORT && f->open_mode != COB_OPEN_INPUT) {
 		if (f->file) {
 			fflush ((FILE *)f->file);
 		}
@@ -6100,7 +6100,7 @@ cob_file_unlock (cob_file *f)
 		}
 		if (f->organization != COB_ORG_INDEXED) {
 #ifndef	WITH_SEQRA_EXTFH
-			if (f->fd >= 0) {
+			if (f->fd >= 0 && f->open_mode != COB_OPEN_INPUT) {
 				fdcobsync (f->fd);
 			}
 #ifdef	HAVE_FCNTL
--- a/build_windows/ChangeLog.txt
+++ b/build_windows/ChangeLog.txt
@@ -1,4 +1,8 @@
 
+2026-08-14  David Declerck <david.declerck@ocamlpro.com>
+
+	* set the UseDebugLibraries property in all project files
+
 2025-03-26  David Declerck <david.declerck@ocamlpro.com>
 
 	* general for cobc: include new gentable.c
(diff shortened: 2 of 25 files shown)


.github/workflows/windows-msys2.yml
```diff
@@ -150,7 +150,7 @@ jobs:
           # command it provides, to workaround a diff bug under the MSYS2 CI ;
           # this no longer seems necessary today
           # PATH="$PATH:/c/Program Files/Git/usr/bin"
-          ../configure $CFGOPT CFLAGS=-std=c17
+          ../configure $CFGOPT CFLAGS=-std=c17 LDFLAGS=-lintl
```
GitMensch:
```suggestion
          ../configure $CFGOPT CFLAGS=-std=gnu17 LDFLAGS=-lintl
```

What do we need the intl part for? Ideally we can add it to the "right"/specific LDFLAGS, instead of the "global" ones

libcob/fileio.c
```diff
@@ -1402,7 +1402,7 @@ cob_sync (cob_file *f)
 #endif
 		return;
 	}
-	if (f->organization != COB_ORG_SORT) {
+	if (f->organization != COB_ORG_SORT && f->open_mode != COB_OPEN_INPUT) {
```
GitMensch, line 1405:
```suggestion
	if (f->open_mode != COB_OPEN_INPUT
	 && f->organization != COB_ORG_SORT) {
```

libcob/fileio.c
```diff
@@ -6100,7 +6100,7 @@ cob_file_unlock (cob_file *f)
 		}
 		if (f->organization != COB_ORG_INDEXED) {
 #ifndef	WITH_SEQRA_EXTFH
-			if (f->fd >= 0) {
+			if (f->fd >= 0 && f->open_mode != COB_OPEN_INPUT) {
```
GitMensch, line 6103:
```suggestion
			if (f->open_mode != COB_OPEN_INPUT
			 && f->fd >= 0) {
```

.. if this code is in another place, please adjust there as well

.github/workflows/windows-msys2.yml
```diff
@@ -185,12 +185,6 @@ jobs:
             _build/tests/cobol85/**/*.out
             _build/tests/cobol85/**/duration.txt
 
-      - name: Adjust testsuite
```
GitMensch, line 188: That just doesn't do anything (same as before)... if possible I'd like to keep it in (but commenting out the step completely) in case we need to ignore something for some time on Win32 (for example during longer development of a feature/fix where "works outside of Windows" is something we may even commit upstream)

(same applies to the change in the other workflow)

GitMensch (changes requested):
I'm totally fine with the changes for upstream - but build_windows/Changelog needs an entry.

For CI (= for now only here) see the comments.

.github/workflows/windows-msys2.yml
GitMensch: I still would like to understand what we need `LDFLAGS=-lintl` for - and if we can set that to something more specific (the global ones will "leak" into cobc calls).

lefessan:
Merged in SVN

Changes to .github have been moved to #310 as they don't target SVN

GitMensch:
As the changes were backported from #301 (written for GC4) - should we not also backport tests?

PR #308, opened 2026-08-22
testing how old deleted branches are still contained

GitMensch:
wow, that restored that branch...


ddeclerck:
Just out of curiosity: what was this PR about ?

GitMensch:
It was about enabling the test "write to internal storage" for MSVC, I think the conclusion was that the sanitizers active kick in before our own signal handling can, and disabling them for this test helps  ... but for MSVC this _may_ have been solved by using the correct runtime - not sure how if the current state of this test looks in MSVC CI.

Also those sanitizer options could be useful when running the test in a debug where configure enabled all sanitizers....

PR #309, opened 2026-08-23
This patch fixes a bug in COPY REPLACING, about spaces surrounding the replacement pseudo-text 

lefessan:
This patch fixes a bug in COPY REPLACING, when spaces are surrounding the replacement pseudo-text, and should be ignored in the non LEADING/TRAILING case.


GitMensch:
Sounds good, it just fails all CI because it contains debug output (as noted before we have COB_DEBUG_LOG macros that would be ideally used instead) and SIGSEGVS.

I'll check again once CI passes - or you drop a review-request :-)
If that is expected to take longer, I'd suggest to mark this as a draft.

lefessan:
The Windows CI again, I think it's good otherwise after fixing some other tests.

cobc/ChangeLog
```diff
@@ -1,4 +1,10 @@
 
+2026-08-23 Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+        * ppparse.y (ppp_replace_list_add): trim spaces before
```
GitMensch:
```suggestion
	 * ppparse.y (ppp_replace_list_add): trim spaces before
```

cobc/ppparse.y
```diff
@@ -160,6 +160,24 @@ ppp_replace_list_add (struct cb_replace_list *list,
...
+			new_text = new_text->next;
+		/* Remove trailing spaces in replacement text */
+		last_text = new_text;
+		curr_text = new_text;
+		while ( curr_text != NULL ){
+			if ( curr_text->text != NULL &&
+			     ! isspace (curr_text->text[0]) )
+				last_text = curr_text;
+			curr_text = curr_text->next;
+		}
```
GitMensch: This seems like reasonable code in the wrong place - should the caller do this already (if there are many, then with a helper function)?

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,34 @@ prog.cob:12: warning: ignoring redundant .
 AT_CLEANUP
 
 
+AT_SETUP([pseudotext replacement trim spaces])
+AT_KEYWORDS([misc copy replace])
+
+AT_DATA([ok.inc], [       01  VARIABLE1 PIC X(01) VALUE "OK".
+       01  VARI(03)ABLE2 PIC X(01) VALUE "OK".
```
GitMensch:
```suggestion
AT_DATA([ok.inc], [
       01  VARIABLE1 PIC X(01) VALUE "OK".
       01  VARI(03)ABLE2 PIC X(01) VALUE "OK".
```

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,34 @@ prog.cob:12: warning: ignoring redundant .
...
+       PROGRAM-ID.      prog.
+
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+           COPY "ok.inc" REPLACING == (01) == BY == (02) ==
+	    == (03) == BY == ==.
+       PROCEDURE DIVISION.
+       MAIN SECTION.
+           DISPLAY VARIABLE1 VARIABLE2.
+	   GO-BACK.
```
GitMensch:
```suggestion
	       GOBACK.
```
fall-through at program end - as the old code had - raises an error on several environments

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,34 @@ prog.cob:12: warning: ignoring redundant .
...
+])
+AT_DATA([prog.cob], [
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+           COPY "ok.inc" REPLACING == (01) == BY == (02) ==
+	    == (03) == BY == ==.
+       PROCEDURE DIVISION.
```
GitMensch, line 3342: shouldn't there be a REPLACE statement above the copybook statement that verifies that, for example with (04) and (05) also contain in the copybook to verify that the same issue does not exist with REPLACE?

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,34 @@ prog.cob:12: warning: ignoring redundant .
...
+       MAIN SECTION.
+           DISPLAY VARIABLE1 VARIABLE2.
+	   GO-BACK.
+])
+
+AT_CHECK([$COMPILE -g prog.cob])
+
+AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [OKOK
```
GitMensch: please add checking the listing to this test as well (just the code, no symbols/headers/summaries)

GitMensch (changes requested):
Thanks for bringing this up.

How did you got to the conclusion of a bug here? Possibly a rule in the standard, I guess.
Does the test code work on IBM/MF/ACU/gcobol?

And: is REPLACE affected as well?
Possibly the listing?

lefessan:
> How did you got to the conclusion of a bug here? Possibly a rule in the standard, I guess.

Some customer code triggered it (mainly because it would introduce a space, i.e. `X(1)` would become an unparsable `X (2)`.

> Does the test code work on IBM/MF/ACU/gcobol?

I could only test it on MF, which is consistent with the fact that our customer's code was coming from MF.

cobc/ppparse.y
lefessan: I can move it to a function, but I think it is actually better to do it here, as it is always needed both for COPY REPLACING and REPLACE.

tests/testsuite.src/syn_misc.at
lefessan: I have added it, but no replacement happens at all, which is independent from my patch. Is it the nominal behavior of listing ?

GitMensch: nope, that's an errror in cobc.c then :-/

lefessan: Yes, but unrelated to this PR. The replacements done in listings has always been broken and inconsistent with the real preprocessing done for the `.i` file. It would need a huge work to fix it, but I don't think it's worth the effort, as developers should use this `.i` file instead of listings to check the real output. Actually, I think it would be much better not to try to do these replacements in listings and only print the raw sources (program and copybooks).

lefessan:
If the only problem now is the listing, I think we should merge this patch in SVN, keeping only an issue that the listing is broken (in general) with replacements.

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,69 @@ prog.cob:12: warning: ignoring redundant .
...
+	                           == (03) == BY == ==.
+       PROCEDURE DIVISION.
+       MAIN SECTION.
+           DISPLAY VARIABLE1 VARIABLE2 VARIABLE3 VARIABLE4.
+	   GOBACK.
+])
+
+AT_CHECK([SOURCE_DATE_EPOCH=1011960000 $COMPILE -Wall -t prog.lst prog.cob])
+
+AT_CHECK([$COBCRUN_DIRECT ./prog], [0], [O1O2O3O4
```
GitMensch, line 3350: please add the findings here to "your" bug issue https://sourceforge.net/p/gnucobol/bugs/831/ - and reference that here

GitMensch (approved):
I agree that we can put that upstream once we added the listing issue to the known problems (it has its own bug already) and reference that in a comment (references like that also help a lot when fixing things and see bunch of failures and inspecting them directly brings up a note "currently wrong, should be this way, see x").

If possible it would be good to check if those issues are fixed now as well:

* crash when replacing by nothing (regression from your PR) https://sourceforge.net/p/gnucobol/bugs/1071/ - the analysis and fix is done, the commit including a testcase is likely missing
* REPLACE statement inside COPY does not work correct https://sourceforge.net/p/gnucobol/bugs/722/ - _should_ be solved with your changes in 3.2
* COPY REPLACING isn't working when replacing with a star https://sourceforge.net/p/gnucobol/bugs/736/ - edge case _possibly_ solved in 3.2
* COPY/REPLACE issues https://sourceforge.net/p/gnucobol/bugs/927/ - could be already solved in 3.x since a while

All that are not fixed yet can still be included in the testsuite as expected failures with the bug reference... but possibly they are solved / tested already or it is "relative easy" to do so.

GitMensch:
side note: I'd prefer @ddeclerck's #303 to be upstreamed before, but if that isn't the case, then it is how it is

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,69 @@ prog.cob:12: warning: ignoring redundant .
...
+	           == (05) == BY == ==.
+           COPY "ok.inc" REPLACING == (01) == BY == (02) ==
+	                           == (03) == BY == ==.
+       PROCEDURE DIVISION.
+       MAIN SECTION.
+           DISPLAY VARIABLE1 VARIABLE2 VARIABLE3 VARIABLE4.
+	   GOBACK.
+])
+
+AT_CHECK([SOURCE_DATE_EPOCH=1011960000 $COMPILE -Wall -t prog.lst prog.cob])
```
GitMensch: you possibly want `-fno-theaders -fno-tmessages -fno-ttimestamp` (the later may not be necessary with suppressing headers)

lefessan:
> side note: I'd prefer @ddeclerck's #303 to be upstreamed before, but if that isn't the case, then it is how it is

Which part of #303 ? I have merged his changes to build_windows/ and libcob/ into the SVN, and created #310 for changes specific to GIT.

GitMensch:
I consider the "303 before" part done :-)

cobc/ChangeLog
```diff
@@ -1,4 +1,9 @@
 
+2026-08-26 Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* ppparse.y (ppp_replace_list_add): fix bug #1041
```
GitMensch: that's 1071 and can be moved way down:
Simon Sobisch 2025-04-04

cobc/ppparse.y
```diff
@@ -189,7 +189,8 @@ ppp_replace_list_add (struct cb_replace_list *list,
 		/* Strictness flag is irrelevant for non-LEADING nor TRAILING
 		   replacements */
 		src->strict = 0;
-	} else {
+	} else
```
GitMensch: either on one line or the `if` indented one tab less

cobc/ChangeLog
lefessan: I fixed the 1041 to 1071, but I don't understand what you mean by moving it way down (I can't find a line with 2025-04-04 and your name).

lefessan:
>     * crash when replacing by nothing (regression from your PR) https://sourceforge.net/p/gnucobol/bugs/1071/ - the analysis and fix is done, the commit including a testcase is likely missing

Fixed in this PR.

>     * REPLACE statement inside COPY does not work correct https://sourceforge.net/p/gnucobol/bugs/722/ - _should_ be solved with your changes in 3.2
> 
>     * COPY REPLACING isn't working when replacing with a star https://sourceforge.net/p/gnucobol/bugs/736/ - edge case _possibly_ solved in 3.2
> 
>     * COPY/REPLACE issues https://sourceforge.net/p/gnucobol/bugs/927/ - could be already solved in 3.x since a while

I think we can close all of them, the first one is similar to 831, and the other ones are, for me, not bugs.

cobc/ChangeLog
GitMensch: I've just meant moving the whole element down (as that change in the issue tracker is from 2025... and also was locally applied before - you know how that is, at least the tracking helped over a year later...)

cobc/ChangeLog
```diff
@@ -1,4 +1,15 @@
...
+2026-08-26 Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* ppparse.y (ppp_replace_list_add): fix bug #1071
+	  "cobc crashes when replacing by nothing"
+
+2026-08-23 Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* ppparse.y (ppp_replace_list_add): trim spaces before
+	  and after the replacement tokens, in the non LEADING/TRAILING
+	  case
```
GitMensch:
```suggestion
	* ppparse.y (ppp_replace_list_add): trim spaces before and after
	  the replacement tokens, for the non LEADING/TRAILING case
```

cobc/ChangeLog
```diff
@@ -1,4 +1,15 @@
 
+2026-08-26 Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
```
GitMensch, line 2: as noted before that could be moved down to 2025, per the original bug issue... but nope, seems you have done it differently, so leave it here

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,104 @@ prog.cob:12: warning: ignoring redundant .
...
+
+])
+
+AT_CLEANUP
+
+
+AT_SETUP([bug 1071 with empty replaced text])
+AT_KEYWORDS([copy replacing trailing])
+
+AT_DATA([prog.cob], [>>SOURCE FREE
```
GitMensch:
```suggestion
AT_SETUP([COPY REPLACING with nothing])
AT_KEYWORDS([copy replacing trailing])

# see bug 1071 - cobc crashed here
AT_DATA([prog.cob], [
```

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,104 @@ prog.cob:12: warning: ignoring redundant .
...
+DATA DIVISION.
+   WORKING-STORAGE SECTION.
+   01 result PIC 9(4).
+
+PROCEDURE DIVISION.
+   COPY fab.
+   DISPLAY "The result is " result.
+STOP RUN.
+])
+AT_DATA([fab.cob], [>>SOURCE IS FREE
```
GitMensch:
```suggestion
AT_DATA([fab.cob], [
```

same as above, not needed because of  `--free` given (and if not given, then we'd auto-set --free nowadays)

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,104 @@ prog.cob:12: warning: ignoring redundant .
...
+>>DEFINE 1aa AS 2
```
GitMensch: I _guess_ you wanted to check if replace also worked in the directive - I don't know if they should - but that isn't tested yet, to do so you need to do anything with the define also in the program (with the changed/unchanged name)

tests/testsuite.src/syn_misc.at
```diff
@@ -3315,6 +3315,104 @@ prog.cob:12: warning: ignoring redundant .
...
+
+AT_SETUP([bug 1071 with empty replaced text])
+AT_KEYWORDS([copy replacing trailing])
+
+AT_DATA([prog.cob], [>>SOURCE FREE
+IDENTIFICATION DIVISION.
+PROGRAM-ID. HELLO.
+
+DATA DIVISION.
+   WORKING-STORAGE SECTION.
```
GitMensch, line 3390: Shouldn't that test (and therefore also the keywords) include `REPLACE` as well?

cobc/ChangeLog
lefessan: I am still confused. Isn't the ChangeLog supposed to show the changes in the order they happen ? If I move it down, the dates wouldn't be correctly ordered ? 

Also, what do you mean by "[it] was locally applied before" , there was already a partial attempt at fixing it ?

GitMensch: yes, per the ticket, but as noted in the review, you've done that different today with better tests, so we should not take "my" old change but yours - and a diff with a local version will show the difference

GitMensch (changes requested):
... yay, I've seem to have missed clicking "Submit" some hours ago, so here we go, mostly with a request on the testing

tests/testsuite.src/syn_misc.at
lefessan: Actually, this file comes from the issue on the bug tracker, I just took the files that were submitted and created a test case with them (only `fob.cob` was created because it was not in the issue)

tests/testsuite.src/syn_misc.at
lefessan: I haven't tried to create a perfect test case, just the one that was in the issue...

lefessan:
If it's ok with you, I am going to merge this PR as is, to prevent it from stalling.

GitMensch (approved):
OK to send upstream, while I'd prefer this testcase to be reinspected - we can set the upstream issue(s) to pending for that while leaving a note that the bug itself is fixed with that commit.

lefessan:
I updated the test to check also REPLACE and remove useless directives.

lefessan:
Merged in SVN

PR #310, opened 2026-08-26
fix GitHub windows workflows

lefessan:
This commit comes from #303 , to separate between changes in GnuCOBOL SVN, and changes specific to the GIT repository. It is supposed to be merged into .gitside-gnucobol-3.x after branch 303 has been committed into SVN.

### DIFF
--- a/.github/workflows/build_nightly.yml
+++ b/.github/workflows/build_nightly.yml
@@ -894,15 +894,6 @@ jobs:
             _build/tests/cobol85/**/*.out
             _build/tests/cobol85/**/duration.txt
 
-      - name: Adjust testsuite
-        run: |
-          rem Skip test as it sometimes works and sometimes not...
-          rem %MSYS_BASH% -lc "sed -i '/used_binaries/{N;/temporary path invalid/{N;N;N;N;s/traceon/traceon; echo \"workflow:1\">\"$at_check_line_file\"; at_fn_check_skip 77/;}}' tests/testsuite"
-
-          rem Change to expected fail if there's an error specific to the CI that we don't want to change
-          rem in the testsuite for expected result (should normally only be a temporary thing)
-          rem sed -i '/run_fundamental/{N;/Compare FLOAT-LONG with floating-point literal/{N;s/at_xfail=no/at_xfail=yes/;}}' tests/testsuite
-
       - name: Run testsuite
         run: |
           rem To work around regular hangs we run NIST first, then the internal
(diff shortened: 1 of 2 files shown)


.github/workflows/windows-msys2.yml
```diff
@@ -150,7 +150,7 @@ jobs:
           # command it provides, to workaround a diff bug under the MSYS2 CI ;
           # this no longer seems necessary today
           # PATH="$PATH:/c/Program Files/Git/usr/bin"
-          ../configure $CFGOPT CFLAGS=-std=c17
+          ../configure $CFGOPT CFLAGS=-std=gnu17 LDFLAGS=-lintl
```
GitMensch, line 153: I still would like to understand what we need LDFLAGS=-lintl for - and if we can set that to something more specific (the global ones will "leak" into cobc calls).

.github/workflows/build_nightly.yml
```diff
@@ -894,15 +894,6 @@ jobs:
...
 
-      - name: Adjust testsuite
-        run: |
-          rem Skip test as it sometimes works and sometimes not...
-          rem %MSYS_BASH% -lc "sed -i '/used_binaries/{N;/temporary path invalid/{N;N;N;N;s/traceon/traceon; echo \"workflow:1\">\"$at_check_line_file\"; at_fn_check_skip 77/;}}' tests/testsuite"
-
-          rem Change to expected fail if there's an error specific to the CI that we don't want to change
-          rem in the testsuite for expected result (should normally only be a temporary thing)
-          rem sed -i '/run_fundamental/{N;/Compare FLOAT-LONG with floating-point literal/{N;s/at_xfail=no/at_xfail=yes/;}}' tests/testsuite
-
```
GitMensch, line 905: That just doesn't do anything (same as before)... if possible I'd like to keep it in (but commenting out the step completely) in case we need to ignore something for some time on Win32 (for example during longer development of a feature/fix where "works outside of Windows" is something we may even commit upstream)

(same applies to the change in the other workflow)

lefessan:
The CI is currently working well on Windows, so I am not sure this PR is still useful...

GitMensch:
Asa the rest of this PR (originally it contained more) only deletes comments + add LDFLAGS that seem to not be used (+switches to gnu standard isntead of iso, which enables some other features), I suggest to close this and directly change c17 to gnu17 (no PR needed),

GitMensch:
one single remark on the MSVC Debug builds originally pointed out in the obsolete #257;

> for debug builds it should also package ucrtbased.dll and matching vcruntime*d.dll as those are commonly not available

PR #312, opened 2026-08-30
libcob: implement dynamic high-water mark screen_term_buff to prevent…

arnab8335:
Fix: heap overflow in field_accept() for ACCEPT fields larger than the fixed screen I/O buffer

Implemented the dynamic buffer approach you suggested, scoped entirely to screenio.c:
Added a file-local static char *screen_term_buff / static size_t screen_buffer_size, separate from the shared cobglobptr->cob_term_buff used by termio.c — so this only affects extended screen I/O, per your note.
ensure_buffer(size_t requested_size): starts at a 256-byte baseline (COB_MINI_BUFF), grows via raw realloc (not cob_malloc/cob_realloc, since those hard-abort on OOM and the requested size comes straight from the COBOL field — in theory up to ~2GB). Never shrinks; grows to exactly the requested size (high-water mark), so repeated ACCEPTs at similar sizes don't re-realloc.
Warns via cob_runtime_warning() (non-fatal) when a grow exceeds COB_TERM_BUFF_WARN_SIZE (32KB, now a #define), and reports via cob_runtime_error() on actual allocation failure — the field is still accepted at full size in the warning case; only real OOM causes field_accept() to bail out of the ACCEPT (checked via screen_buffer_size < size_accept after calling ensure_buffer, so a failed grow can't fall through into writing past the old, smaller buffer).
Wired into field_accept(): replaced every use of the COB_TERM_BUFF macro within that function with a local term_buff pointer from ensure_buffer(). COB_TERM_BUFF itself is untouched for termio.c.
Freed in cob_exit_screen(), alongside the existing cob_base_inp cleanup, and reset to NULL/0 so re-entering screen mode later in the same process starts clean.

Testing:
Added a regression test to tests/testsuite.src/run_manual_screen.at (ACCEPT field larger than internal I/O buffer) using a PIC X(9000) screen field, past the old 8192-byte ceiling.
Verified with AddressSanitizer: rebuilt screenio.c under ASan, ran the 9000-byte ACCEPT through cobcrun — no overflow/crash, versus the old fixed-buffer code which would have overflowed on this input.


libcob/screenio.c
```diff
@@ -3612,15 +3650,20 @@ field_accept (cob_field *f, cob_flags_t fattr, const int sline, const int scolum
 		} else {
 			size_accept = f->size;
 		}
+		term_buff = (unsigned char *)ensure_buffer (size_accept);
+		if (screen_buffer_size < size_accept) {
+			/* allocation failure */
+			return;
```
GitMensch: We should raise an exception here (add to exceptions.def, then use the define here; possibly a new, non-fatal one `EC-SCREEN-IMP-STORAGE`?) and only return if the size is zero - otherwise set the size_accept to the actual size.

libcob/screenio.c
```diff
@@ -3612,15 +3650,20 @@ field_accept (cob_field *f, cob_flags_t fattr, const int sline, const int scolum
 		} else {
 			size_accept = f->size;
 		}
+		term_buff = (unsigned char *)ensure_buffer (size_accept);
```
GitMensch:
```suggestion
	}
	if (size_accept) {	/* non-zero-sized field */
		term_buff = (unsigned char *)ensure_buffer (size_accept);
```

this effectively handles a variable-length field of size zero like `ACCEPT OMITTED` - and we don't have to care for zero-size buffer later

libcob/screenio.c
```diff
@@ -3530,6 +3532,41 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+			return screen_term_buff;
+		}
+		screen_term_buff = tmp;
+		screen_buffer_size = requested_size;
+	} else if (screen_term_buff == NULL) {
+		screen_term_buff = realloc (screen_term_buff, screen_buffer_size);
+		if (screen_term_buff == NULL) {
+			cob_runtime_error (_("could not allocate %lu bytes of memory"),
+					    (unsigned long)screen_buffer_size);
+		}
```
GitMensch:
```suggestion
		screen_term_buff = cob_malloc (screen_buffer_size);
```

we only get in here once and expect that to work

libcob/screenio.c
```diff
@@ -3530,6 +3532,41 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+	if (screen_buffer_size == 0) {
+		screen_buffer_size = COB_MINI_BUFF;
+	}
+
+	if (requested_size > screen_buffer_size) {
+		char *tmp;
+
+		if (requested_size > COB_TERM_BUFF_WARN_SIZE) {
+			cob_runtime_warning (_("ACCEPT/DISPLAY of unusually large field (%lu bytes)"),
+					     (unsigned long)requested_size);
```
GitMensch: Better move the size casting outside, for example by making it `unsigned long requested_size`.

I'm not _sure_ if it is better to have that warning only once per increase - as done here - or at the start of this function (both options have benefits and issues); I _tend_ to the second (otherwise a bit reorderning is needed to do that before the `== 0` case above

libcob/screenio.c
```diff
@@ -3530,6 +3532,41 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
 	refresh_mywin (mywin);
 }
 
+static char *
+ensure_buffer (size_t requested_size)
+{
+	if (screen_buffer_size == 0) {
+		screen_buffer_size = COB_MINI_BUFF;
+	}
```
GitMensch:
```suggestion
	if (screen_buffer_size == 0) {
		/* ensure minimal buffer at first use */
		if (requested_size > COB_MINI_BUFF) {
			screen_buffer_size = requested_size;
		} else {
			screen_buffer_size = COB_MINI_BUFF;
		}
		screen_term_buff = cob_fast_malloc (screen_buffer_size);
		return screen_term_buff;
	}
```

Note: I've moved the allocation from down there here, and don't do an extra-check of the initial cob_fast_malloc... but we may should do that, given that the first allocation can be huge as well

libcob/screenio.c
GitMensch:
```suggestion
```
new code checks that up-front

libcob/screenio.c
```diff
@@ -3538,6 +3575,7 @@ field_accept (cob_field *f, cob_flags_t fattr, const int sline, const int scolum
 {
 	unsigned char	*p;
 	unsigned char	*p2;
+	unsigned char	*term_buff;
```
GitMensch: if we use the static buffer only here then we should use the static var in this function as well, not a local copy

libcob/screenio.c
GitMensch: by using the static var, this function can be changed to return 0 on success, 1 on failure and we can check the sucess this way

libcob/screenio.c
```diff
@@ -3530,6 +3532,41 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+		tmp = realloc (screen_term_buff, requested_size);
+		if (tmp == NULL) {
+			cob_runtime_error (_("could not allocate %lu bytes of memory"),
+					    (unsigned long)requested_size);
+			return screen_term_buff;
+		}
+		screen_term_buff = tmp;
+		screen_buffer_size = requested_size;
+	} else if (screen_term_buff == NULL) {
+		screen_term_buff = realloc (screen_term_buff, screen_buffer_size);
```
GitMensch: as we already have an explicit path for "did not work" and don't need to retain the old content: please use cob_fast_malloc + cob_free here instead of plain realloc...

And yes, you're right that cob_malloc would potentially abort if too big (which is unlikely but still possible) - there is a pending TODO to work on this by providing a non-aborting version - feel free to add a special cob_fast_malloc version which never aborts [using that will mean the caller need to check, like done here)

libcob/screenio.c
```diff
@@ -4713,6 +4756,11 @@ cob_exit_screen (void)
 			cob_free (cob_base_inp);
 			cob_base_inp = NULL;
 		}
+		if (screen_term_buff) {
+			free (screen_term_buff);
```
GitMensch:
```suggestion
			cob_free (screen_term_buff);
```

tests/testsuite.src/run_manual_screen.at
```diff
@@ -4159,3 +4159,43 @@ AT_CHECK([$COMPILE prog.cob], [0], [], [])
...
+       PROGRAM-ID.      prog.
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       01  big-field    PIC X(9000).
+
+       SCREEN           SECTION.
+       01  scr.
+           03  LINE 1 VALUE
+               'Type a few characters, then press ENTER.'.
+           03  LINE 3 COL 1 PIC X(9000) USING big-field.
```
GitMensch: the test is good; for full test coverage we'd need an additional accept > warning threshold as well (then checking its diagnostic as well)

GitMensch (changes requested):
The general approach is good, thanks for the PR!

TODOs:
* COB_TERM_BUFF (cobolocal.h) should be removed in this PR, as well as the struct element it points to
* exception to be added
* libcob/Changelog needs an entry

Please note https://sourceforge.net/p/gnucobol/wiki/Copyright%20reassignment/ as we're reaching the threshold of legally significance - especially if you can see yourself doing more contributions.

arnab8335:
Hello, @GitMensch 
I want to be completely honest, the structural concept and reasoning behind this fix are entirely mine: I used an AI tool to help write the initial memory management code blocks.How should we proceed with the copyright assignment form in accordance with GNU's current guidelines since you indicated that we are approaching the threshold of legal significance?
I'm prepared to make all of the refactoring adjustments you recommended. I would also be more than pleased to manually rewrite the code from scratch based on your input if necessary to adhere to the project's copyright requirements. Tell me the best course of action, please!

GitMensch:
Hi @arnab8335. Thanks for your honesty - it is good to know where the current guidelines are.

The symbol-renaming part is insignificant - so "most" LOC aren't.

For the other parts: if you rewrite them per review comments, I'm quite sure that those more significant memory blocks are also rewritten, so that's fine with me and I don't mind if you work from the current PR's state or from scratch.

We'd still need the copyright papers (or at least disclaimers) which can take a while so you may consider starting that procedure, especially if you can see yourself to stay (we have a lot of easy to complex open tasks. For regular contributors we do not only offer (hopefully clear and informative) reviews but also mentoring, if wanted.

arnab8335:
Hi @GitMensch 
Thank You so much for understanding and the encouraging response. I genuinely appreciate your guidance on how to navigate the code.
Nevertheless, I will go ahead and fix the review comments that you have provided to me, since the changes will rewrite those sections, I am glad to know it keeps the contribution legally cleaned complaint.
Thank You for your offer regarding mentoring, I would genuinely love to learn more from you and continue contributing to open tasks.
Also, I will get started on the FSF copyright assignment steps right away , so its ready.
I will push the code changes shortly for your review.
Once again, Dankeschön!!

arnab8335:
Hi @GitMensch 
One Quick Question, 
Do you want me to move the size casting outside here
/* Expanding the Buffer */
		tmp = realloc (screen_term_buff, requested_size);
		if (tmp == NULL) {
			cob_runtime_error (_("could not allocate %lu bytes of memory"),
					    (unsigned long)requested_size);
			return screen_term_buff;
		}

We can make like 
/* Expanding the Buffer */
		tmp = realloc (screen_term_buff, requested_size);
		if (tmp == NULL) {
                        const unsigned long warning_size = (unsigned long)requested_size
			cob_runtime_error (_("could not allocate %lu bytes of memory"),
					    warning_size);
			return screen_term_buff;
		}

Does it sound good ?

arnab8335:
Hello @GitMensch 
One more clarification question :
For `ensure_buffer()`, should I change its return type to int (0 success, 1 failure), allocate with `cob_fast_malloc()`, and replace the old buffer with `cob_free()` only after successful allocation?

On failure, should `screen_buffer_size` remain unchanged so `field_accept()` can raise `COB_EC_SCREEN_IMP_STORAGE`, return only when it is zero, and otherwise truncate size_accept to the existing buffer size?

GitMensch:
For the casting: if we always cast inside the function, then the function should better not take the size_t argument in the first place - but let caller cast (or use the appropriate data type).

arnab8335:
> For the casting: if we always cast inside the function, then the function should better not take the size_t argument in the first place - but let caller cast (or use the appropriate data type).

@GitMensch Sorry for sounding stupid,
But to confirm, should I change ensure_buffer()’s parameter from size_t to unsigned long, and cast at the call site:
`ensure_buffer ((unsigned long) size_accept);`
This would let the warning pass requested_size directly to %lu. Or would you prefer keeping size_t and using a size-appropriate format instead?

GitMensch:
No problem - the first of those two... and you may find that unsigned long is also fine for the caller's size_accept.

arnab8335:
Hi @GitMensch 
One more question:
I understand that `ensure_buffer()` must report allocation failure so `field_accept()` can raise `EC-SCREEN-IMP-STORAGE` and either return or truncate the field. Since `cob_fast_malloc()` aborts on failure, should I add an exported non-aborting variant, for example `cob_fast_malloc_no_abort()`, that simply returns malloc(size) and returns NULL?

Should this variant be declared in common.h and implemented beside `cob_fast_malloc()` in `common.c`, then used only by `ensure_buffer()`? Or would you prefer using `malloc()` directly in ensure_buffer()?

GitMensch:
add a `cob_fast_malloc_or_null above` the existing function - and place the definition in coblocal.h (we don't want to export that "outside" of libcob

reasoning: a grep fro cob_fast_malloc will still show its use while directly telling the caller to inspect the null return

arnab8335:
Hi @GitMensch 
I have made the necessary changes , please review when you get time and please let me know review comments if any.

libcob/exception.def
```diff
@@ -565,6 +564,10 @@ COB_EXCEPTION  (0F04, COB_EC_SCREEN_LINE_NUMBER,
 COB_EXCEPTION  (0F05, COB_EC_SCREEN_STARTING_COLUMN,
 		"EC-SCREEN-STARTING-COLUMN", 0)
 
+/*Insufficent Storage for Screen Field*/
```
GitMensch, line 567:
```suggestion
/ *Insufficient Storage for Screen Field */
```

libcob/screenio.c
```diff
@@ -3530,6 +3532,39 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+	size_t buffer_size;
+	unsigned char *tmp;
+
+	if(requested_size > COB_TERM_BUFF_WARN_SIZE){
+		cob_runtime_warning (
+			_("ACCEPT/DISPLAY of unusually large field (%lu bytes)"),
+			requested_size);
+	}
+
+	if(requested_size <= screen_buffer_size){
```
GitMensch, line 3547:
```suggestion
	if (requested_size <= screen_buffer_size) {
```
please edit the view spacing style issues in this function similar

libcob/screenio.c
```diff
@@ -3607,20 +3642,29 @@ field_accept (cob_field *f, cob_flags_t fattr, const int sline, const int scolum
 			size_accept = cob_get_int (size_is);
 			/* SIZE ZERO is ignored */
 			if (size_accept < 1) {
-				size_accept = (int)f->size;
+				size_accept = f->size;
 			}
 		} else {
 			size_accept = f->size;
+		}		
+		if (size_accept && ensure_buffer(size_accept) != 0) {
```
GitMensch, line 3650:
```suggestion
		/* Note; may be a dynamic zero-size field */
		if (size_accept && ensure_buffer (size_accept) != 0) {
```

libcob/screenio.c
```diff
@@ -4254,7 +4298,7 @@ field_accept (cob_field *f, cob_flags_t fattr, const int sline, const int scolum
 		cob_move (&temp_field, f);
 		cob_move_cursor (sline, right_pos + 1);
 #if 0	/* possible cleanup to not "leak" input data */
-		memset (COB_TERM_BUFF, ' ', size_accept);
+		memset (screen_term_buff, ' ', size_accept);
```
GitMensch, line 4301: not checked the long branch above - can this be null - then we'd need a guard

libcob/termio.c
```diff
@@ -93,14 +98,14 @@ display_numeric (cob_field *f, FILE *fp)
...
-		COB_FIELD_INIT (size, COB_TERM_BUFF, &attr);
+		COB_FIELD_INIT (size, termio_buff, &attr);
 
 		cob_move (f, &field);
 	}
 
 	/* output of data to viewport */
 	{
-		register unsigned char *q = COB_TERM_BUFF;
+		register unsigned char *q = termio_buff;
```
GitMensch, line 108: this function should use a function-local fix-sized array of smaller size

libcob/termio.c
```diff
@@ -195,7 +200,7 @@ pretty_display_numeric (cob_field *f, FILE *fp)
 
 		/* output of data to viewport */
 		{
-			register unsigned char *q = COB_TERM_BUFF;
+			register unsigned char *q = termio_buff;
```
GitMensch, line 203: function-local fixed-size array (but with a bigger size than the non-pretty display)

libcob/termio.c
GitMensch, line 1100: we definitely need to guard against termio_buff's size here, for example by only increasing size if there is still space (< sizeof buffer)...

libcob/termio.c
```diff
@@ -1072,7 +1077,7 @@ cob_accept (cob_field *f)
 		}
 		return;
 	}
-	p = COB_TERM_BUFF;
+	p = termio_buff;
```
GitMensch, line 1080: This is the only place left... so maybe rename that to just "buffer" and move the static buffer inside this function (to reduce visibility - still static to no put too much pressure on the stack)?

tests/testsuite.src/run_manual_screen.at
```diff
@@ -4159,3 +4159,91 @@ AT_CHECK([$COMPILE prog.cob], [0], [], [])
...
+# output for a line such as:
+#   libcob: warning: ACCEPT/DISPLAY of unusually large field (40000 bytes)
+AT_DATA([prog.cob], [
+       IDENTIFICATION   DIVISION.
+       PROGRAM-ID.      prog.
+       DATA             DIVISION.
+       WORKING-STORAGE  SECTION.
+       01  big-field    PIC X(40000).
+
+       PROCEDURE        DIVISION.
```
GitMensch, line 4229: no need for a manual check - we should do that directly like with

```
MANUAL_CHECK([$COBCRUN_DIRECT ./prog], [0], [],
[libcob: warning: ACCEPT/DISPLAY of unusually large field (40000 bytes)
])
```

I guess the output will also contain the source line

tests/testsuite.src/run_manual_screen.at
GitMensch, line 1: please execute the test once with `make checkmanual TESTSUITEFLAGS="-k accept"` - see run_prog_manual.sh under tests (in your build directory) and adjust as needed for the test runner

GitMensch (changes requested):
A good approach with some minor things in screenio and a mayor one in termio... and a test request

arnab8335:
@GitMensch 
I have one question , I have already changed from `COB_TERM_BUFF `to `term_buff`
But why is it not reflecting in your suggestions.
In my commit also it is visible.

GitMensch:
That's an easy one. GitHub renders suggestions and comments similar, you can mostly differentiate them by the "apply suggestion" button.
In case of suggestions: these are a showcase of a change between your version and something proposed (even if not complete that can be much more clear than using words to describe them).
In case of comments: these show the difference between the original code and your commits.

arnab8335:
Hi @GitMensch ,
I have fixed the changes suggested by you.
Please have a look.

arnab8335:
Also @GitMensch I have a suggestion, came to my mind.
Can it be like if I start contributing in some other issues (say) will it make sense to provide you a design how I seem that needs to be solved, then you or anyone can provide his/her inputs and once the design is finalized we can focus on the implementation.
Does it seem right ?
Because I feel then the coding would be much more cleaner and easier, just putting the blocks together.
Thank You!

GitMensch:
> Can it be like if I start contributing in some other issues (say) will it make sense to provide you a design how I seem that needs to be solved, then you or anyone can provide his/her inputs and once the design is finalized we can focus on the implementation.

That all depends on how familiar with that part of the codebase you are - as currently nearly all is "new", it would be general reasonable to discuss an implementation for a fix / feature-request on the issue that handles it in more detail. This also provides you with the option to ask up-front about the code you've glanced over for that and is not clear.

arnab8335:
> > Can it be like if I start contributing in some other issues (say) will it make sense to provide you a design how I seem that needs to be solved, then you or anyone can provide his/her inputs and once the design is finalized we can focus on the implementation.
> 
> That all depends on how familiar with that part of the codebase you are - as currently nearly all is "new", it would be general reasonable to discuss an implementation for a fix / feature-request on the issue that handles it in more detail. This also provides you with the option to ask up-front about the code you've glanced over for that and is not clear.

Thanks for raising your point and actually what you are telling makes sense.

libcob/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2026-09-10  Arnab Dan <arnabdan.edu@gmail.com>
+
+	* screenio.c, termio.c: use private terminal buffers and handle
+	  insufficient screen buffer storage with EC-SCREEN-IMP-STORAGE.
+	* common.c, coblocal.h: add non-aborting allocation helper.
+	* common.h: remove the obsolete global screen buffer.
+	* exception.def: add EC-SCREEN-IMP-STORAGE.
```
GitMensch, line 8: minor formatting: we only use sentences (rarely) if there are longer texts, please drop the `.` here

libcob/ChangeLog
```diff
@@ -1,4 +1,12 @@
 
+2026-09-10  Arnab Dan <arnabdan.edu@gmail.com>
+
+	* screenio.c, termio.c: use private terminal buffers and handle
+	  insufficient screen buffer storage with EC-SCREEN-IMP-STORAGE.
+	* common.c, coblocal.h: add non-aborting allocation helper.
```
GitMensch, line 6:
```suggestion
	* common.c (cob_fast_malloc_or_null), coblocal.h: add non-aborting
     allocation helper
```

libcob/ChangeLog
GitMensch, line 8:
```suggestion
	* exception.def (EC-SCREEN-IMP-STORAGE): new nonfatal exception
```

libcob/coblocal.h
```diff
@@ -541,7 +540,7 @@ COB_HIDDEN int		cob_check_env_false	(char*);
 COB_HIDDEN const char	*cob_get_last_exception_name	(void);
 COB_HIDDEN void		cob_parameter_check	(const char *, const int);
 COB_HIDDEN char*        cob_get_strerror (void);
-
+COB_HIDDEN void	*cob_fast_malloc_or_null (const size_t);
```
GitMensch, line 543:
```suggestion
COB_HIDDEN void		*cob_fast_malloc_or_null (const size_t);
```

libcob/exception.def
GitMensch, line 2:
```suggestion
   Copyright (C) 2003-2012, 2014-2015, 2018-2021, 2025-2026 Free Software Foundation, Inc.
```

libcob/screenio.c
```diff
@@ -3607,20 +3642,30 @@ field_accept (cob_field *f, cob_flags_t fattr, const int sline, const int scolum
...
+				size_accept = f->size;
 			}
 		} else {
 			size_accept = f->size;
+		}		
+		/* Note; Maybe a dynamic zero sized field */
+		if (size_accept && ensure_buffer(size_accept) != 0) {
+			cob_set_exception (COB_EC_SCREEN_IMP_STORAGE);
+
+			if(screen_buffer_size == 0){
```
GitMensch, line 3654:
```suggestion
			if (screen_buffer_size == 0) {
```

libcob/screenio.c
```diff
@@ -3530,6 +3532,39 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+	
+	buffer_size = requested_size > COB_MINI_BUFF 
+		? (size_t)requested_size : COB_MINI_BUFF;
+
+	tmp = cob_fast_malloc_or_null(buffer_size);
+	if(tmp == NULL){
+		return 1;
+	}
+
+	if(screen_term_buff){
```
GitMensch, line 3559:
```suggestion
	if (screen_term_buff) {
```

libcob/screenio.c
```diff
@@ -3530,6 +3532,39 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+	}
+
+	if (requested_size <= screen_buffer_size) {
+		return 0;
+	}
+	
+	buffer_size = requested_size > COB_MINI_BUFF 
+		? (size_t)requested_size : COB_MINI_BUFF;
+
+	tmp = cob_fast_malloc_or_null(buffer_size);
```
GitMensch, line 3554:
```suggestion
	tmp = cob_fast_malloc_or_null (buffer_size);
```

libcob/screenio.c
```diff
@@ -3530,6 +3532,39 @@ field_display (cob_field *f, cob_flags_t fattr, const int line, const int column
...
+	if(tmp == NULL){
```
GitMensch, line 3555:
```suggestion
	if (tmp == NULL) {
```

libcob/termio.c
```diff
@@ -1092,7 +1097,9 @@ cob_accept (cob_field *f)
 		} else if (ipchr == '\n') {
 			break;
 		}
```
GitMensch, line 1099: we don't need to implement that here (this would belopng to another PR and likely would make some people happy), but should leave a comment - if the terminal is in raw mode we may need similar code to screenio.c (handling at least backspace + for numeric fields "prevent" the use of numeric fields), we may also handle noecho mode . .  or try to set those for the duration of that code (tcgetattr/tcsetattr)

Maybe leave two comments here  before the loop that the code expects a line-input, and a second one (could be  combined) a TODO that tcgetattr/tcsetattr should be inspected, to add a similar code like in screenio

GitMensch (changes requested):
that looks better again; I've left some code suggestions (you should be able to get them in all at once using the UI) and some general comments, the test looks like it needs most adjustment

any update from the copyright paper side?

arnab8335:
> that looks better again; I've left some code suggestions (you should be able to get them in all at once using the UI) and some general comments, the test looks like it needs most adjustment
> 
> any update from the copyright paper side?

Hi @GitMensch 
I am really sorry that i did not start the process regarding the copyright papers. I will start it today and let you know if i have any update in the following days.

arnab8335:
Hi @GitMensch 

Thank You so much for taking the time to review this - really appreciate it. I'm going to close the PR for now: I started the FSF copyright assignment process, but as an employed software engineer, FSF requires a signed employer disclaimer to proceed, and that isn't something I can practically arrange right now. My employer's own open-source policy does confirm this kind of private contribution doesn't need approval, but the formal disclaimer itself is a separate, bigger ask. 

Sorry for the back and forth, and thanks again for your time - I'd love to revisit this down the line if things change.

Thanks for the review!

PR #314, opened 2026-09-04
Implement bidirectional pipes on Windows

ddeclerck:
This PR implements the missing Win32 counterpart of PIPE I/O ; allows to pass two more test cases.

### DIFF
--- a/.github/workflows/windows-msys2.yml
+++ b/.github/workflows/windows-msys2.yml
@@ -150,7 +150,7 @@ jobs:
           # command it provides, to workaround a diff bug under the MSYS2 CI ;
           # this no longer seems necessary today
           # PATH="$PATH:/c/Program Files/Git/usr/bin"
-          ../configure $CFGOPT CFLAGS=-std=c17
+          ../configure $CFGOPT CFLAGS=-std=c17 LDFLAGS=-lintl
 
       - name: Upload config.log
         if: ${{ !cancelled() }}
@@ -193,13 +193,11 @@ jobs:
           # Skip test as it sometimes works and sometimes not...
           # sed -i '/AT_SETUP(\[temporary path invalid\])/a AT_SKIP_IF(\[true\])' used_binaries.at
 
-          # The following 13 fail (identical to MSYS1)
+          # The following 11 fail (identical to MSYS1)
           # Note LINE SEQUENTIAL one Record works with the clang64 environment
           sed -i '/AT_SETUP(\[Test dump feature (4)\])/a AT_SKIP_IF(\[true\])' run_misc.at
           sed -i '/AT_SETUP(\[OUTPUT on INDEXED file to missing directory\])/a AT_SKIP_IF(\[true\])' run_file.at
           sed -i '/AT_SETUP(\[INDEXED undeclared keys\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[PIPE I\/O\])/a AT_SKIP_IF(\[true\])' run_file.at
-          sed -i '/AT_SETUP(\[STDOUT STDERR I\/O\])/a AT_SKIP_IF(\[true\])' run_file.at
           sed -i '/AT_SETUP(\[LINE SEQUENTIAL one Record\])/a AT_SKIP_IF(\[true\])' run_file.at
           sed -i '/AT_SETUP(\[INDEXED COMMIT\/ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
           sed -i '/AT_SETUP(\[SEQUENTIAL\/RELATIVE COMMIT\/ROLLBACK\])/a AT_SKIP_IF(\[true\])' run_file.at
--- a/libcob/common.h
+++ b/libcob/common.h
@@ -1682,6 +1682,9 @@ typedef struct __cob_file {
 	const char			*xfdname;		/* Name for SQL table */
 	const char			*xfdschema;		/* Override of COB_SCHEMA_DIR for this file */
 	long				file_pid;		/* Process id of other end of pipe */
+#ifdef _WIN32
+	void *				file_handle;		/* Win32 hproc: handle of the process at end of (bidirectional) pipe */
+#endif
 	void				*fileout;		/* output side of bi-directional pipe 'FILE*' */
 	int					fdout;			/* output side of bi-directional pipe 'fd' */
 	int					limitreads;		/* Database should LIMIT rows read */
--- a/libcob/fileio.c
+++ b/libcob/fileio.c
@@ -4347,6 +4347,9 @@ cob_file_open (cob_file_api *a, cob_file *f, char *filename,
 	nonexistent = 0;
 	errno = 0;
 	f->file_pid = 0;
+#ifdef _WIN32
+	f->file_handle = NULL;
+#endif
 	f->flag_is_pipe = 0;
 	if (filename[0] == '>') {
 		if (mode != COB_OPEN_OUTPUT)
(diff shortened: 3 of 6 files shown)


libcob/fileio.c
```diff
@@ -4503,6 +4506,87 @@ cob_file_open (cob_file_api *a, cob_file *f, char *filename,
 			fflush(stderr);
 			exit(-1);
 		}
+#elif defined (_WIN32)
+		HANDLE				p_read, p_write; /* parent to child (parent writes, child reads)  */
```
GitMensch, line 4442: this should be in a new block (then the definitions of new vars is totally fine)

.github/workflows/windows-msys2.yml
```diff
@@ -150,7 +150,7 @@ jobs:
           # command it provides, to workaround a diff bug under the MSYS2 CI ;
           # this no longer seems necessary today
           # PATH="$PATH:/c/Program Files/Git/usr/bin"
-          ../configure $CFGOPT CFLAGS=-std=c17
+          ../configure $CFGOPT CFLAGS=-std=c17 LDFLAGS=-lintl
```
GitMensch, line 153: What did we need -lintl for?
Why not using -std=gnu17?

libcob/fileio.c
```diff
@@ -4503,6 +4506,87 @@ cob_file_open (cob_file_api *a, cob_file *f, char *filename,
...
+		/* Cleanup objects that are no longer needed  */
+		cob_free (cmdline);
+		CloseHandle (p_read);
+		CloseHandle (c_write);
+		CloseHandle (pi.hThread);
+
+		/* Update the cob_file structure accordingly */
+		f->fdout = _open_osfhandle ((intptr_t)p_write, 0);
+		f->fd = _open_osfhandle ((intptr_t)c_read, _O_RDONLY);
+		errno = 0;
```
GitMensch: That line isn't needed, is it?

libcob/fileio.c
```diff
@@ -4503,6 +4506,87 @@ cob_file_open (cob_file_api *a, cob_file *f, char *filename,
...
+		f->file = (void *)fdopen (f->fd, "r");
+		errno = 0;
```
GitMensch: I'm often confused when it comes to win32 and setting errno.
Is that fine or do we need `_set_errno(0)` as described in MSDN?

Does it make a difference at all, given, that we don't use `perror` or similar here?

libcob/fileio.c
```diff
@@ -4503,6 +4506,87 @@ cob_file_open (cob_file_api *a, cob_file *f, char *filename,
 			fflush(stderr);
 			exit(-1);
 		}
+#elif defined (_WIN32)
```
GitMensch: we _may_ move that up, first the code for WIN32, then for unistd_h

libcob/fileio.c
```diff
@@ -4794,9 +4878,24 @@ cob_file_close (cob_file_api *a, cob_file *f, const int opt)
 				}
 #elif defined (_WIN32)
 				{
-					char buff[COB_MINI_BUFF];
-					snprintf ((char *)&buff, COB_MINI_MAX, "TASKKILL    /PID %d", f->file_pid);
-					system (buff);
+					HANDLE hproc = (HANDLE)f->file_handle;
+					DWORD exit_code = 0;
+					if (hproc != NULL) {
```
GitMensch, line 4864: Should we fallback to taskkill if PID is set but hproc is null or the terminate didn't work?
Just wondering...

libcob/common.h
```diff
@@ -1682,6 +1682,9 @@ typedef struct __cob_file {
 	const char			*xfdname;		/* Name for SQL table */
 	const char			*xfdschema;		/* Override of COB_SCHEMA_DIR for this file */
 	long				file_pid;		/* Process id of other end of pipe */
+#ifdef _WIN32
+	void *				file_handle;		/* Win32: handle of the process at end of pipe */
```
GitMensch: What do we have to include to make it a `HANDLE` directly?
Do you see any issues of that (would remove the casting).

GitMensch (changes requested):
Nice - that's _nearly_ fine for upstream.

Depending on what we say in our docs (gnucobol.texi, NEWS) we may need an adjustment/entry there.
We definitely need a Changelog entry :-)

.github/workflows/windows-msys2.yml
ddeclerck: This is just a temporary workaround.
This is aleady taken care of in PR 301, cf https://github.com/OCamlPro/gnucobol/pull/301#discussion_r3903112833.

libcob/fileio.c
ddeclerck: For this part I just mirrored the existing Linux code without really thinking.

libcob/fileio.c
ddeclerck: Do you have any scenario where PID would be set but hproc is null ?

GitMensch:
How much work would it be to backport the whole feature to 3.x?

libcob/common.h
ddeclerck: We'd need to include windows.h - but my attempt at doing so generated errors elsewhere (I think it was in the parser).

libcob/fileio.c
GitMensch: I just was wondering as I haven't followed the complete code flow.
Also I'm not sure if there's a reason that Terminate(hproc) won't work but taskkill would...

ddeclerck:
> How much work would it be to backport the whole feature to 3.x?

Probably not too much, if someone is willing to do this.

libcob/common.h
GitMensch: let's just leave it as is, maybe change the comment to `Win32 hproc: handle of the process at end of (bidirectional) pipe */

libcob/fileio.c
ddeclerck: Note we're already in a block :
```
	if (filename[0] == '|') {
#if defined (HAVE_UNISTD_H) && !(defined (_WIN32))
...
#elif defined (_WIN32)
...
#endif
	}
```
(the POSIX counterpart does not open a new block)

GitMensch: sorry for the misleading comment - you're right, this whole part is in a new block, which starts in line 4440

libcob/fileio.c
ddeclerck: Unless I missed something, the current code ensures PID and handle are always in sync, so we should never end up with PID = something while handle = null.
As for TerminateProcess, since we're terminating our own child and not some random process (which would have different execution privileges for instance), the only reason I can think of that would make it to fail would be if the process is already terminated.

libcob/fileio.c
ddeclerck: Apparently on Windows `errno` is just a macro that returns a (dereferenced) thread-local pointer (obtained bia a function call) ; it seems just fine to use `errno`.

But anyways, as you mention, we don't even check for errno, nor we use `perror`, so I might as well just remove these lines (also in the POSIX case).

utam-1: I hope it's fine if I comment here 😅, I had worked on a small personal project that used windows API so I'm just sharing what I know.
The only issue we may come across with `errno` is that local windows function like `CreateFileA` for example rely on specific windows error code and don't update `errno`.
Otherwise, it's fine to use this.

GitMensch (approved):
Good for upstream 4.x - would welcome a 3.x backport of the pipe file feature if someone provides this.

ddeclerck:
Merged in SVN @ 5704.

PR #315, opened 2026-09-07
Update ltmain script

Halbaroth:
The libtool scripts of the source tree are generated by both 2.5.3 and 2.5.4. As a result, the compilation fails with the following error:
```console
libtool: Version mismatch error.  This is libtool 2.5.3, but the
libtool: definition of this LT_INIT comes from libtool 2.5.4.
libtool: You should recreate aclocal.m4 with macros from libtool 2.5.3
libtool: and run autoconf again.
```

The new scripts were generated by `libtoolize -c` with libtool 2.5.4.

To test this PR, run: 
```console
./autogen.sh
./configure CFLAGS='-std=c17'
make
```

### DIFF
--- a/build_aux/ltmain.sh
+++ b/build_aux/ltmain.sh
@@ -2,7 +2,7 @@
 ## DO NOT EDIT - This file generated from ./build-aux/ltmain.in
 ##               by inline-source v2019-02-19.15
 
-# libtool (GNU libtool) 2.5.3
+# libtool (GNU libtool) 2.5.4
 # Provide generalized library-building support services.
 # Written by Gordon Matzigkeit <gord@gnu.ai.mit.edu>, 1996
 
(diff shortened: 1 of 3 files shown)


GitMensch:
Thanks for taking the time to write a PR with understandable explanation!

While we may update the libtool version the issue seems that your git checkout does not preserve the timestamps (or, also possible but less likely, they are broken in this git mirror) and therefore make tries to rebuild libtool.

Fix if you only want to build GnuCOBOL: use one of the nightly tarballs from CI instead of git checkout - those also do not need any of bison/flex/automake/libtool/help2man/texinfo/... locally installed.

Likely fix: check out or restore mtime via git tools.

Workaround (which should be noted in HACKING); `./autogen.sh --install` (should install the versions that are locally installed to the checked out repository), so rebuilding libtool will then have a matching version

Halbaroth:
> Workaround (which should be noted in HACKING); ./autogen.sh --install (should install the versions that are locally installed to the checked out repository), so rebuilding libtool will then have a matching version

I knew this workaround, thanks ;)

PR #313, opened 2026-09-01
Various GC3 fixes

ddeclerck:
Follow-up of #303, which was merged too soon.

PR #316, opened 2026-09-08
Fix COB_SYNC

ddeclerck:
This PR fixes two things:
- a first bug prevented COB_SYNC from actually working (dead condition)
- fixing it revealed another bug, where the use of COB_SYNC caused incomplete reads (actualy PR #313 attempted to fix this, but there was a missing part - ie calling `fflush` only after write operations)

### DIFF
--- a/libcob/ChangeLog
+++ b/libcob/ChangeLog
@@ -1,4 +1,16 @@
 
+2026-09-08  David Declerck <david.declerck@ocamlpro.com>
+
+	* fileio.h (cob_file_operation, cob_file): new enum
+	  cob_file_operation to memorize the last file operation
+	* fileio.c: use the new file field last_operation instead
+	  of last_operation_open
+	* fileio.c (cob_file_fcd_sync): move reset of last_operation
+	  to save_status to fix a bug (uneffetive COB_SYNC)
+	* fileio.c (cob_sync): add missing condition - fflush
+	  should only be called when the last operation is a write
+	  (follow-up to 2026-08-13)
+
 2026-09-01  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
 
 	* profiling.c (print_stack): export/debug option with environment
--- a/libcob/fileio.c
+++ b/libcob/fileio.c
@@ -542,7 +542,6 @@ static cob_settings	*cobsetptr = NULL;
 static unsigned int	eop_status = 0;
 static unsigned int	check_eop_status = 0;
 static int		cob_vsq_len = 0;
-static int		last_operation_open = 0;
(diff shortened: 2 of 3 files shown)


libcob/ChangeLog
```diff
@@ -1,4 +1,16 @@
 
+2026-09-08  David Declerck <david.declerck@ocamlpro.com>
+
+	* fileio.h (cob_file_operation, cob_file): new enum
+	  cob_file_operation to memorize the last file operation
```
GitMensch, line 5: that should be common.h - I _guess_ that part is a backport of 4.x - best would be to leave a merge note here and add the original 4.x changelog note down where it belongs (done in other places already, search for fileio.h or fbdb.c); doing so may show other very reasonable and relative easy parts to backport (_maybe_ also the option to do that via svn)

GitMensch (changes requested):
please add test cases that are currently broken to this PR

ddeclerck:
> please add test cases that are currently broken to this PR

This comes from customer feedback and is hard to test reliably (needs one process that writes and one that reads the same file, timing-sensitive)

libcob/fileio.c
```diff
@@ -6607,6 +6623,7 @@ cob_read (cob_file *f, cob_field *key, cob_field *fnstatus, const int read_opts)
 
 	/* Sequential read at the end of file is an error */
 	if (key == NULL) {
+		f->last_operation = COB_LAST_READ_SEQ;
```
GitMensch, line 6626: note: this is not test-covered yet

libcob/fileio.c
```diff
@@ -1403,7 +1402,10 @@ cob_sync (cob_file *f)
 		return;
 	}
 	if (f->organization != COB_ORG_SORT && f->open_mode != COB_OPEN_INPUT) {
-		if (f->file) {
+		if (f->file
+		 && (f->last_operation == COB_LAST_WRITE
+		  || f->last_operation == COB_LAST_REWRITE
+		  || f->last_operation == COB_LAST_DELETE)) {
```
GitMensch, line 1408: this (the DELETE) is not covered

GitMensch:
I see. Ron added a bunch of multi-process tests for file io to 4.x - not sure if there is a COB_SYNCH one that we can at least backport to get coverage of that code + actual multi-process tests (I think we disabled a bunch of them later as there were portability issues, can't remember if they were solved and enabled again later),
Note: the changed code itself is nearly test-covered, possibly not with multiple processes, will note the missing ones next.

GitMensch (commented):
just checked - nearly covered completely

PR #319, opened 2026-09-17
fix parser state reset

lefessan:
When calling `cobc -x -o prog.exe x1.cob x2.cob` with `x1.cob` incorrect and `x2.cob` correct, the compiler displays errors on `x2.cob` that are actually early errors of `x1.cob`. The reason is that the state of the parser is only partially reset between the two files.


tests/testsuite.src/syn_misc.at
```diff
@@ -8711,3 +8711,30 @@ AT_CHECK([$GREP "#INCLUDE <file3.h>" prog.c], [1], ignore, [])
...
+])
+AT_DATA([prog2.cob], [
+       IDENTIFICATION DIVISION.
+       PROGRAM-ID. prog1.
+       PROCEDURE DIVISION.
+           DISPLAY "prog1".
+           GOBACK.
+])
+
+AT_CHECK([$COMPILE prog1.cob prog2.cob], [1], [],
```
nberth, line 8733: `COMPILE_ONLY`?

lefessan: I prefer checking as much as possible

GitMensch: It is a syntax test and we don't expect to execute something, so $COMPILE_ONLY is the right thing to do.

tests/testsuite.src/syn_misc.at
```diff
@@ -8711,3 +8711,30 @@ AT_CHECK([$GREP "#INCLUDE <file3.h>" prog.c], [1], ignore, [])
 AT_CHECK([$GREP "#INCLUDE <file4.h>" prog.c], [1], ignore, [])
 
 AT_CLEANUP
+
+AT_SETUP([cobc multiple files with errors])
```
GitMensch, line 8715:
```suggestion

AT_SETUP([cobc multiple files with errors])
```

cobc/ChangeLog
```diff
@@ -1,4 +1,10 @@
 
+2026-09-17  Fabrice Le Fessant <fabrice.le_fessant@ocamlpro.com>
+
+	* scanner.l (ylex_call_destroy): reset more state variables
+	  between two calls to the parser
```
GitMensch, line 5: the change looks good, but the test doesn't match

GitMensch (changes requested):
The test and changes are not wrong "per se", but the test passes nicely with the current revision - please provide a test that shows the actual error you fix

lefessan:
Indeed, this error is actually caused by another patch, that relaxes a little the syntax at the beginning for better compatibility with MF. I still think that the fix is needed (full reset of all variables).
Shall I close the PR ? or remove the test and the comment ?

GitMensch:
This PR should be extended to add the other patch - this way the new test verifies that there is no new error coming up and we have the other feature - most likely guarded by -frelaxed-syntax-checks - in as well.

PR #320, opened 2026-09-17
pre-load adjustment

GitMensch:
libcob:
* call.c (pre_load_once): extracted from cob_init_call
*  call.c (pre_load_once): elements in COB_PRE_LOAD are always loaded with a global scope request, independent of COB_LOAD_GLOBAL setting
* common.c: minimal refactor


GitMensch:
all fine, just opened for CI

PR #321, opened 2026-09-17
Fix ignored END-EVALUATE after a WHEN without imperative statements

GBertholon:
Changes in cobc:
* parser.y (rules evaluate_when and evaluate_other): repeat the END-EVALUATE token if consumed by the empty imperative statement rule.


tests/testsuite.src/run_misc.at
```diff
@@ -14327,6 +14327,35 @@ E: OK OTHER option taken
...
+               EVALUATE B
+                   WHEN 1
+               END-EVALUATE
+      * If END-EVALUATE is ignored, WHEN 2 will be attached to EVALUATE B
+           WHEN 2
+               DISPLAY "FAIL"
+       END-EVALUATE.
+])
+
+AT_CHECK([$COMPILE -fmissing-statement=ok prog.cob], [0], [], [])
```
GitMensch:
```suggestion
# explicit check that we only get only a single warning for WHEN
# and no wrong warnings for the END-EVALUATE terminator
AT_CHECK([$COMPILE -fmissing-statement=warning -Wall prog.cob], [0], [],
[prog.cob:14: warning: WHEN without imperative statement used
prog.cob:17: warning: DISPLAY statement not terminated by END-DISPLAY
])
```

GitMensch (changes requested):
Nice find! I kind of feel dumb because that strange terminator warning kind of foreshaddowed that.

Also a well done fix - only a request for changing the testsuite, can go upstream afterwards, so I already mark it that way (enough to test locally, if that works, feel free to directly commit upstream).

ddeclerck (approved):

tests/testsuite.src/run_misc.at
GBertholon: I added the warning check but without the `DISPLAY statement not terminated by END-DISPLAY` because it is not emitted on the example.

GitMensch: oh, it is, as soon as you add `-Wall` as suggested :-)

GBertholon: Well no the cb_warn_terminator warnings are activated only with `-Wextra`... Since checking this kind of warning is important for this test we probably should add `-Wterminator` instead, no?

For another PR, this `END-DISPLAY` warning is really weird if there is no exception statements in the DISPLAY. I think we should remove it in such cases even when `-Wterminator` is activated.

GitMensch: For -Wterminator here: yes, that's the right approach, so this PR is ready-for-svn-now :-)

For "not warn if no exception statements are used" -> that's the reason that -Wterminator was moved from -Wall to -Wextra :-) I agree that it would be good to get a separate PR for:
* split -Wterminator - keeping all of the "reasonable" ones (also for CALL and others) in that warning and move the "less useful" ones out to -Wterminator-all
* let -W[no-]terminator-all enable/disable -Wterminator automatically
* keep -Wterminator in -Wextra, moving -Wterminator back to -Wall

Any chance you can have a look at that once this PR is upstream?

Do you want to do the upstreaming yourself or should someone else do that?

GBertholon: You can do the upstreaming, I am not familiar with SVN.
For -Wterminator I added it in my low-priority task backlog.

PR #322, opened 2026-09-18
Fix temporary file collisions between users

ddeclerck:
This patch was made for a customer who had collision problems with temporary filenames.

It does the following:
- put temporary files in per-user subdirectories of the temporary directory
- use random numbers in generated file names (in addition to the pid)
- cleanup temporary files even on unexpected exit


cobc/cobc.c
```diff
@@ -2214,6 +2201,31 @@ cobc_check_action (const char *name)
 	}
 }
 
+/* atexit handler to remove any leftover temporary file
+   (for code paths that do not reach cobc_clean_up) */
```
GitMensch, line 2205: Which code-paths are that?
... and shouldn't we add the cleanup there?

libcob/common.c
```diff
@@ -6191,48 +6233,104 @@ cob_gettmpdir (void)
...
 #undef TEMP_EXT_SCHEMA
 #undef TEMP_SORT_SCHEMA
 }
 
+/* no longer needed but kept for binary compatibility (exported symbol) */
 void
 cob_incr_temp_iteration (void)
 {
-	cob_temp_iteration++;
 }
```
GitMensch, line 6334: this is only "half public" API - I've added it only for cobc, if we don't use it there anymore then both the declaration and now empty function should be deleted

libcob/common.c
```diff
@@ -6191,48 +6233,104 @@ cob_gettmpdir (void)
...
+#endif
+
 	/* get the pointer to the environment copy - as this may point to a different place -
 	   store it for subsequent calls and finally return it to the caller */
 	tmpdir = getenv ("TMPDIR");
+
 	return tmpdir;
 }
 
+/* Gennerate a random number for use with temporary file names */
```
GitMensch, line 6264:
```suggestion
/* Generate a random number for use with temporary file names */
```

libcob/ChangeLog
```diff
@@ -1,12 +1,15 @@
...
-	  that setjmp / longjmp can be executed only once as a user may store
-	  multiple places and also may jump to them multiple times - it is up
-	  to the user to ensure a correct stack (= never jump "forward")
+2026-09-18  David Declerck <david.declerck@ocamlpro.com>
+
+	* common.c (get_user_tmpdir): new function to create a private
+	  per-user sub-directory int the temporary directory
+	* common.c (cob_gettmpdir): use directory returned by get_uer_tmpdir
+	* common.c (get_temp_random): new function to return a random
+	  number of tuse in temporary file name
```
GitMensch, line 8:
```suggestion
	  number to be used for temporary file name generation
```

libcob/ChangeLog
```diff
@@ -1,12 +1,15 @@
...
-	* call.c (cob_setjmp), common.h: cob_setjmp is now a proper function
-	  instead of a plain define, and therefore potentially CALLable
-	* call.c: drop internal marker for "longjmp available" and limitation
-	  that setjmp / longjmp can be executed only once as a user may store
-	  multiple places and also may jump to them multiple times - it is up
-	  to the user to ensure a correct stack (= never jump "forward")
+2026-09-18  David Declerck <david.declerck@ocamlpro.com>
+
+	* common.c (get_user_tmpdir): new function to create a private
+	  per-user sub-directory int the temporary directory
```
GitMensch, line 5:
```suggestion
	  per-user sub-directory below the temporary directory
```

libcob/common.c
```diff
@@ -6145,13 +6151,49 @@ check_valid_env_tmpdir (const char *envname)
...
+
+	if (mkdir (dir, 0700) != 0) {
+		/* ensure it is a real directory owned by the current user */
+		if (errno != EEXIST
+		 || lstat (dir, &sb) != 0
+		 || !S_ISDIR (sb.st_mode)
+		 || sb.st_uid != uid) {
+			cob_runtime_warning (_("temporary directory %s is not usable"), dir);
+			cob_free (dir);
+			return NULL;
```
GitMensch: I'd tend to change the ABI to always return a temporary dir - so in the error case here tmpdir --> then no caller needs to do an explicit check and does it on his own

libcob/common.c
```diff
@@ -6191,48 +6233,104 @@ cob_gettmpdir (void)
...
+
+#if !defined (_WIN32) && !defined (HAVE_8DOT3_FILENAMES)
+	user_dir = get_user_tmpdir (tmpdir);
+	if (user_dir != NULL) {
+		tmpdir = user_dir;
+	}
+#endif
+
 	/* ensure TMPDIR is set for called tools (which partially break hard otherwise) */
 	(void)cob_setenv ("TMPDIR", tmpdir, 1);
```
GitMensch, line 6245: This code gets TMPDIR above, adds username then sets TMPDIR below... and on the next iteration adds the user again, no?

GitMensch (changes requested):
Please explain _how_ it would be possible to have the same pid and the same iteration "in parallel"?
I don't think that is possible (if it is, then we have to fix the bug) so I guess that's "user a created a temp file which was left around because of X, then user b comes after the PID had its reset [because of many PIDs created or because of a reboot] and does not have the necessary rights to the file"

Is that right? If that is the case then we don't use the random hex _and_ the user (as noted in the review the user setting has some issues, so possibly get rid of that?), right?
Note: if we _really_ want the user to distinguish further, then I'd tend to add it to the temporary filenames (cob_temp_name) instead of tinkering with TMPDIR even more.

But that leaves the question: wouldn't it be better to just check if the temp_name already exists, and if this is the case call cob_incr_temp_iteration? That would also be a single function change without changing the names at all.

ddeclerck:
> Please explain _how_ it would be possible to have the same pid and the same iteration "in parallel"?

I don't think I used the term "parallel".

> I don't think that is possible (if it is, then we have to fix the bug) so I guess that's "user a created a temp file which was left around because of X, then user b comes after the PID had its reset [because of many PIDs created or because of a reboot] and does not have the necessary rights to the file"

> Is that right? If that is the case then we don't use the random hex _and_ the user (as noted in the review the user setting has some issues, so possibly get rid of that?), right?

My understanding is that AIX reuses pids agressively compared to Linux. As soon a process terminates, its pid can be reused for a new process - which might be by another user. The customer had two problems :
- an easy to reproduce one : leftover files with different permissions (many users work at the same time on this machine), which the next call to cobc with a reused pid would fail to overwrite
- a very rare one, which we are not exactly sure about the cause : with fast consecutive calls to cobc by the same user (build script), likely recycling the pid, it "seems" that sometimes cobc fails to write to a file if a previous cobc invocation already produced a file by that name

> Note: if we _really_ want the user to distinguish further, then I'd tend to add it to the temporary filenames (cob_temp_name) instead of tinkering with TMPDIR even more. 

Having a per-user temp directory seems to be a common idiom, but indeed to make it simpler having the user id directly in the temporary file name looke like a good compromise - less modifications to the existing code.

> But that leaves the question: wouldn't it be better to just check if the temp_name already exists, and if this is the case call cob_incr_temp_iteration? That would also be a single function change without changing the names at all.

That might work.

cobc/cobc.c
ddeclerck: I'm not sure, but the fact is we had leftover files. Maybe after crashes ?

libcob/common.c
ddeclerck: No, once tmpdir is set, we never change it again, as stated above :
```
if (tmpdir != NULL) {
		return tmpdir;
}
```

cobc/cobc.c
GitMensch: a real crash should of course not happen, but even a CTRL+C you may have those - but then at_exit isn't called, I'd guess

